PR #1, opened 2024-05-01
Add vertex evaluator information to docs

tagboola:


schnecle (approved):

schnecle:
Pavel added Kevin, copying that over here.

docs/plugins/vertex-ai.md
```diff
@@ -147,3 +156,36 @@ const llmResponse = await generate({
...
+      },
+    }),
+  ],
+  // ...
+});
+```
+
+The configuration above adds evaluators for the `Safety` and `ROUGE` metrics. The example shows two approaches- the `Safety` metric uses the default specification, whereas the `ROUGE` metric provides a customized specification that sets the rouge type to `rougeLsum`.
+
+Both evaluators can be run using the `genkit eval:run` with a compatible dataset i.e. a datset with `output` and `reference` fields. The `Safety` evaluator can also be run using the `genkit eval:flow -e vertexai/safety` command since it only requires an `output`.
```
kevinthecheung, line 191:
```suggestion
Both evaluators can be run using the `genkit eval:run` command with a compatible dataset: that is, a dataset with `output` and `reference` fields. The `Safety` evaluator can also be run using the `genkit eval:flow -e vertexai/safety` command since it only requires an `output`.
```

kevinthecheung (approved):

PR #22, opened 2024-05-03
Support system prompt natively for Gemini 1.5

MichaelDoyle:
Gemini 1.5 supports system instructions natively, but we do not utilize it currently. This proposal would fix that.

The main gotcha here, is that Gemini implemented `systemInstructions` as a new input parameter, rather than leveraging messages with the `role=system`. Since our `GenerateRequest` abstraction is currently predicated on this paradigm (`role=system`) I had to do a bit of legwork. This proposal will:

- If system instructions are supported, look to see if the first message has `role=system`
- Transform it into a message with `role=user` and pass to the model as `systemInstructions`
- Splice the system message off of the history
- Throw an Error any time a message with `role=system` appears anywhere else in the request

_Alternatively, we could change GenerateRequest to accept_ `systemInstructions` _directly._


js/plugins/googleai/src/gemini.ts
```diff
@@ -279,10 +312,26 @@ export function googleAIModel(
...
+
+      // Gemini does not support messages with role system and instead
+      // expects system instructions to be provided as a separate input.
+      // By convention, system message are expected to be in the "first"
+      // position, so we look for it there. System messages anywhere else
+      // are considered "exceptional".
+      let systemInstruction;
+      if (messages[0].role === 'system') {
+        systemInstruction = toGeminiSystemInstruction(messages[0]);
+      }
```
mbleigh, line 327: I'd recommend just detecting and splicing off the first system message. Something like:

```ts
const systemMessage = messages.find(m => m.role === 'system');
if (systemMessage) messages.splice(messages.indexOf(systemMessage), 1);
const systemInstruction = toGeminiSystemInstruction(systemMessage);
```

mbleigh (commented):
Food for thought: I intended ModelInfo to be user-facing indicators of what's supported. So a model that has a synthetic system prompt would still support a system prompt as far as ModelInfo is concerned. I'm not sure how to rationalize that with the need to change model factory behavior based on underlying capability tho...

js/plugins/googleai/src/gemini.ts
mbleigh: How about above you do `const messages = [...request.messages]` to copy the array, then it won't be modified in place.

MichaelDoyle:
> Food for thought: I intended ModelInfo to be user-facing indicators of what's supported. So a model that has a synthetic system prompt would still support a system prompt as far as ModelInfo is concerned. I'm not sure how to rationalize that with the need to change model factory behavior based on underlying capability tho...

Yes, this is a _great_ point. I'll noodle on this some more.

MichaelDoyle:
Updated `supports: systemRole` to be a user-facing indicator.

js/plugins/vertexai/src/gemini.ts
```diff
@@ -94,27 +96,45 @@ export const gemini15ProPreview = modelRef({
...
     case 'user':
       return 'user';
     case 'model':
       return 'model';
     case 'system':
-      throw new Error('system role is not supported');
+      if (model && SUPPORTED_V15_MODELS[model.name]) {
+        throw new Error(
+          'system role is only supported for a single message in the first position'
+        );
```
pavelgj, line 134: I don't understand the combination of this condition and this error message... is it somehow implied in the context of this function that we're not in "single message in the first position"? Doesn't look like it...

MichaelDoyle: It's not implied, you are correct. It's loosely coupled based on the fact that we pull off the first system message and leave the rest. 

Technically what could happen is we get multiple messages of role system. e.g.)

```
system: always be friendly
user: say hello
model: oh hello my very best friend in the whole world
system: but not tooooo friendly
user: say hello
...
```

Right now, we pick the "first one", and leave the rest in place. The question is, what to do w/ the rest? Would it be better just to throw right away if we detect this? It's not super clear cut.

- If we pull them out, we can't guarantee the messages will be in order (gemini expects user/model turns only).
- If we concat them all, it's not clear the prompt will behave as expected
- etc

js/plugins/vertexai/src/gemini.ts
```diff
@@ -94,27 +96,45 @@ export const gemini15ProPreview = modelRef({
...
+): string {
   switch (role) {
     case 'user':
       return 'user';
     case 'model':
       return 'model';
     case 'system':
-      throw new Error('system role is not supported');
+      if (model && SUPPORTED_V15_MODELS[model.name]) {
+        throw new Error(
```
pavelgj, line 132:
```suggestion
      if (model && SUPPORTED_V15_MODELS[model.name]) {
        // we should've already pulled out the supported system messages, anything
        // else left in messages we don't support, throw an error.
        throw new Error(
```

pavelgj (approved):

PR #35, opened 2024-05-06
Introduced defineAction which handles registry for actions under the hood

pavelgj:
```ts
const geminiPro = defineAction({
  actionType: 'model',
  name: 'googleai/gemini-pro',
  // ...
}, (input) => {
  // ...
})
```

or

```ts
const geminiPro = defineAction({
  actionType: 'model',
  name: {
    pluginId: 'googleai',
    actionId: 'gemini-pro',
  },
  // ...
}, (input) => {
  // ...
})
```

this will register this action in the registry under `/model/googleai/gemini-pro`

Formalized action naming -- there are two types of actions
    1. actions provided by plugins - they must be named `pluginId/actionId`
    2. actions not provided by plugins - their name cannot contain slashes (/)

actions not associated with a plugin

```ts
const geminiPro = defineAction({
  actionType: 'model',
  name: 'myCustomModel',
  // ...
}, (input) => {
  // ...
})
```


mbleigh (approved):
Only had time for a quick glance, but 👍  on the direction.

PR #42, opened 2024-05-06
Use dotprompt in evaluators

ssbushi:


schnecle (approved):

docs/dotprompt.md
```diff
@@ -287,10 +287,13 @@ import {
...
   defineDotprompt,
 } from '@genkit-ai/dotprompt';
+import path from 'path';
 import { z } from 'zod';
 
 // Load a prompt from a file
-const myPrompt = await loadPromptFile('./path/to/my_prompt.prompt');
+const myPrompt = await loadPromptFile(
+  path.resolve(__dirname, './path/to/my_prompt.prompt')
+);
```
ssbushi, line 296: Not sure if this the preferred way or if this should be mentioned as an alternative.

pavelgj: yeah, I'm a bit torn, but I think this way is more flexible, if a little more verbose. I think it's good.

pavelgj:
Ugh... I have a really bad headache.... instead of "approve" I hit "merge"...

PR #46, opened 2024-05-06
Make node version requirement explicit in genkit CLI

tonybaroneee:


genkit-tools/package.json
```diff
@@ -33,6 +33,9 @@
     "url": "https://github.com/firebase/genkit.git",
     "directory": "genkit-tools"
   },
+  "engines": {
+    "node": ">=20.0.0"
```
pavelgj, line 37: Is it only a requirement in dev or runtime as well?

tonybaroneee: I believe runtime as well, Jonathan and Ian were getting those errors when trying to use the CLI runner js with Node v18.

tonybaroneee: Maybe we just remove this though and people can try with lower versions if they want (with the caveat that it might not work). I can re-phrase the contributing.md change for this.

tonybaroneee:
@pavelgj yeah or nay on this?

pavelgj (approved):

PR #57, opened 2024-05-07
Create code for menu/basic example

MaesterChestnut:
https://github.com/firebase/genkit/issues/56

Adding a menu/basic example to link to from our docs. This will allow users to go code and immediately be able to run an example of what they just read in the docs.

image


schnecle (approved):

maxl0rd (changes requested):
I think it's confusing to have this nested inside the existing sample. Could we put up a level it in it's own samples/xxxxx folder. When it supersedes the existing sample we can just remove that one?

schnecle (commented):
It looks like we have menu-example/basic/basic - should that just be menu-example/basic?

js/samples/menu-example/basic/basic/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  {
+    name: 'menuQAFlow',
+    inputSchema: z.string(),
+    outputSchema: z.string(),
+  },
+  async (subject) => {
+    const llmResponse = await generate({
+      model: geminiPro,
+      prompt: `Our menu today includes burgers, spinach, and cod.
+      Tell me if ${subject} can be found on the menu`,
```
MichaelDoyle, line 41: I am not 💯 on how it ties in with your broader plans, but what we discussed in-person was perhaps turning this into something more creative/generative in nature, like a menu generator, or similar.

maxl0rd (approved):

MichaelDoyle (approved):

PR #58, opened 2024-05-07
[Go] Added missing TraceID to go TraceData and made timeEvent optional

pavelgj:


jba (approved):
Thanks for fixing our code!

PR #61, opened 2024-05-07
Adds context as first-class feature of Model.

mbleigh:
This adds the ability to pass context (an array of `DocumentData[]`) to a model and adds default support for prompt augmentation if a model does not declare specific support for context. Usage:

```ts
const query = "What do cats like to eat?";
const docs = await retrieveStuff({query});

generate({
  prompt: `Answer this question: ${query}`,
  context: docs
});
```

This will augment the prompt to look something like this:

```
Answer this question: What do cats like to eat?

Use the following information to complete your task:

- [0]: Cats eat a variety of foods including dry kibble and wet foods that can be purchased at a pet store.
- [1]: Cats will hunt mice and birds for fun and for food.
```

### DIFF
--- a/js/dotprompt/src/prompt.ts
+++ b/js/dotprompt/src/prompt.ts
@@ -25,9 +25,9 @@ import {
   toGenerateRequest,
 } from '@genkit-ai/ai';
 import { GenerationCommonConfigSchema, MessageData } from '@genkit-ai/ai/model';
-import { DocumentData, DocumentDataSchema } from '@genkit-ai/ai/retriever';
+import { DocumentData } from '@genkit-ai/ai/retriever';
 import { GenkitError } from '@genkit-ai/core';
-import { parseSchema, toJsonSchema } from '@genkit-ai/core/schema';
+import { parseSchema } from '@genkit-ai/core/schema';
 import { createHash } from 'crypto';
 import fm, { FrontMatterResult } from 'front-matter';
 import z from 'zod';
@@ -153,7 +153,7 @@ export class Dotprompt<Variables = unknown> implements PromptMetadata {
           prompt: this.toJSON(),
         },
       },
-      async (input?: Variables) => toGenerateRequest(this.render({input}))
+      async (input?: Variables) => toGenerateRequest(this.render({ input }))
     );
   }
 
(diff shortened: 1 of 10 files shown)


js/dotprompt/src/prompt.ts
```diff
@@ -9,7 +9,7 @@
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+cd * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
maxl0rd, line 12: oops

maxl0rd (commented):
I think this looks good. I like how dotprompt will let you position the context in the template.

Made one comment where I think you could thread it through to vanilla prompts too.

pavelgj (approved):

js/dotprompt/src/template.ts
```diff
@@ -41,6 +48,23 @@ function mediaHelper(options: Handlebars.HelperOptions) {
...
+//       text += `[${d.metadata[options.hash.cite]}]`;
+//     }
+//     return Promptbars.escapeExpression(text);
+//   });
+
+//   return new Promptbars.SafeString(`<<<dotprompt:section context>>>
+// - ${items.join('\n- ')}`);
+// }
+// Promptbars.registerHelper('context', contextHelper);
+
```
ssbushi, line 67: Remove?

js/dotprompt/src/template.ts
```diff
@@ -103,16 +134,19 @@ export function compile<Variables = any>(
       media: true,
       role: true,
       history: true,
+      // context: true,
```
ssbushi, line 137: Remove.

js/dotprompt/tests/template_test.ts
```diff
@@ -106,6 +106,69 @@ describe('compile', () => {
       template: '{{json . indent=2}}',
       want: [{ role: 'user', content: [{ text: '{\n  "test": true\n}' }] }],
     },
+    // {
+    //   should: 'allow defining context',
```
ssbushi, line 110: Remove?

ssbushi (approved):

PR #62, opened 2024-05-07
Update genkit init templates and docs

MaesterChestnut:
https://github.com/firebase/genkit/issues/56

Replace genkit init output with "menuSuggestionFlow" & modify documentation for Getting Started 
image


genkit-tools/config/nextjs.genkit.ts.template
```diff
@@ -38,3 +34,8 @@ const jokeFlow = defineFlow(
     return llmResponse.text();
   }
 );
+
+export async function callMenuQAFlow() {
```
schnecle, line 38: I think we can leave this off.

schnecle (approved):

docs/get-started.md
```diff
@@ -71,36 +71,32 @@ Node.js 18 or later.
     1. Choose default answers to the rest of the questions, which will
        initialize your project folder with some sample code.
 
-    The `genkit init` command created two sample source files:
+    The `genkit init` command creates a sample source file:
 
-    - `genkit.config.ts`: this is where you configure Genkit for your project,
-      and select and configure the plugins you want to load. The sample config
-      loads a plugin to support the model provider you chose earlier.
+    - `index.ts`: this is your project's entry point, where you configure Genkit for your project,
```
kevinthecheung, line 76: "... a sample source file, index.ts. This is your..."

(and fix indentation after)

kevinthecheung (approved):

docs/get-started.md
```diff
@@ -71,36 +71,32 @@ Node.js 18 or later.
...
-      flows and other resources you've defined. The sample file contains a
-      single flow, `jokeFlow`, that simply calls the model provider's API with a
-      simple prompt and returns the result.
-
-      ```js
-      export const jokeFlow = defineFlow(
+      export const menuQAFlow = defineFlow(
         {
-          name: 'jokeFlow',
+          name: 'menuQAFlow',
```
MichaelDoyle, line 91: What do you think about making this more descriptive? QA could mean question/answer. It could mean quality assurance, etc.

MaesterChestnut: good feedback; thoughts on `menuQuestionFlow` ?

docs/get-started.md
```diff
@@ -71,36 +71,32 @@ Node.js 18 or later.
...
+    The `genkit init` command creates a sample source file:
 
-    - `genkit.config.ts`: this is where you configure Genkit for your project,
-      and select and configure the plugins you want to load. The sample config
-      loads a plugin to support the model provider you chose earlier.
+    - `index.ts`: this is your project's entry point, where you configure Genkit for your project,
+      configure the plugins you want to load and export your AI flows and other resources you've defined.
+      The sample file contains a config that loads a plugin to support the model provider you chose earlier.
+      It also contains a single flow, `menuQAFlow`, that simply calls the model provider's API with a
+      simple prompt and returns the result.
```
MichaelDoyle, line 80: Perhaps something like:

> It also contains a single flow, `<new name here>`, that prompts an LLM to determine whether or not a given entree is offered on a restaurant's menu.

genkit-tools/config/nextjs.genkit.ts.template
MaesterChestnut: Pavel originally included this because nextjs server code can only export async functions. We can't export the flow for nextjs to use so we generate a sample async function that they should call as a server action.

docs/get-started.md
MaesterChestnut: Modified; thanks for the feedback!

docs/get-started.md
MaesterChestnut: moved to a more useful flow `menuSuggestionFlow` like was discussed

docs/get-started.md
MaesterChestnut: Updated

PR #63, opened 2024-05-07
Support references in `eval:flow`

ssbushi:


genkit-tools/src/commands/eval-flow.ts
```diff
@@ -178,25 +185,31 @@ export const evalFlowRun = new Command('eval:flow')
...
+): Promise<EvalFlowInput> {
   const parsedData = JSON.parse(
     data ? data : await readFile(filePath!, 'utf8')
   );
   if (Array.isArray(parsedData)) {
     return parsedData as any[];
   }
 
-  return [parsedData];
+  return EvalFlowInputSchema.parse(parsedData);
```
tagboola, line 199: Consider showing a easy to read error message if the parse fails. I'm not sure with the zod error message is, but I assume it may be hard for folks to understand exactly what caused the failure.

genkit-tools/common/src/types/eval.ts
```diff
@@ -21,6 +21,33 @@ import { ListEvalKeysRequest, ListEvalKeysResponse } from './apis';
  * This file defines schema and types that are used by the Eval store.
  */
 
+/**
+ * Structured input for eval:flow
+ */
+export const EvalFlowStructuredInputSchema = z.object({
+  cases: z.array(
```
tagboola, line 28: Cases is pretty generic. Are we happy with this? Or should we think of something more specific?

tagboola (commented):
Looks good! Just a few small comments 🙂

genkit-tools/common/src/types/eval.ts
ssbushi: Renamed to `samples`

genkit-tools/src/commands/eval-flow.ts
ssbushi: Done.

genkit-tools/src/commands/eval-flow.ts
```diff
@@ -178,25 +187,37 @@ export const evalFlowRun = new Command('eval:flow')
...
   if (Array.isArray(parsedData)) {
     return parsedData as any[];
   }
 
-  return [parsedData];
+  try {
+    return EvalFlowInputSchema.parse(parsedData);
+  } catch (e) {
+    throw new Error(
+      `Error parsing the input. Please provide an array of inputs for the flow or a ${EVAL_FLOW_SCHEMA} object`
```
tagboola, line 205: Should we also include the error details in the message so they know why it failed?

tagboola (approved):

genkit-tools/src/commands/eval-flow.ts
ssbushi: Done.

PR #68, opened 2024-05-08
Respect log levels

ssbushi:
Note: The CLI is not affected by this change.

### DIFF
--- a/js/core/src/logging.ts
+++ b/js/core/src/logging.ts
@@ -16,6 +16,8 @@
 
 import { LoggerConfig } from './telemetryTypes.js';
 
+const LOG_LEVELS = ['debug', 'info', 'warn', 'error'];
+
 class Logger {
   logger: {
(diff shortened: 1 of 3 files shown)


ssbushi:
Output on ` genkit flow:run jokeFlowSimple` if logging level is set to error:

```
Starting app at `lib/index.js`...
Warn: Port 3100 not available, using 3101 instead.
Reflection API running on http://localhost:3101
Running '/flow/jokeFlowSimple' (stream=false)...
Flow operation:
{
  "name": "c9f494b8-0cb7-44dc-98e4-564b37d38da1",
  "done": true,
  "result": {
    "response": "What do you call a banana that's been in the sun too long?\n\nA banana split."
  }
}
App process exited with code 0
```

MichaelDoyle (approved):

pavelgj (approved):

PR #71, opened 2024-05-08
[Go] adding metadata to models, etc.

pavelgj:


go/genkit/action.go
```diff
@@ -65,21 +65,22 @@ type Action[I, O, S any] struct {
 // See js/common/src/types.ts
 
 // NewAction creates a new Action with the given name and non-streaming function.
-func NewAction[I, O any](name string, fn func(context.Context, I) (O, error)) *Action[I, O, struct{}] {
-	return NewStreamingAction(name, func(ctx context.Context, in I, cb NoStream) (O, error) {
+func NewAction[I, O any](name string, metadata map[string]any, fn func(context.Context, I) (O, error)) *Action[I, O, struct{}] {
```
ianlancetaylor, line 68: How do you see `metadata` being used?  Who is the producer and who is the consumer?  Typically in Go we would use specific types here (and make the `metadata` parameter type `any`), but if this is truly general, or user-controlled, than a map makes sense.

pavelgj: BTW, for the generator I defined structs: https://github.com/firebase/genkit/pull/71/files#diff-c8c738aff67a5f8e34971b78f75941a99020e41e6ac1c535bef6e83b4c998278

ianlancetaylor: That suggests that the `metadata` parameter to `NewAction` should be type `any`, and the actual action will type assert to the desired value.

jba: Without getting into too much detail, I think the question here comes down to whether this value will ever be used in Go code, or will it just be written to JSON for the benefit of the tooling?

If no Go code needs to access the metadata except to JSON-marshal it, then Ian's way is more type-safe and Go-like. With an `any` parameter, a user can still pass in a `map[string]any`, or they can pass in a struct that is JSON-marshalable (like the GeneratorCapabilities one you've defined). Both a `map[string]any` and a struct with the right field names or struct tags will marshal to the same JSON, but the latter will be easier for Go programmers to write and won't need to be hand-converted to a map in the code.

pavelgj: Go code might need/want to use it. This example: https://github.com/firebase/genkit/blob/9445f92a749f539a8a00b1e326f53e26853e3f82/js/ai/src/generate.ts#L547

Maybe in that case it could be added to the [Generator](https://github.com/firebase/genkit/blob/main/go/ai/generator.go#L27) interface?

ianlancetaylor: Using specific types for generators, as you've already written, is consistent with having the `metadata` parameter of `NewAction` by type `any`.  Generators can type assert the metadata as needed.  Perhaps I misunderstand.

jba: Yes, it should be possible to get the `GeneratorCapabilities` from a `Generator` without going into the action.

pavelgj: I tried changing the metadata type on NewAction to any, but I don't think it's possible... I'll leave that to you. :)

Basically:
https://github.com/firebase/genkit/blob/main/go/genkit/action.go#L62
and
https://github.com/firebase/genkit/blob/main/go/genkit/action.go#L173

PR #75, opened 2024-05-08
Rename toc titles to be more consistent with expected behaviors

schnecle:


docs/models.md
```diff
@@ -1,4 +1,8 @@
-# Models
+# Generating Content
```
kevinthecheung, line 1:
```suggestion
# Generating content
```

kevinthecheung (approved):

PR #80, opened 2024-05-08
Moving the dotprompt package into a plugin

maxl0rd:
- Dotprompt should now be configured as a plugin but has fallback behavior to look in a default prompts folder so no existing code should break.
- Async init used to load the entire prompts folder even if prompts are not explicitly referenced in code.


js/plugins/dotprompt/src/registry.ts
```diff
@@ -0,0 +1,103 @@
...
+import { Dotprompt } from './prompt.js';
+
+export async function lookupPrompt(
+  dir: string,
+  name: string,
+  variant?: string
+): Promise<Dotprompt> {
+  // Expect to find the prompt in the registry
+  const registryPrompt = (await lookupAction(
+    `/prompt/${name}${variant ? `.${variant}` : ''}`
```
pavelgj, line 31: why not register .prompts under `/prompt/dotprompt/${name}${variant ? `.${variant}` : ''}`? This way `lookupAction` will handle plugin initialization for you.

maxl0rd: Ok, I wasn't aware of that, but I see how it resolves some issues. I'll try that.

js/plugins/dotprompt/src/registry.ts
```diff
@@ -0,0 +1,103 @@
...
+  const expectedFileName = `${name}${variant ? `.${variant}` : ''}.prompt`;
+  const promptFolder = resolve(dir);
+  const promptExists = existsSync(join(promptFolder, expectedFileName));
+  if (promptExists) {
+    return loadPrompt(promptFolder, expectedFileName);
+  } else {
+    throw new GenkitError({
+      source: 'dotprompt',
+      status: 'NOT_FOUND',
+      message: `Could not find '${name}.prompt' in the prompts folder.`,
```
mbleigh, line 56: To cover variants don't you want

```suggestion
      message: `Could not find '${expectedFileName}' in the prompts folder.`,
```

js/plugins/dotprompt/src/index.ts
```diff
@@ -14,14 +14,53 @@
...
 import { basename } from 'path';
-
 import { defineDotprompt, Dotprompt } from './prompt.js';
-import { lookupPrompt } from './registry.js';
+import { loadPromptFolder, lookupPrompt } from './registry.js';
 
 export { defineDotprompt, Dotprompt };
 
+export interface DotpromptPluginOptions {
+  promptDir: string;
```
mbleigh, line 27:
```suggestion
  dir: string;
```

It's the Dotprompt plugin, so the directory is naturally for prompts. Also this should have a default value of `./prompts`.

js/plugins/dotprompt/src/index.ts
```diff
@@ -14,14 +14,53 @@
...
+}
+
+let pluginOptions: DotpromptPluginOptions;
```
mbleigh, line 30: This is weird global state stuff. Will this go away once you fix async init?

js/plugins/dotprompt/src/registry.ts
maxl0rd: Done

js/plugins/dotprompt/src/index.ts
maxl0rd: Done

js/plugins/dotprompt/src/index.ts
maxl0rd: Yes, we don't need this any more!

js/plugins/dotprompt/src/registry.ts
maxl0rd: This is working now!

js/plugins/dotprompt/src/index.ts
mbleigh: Can you remove it then?

maxl0rd: Yes, it's removed. I just forgot to push it.

js/plugins/dotprompt/src/registry.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+  }
+}
+
+export async function loadPromptFolder(
+  dir: string = './prompts'
+): Promise<void> {
+  const promptsPath = resolve(dir);
+  return new Promise<void>((resolve, reject) => {
+    readdir(
```
pavelgj, line 74: there's a `readdirSync` variant... that you can use to avoid all this boilerplate with Promises.

pavelgj (approved):

MichaelDoyle:
Any samples that need to get updated based on this new plugin?

js/plugins/dotprompt/src/registry.ts
maxl0rd: I looked at that, but it was slightly harder to resolve the initialization at the right place.

PR #81, opened 2024-05-08
[Go] add vertexai embedder

ianlancetaylor:


go/plugins/vertexai/embed.go
```diff
@@ -0,0 +1,124 @@
...
+
+	endpoint := fmt.Sprintf("projects/%s/locations/%s/publishers/google/models/%s", e.projectID, e.location, e.model)
+
+	instances := make([]*structpb.Value, 0, len(req.Document.Content))
+	for _, part := range req.Document.Content {
+		fields := map[string]any{
+			"content": part.Text(),
+		}
+		if title != "" {
+			fields["title"] = title
```
randall77, line 65: It seems weird that these get set on every part, instead of once somewhere. I guess that's just how the proto wants them?

randall77 (approved):

go/plugins/vertexai/embed.go
ianlancetaylor: I agree that it is weird, but it is what the API examples do.

PR #84, opened 2024-05-08
Fix genkit init to work with new cli directory

MaesterChestnut:
Also fixes Lint error introduced by recent template change


pavelgj:
Hmm... just sent #83 but I think your fix is better

genkit-tools/cli/config/firebase.index.ts.template
```diff
@@ -26,7 +26,8 @@ export const menuSuggestionFlow = onFlow(
     }),
   },
   async (subject) => {
-    const prompt = `Suggest an item for the menu of a ${subject} themed restaurant`;
+    const prompt = `Suggest an item for the menu of a
+     ${subject} themed restaurant`;
```
pavelgj, line 30: is this intentional?

MaesterChestnut: Yes, but open to a different way of solving the max-len of 80 lint error for Firebase functions

sam-gc: I'd recommend this instead:
```ts
    const prompt =
        `Suggest an item for the menu of a ${subject} themed restaurant`;
```

genkit-tools/cli/config/firebase.index.ts.template
```diff
@@ -26,7 +26,8 @@ export const menuSuggestionFlow = onFlow(
...
 
     const llmResponse = await generate({
       model: $GENKIT_MODEL,
```
sam-gc, line 33: This file also needs a newline at the end I believe

sam-gc (changes requested):

tonybaroneee (approved):

sam-gc (approved):

PR #93, opened 2024-05-09
Added harness mode for Next.js.

apascal07:


genkit-tools/common/src/runner/harness.ts
```diff
@@ -0,0 +1,14 @@
+import { glob } from 'glob';
+import * as path from 'path';
+import { logger } from '../utils';
+
+async function main() {
```
mbleigh, line 5: Will this work with loading/transpiling TS files? Does it automatically set `GENKIT_ENV=dev`?

mbleigh (commented):
So the usage for this is:

```
genkit start src/ai/*.ts --mode harness
```

is that right?

apascal07:
> So the usage for this is:
> 
> ```
> genkit start src/ai/*.ts --mode harness
> ```
> 
> is that right?

Not quite.

You add this to your genkit-tools.conf.ts:

```
module.exports = {
  runner: {
    mode: 'harness',
    files: ['./src/*.ts'],
  }
};```

And then run normally with `genkit start`.

genkit-tools/common/src/runner/harness.ts
apascal07: Tested it with .ts, it worked. The envs are set automatically like in the default app loading.

mbleigh (approved):
Love it! Can you adjust `genkit init` for next.js to:

1. Create `genkit.config.ts` inside a `src/ai` directory
2. Automatically configure `genkit start` for harness mode

apascal07:
> Love it! Can you adjust `genkit init` for next.js to:
> 
> 1. Create `genkit.config.ts` inside a `src/ai` directory
> 2. Automatically configure `genkit start` for harness mode

Yep, will do in follow up PR.

PR #94, opened 2024-05-09
[Go] update dotprompt model strings to use provider/model

ianlancetaylor:
Also update coffee-shop to use GOOGLE_API_KEY rather than GEMINI_API_KEY.


jba (approved):

jba:
nit: commit message is slightly wrong: key is GOOGLE_GENAI_API_KEY.

PR #96, opened 2024-05-09
Replace jokeFlow with menuSuggestionFlow in flows.md

tagboola:


docs/flows.md
```diff
@@ -23,16 +28,16 @@ Input and output schemas for flows can be defined using `zod`.
...
   {
-    name: 'myFlow',
-    inputSchema: z.object({ subject: z.string() }),
+    name: 'menuSuggestionFlow',
+    inputSchema: z.object({ cuisine: z.string() }),
     outputSchema: z.string(),
   },
   async (input) => {
-    const output = doSomethingCool(input.subject);
+    const suggestion = makeMenuSuggestion(input.cuisine);
```
MichaelDoyle, line 38: Just the input get wrapped in an `input` object? Just trying to reconcile some examples here have it and others do not.

docs/flows.md
```diff
@@ -98,7 +103,7 @@ If the flow does not implement streaming `streamFlow` will behave identically to
 You can use the CLI to stream flows as well:
 
 ```posix-terminal
-genkit flow:run jokeFlow '"banana"' -s
+genkit flow:run streamer '"banana"' -s
```
MichaelDoyle, line 106: `streamer '"banana"'`? 

This looks off. should this be: `genkit flow:run menuSuggestionFlow '"French"' -s`

docs/flows.md
```diff
@@ -112,19 +117,17 @@ observability).
...
 import { onFlow, noAuth } from '@genkit-ai/firebase/functions';
 
-export const jokeFlow = onFlow(
+export const menuSuggestionFlow = onFlow(
   {
-    name: 'jokeFlow',
-    inputSchema: z.string(),
-    outputSchema: z.string(),
+    name: 'menuSuggestionFlow',
     authPolicy: noAuth(),
```
MichaelDoyle, line 128: Add a comment that this should probably be replaced by a "real" auth policy?

docs/flows.md
```diff
@@ -137,27 +140,42 @@ To deploy flows using Cloud Run and similar services, define your flows using `d
...
 
 ```js
 import { defineFlow, startFlowsServer } from '@genkit-ai/flow';
 
-export const flowA = defineFlow({ name: 'flowA' }, async (subject) => {
-  // ....
-});
+export const flowA = defineFlow(
+  {
+    name: 'menuSuggestionFlowA',
```
MichaelDoyle, line 164: `flowA` / `flowB` works because its generic in nature. I am not sure `menuSuggestionFlowA` / `menuSuggestionFlowB` reads the same way. Should we stick with flowA/flowB? or introduce a second flow name into the mix?

docs/flows.md
```diff
@@ -175,13 +193,20 @@ Sometimes when using 3rd party SDKs that that are not instrumented for observabi
 ```js
 import { defineFlow, run } from '@genkit-ai/flow';
 
-export const myFlow = defineFlow(
-  { name: 'myFlow', inputSchema: z.string(), outputSchema: z.string() },
+export const menuSuggestionFlow = defineFlow(
+  {
+    name: 'menuSuggestionFlow',
```
MichaelDoyle, line 198: Is this the right name for this flow? perhaps `findSimilarCuisinesFlow`?

docs/flows.md
```diff
@@ -175,13 +193,20 @@ Sometimes when using 3rd party SDKs that that are not instrumented for observabi
...
+    inputSchema: z.object({ cuisine: z.string() }),
+    outputSchema: z.string(),
+  },
   async (input) => {
-    const output = await run('step-name', async () => {
-      return await doSomething(input);
+    const cuisines = await run('find-similar-cusines', async () => {
```
MichaelDoyle, line 203: spell check; should be `find-similar-cuisines`

docs/flows.md
```diff
@@ -23,16 +28,16 @@ Input and output schemas for flows can be defined using `zod`.
...
+    name: 'menuSuggestionFlow',
+    inputSchema: z.object({ cuisine: z.string() }),
     outputSchema: z.string(),
   },
   async (input) => {
-    const output = doSomethingCool(input.subject);
+    const suggestion = makeMenuSuggestion(input.cuisine);
 
-    return output;
+    return llmResponse.text();
```
MichaelDoyle, line 40: `suggestion.text()`?

MichaelDoyle: Actually, simply `suggestion`? I don't think this code is calling an LLM. `makeMenuSuggestion` appears to be a user defined function call?

docs/flows.md
```diff
@@ -10,11 +10,16 @@ Firebase Genkit provides CLI and Developer UI tooling for working with flows
...
 import { defineFlow } from '@genkit-ai/flow';
 
-export const myFlow = defineFlow({ name: 'myFlow' }, async (input) => {
-  const output = doSomethingCool(input);
+export const menuSuggestionFlow = defineFlow(
+  {
+    name: 'menuSuggestionFlow',
+  },
+  async (cuisine) => {
+    const suggestion = makeMenuSuggestion(input);
```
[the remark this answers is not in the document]
tagboola: Done.

docs/flows.md
tagboola: Yeah this was a copy-pasta error, it should just be `suggestion`. Done.

docs/flows.md
tagboola: I actually didn't update the streaming example above since we don't have a menu-example for streaming yet.

Here's the
Screenshot 2024-05-09 at 3 36 05 PM
 sample code it's referenceing.

docs/flows.md
tagboola: Decided to just replace this with a real auth policy.

docs/flows.md
tagboola: Went back to `flowA` and `flowB`.

docs/flows.md
tagboola: Done.

docs/flows.md
tagboola: Yeah, but on second thought this seems like overkill. Just gonna do `z.string()` for simplicity.

docs/flows.md
tagboola: The idea is to use the same base `menuSuggestionFlow` and make small modifications to for each scenario. On second thought, I don't think the whole cuisine makes a ton of sense. I updated the exampels to be more aligned with the `menu/basic` sample.

docs/flows.md
MichaelDoyle: Oh - no worries. 2 things then...

- Banana is not a valid input for the streamer flow you cited (it expects a number)
- I am primarily just noticing that the flow (sorry no pun intended) of the docs was to simply enable streaming (-s) for the previous flow (jokeFlow). So we aught to do the same in this snippet? Or if you really wanted, menuSuggestionFlowStreaming or something like that? But I think the important thing being documented here is the CLI capability (`-s`).

MichaelDoyle: Aha, i didn't see the sample code immediately above . :)

Discussed offline - we are waiting for waiting for better sample code to incorporate that is on theme.

docs/flows.md
```diff
@@ -10,11 +10,16 @@ Firebase Genkit provides CLI and Developer UI tooling for working with flows
...
+    name: 'menuSuggestionFlow',
+  },
+  async (restaurantTheme) => {
+    const suggestion = makeMenuItemSuggestion(restaurantTheme);
 
-  return output;
-});
+    return suggestion;
+  }
+);
```
MaesterChestnut, line 22: I'm not sure this example _needs_ to change from the original

docs/flows.md
```diff
@@ -23,16 +28,16 @@ Input and output schemas for flows can be defined using `zod`.
 import { defineFlow } from '@genkit-ai/flow';
 import * as z from 'zod';
 
-export const myFlow = defineFlow(
+export const menuSuggestionFlow = defineFlow(
```
MaesterChestnut, line 31: I'm not sure this example _needs_ to change from the original

MaesterChestnut (approved):

docs/flows.md
```diff
@@ -175,13 +179,21 @@ Sometimes when using 3rd party SDKs that that are not instrumented for observabi
...
-    return output;
+
+    const suggestions = makeMenuItemSuggestions(themes);
+
+    return suggestions;
   }
 );
 ```
+
+See https://github.com/firebase/genkit/blob/main/js/samples/menu-example to view the sample app that uses the `menuSuggestionFlow`.
```
kevinthecheung, line 199:
```suggestion
See the [sample app](https://github.com/firebase/genkit/blob/main/js/samples/menu-example) that uses the `menuSuggestionFlow`.
```

kevinthecheung (approved):

docs/flows.md
tagboola: Do you think I should revert it back? I figured we would just use the menu suggestion theme throughout.

docs/flows.md
tagboola: Do you think I should revert it back? I figured we would just use the menu suggestion theme throughout.

docs/flows.md
tagboola: Took a pass at changing this so it's menu suggestion themed.

docs/flows.md
MaesterChestnut: My hesitancy comes from this `myFlow` example not necessarily being it's own theme; on it's own it's pretty generic. @kevinthecheung do you have an opinion here?

MaesterChestnut: on further thought; @kevinthecheung already approved this change; let's ship it

PR #98, opened 2024-05-09
Added defaultModel to config that allows setting a global default model and config

pavelgj:
```ts
configureGenkit({
  plugins: [vertexAI()],
  defaultModel: {
    name: geminiPro,
    config: {
      temperature: 0.6,
      stopSequences: ['sorry'],
    },
  },
});
```

can then call `generate` (including dotprompt) without having to specify a model.

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -490,10 +490,10 @@ async function resolveModel(options: GenerateOptions): Promise<ModelAction> {
           : genkitConfig.options.defaultModel.name.name;
       if (
         (!options.config || Object.keys(options.config).length === 0) &&
-        genkitConfig?.options.defaultModel.config
+        genkitConfig.options.defaultModel.config
       ) {
         // use configured global config
-        options.config = genkitConfig?.options.defaultModel.config;
+        options.config = genkitConfig.options.defaultModel.config;
       }
     } else {
       throw new Error('Unable to resolve model.');
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -483,7 +483,7 @@ const isValidCandidate = (
 async function resolveModel(options: GenerateOptions): Promise<ModelAction> {
   let model = options.model;
   if (!model) {
-    if (genkitConfig?.options.defaultModel) {
+    if (genkitConfig?.options?.defaultModel) {
       model =
         typeof genkitConfig.options.defaultModel.name === 'string'
           ? genkitConfig.options.defaultModel.name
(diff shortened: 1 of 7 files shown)


js/ai/src/generate.ts
```diff
@@ -479,9 +479,39 @@ const isValidCandidate = (
...
+}
+
+export function setGlobalDefaultModel<
+  CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
+>(model: ModelArgument<CustomOptions>, config?: z.infer<CustomOptions>) {
+  global[DEFAULT_MODEL_GLOBAL_KEY] = {
+    model,
+    config,
+  } as GlobalModelRef;
+}
```
mbleigh, line 496: I'm curious - why not make this part of `configureGenkit`? Is it because models aren't a concept in core?

I ask because it could be kinda cool if `genkit init` set default models (e.g. Gemini Pro for Vertex/Google AI)

pavelgj: Yeah. I can't strongly type it in the config (circular dep with AI package). We can still make genkit init set it this way.

pavelgj: Can take a string model ref in the config....

mbleigh: I think I'd probably prefer string model ref in config. Honestly I think we should consider merging `ai` and `core` packages in the future.

pavelgj: Sure. Do you want config as well?

mbleigh: Yes please, how about:

```ts
configureGenkit({
  defaults: {
    model: {ref: string | {name: string}, config: any}
  }
})
```

mbleigh: How about:

```ts
interface {
  defaultModel?: {
    name: string | {name: string};
    config: Record<string,any>;
  }
}
```

pavelgj: Sg

js/ai/src/generate.ts
```diff
@@ -479,34 +480,20 @@ const isValidCandidate = (
...
+          ? genkitConfig?.options.defaultModel.name
+          : genkitConfig?.options.defaultModel.name.name;
       if (
         (!options.config || Object.keys(options.config).length === 0) &&
-        globalModel.config
+        genkitConfig?.options.defaultModel.config
       ) {
         // use configured global config
-        options.config = globalModel.config;
+        options.config = genkitConfig?.options.defaultModel.config;
```
mbleigh, line 496: Just double-checking - this gets overridden if specified by the user, right?

mbleigh (approved):

js/ai/src/generate.ts
pavelgj: yes, the condition above that this is only applied if `options.config` is not provided (or is `{}`).

js/ai/src/generate.ts

PR #107, opened 2024-05-10
Menu/generation example

MaesterChestnut:
Adding in a few flows for developers to try streaming, history, and image input quickly

image


js/samples/docs-menu-generation/README.md
```diff
@@ -0,0 +1,30 @@
+## Build it
+
+```
+pnpm build
+```
+
+or if you need to, build everything:
```
MichaelDoyle, line 7: Maybe we can set this up as a prerequisite step (build the entire repo first) that we note can be skipped if already done. otherwise, how does one answer the question "build if you need to?"

js/samples/docs-menu-generation/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+import { geminiPro, geminiProVision, googleAI } from '@genkit-ai/googleai';
+import * as z from 'zod';
+
+configureGenkit({
+  plugins: [googleAI()],
+  logLevel: 'debug',
+  enableTracingAndMetrics: true,
+});
+
+export const menuStreamingSuggestionFlow = defineFlow(
```
MichaelDoyle, line 31: kind of a nit, but its weird to split the name like this: `menu<Streaming>SuggestionFlow`. alternatives:

- `streamingMenuSuggestionFlow`.
- `menuSuggestionFlowStreaming`.
- `menuSuggestionStreamingFlow`.

js/samples/docs-menu-generation/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  {
+    name: 'menuStreamingSuggestionFlow',
```
MichaelDoyle, line 33: Same nit as above.

js/samples/docs-menu-generation/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+    inputSchema: z.string(),
+    outputSchema: z.void(),
+  },
+  async (subject) => {
+    const { response, stream } = await generateStream({
+      prompt: `Suggest many items for the menu of a ${subject} themed restaurant`,
```
MichaelDoyle, line 39: maybe specify exactly how many, or provide a range?

js/samples/docs-menu-generation/src/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export const menuHistoryFlow = defineFlow(
+  {
+    name: 'menuHistoryFlow',
+    inputSchema: z.string(),
+    outputSchema: z.string(),
+  },
+  async (subject) => {
+    // lets do few-shot examples: generate a few different generations, keep adding the history
+    // before generating an item for the menu of a themed restaurant
```
MichaelDoyle, line 65: Since this is actually going to call the model, it's probably not a good example of a "few-shot" prompt, which would typically have hardcoded examples.

MichaelDoyle: This looks more like a pre-scripted chat exchange, but will be non-deterministic. We should probably talk through more what you want to be the outcome here.

js/samples/docs-menu-generation/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+      prompt: `Suggest many items for the menu of a ${subject} themed restaurant`,
+      model: geminiPro,
+      config: {
+        temperature: 1,
+      },
+    });
+
+    for await (let chunk of stream()) {
+      for (let content of chunk.content) {
+        console.log(content.text);
```
tagboola, line 50: Consider passing the contents of the chunk to the `streamingCallback` if it's defined. This will let developers know that they can provide a way to process the stream.

js/samples/docs-menu-generation/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+    }
+
+    console.log((await response()).text());
+  }
+);
+
+let historyMap: MessageData[] = sampleMenuHistory;
```
tagboola, line 58: nit: Drop the `Map` since it's an array

tagboola (approved):

PR #117, opened 2024-05-10
Write tools config in `genkit init`.

apascal07:


genkit-tools/cli/src/commands/init.ts
```diff
@@ -325,6 +329,28 @@ async function updatePackageJson() {
...
+/**
+ * Generates an appropriate tools config for the given platform.
+ * @param platform platform
+ */
+function generateToolsConfig(platform: Platform) {
+  if (platform === 'nextjs') {
+    const templatePath = path.join(__dirname, nextjsToolsConfigTemplatePath);
+    let template = fs.readFileSync(templatePath, 'utf8');
+    if (fs.existsSync('src/app')) {
+      template = template.replace('$GENKIT_HARNESS_FILES', `'./src/app/*.ts'`);
```
pavelgj, line 341: I asked about this in chat whether we want/need to limit this more to include only genkit related files... but it seems to work like this, so perhaps not necessary.

pavelgj (approved):

PR #123, opened 2024-05-10
Generate API reference with typedoc

kevinthecheung:


kevinthecheung:
I think this is a good foundation for us to start incrementally improving the doc strings. PTAL

PR #125, opened 2024-05-11
Adds defineSimpleRetriever and docs.

mbleigh:
Building out my demo I found it kind of annoying to write a retriever because converting things into Documents is tedious. This provides a "simple retriever" that makes it very easy to transform arbitrary data into documents. For example, here is the Data Connect retriever from my demo implemented as a simple retriever:

```ts
export const destinationRetriever = defineSimpleRetriever(
  {
    name: 'destinationRetriever',
    configSchema: z.object({ limit: z.number() }),
    content: 'knownFor',
  },
  async (query, { limit }) => {
    const result = await searchDestinations(dc, {
      query: query.text(),
      limit: limit,
    });
    return result.data.destinations;
  },
);
```

Pretty nice, right?


docs/rag.md
```diff
@@ -303,10 +303,39 @@ export const ragFlow = defineFlow(
...
+
+Simple retrievers let you easily convert existing code into retrievers:
+
+```javascript
+import {
+  defineSimpleRetriever,
+  retrieve
+} from "@genkit-ai/ai/retriever";
+import { searchEmails } from "./db";
+import { z } from "zod";
```
pavelgj, line 321:
```suggestion
} from '@genkit-ai/ai/retriever';
import { searchEmails } from './db';
import { z } from 'zod';
```

js/ai/src/retriever.ts
```diff
@@ -294,3 +294,106 @@ export function indexerRef<
...
+
+/**
+ * defineSimpleRetriever makes it easy to map existing data into documents that
+ * can be used for prompt augmentation.
+ *
+ * @param options Configuration options for the retriever.
+ * @param handler A function that queries a datastore and returns items from which to extract documents.
+ * @returns A Genkit retriever.
+ */
+export function defineSimpleRetriever<
```
pavelgj, line 375: This is neat and I think it should exist, just wondering if it should exist here... the more I think about it the more I think it's probably OK that it's exported via `@genkit-ai/ai/retriever`... not sure, just saying it out loud...  feel free to ignore.

pavelgj (approved):

js/ai/src/retriever.ts
mbleigh: I think it's the right place for it, but fair to ask.

PR #132, opened 2024-05-12
Moves firestoreTraceStore to firebase plugin, cleans up unused deps and outdated examples.

mbleigh:


pavelgj (approved):
Thanks for doing this! It fell off my TODO somewhere along the way.

PR #137, opened 2024-05-12
Moved menu and coffee shop samples into top level samples folder

pavelgj:
and made them depend on npm versions of genkit packages and easily runnable with npm (pnpm not required)


MichaelDoyle (approved):
Looks good to me, assuming we get some consensus on this. That way we can update docs, readme's, etc as a follow on.

samples/js-coffee-shop/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "dependencies": {
+    "@genkit-ai/ai": "~0.5.0-rc.11",
+    "@genkit-ai/core": "~0.5.0-rc.11",
+    "@genkit-ai/dotprompt": "~0.5.0-rc.11",
+    "@genkit-ai/flow": "~0.5.0-rc.11",
+    "genkitx-chromadb": "~0.5.0-rc.11",
+    "@genkit-ai/dev-local-vectorstore": "~0.5.0-rc.11",
+    "@genkit-ai/firebase": "~0.5.0-rc.11",
+    "@genkit-ai/googleai": "~0.5.0-rc.11",
+    "genkitx-ollama": "~0.5.0-rc.11",
```
i2amsam, line 26: This makes sense to me, thinking through the process, if someone is just testing they can `genkit start` here and build and run using the published versions.  If they're doing local development the best process would be to bump their local `package.json` file to the next patch/RC version and the validate that it's getting picked up in these samples, but they shouldn't have to adjust the imports since they're getting ~'d in, is that right?

PR #140, opened 2024-05-13
Added (experimental) non-interactive mode option for `genkit init`

pavelgj:


genkit-tools/cli/src/commands/init.ts
```diff
@@ -166,6 +168,10 @@ export const init = new Command('init')
     '-m, --model <model>',
     'Model provider (googleai, vertexai, ollama, or none)'
   )
+  .option(
+    '--non-interactive',
+    'Run init in non-interactive mode (experimental)'
+  )
```
MichaelDoyle, line 174: What do you think about doing `--interactive` with a default setting of "true"?

Might help with some of the double-negatives later on.

`!nonInteractive`

and

`!process.argv.includes('--non-interactive')`

MichaelDoyle (approved):
Looks good. 

Left a small idea for you in case it helps wrt booleans.

genkit-tools/cli/src/commands/init.ts
MichaelDoyle: ok 😄  I could see why (`--non-interactive` vs `--interactive=false`) it's just a readability thing in the code itself to consider from my perspective.

alternatively, could try and come up w/ something "positive" like `--skip-prompts`, but let's not spin our wheels on it.

Issue #130, opened 2024-05-12
[JS] @genkit-ai/evaluator package is not ESM compatible


pavelgj:
**Describe the bug**

```
node:internal/url:1467
    throw new ERR_INVALID_ARG_TYPE('path', ['string', 'URL'], path);
          ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of URL. Received undefined
    at fileURLToPath (node:internal/url:1467:11)
    at getFilename (file:///Users/pavel/esmtest/node_modules/@genkit-ai/evaluator/lib/chunk-TXBHYK5M.mjs:24:25)
    at getDirname (file:///Users/pavel/esmtest/node_modules/@genkit-ai/evaluator/lib/chunk-TXBHYK5M.mjs:25:37)
    at file:///Users/pavel/esmtest/node_modules/@genkit-ai/evaluator/lib/chunk-TXBHYK5M.mjs:26:33
    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:120:12) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

**To Reproduce**
use rag sample, set `"type": "module"`, in `js/samples/rag/package.json` and delete `js/samples/rag/genkit-tools.conf.js`

ssbushi:
Fixed in #158

PR #158, opened 2024-05-14
Make Evaluators plugin work with ESM

ssbushi:
- Fixes Issue #130 
- Also fix minor bug in maliciousness.


pavelgj:
Can you please double check that it still works properly with CJS?

pavelgj (approved):

ssbushi:
Yup, I tested it using the DevUI. Works!

PR #175, opened 2024-05-15
fix sample code of firebase

laiso:
menuSuggestionFlow() is missing


MichaelDoyle (approved):
Great catch, thank you!

PR #177, opened 2024-05-15
Docs/firebase fixes

MaesterChestnut:
Modify prettier rules for docs/firebase since Firebase Lint requires strings use doublequotes and no spaces next to curly braces

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,4 +4,7 @@ public/
 .prettierrc.yaml
 pnpm-lock.yaml
 docs/_guides.yaml
+docs/firebase.md
+docs/auth.md
+docs/plugins/firebase.md
 genkit-tools/genkit-schema.json
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,7 +4,4 @@ public/
 .prettierrc.yaml
 pnpm-lock.yaml
 docs/_guides.yaml
-docs/firebase.md
-docs/auth.md
-docs/plugins/firebase.md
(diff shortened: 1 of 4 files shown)


kevinthecheung (approved):

.prettierignore
```diff
@@ -4,4 +4,7 @@ public/
 .prettierrc.yaml
 pnpm-lock.yaml
 docs/_guides.yaml
+docs/firebase.md
+docs/auth.md
+docs/plugins/firebase.md
```
tonybaroneee, line 9: Remove these, you can mark a code block to be ignored from prettier in the markdown file so that you still can get markdown formatting outside of the code block. See other docs for an example of that (search "prettier" in our docs for the proper ignore comment).

MaesterChestnut: Thanks for the tip; will do!

MichaelDoyle:
This looks good. The question I have going forward is, how to ensure these snippets continue to meet the standard, wrt the firebase linter.

MichaelDoyle (approved):

MaesterChestnut:
Good call @MichaelDoyle; I added comments about why the prettier-ignore exists which should help the dev who comes next. I think a more sustainable solution will be to move the snippet code into a different repository that has Firebase ESLint running on it, similar to what [Firestore has](https://firebase.google.com/docs/firestore/manage-data/transactions). @kevinthecheung not sure if we plan on extracting code snippets for Genkit in the future, but I think we have a good case to push us that way

PR #193, opened 2024-05-18
Added option to pass custom body parser options to modify limit

tzahush:
Added option to pass custom body parser options to modify limit


pavelgj:
Thanks for the PR. Sorry for the very late response! We actually want to switch to using `express.json()` (https://expressjs.com/en/api.html#express.json) middleware (which I believe it based on body-parser). Do you mind updating the PR to use `express.json()`?

pavelgj:
I merged this as part of https://github.com/firebase/genkit/pull/570
Thanks!

PR #196, opened 2024-05-19
[JS] feat: add path prefix option to startFlowsServer

debkanchan:
Allows you to specify an optional path prefix for flows server.

if the flow name is `suggestMenuItems` and path prefix is `flows` then the server will be available on `/flows/suggestMenuItems`

Applies to all flows

### DIFF
--- a/js/flow/src/flow.ts
+++ b/js/flow/src/flow.ts
@@ -847,7 +847,7 @@ export function startFlowsServer(params?: {
   const flows = params?.flows || createdFlows();
   logger.info(`Starting flows server on port ${port}`);
   flows.forEach((f) => {
-    const flowPath = `${pathPrefix}/${f.name}`;
+    const flowPath = `/${pathPrefix}${f.name}`;
     logger.info(` - ${flowPath}`);
     // Add middlware
     f.middleware?.forEach((m) => {
--- a/js/flow/src/flow.ts
+++ b/js/flow/src/flow.ts
@@ -378,7 +378,7 @@ export class Flow<
 
     throw new Error(
       'Unexpected envelope message case, must set one of: ' +
-      'start, schedule, runScheduled, resume, retry, state'
+        'start, schedule, runScheduled, resume, retry, state'
     );
   }
 
@@ -710,8 +710,8 @@ export function streamFlow<
     start(controller) {
       chunkStreamController = controller;
     },
-    pull() { },
-    cancel() { },
+    pull() {},
+    cancel() {},
   });
 
   const operationPromise = flow
@@ -839,7 +839,7 @@ export function startFlowsServer(params?: {
 }) {
   const port =
     params?.port || (process.env.PORT ? parseInt(process.env.PORT) : 0) || 3400;
-  const pathPrefix = params?.pathPrefix ? `/${params?.pathPrefix}` : "";
+  const pathPrefix = params?.pathPrefix ? `/${params?.pathPrefix}` : '';
   const app = express();
   app.use(bodyParser.json());
   app.use(cors(params?.cors));
--- a/js/flow/src/flow.ts
+++ b/js/flow/src/flow.ts
@@ -378,7 +378,7 @@ export class Flow<
 
     throw new Error(
       'Unexpected envelope message case, must set one of: ' +
-        'start, schedule, runScheduled, resume, retry, state'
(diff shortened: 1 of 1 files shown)


js/flow/src/flow.ts
```diff
@@ -835,22 +835,24 @@ export function startFlowsServer(params?: {
   flows?: Flow<any, any, any>[];
   port?: number;
   cors?: CorsOptions;
+  pathPrefix?: string;
 }) {
   const port =
     params?.port || (process.env.PORT ? parseInt(process.env.PORT) : 0) || 3400;
+  const pathPrefix = params?.pathPrefix ? `/${params?.pathPrefix}` : '';
```
pavelgj, line 842:
```suggestion
  const pathPrefix = params?.pathPrefix ?? '';
```

js/flow/src/flow.ts
```diff
@@ -835,22 +835,24 @@ export function startFlowsServer(params?: {
...
   const app = express();
   app.use(bodyParser.json());
   app.use(cors(params?.cors));
 
   const flows = params?.flows || createdFlows();
   logger.info(`Starting flows server on port ${port}`);
   flows.forEach((f) => {
-    logger.info(` - /${f.name}`);
+    logger.info(` - ${pathPrefix}/${f.name}`);
```
pavelgj, line 850:
```suggestion
    const flowPath = `${pathPrefix}${f.name}`;
    logger.info(` - ${flowPath}`);
```

js/flow/src/flow.ts
```diff
@@ -835,22 +835,24 @@ export function startFlowsServer(params?: {
...
     // Add middlware
     f.middleware?.forEach((m) => {
-      app.post(`/${f.name}`, m);
+      app.post(`${pathPrefix}/${f.name}`, m);
```
pavelgj, line 853:
```suggestion
      app.post(flowPath, m);
```

js/flow/src/flow.ts
```diff
@@ -835,22 +835,24 @@ export function startFlowsServer(params?: {
...
     });
-    app.post(`/${f.name}`, f.expressHandler);
+    app.post(`${pathPrefix}/${f.name}`, f.expressHandler);
```
pavelgj, line 855:
```suggestion
    app.post(flowPath, f.expressHandler);
```

pavelgj (changes requested):

js/flow/src/flow.ts
debkanchan: @pavelgj I think you meant to say ```const flowPath = `${pathPrefix}/${f.name}`;```? without the `/` it won't work properly

pavelgj: The suggestion was that `params.pathPrefix` defaults to `/`, so `${pathPrefix}${f.name}` would be correct. The idea is that prefix doesn't necessarily need to have a `/` as separator. For example, if I want to a path like `/customPrefix_flowName` I'd just need to specify `/customPrefix_` as prefix. 

I'm thinking if it's required that the path starts with `/` then we can validate it....

debkanchan: This feels very counter intuitive. I feel very rarely people would like to add a proper "prefix" to their paths rather than a parent route. having `/flows/joke` will be way more common than `/flows_joke`.

One thing we can do maybe is, have a proper "prefix" like what you're suggesting, but also through #198 we can enable devs to add parent routes like what i'm envisioning

debkanchan: So there could be 4 options
- flow name only ex: `/joke`
- prefix + flow name ex: `/prefixjoke` (path has to start with `/` for express to work iirc)
- parent route + flow name ex: `/flows/joke` (through #198 )
- parent + prefix + flow name ex: `/flows/prefixjoke` (through #198 )

pavelgj: Yes, and my thinking was that this is nicely covered by:

```js
const pathPrefix = params?.pathPrefix ?? '/';
if (pathPrefix.chatAt(0) !== '/') {
  pathPrefix = '/' + pathPrefix;
  // or throw new Error('pathPrefix must start with /')
}
const flowPath = `${pathPrefix}${f.name}`;
```

debkanchan: @pavelgj made leading / default. It will now be `/flowName` or `/prefixflowName`

pavelgj (approved):

PR #198, opened 2024-05-19
[JS] feat: standalone express router

debkanchan:
create a new function called createFlowsServer that creates a flow server but does not automatically start it. Instead it returns an express server so that it could be used alongside non flow routes


debkanchan:
This pr is based on top of [this pr](https://github.com/firebase/genkit/pull/196). Merge that pr before this pr to avoid unwanted changes.

pavelgj:
I'm going to close this PR. #196 was merged, but we discussed this in chat and I think this can easily be achieved by defining a custom express server.

PR #202, opened 2024-05-20
[Go] POC: redo dotprompt API

jba:
Attempt to make the dotprompt API a little more user-friendly.

The main change is to replace the exported Frontmatter type
with a type (called Config) that is easier to write a literal for.

Miscellaneous other changes:

- Unexport the Prompt.Hash field.

- Add `New` to create a prompt without registering it,
  and use `Define` to mean create + register, as it does elsewhere.

- Remove the generator as an argument; I believe it would only be
  used for tests.


ianlancetaylor (approved):
Thanks!

PR #229, opened 2024-05-23
[Go] added action type and subtype span metadata to actions

pavelgj:
This is to make sure that spans are properly rendered in the Dev UI

image


jba (approved):

jba:
Thanks.

PR #231, opened 2024-05-23
Improving dotprompt plugin

maxl0rd:
Addresses friction with defineDotprompt and ESM modules.

- Use of defineDotprompt() no longer requires installing the plugin
- The plugin still needs to be installed to load prompt files
- Validated compatibility with projects built as ESM modules

### DIFF
--- a/js/testapps/menu/package.json
+++ b/js/testapps/menu/package.json
@@ -2,7 +2,7 @@
   "name": "menu",
   "version": "1.0.0",
   "description": "A Genkit test app for understanding a restaurant menu",
-  "main": "lib/index.js",
+  "main": "./lib/src/index.js",
   "scripts": {
     "start": "node lib/index.js",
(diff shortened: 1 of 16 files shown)


js/testapps/menu/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "menu",
   "version": "1.0.0",
   "description": "A Genkit test app for understanding a restaurant menu",
-  "main": "lib/index.js",
+  "main": "./lib/src/index.js",
```
pavelgj, line 5: is this right?

pavelgj (approved):

js/testapps/menu/package.json
maxl0rd: Possibly because we are importing some json modules, it creates lib/data and lib/src.

PR #233, opened 2024-05-23
[Go] production server

jba:
Implement NewFlowServeMux, which returns an http.ServeMux
that has a route for each defined flow.

Provide convenience functions for using that ServeMux as a server,
and also for selecting which server to use based on the current
development environment.


go/genkit/servers.go
```diff
@@ -280,6 +229,189 @@ type listFlowStatesResult struct {
 	ContinuationToken string       `json:"continuationToken"`
 }
 
+// Copyright 2024 Google LLC
```
ianlancetaylor, line 232: What is this copyright message doing down here?

jba:
@pavelgj WDYT of the API here?

 - `StartServer` to start a server depending on `GENKIT_ENV`
 - `StartDevServer` and `StartProdServer` if you know what you want
 - `NewFlowServeMux` to get the prod router piece that has the flow routes, so you can make it part of a larger server.

I know the names don't quite align; do you have suggestions?
(BTW, for the js, I think "flowServer" is better than "flowsServer" because it doesn't have the double 's'.)

ianlancetaylor (commented):
This seems to do much more than rename some things.  Not clear on what is happening here.

jba:
I tried to do a poor person's CL stack by making two separate commits. It took the message from the first instead of the second. I edited the top comment.

I meant to suggest reviewing the commits separately.

go/genkit/servers.go
jba: Fixed.

go/genkit/servers.go
```diff
@@ -39,48 +43,39 @@ import (
...
+// StartServer always returns a non-nil error, the one returned by http.ListenAndServe.
+func StartServer(addr string) error {
+	switch e := currentEnvironment(); e {
+	case EnvironmentDev:
+		return StartDevServer(addr)
+	case EnvironmentProd:
+		return StartProdServer(addr)
+	default:
+		return fmt.Errorf("unknown environment value: %q", e)
+	}
```
pavelgj, line 57: the idea is that `StartProdServer` should always start (regardless of `currentEnvironment()`), even in dev mode. Dev (reflection API) server should only run in dev mode.

So the logic here should be:

```
if currentEnvironment() == EnvironmentDev {
  StartDevServer(addr)
}
StartProdServer(addr)
```

jba: Ah, I see. I was going about this wrong.
I removed StartServer and changed StartProdServer to StartFlowServer.
When you call StartFlowServer, you'll get a dev server too.

ianlancetaylor (approved):

PR #234, opened 2024-05-23
Recursively look for `.prompt` files in specified dir

tonybaroneee:
Recursively looks for any `.prompt` file within the configured prompt directory, instead of just in the immediate folder.

image

image

Fixes https://github.com/firebase/genkit/issues/210.

### DIFF
--- a/js/plugins/dotprompt/src/registry.ts
+++ b/js/plugins/dotprompt/src/registry.ts
@@ -80,7 +80,7 @@ export async function loadPromptFolder(
         promptsPath,
         {
           withFileTypes: true,
-          recursive: false,
+          recursive: true,
         },
         (err, dirEnts) => {
           if (err) {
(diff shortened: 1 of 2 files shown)


js/plugins/dotprompt/src/registry.ts
```diff
@@ -80,7 +80,7 @@ export async function loadPromptFolder(
         promptsPath,
         {
           withFileTypes: true,
-          recursive: false,
+          recursive: true,
```
pavelgj, line 83: how would this handle:
```
prompts
  |- foo.prompt
  |- subfolder
    |- foo.prompt
```

tonybaroneee: Good call, let me update the namespace to include the sub-directory path (so this would be registered as `dotprompt/subfolder/foo`)

tonybaroneee: Done.

js/plugins/dotprompt/src/registry.ts
```diff
@@ -80,15 +80,23 @@ export async function loadPromptFolder(
...
               if (dirEnt.isFile() && dirEnt.name.endsWith('.prompt')) {
-                loadPrompt(dirEnt.path, dirEnt.name);
+                // If this prompt is in a subdirectory, we need to include that
+                // in the namespace to prevent naming conflicts.
+                let subDir = '';
+                if (promptsPath !== dirEnt.path) {
+                  subDir = dirEnt.path
+                    .replace(`${promptsPath}/`, '')
+                    .replace(/\//g, '-');
+                }
```
pavelgj, line 98: would this make sense to be inside `loadPrompt`? I mean, you're passing it as `subDir`, but it's not more of a prefix... alternatively maybe just rename `subDir` to `prefix`?

tonybaroneee: We would need to pass in `promptsPath` for the extra logic, so it's a sixer to me. Good call on the naming, updated to `prefix`.

pavelgj (approved):

PR #238, opened 2024-05-23
Update plugin docs to include explicit imports

i2amsam:


kevinthecheung:
I don't oppose this but we got internal feedback that the way the snippets looked with import statements implied that the snippets could be copy pasted as complete working examples which is often not the case.

In other languages, I would just rely on IDE auto-completion, but it seems to rarely work for typescript/vs code, at least for me.

i2amsam:
> I don't oppose this but we got internal feedback that the way the snippets looked with import statements implied that the snippets could be copy pasted as complete working examples which is often not the case.
> 
> In other languages, I would just rely on IDE auto-completion, but it seems to rarely work for typescript/vs code, at least for me.



> I don't oppose this but we got internal feedback that the way the snippets looked with import statements implied that the snippets could be copy pasted as complete working examples which is often not the case.
> 
> In other languages, I would just rely on IDE auto-completion, but it seems to rarely work for typescript/vs code, at least for me.

Hmmmm, that's a fair point.  I just did some testing with a sample project and was able to get too autocompletion for these pretty smoothly.  Let me dig in more

pavelgj (approved):

i2amsam:
> > I don't oppose this but we got internal feedback that the way the snippets looked with import statements implied that the snippets could be copy pasted as complete working examples which is often not the case.
> > In other languages, I would just rely on IDE auto-completion, but it seems to rarely work for typescript/vs code, at least for me.
> 
> > I don't oppose this but we got internal feedback that the way the snippets looked with import statements implied that the snippets could be copy pasted as complete working examples which is often not the case.
> > In other languages, I would just rely on IDE auto-completion, but it seems to rarely work for typescript/vs code, at least for me.
> 
> Hmmmm, that's a fair point. I just did some testing with a sample project and was able to get too autocompletion for these pretty smoothly. Let me dig in more


In my testing I was able to get "Quick fixes" in VSCode for these imports, so closing this for now

PR #241, opened 2024-05-23
[Go] move dotprompt to plugins

ianlancetaylor:


ianlancetaylor:
@jba We discussed moving dotprompt parallel to genkit, but I see from other genkit discussion that the reason that TypeScript moved their version to plugins was in case other prompt libraries were added.  So this moves dotprompt to plugins on the Go side.  I can change to be parallel to genkit if you still think that is better.

jba:
Fine with me.

jba (approved):

jba:
Submitting because my current work depends on it.

PR #243, opened 2024-05-24
[Go] fix go generate

jba:
Some new types were added, and a new aspect of JSON Schema
that the code generator didn't know about.

The new aspect is that a "$ref" can refer to a property
inside a schema. The code generator still doesn't understand
these, but the config file handles them by ignoring them or
replacing them with another type.

### DIFF
--- a/go/ai/gen.go
+++ b/go/ai/gen.go
@@ -28,6 +28,20 @@ type Candidate struct {
 	Usage         *GenerationUsage `json:"usage,omitempty"`
 }
 
+type CandidateError struct {
+	Code    CandidateErrorCode `json:"code,omitempty"`
+	Index   float64            `json:"index,omitempty"`
+	Message string             `json:"message,omitempty"`
+}
+
+type CandidateErrorCode string
+
+const (
+	CandidateErrorCodeBlocked CandidateErrorCode = "blocked"
+	CandidateErrorCodeOther   CandidateErrorCode = "other"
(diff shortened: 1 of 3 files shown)


go/ai/gen.go
```diff
@@ -39,10 +53,16 @@ const (
 	FinishReasonUnknown FinishReason = "unknown"
 )
 
+type DataPart struct {
```
ianlancetaylor, line 56: Should this be `dataPart`?  The other `Part` types (`textPart`, `mediaPart`) are not exported.

go/genkit/schemas.config
```diff
@@ -144,16 +148,27 @@ ToolResponsePartToolResponse	pkg ai
 Part							pkg ai
 TextPart						pkg ai
 TextPart						name textPart
+TextPart.data					omit
```
ianlancetaylor, line 151: Any idea what these new `Part` fields are used for?  Should we be using them in document.go somehow?

ianlancetaylor (approved):

go/ai/gen.go
jba: See below.

go/genkit/schemas.config
jba: I've asked Keith to look into this. I just got things working so I could continue my API refactoring.

randall77: I think the js side just split the part that we call "blob" into two, "media" and "data". I will make that same change in our code.

PR #244, opened 2024-05-24
[Go] move most of genkit package into core

jba:
This PR greatly simplifies the genkit package, limiting it to
the symbols that Genkit app developers, as exemplified by
the programs in the "samples" directory, would need.

To accomplish this, it moves most of the code to a new package, named
core. The core package, rather than the genkit package, is now imported
by the ai package and plugins. End-user applications should not normally
require it.

This overlapped with #229, so unfortunately those (minor) changes
are incorporated here as well, in a slightly different form.


jba:
Here is the output of `go doc -all` on the genkit package:

package genkit // import "github.com/firebase/genkit/go/genkit"

Package genkit provides Genkit functionality for application developers.

FUNCTIONS

func DefineFlow[I, O, S any](name string, fn core.Func[I, O, S]) *core.Flow[I, O, S]
    DefineFlow creates a Flow that runs fn, and registers it as an action.

func NewFlowServeMux() *http.ServeMux
    NewFlowServeMux constructs a net/http.ServeMux where each defined flow is
    a route. All routes take a single query parameter, "stream", which if true
    will stream the flow's results back to the client. (Not all flows support
    streaming, however.)

    To use the returned ServeMux as part of a server with other routes, either
    add routes to it, or install it as part of another ServeMux, like so:

        mainMux := http.NewServeMux()
        mainMux.Handle("POST /flow/", http.StripPrefix("/flow/", NewFlowServeMux()))

func Run[T any](ctx context.Context, name string, f func() (T, error)) (T, error)
    Run runs the function f in the context of the current flow. It returns an
    error if no flow is active.

    Each call to Run results in a new step in the flow. A step has its own span
    in the trace, and its result is cached so that if the flow is restarted,
    f will not be called a second time.

func RunFlow[I, O, S any](ctx context.Context, flow *core.Flow[I, O, S], input I) (O, error)
    RunFlow runs flow in the context of another flow. The flow must run to
    completion when started (that is, it must not have interrupts).

func StartFlowServer(addr string) error
    StartFlowServer starts a server serving the routes described in
    NewFlowServeMux. It listens on addr, or if empty, the value of the PORT
    environment variable, or if that is empty, ":3400".

    In development mode (if the environment variable GENKIT_ENV=dev), it also
    starts a dev server.

    StartFlowServer always returns a non-nil error, the one returned by
    http.ListenAndServe.

func StreamFlow[I, O, S any](ctx context.Context, flow *core.Flow[I, O, S], input I) func(func(*StreamFlowValue[O, S], error) bool)
    StreamFlow runs flow on input and delivers both the streamed values and
    the final output. It returns a function whose argument function (the "yield
    function") will be repeatedly called with the results.

    If the yield function is passed a non-nil error, the flow has failed with
    that error; the yield function will not be called again. An error is also
    passed if the flow fails to complete (that is, it has an interrupt).
    Genkit Go does not yet support interrupts.

    If the yield function's StreamFlowValue argument has Done == true, the
    value's Output field contains the final output; the yield function will not
    be called again.

    Otherwise the Stream field of the passed StreamFlowValue holds a streamed
    result.


TYPES

type NoStream = core.NoStream

type StreamFlowValue[O, S any] struct {
	Done   bool
	Output O // valid if Done is true
	Stream S // valid if Done is false
}
    StreamFlowValue is either a streamed value or a final output of a flow.

ianlancetaylor (approved):

PR #247, opened 2024-05-24
[Go] move tracing under core

jba:
Move internal/tracing to core/tracing, so third-party plugins
can use it.


go/core/flow.go
```diff
@@ -485,10 +484,9 @@ func Run[T any](ctx context.Context, name string, f func() (T, error)) (T, error
 	// as in the js.
 	return tracing.RunInNewSpan(ctx, fc.tracingState(), name, "flowStep", false, 0, func(ctx context.Context, _ int) (T, error) {
 		uName := fc.uniqueStepName(name)
-		spanMeta := tracing.SpanMetaKey.FromContext(ctx)
```
randall77, line 488: Is there a reason why this moved from creating a separate object from the context and adding attributes using methods? I'm not sure if that's worse or better than using global functions. But it is certainly more than just a move from one package to another.

jba: I took the opportunity to make a couple of small changes. One was to avoid exporting the context key in favor of a couple of functions. That felt like I was reducing the API a bit. For instance, the thing you get back from `SpanMetaKey.FromContext` is an unexported type, and getting an unexported symbol from an exported one always feels odd to me.

Happy to reconsider.

(The second small change was removing NewDevStore.)

randall77 (approved):

PR #251, opened 2024-05-26
[Go] Added input/output validation for flows and actions and model output conformance.

apascal07:
Summary of changes:

- Flattened input schemas so the dev UI can auto-fill values for required fields.
- Made `Part` fields exported because `inferJSONSchema` wasn't generating schema for them correctly which failed validation.
- Added input validation pre-JSON unmarshaling to `runJSON` for both flows and actions. This is necessary because the unmarshaling throws away unknown/invalid fields and we want to enforce clean correct inputs.
- Added input/output validation to the core `Run` method for actions and `Execute` method for flows.
- Added output conformance to `Generate` methods for model actions.
- Removed special case for empty struct (void) input because it failed input validation in conflicting ways between action input and flow input. No special case works fine.

To dos:

- [ ] `Config.OutputSchema` needs to be `*jsonschema.Schema` instead of `map[string]any`. Tried replacing it but it seemed to cause a panic, but `Config.InputSchema` is already that type so there must be more to it...
- [x] Doesn't support `null` as an input right now; seems to convert it to an object at some point.
- [ ] Needs a friendlier API for providing output schema when starting from a struct.
- [ ] Input validation errors at the `runJSON` stage display terribly (unreadably, need to look in terminal) in dev UI right now since they run outside of the tracing and dev UI only reads from tracing right now.


go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+// conformOutput appends a message to the request indicating conformance to the expected schema.
+func conformOutput(input *GenerateRequest) error {
+	if len(input.Output.Schema) > 0 && len(input.Messages) > 0 {
+		jsonBytes, err := json.Marshal(input.Output.Schema)
+		if err != nil {
+			return fmt.Errorf("expected schema is not valid: %w", err)
+		}
+
+		jsonStr := string(jsonBytes)
+		escapedJSON := strconv.Quote(jsonStr)
```
jba, line 170: You can do this in one line and eliminate `jsonStr`.

go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+		part := &Part{
+			text: fmt.Sprintf("Output should be in JSON format and conform to the following schema:\n\n```%s```", escapedJSON),
+		}
+		input.Messages[len(input.Messages)-1].Content = append(input.Messages[len(input.Messages)-1].Content, part)
+	}
+	return nil
+}
+
+// findValidCandidates finds all candidates that match the expected schema.
+func findValidCandidates(ctx context.Context, resp *GenerateResponse) []*Candidate {
```
jba, line 180: You could just call this `validCandidates`.

go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+	candidates := []*Candidate{}
```
jba, line 181: var candidates []*Candidate

go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+	}
+
+	text, err := candidate.Text()
+	if err != nil {
+		return err
+	}
+
+	text = stripJsonDelimiters(text)
+
+	var jsonData interface{}
```
jba, line 207: any

go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+	err = json.Unmarshal([]byte(text), &jsonData)
```
jba, line 208: if err := ...; err != nil

go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+	jsonLoader := gojsonschema.NewGoLoader(jsonData)
+	result, err := gojsonschema.Validate(schemaLoader, jsonLoader)
+	if err != nil {
+		return fmt.Errorf("failed to validate expected schema: %w", err)
+	}
+
+	if !result.Valid() {
+		var errMsg string
+		for _, err := range result.Errors() {
+			errMsg += fmt.Sprintf("- %s\n", err)
```
jba, line 228: If `result.Errors` is a `[]error`, you can write
```
err := errors.Join(result.Errors()...)
```

go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+		}
+		return fmt.Errorf("candidate did not match expected schema:\n%s", errMsg)
+	}
+
+	return nil
+}
+
+// stripJsonDelimiters strips JSON delimiters that may come back in the response.
+func stripJsonDelimiters(s string) string {
```
jba, line 237: `stripJSONDelimiters`

go/ai/generator.go
```diff
@@ -134,6 +158,86 @@ func (ga *generatorAction) Generate(ctx context.Context, input *GenerateRequest,
...
+		for _, err := range result.Errors() {
+			errMsg += fmt.Sprintf("- %s\n", err)
+		}
+		return fmt.Errorf("candidate did not match expected schema:\n%s", errMsg)
+	}
+
+	return nil
+}
+
+// stripJsonDelimiters strips JSON delimiters that may come back in the response.
```
jba, line 236: Mention in this comment that these are Markdown.

go/samples/coffee-shop/main.go
```diff
@@ -151,6 +165,51 @@ func main() {
 		return text, nil
 	})
 
+	r := &jsonschema.Reflector{
```
jba, line 168: We should design an API to do this for people. Like maybe another field in dotprompt.Config that you can set to a value to reflect, instead of setting InputSchema. WDYT?

jba (changes requested):
Thanks Alex, great stuff. Especially the extensive tests.

go/ai/generator.go
apascal07: It is `[]gojsonschema.ResultError` unfortunately.

go/ai/generator_test.go
```diff
@@ -0,0 +1,193 @@
...
+					NewTextPart(JSONMarkdown(`{"name": "John", "age": 30`)), // Missing trailing }.
+				},
+			},
+		}
+		outputSchema := &GenerateRequestOutput{
+			Format: OutputFormatJSON,
+		}
+		_, err := validCandidate(candidate, outputSchema)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "data is not valid JSON")
```
jba, line 122: I believe this will panic if err == nil. See my comment above.

jba: Just to be clear, it is fine to write helper functions to make it easier to write tests. But they are best kept local.
For example, you have a lot of tests that look for strings in the error message. (Which we also frown upon, but that's another story.) You could write a helper:
```
func errorContains(t *testing.T, err error, want string) {
   t.Helper()
   if err == nil {
       t.Error("got nil, want error")
   } else if !strings.Contains(err.Error(), want) {
      t.Errorf("got error message %q, want it to  contain %q", err, want)
   }
}
```

go/core/flow.go
```diff
@@ -164,6 +171,27 @@ type retryInstruction struct {
 	FlowID string `json:"flowId,omitempty"`
 }
 
+// flowInstructioner is the common type of all flowInstruction[I] types.
+type flowInstructioner interface {
+	IsFlowInstruction()
```
jba, line 176: I think all these methods can be unexported.

go/core/flow.go
```diff
@@ -370,7 +407,19 @@ func (f *Flow[I, O, S]) execute(ctx context.Context, state *flowState[I, O], dis
...
 		start := time.Now()
-		output, err := f.fn(ctx, input, cb)
+		var err error
+		if err = ValidateObject(input, f.inputSchema); err != nil {
+			err = fmt.Errorf("invalid input: %w", err)
+		}
+		var output O
+		if err == nil {
+			output, err = f.fn(ctx, input, cb)
+			if err != nil {
```
jba, line 417: if err == nil, I think

go/core/validation.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/invopop/jsonschema"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+// ValidateObject will validate any object against the expected schema.
+// It will return an error if it doesn't match the schema, otherwise it will return nil.
+func ValidateObject(data any, schema *jsonschema.Schema) error {
```
jba, line 27: ValidateValue (we don't call them objects in Go).

go/core/validation.go
```diff
@@ -0,0 +1,71 @@
...
+// ValidateRaw will validate JSON data against the JSON schema.
+// It will return an error if it doesn't match the schema, otherwise it will return nil.
+func ValidateRaw(dataBytes json.RawMessage, schemaBytes json.RawMessage) error {
+	var data any
+	// Do this check separately from below to get a better error message.
+	if err := json.Unmarshal(dataBytes, &data); err != nil {
+		return fmt.Errorf("data is not valid JSON: %w", err)
+	}
+
+	schemaLoader := gojsonschema.NewBytesLoader(schemaBytes)
```
jba, line 54: I see, so you have to convert a jsonschema.Schema into a gojsonschema.Schema? That's unfortunate. Maybe @ianlancetaylor can help with that.

go/core/validation.go
```diff
@@ -0,0 +1,71 @@
...
+	documentLoader := gojsonschema.NewBytesLoader(dataBytes)
+
+	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
+	if err != nil {
+		return fmt.Errorf("failed to validate data against expected schema: %w", err)
+	}
+
+	if !result.Valid() {
+		var errors string
```
jba, line 63: If you can't use errors.Join then instead do this:
```
var errors []string
for _, err := range result.Errors() {
   errors = append(errors, fmt.Sprintf("- %s", err)
}
return fmt.Errorf(".....:\n%s", strings.Join(errors))
```
That will prevent quadratic behavior from repeated `+=` on a string.

jba (changes requested):
Again, thanks for all this. Sorry it's such a beast.
And also sorry for our API updates, which will give you some merge headaches.
I would try to get this in quickly even if incomplete, to avoid future merge pain.

go/core/validation.go
ianlancetaylor: Looking at it but this is likely the best we can do for now.

go/core/flow.go
apascal07: This is actually not needed at all anymore. I was playing around with something before but it's not necessary. Reverted.

go/core/flow.go
apascal07: Good catch!

go/ai/generator_test.go
apascal07: Used the helper here, thanks!

go/ai/generator_test.go
```diff
@@ -0,0 +1,193 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ai
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
apascal07: Got rid of the assert, thanks!

go/ai/document.go
```diff
@@ -31,11 +31,11 @@ type Document struct {
 // A Part is one part of a [Document]. This may be plain text or it
 // may be a URL (possibly a "data:" URL with embedded data).
 type Part struct {
-	kind         partKind
-	contentType  string        // valid for kind==blob
-	text         string        // valid for kind∈{text,blob}
-	toolRequest  *ToolRequest  // valid for kind==partToolRequest
-	toolResponse *ToolResponse // valid for kind==partToolResponse
+	Kind         partKind      `json:"kind,omitempty"`
```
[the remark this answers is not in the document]
apascal07: Ah yeah, missed this. Unfortunately, had to undo all of this nice visibility control here because there's no way to generate the schema correctly any other way from what I can tell.

go/ai/generator.go
```diff
@@ -175,77 +175,63 @@ func conformOutput(input *GenerateRequest) error {
...
+	return candidates, nil
+}
 
-	if !result.Valid() {
-		var errMsg string
-		for _, err := range result.Errors() {
-			errMsg += fmt.Sprintf("- %s\n", err)
+// validCandidate will validate the candidate's response against the expected schema.
+// It will return an error if it does not match, otherwise it will return a candidate with pure JSON content.
+func validCandidate(candidate *Candidate, outputSchema *GenerateRequestOutput) (*Candidate, error) {
```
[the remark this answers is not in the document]
apascal07: Renamed this and a few other things to be consistent with convention/other functions.

apascal07:
> Again, thanks for all this. Sorry it's such a beast. And also sorry for our API updates, which will give you some merge headaches. I would try to get this in quickly even if incomplete, to avoid future merge pain.

No worries! Thanks for the thorough review, this guidance was very helpful.

go/core/validation.go
apascal07: Yeah... this is not ideal. The package you were using is the better (and much newer) of the two but it doesn't provide any validation.

go/samples/coffee-shop/main.go
apascal07: This is a good idea. This PR is getting big so I added it to the 3 other TODOs to do in a follow up.

PR #269, opened 2024-05-28
[Go] rename type params in core package

jba:


go/core/flow.go
```diff
@@ -172,7 +172,7 @@ type flowState[I, O any] struct {
 	FlowName string `json:"name,omitempty"`
 	// start time in milliseconds since the epoch
 	StartTime tracing.Milliseconds `json:"startTime,omitempty"`
-	Input     I                  `json:"input,omitempty"`
+	Input     I                    `json:"input,omitempty"`
```
ianlancetaylor, line 175: Shouldn't `flowState` change to `In, Out` ?

ianlancetaylor (approved):

go/core/flow.go
jba: Done.

PR #273, opened 2024-05-29
feat: added `toolRequests` helper to GenerateResponse to make it easier to work with tools

pavelgj:
Checklist (if applicable):


js/ai/src/generate.ts
```diff
@@ -104,6 +105,16 @@ export class Message<T = unknown> implements MessageData {
...
 
+  /**
+   * Returns all tool request found in this message.
+   * @returns Array of all tool request found in this message.
+   */
+  toolRequests(): ToolRequestPart[] {
+    return this.content.filter(
+      (part) => !!part.toolRequest
+    ) as ToolRequestPart[];
+  }
```
mbleigh, line 116: Should this actually return the tool request itself instead of just the part? i.e. if I have: `[{text: "foo", {toolRequest: {foo: "bar"}]` should this return `[{toolRequest: {foo: "bar"}]` or should it return `[{foo: "bar"}]`?

Oh wait, parts can have metadata, nevermind.

mbleigh (approved):

huangjeff5 (approved):

js/ai/src/generate.ts
pavelgj: wait, what metadata? I think your suggestion actually makes sense...

pavelgj: oh, EmptyPartSchema has the metadata.... so yeah, metadata is at the part level... ugh... undo!

PR #276, opened 2024-05-29
[Go] change embedder plugins to use Init and New

ianlancetaylor:
Change the pinecone and localvec plugins to use a New function that returns an ai.Retriever, and an Init function that registers one with genkit.

This makes the pinecone and localvec plugins consistent with each other, and with other plugins such as googleai.


randall77 (approved):

ianlancetaylor:
Argh, I did not mean to include the menu samples.

PR #284, opened 2024-05-30
[Go] add general prompt support to ai package

ianlancetaylor:
Also rationalize the dotprompt package and make it easier to use.


apascal07 (approved):
This is much better, thanks!

PR #285, opened 2024-05-30
[Go] Fixed streaming.

apascal07:
Streamed responses need to be flushed otherwise they will all be returned at the same time at the end. Also, streaming relies on newline characters so response JSON cannot be pretty-printed.


genkit-tools/common/src/runner/runner.ts
```diff
@@ -418,21 +418,31 @@ export class Runner {
...
-        resolver = r;
+      let rejecter: (err: Error) => void;
+      const promise = new Promise<RunActionResponse>((resolve, reject) => {
+        resolver = resolve;
+        rejecter = reject;
       });
       stream.on('end', () => {
         resolver(RunActionResponseSchema.parse(JSON.parse(buffer)));
       });
+      stream.on('error', (err: Error) => {
```
pavelgj, line 443: Does this by any chance also fix https://github.com/firebase/genkit/issues/271?

apascal07: No :( I didn't realize you filed an issue for this in JS but I've been running into the same issue with Go. Getting closer to fixing it but not quite there yet...

pavelgj: Hmm... you tried or guessing? I'm thinking it might...

apascal07: I didn't try with your case but the case I'm trying to fix similarly crashes the process due to schema validation and it didn't fix it. We're throwing an exception and it's not caught anywhere so it makes sense that it crashes it, no?

go/core/servers.go
```diff
@@ -122,7 +122,13 @@ func (s *devServer) handleRunAction(w http.ResponseWriter, r *http.Request) erro
 		// Stream results are newline-separated JSON.
 		callback = func(ctx context.Context, msg json.RawMessage) error {
 			_, err := fmt.Fprintf(w, "%s\n", msg)
-			return err
+			if err != nil {
+				return err
+			}
+			if f, ok := w.(http.Flusher); ok {
+				f.Flush()
```
pavelgj, line 129: I'm surprised `Transfer-Encoding: chunked` header is not required... but appears to be the case: https://stackoverflow.com/questions/26769626/send-a-chunked-http-response-from-a-go-server

pavelgj (approved):

PR #286, opened 2024-05-30
[Go] Reuse the same file in localvec when indexing

ssbushi:
Make LocalVec use the same DB file across runs.

Checklist (if applicable):

### DIFF
--- a/go/plugins/localvec/localvec.go
+++ b/go/plugins/localvec/localvec.go
@@ -134,7 +134,7 @@ func (r *retriever) Index(ctx context.Context, req *ai.IndexerRequest) error {
 	}
 
 	// Update the file every time we add documents.
-	tmpname := r.filename + ".tmp"
+	tmpname := r.filename
 	f, err := os.Create(tmpname)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


go/plugins/localvec/localvec.go
```diff
@@ -134,7 +134,7 @@ func (r *retriever) Index(ctx context.Context, req *ai.IndexerRequest) error {
 	}
 
 	// Update the file every time we add documents.
-	tmpname := r.filename + ".tmp"
+	tmpname := r.filename
```
jba, line 137: I don't see how this "makes locavec use the same file across runs" more than it did before. Either way it's a fixed filename, right?

ssbushi: Well, index updates are made in a different file and do not appear in the retrieved docs if you recreate the retriever. It is a filename fix though. I was describing the effect of the bug.

ianlancetaylor: Argh, the code is missing a rename of the .tmp file to the original file.  It should write to the temporary file and then call `os.Rename` to rename that to the real file.  Otherwise we are in trouble if the program crashes while writing to the temporary file.  Sorry about that.  I can send a fix tomorrow if nobody does it before then.

ssbushi: I can do that. Thanks!

jba (approved):

PR #299, opened 2024-05-31
[Go] add menu sample

ianlancetaylor:
This is based closely on the TypeScript code.

The s05 tests don't work yet. They get an error

LookupGeneratorAction: no generator action named "google-vertexai"/"gemini-1.0-pro-vision"

Using the gemini-1.0-pro action (without "-vision") doesn't work,
saying that it doesn't support the input type.


ianlancetaylor:
I don't understand how to update this for the changes on the main branch, so closing this pull request in favor of #300.

jba:
Pull and rebase as you would in gerrit, then `git push -f`.
I think that might mess with comments on defunct commits, but there weren't any in this case.

ianlancetaylor:
In Gerrit the pull and rebase is handled for me via `git codereview sync`.  Would that work on a non-Gerrit repo?

jba:
I don't know. I've never used that command (pure fear). I do it manually. 
One concern is that in gerrit, each CL is one commit, so `git codereview sync` probably always rebases just one commit.
If your PR consists of multiple commits, you have to rebase them all onto main.
For example, if there are three commits, the command, when you are on the PR branch, is
```
git rebase HEAD~3 --onto main
```

PR #322, opened 2024-06-04
[Go] support sending images to googleai/vertexai

ianlancetaylor:
The googleai and vertexai API support gs URIs or raw image bytes. The genkit flow doesn't permit raw bytes, so use base64 encoding of data URIs as the TypeScript code does.

This revealed that the JSON validation code didn't support Part correctly, as the JSON marshaling didn't match the type as seen by the JSON reflection code. Fortunately the JSON reflection package makes it easy to specify a different schema to use for any type.

Somehow menu.jpeg was submitted as an empty file, so fix that.

Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


go/plugins/vertexai/vertexai.go
```diff
@@ -180,35 +191,72 @@ func NewGenerator(ctx context.Context, model, projectID, location string) (ai.Ge
...
 // convertPart converts a *ai.Part to a genai.Part.
-func convertPart(p *ai.Part) genai.Part {
+func convertPart(p *ai.Part) (genai.Part, error) {
 	switch {
 	case p.IsText():
-		return genai.Text(p.Text)
+		return genai.Text(p.Text), nil
 	case p.IsMedia():
-		return genai.Blob{MIMEType: p.ContentType, Data: []byte(p.Text)}
+		url := p.Text
```
jba, line 212: Can this part be deduped with the googleai code? Perhaps it returns (string, []byte).
It could live in plugins/internal.
Also I think it's complicated enough to warrant a test.

jba (changes requested):
Thanks for figuring this out.

go/plugins/vertexai/vertexai.go
ianlancetaylor: Done.

jba (approved):

PR #324, opened 2024-06-04
Normalize to type/provider/name pattern for action names.

randall77:
Tools were not working because they were being inserted as
tool/tool/name and looked up as tool/name/name.

### DIFF
--- a/go/ai/embedder.go
+++ b/go/ai/embedder.go
@@ -35,8 +35,8 @@ type EmbedRequest struct {
 
 // DefineEmbedder registers the given embed function as an action, and returns an
 // [Embedder] whose Embed method runs it.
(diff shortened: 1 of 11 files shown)


jba:
Thanks, I probably broke this when I did my "Define" reorg.
I still don't understand why some components take a provider and some do not. But as long as we match the JS I'm fine.

jba (approved):

randall77:
I don't know that we're matching the js. I just tried to make the Go part self-consistent.
It's always actiontype/provider/name. In some cases where we don't have a provider I used "local" as the provider, like for tools. DocumentStore is also weird. It has providers for each of its sub-indexer and sub-retreiver, but has no provider itself.

PR #326, opened 2024-06-05
[Go] googleai: revised plugin design

jba:
This illustrates the new plugin design:

- Plugins are initialized with an Init function that holds
  configuration.

- All models and other entities are defined in Init.

- Plugins have global state.

- Configured values can be retrieved by top-level functions that
  take strings as arguments.


go/plugins/googleai/googleai.go
```diff
@@ -16,17 +16,154 @@ package googleai
...
+		for {
+			mi, err := iter.Next()
+			if err == iterator.Done {
+				break
+			}
+			if err != nil {
+				return err
+			}
+			// Model names are of the form "model/name".
+			name := path.Base(mi.Name)
```
ianlancetaylor, line 76: This isn't really a path, so I think it would be easier to understand as

    name, _, _ := strings.Cut(mi.Name, "/")

go/plugins/googleai/googleai.go
```diff
@@ -16,17 +16,154 @@ package googleai
...
+	})
+}
+
+// Generator returns the generator with the given name.
+// It panics if the generator was not configured.
+func Generator(name string) ai.Generator {
+	if g, ok := state.models[name]; ok {
+		return g
+	}
+	panic(fmt.Sprintf("googleai: generator %q not configured", name))
```
ianlancetaylor, line 147: Calling `panic` here seems pretty unfriendly.  Why not return `nil`?

go/plugins/googleai/googleai_test.go
```diff
@@ -165,35 +81,100 @@ func init() {
...
-	}
+		}
+
+		resp, err := g.Generate(ctx, req, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		out := resp.Candidates[0].Message.Content[0].Text
+		if out != "France" {
+			t.Errorf("got \"%s\", expecting \"France\"", out)
```
ianlancetaylor, line 123: Use %q.

go/plugins/googleai/googleai_test.go
```diff
@@ -165,35 +81,100 @@ func init() {
...
+		_, err = googleai.Generator(generativeModel).Generate(ctx, req, func(ctx context.Context, c *ai.Candidate) error {
+			parts++
+			out += c.Message.Content[0].Text
+			return nil
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !strings.Contains(out, "Golden") {
+			t.Errorf("got \"%s\", expecting it to contain \"Golden\"", out)
```
ianlancetaylor, line 151: Use %q.

go/plugins/googleai/googleai.go
```diff
@@ -16,17 +16,154 @@ package googleai
...
+func Embedder(name string) ai.Embedder {
+	if e, ok := state.embedders[name]; ok {
+		return e
+	}
+	panic(fmt.Sprintf("googleai: embedder %q not configured", name))
+}
+
+// Generators returns the names of the configured generators.
+func Generators() []string {
+	return maps.Keys(state.models)
```
ianlancetaylor, line 161: At this point all the generators are actually defined.  So why not return them, rather than returning strings and requiring the user to call `Generator`?

Or, another approach would be to have `Generator` call `defineModel` itself if necessary, using the mutex and client in `state`.

ianlancetaylor (commented):
It feels a bit weird to me.  We set up state initially and then look it up again later.  Why not just return it?  But I guess if this is what Genkit wants, it's not terrible.

go/plugins/googleai/googleai.go
jba: (a) It sort of is. [GCP resource names are slash-separated paths](https://cloud.google.com/apis/design/resource_names).

(b) We want the last component, not the first, so the code wouldn't be as nice.

I think this is fine.

go/plugins/googleai/googleai_test.go
jba: Done.

go/plugins/googleai/googleai.go
jba: > At this point all the generators are actually defined. So why not return them, rather than returning strings and requiring the user to call Generator?

Maybe. We'd also need a way to get the name from a Generator, in case someone wanted it.

> Or, another approach would be to have Generator call defineModel itself if necessary, using the mutex and client in state.

That brings us back to the problem that Init is trying to solve, which is to register all actions up front so the genkit UI can show them. Your way, like my original idea, would make that list control-flow dependent.

jba: Or maybe `Generators` returns (a copy of) the internal map in state. In which case who needs `Generate`?

The reason I added `Generators` in the first place was to handle the case where you provide the empty list of models and let the List RPC fill it in for you. There would be no other way to know what models were available. That's also why I just returned names; you only care about the names for that use case.

go/plugins/googleai/googleai.go
jba: Done.

jba:
> We set up state initially and then look it up again later. Why not just return it?

I agree with you, but the JS folks are more comfortable with action-at-a-distance than we are. They want the convenience of not having to plumb the state through the program. I can see their point of view.

go/plugins/googleai/googleai.go
```diff
@@ -16,17 +16,148 @@ package googleai
...
+				cfg.Models = append(cfg.Models, name)
+			}
+			if needEmbedders && slices.Contains(mi.SupportedGenerationMethods, "embedContent") {
+				cfg.Embedders = append(cfg.Embedders, name)
+			}
+		}
+	}
+	state.models = map[string]ai.Generator{}
+	for _, name := range cfg.Models {
+		state.models[name] = defineModel(name, state.client)
```
pavelgj, line 87: I'm worried maintaining state in the plugin is passing on the burden to the plugin developer. As framework we really focus on ease of use for both plugins users and plugin developers. I think we should avoid state in the plugin.

When you call `ai.DefineGenerator` the defined generator gets added to the registry which already maintains the global state. `googleai.Generator` can just lookup the generator from registry directly... we could even make it `ai.Generator`, except the user would have to provide fully qualified model names `ai.Generate("googleai/gemini-1.5-flash")`

jba: Done.
And guess what, we do need LookupEmbedder, etc. after all!

go/plugins/googleai/googleai.go
```diff
@@ -16,17 +16,148 @@ package googleai
...
+			return nil, err
+		}
+		return res.Embedding.Values, nil
+	})
+}
+
+// Generator returns the generator with the given name.
+// It returns nil if the generator was not configured.
+func Generator(name string) ai.Generator {
+	return state.models[name]
```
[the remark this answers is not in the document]
jba: Sorry, I forgot to push the latest commit. If you look now, the state is gone.

I'll address the other changes on #331, but I think they are orthogonal.

go/ai/generator.go
```diff
@@ -76,6 +76,16 @@ func DefineGenerator(provider, name string, metadata *GeneratorMetadata, generat
...
 }
 
+// LookupGenerator looks up a [Generator] registered by [DefineGenerator].
+// It returns nil if the Generator was not defined.
+func LookupGenerator(provider, name string) Generator {
+	action := core.LookupActionFor[*GenerateRequest, *GenerateResponse, *Candidate](core.ActionTypeModel, provider, name)
+	if action == nil {
+		return nil
+	}
+	return generator{action}
```
pavelgj, line 86: For whatever reason I'm trying to avoid doing this additional wrapping... maybe I'm overreacting? Just tell me it's OK and I'll let it go. :)

pavelgj (approved):

go/ai/generator.go
jba: resolved on followup

go/plugins/googleai/googleai.go
jba: I removed `Generators` and `Embedders` for now.

PR #334, opened 2024-06-06
[Go] googleai: revised plugin design

jba:
This illustrates the new plugin design:

- Plugins are initialized with an Init function that holds
  configuration.

- All models and other entities are defined in Init.

- Plugins have global state.

- Configured values can be retrieved by top-level functions that
  take strings as arguments.


jba:
meant to stack this

PR #347, opened 2024-06-07
[Go] change vertexai plugin to new style

jba:
Same change as to googleai.
Add Init function, and Model and Embedder functions to retrieve them.


go/plugins/vertexai/vertexai.go
```diff
@@ -16,15 +16,97 @@ package vertexai
...
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/plugins/internal/uri"
+	"google.golang.org/api/option"
 )
 
-func newClient(ctx context.Context, projectID, location string) (*genai.Client, error) {
-	return genai.NewClient(ctx, projectID, location)
+const provider = "google-genai"
+
+// Config configures the plugin.
```
randall77, line 32: Config provides configuration options for the Init function.

go/plugins/vertexai/vertexai.go
```diff
@@ -16,15 +16,97 @@ package vertexai
...
+		return errors.New("need at least one model or embedder")
+	}
+
+	// Client for Gemini SDK.
+	gclient, err := genai.NewClient(ctx, cfg.ProjectID, cfg.Location)
+	if err != nil {
+		return err
+	}
+
+	// Client for prediction SDK, needed for embedders.
```
randall77, line 69: This code seems to alternate between model setup and embedder setup. Maybe we should do all one, then all of the other. Maybe in separate functions?

randall77 (approved):

go/plugins/vertexai/vertexai.go
jba: Done.

go/plugins/vertexai/vertexai.go
jba: Done.

PR #348, opened 2024-06-07
[Go] make googlecloud.Init conform

jba:
Change the signature of googlecloud.Init to make it like the other
plugins.

### DIFF
--- a/go/plugins/googlecloud/googlecloud.go
+++ b/go/plugins/googlecloud/googlecloud.go
@@ -37,7 +37,7 @@ import (
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 )
 
-// Config is configuration for this plugin.
+// Config provides configuration options for the Init function.
 type Config struct {
 	// ID of the project to use. Required.
 	ProjectID string
--- a/go/plugins/googlecloud/googlecloud.go
+++ b/go/plugins/googlecloud/googlecloud.go
@@ -20,6 +20,8 @@ package googlecloud
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"log/slog"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


go/plugins/googlecloud/googlecloud.go
```diff
@@ -50,25 +55,31 @@ type Options struct {
...
 // In the dev environment, this does nothing unless [Options.ForceExport] is true.
-func Init(ctx context.Context, projectID string, opts *Options) error {
-	if opts == nil {
-		opts = &Options{}
+func Init(ctx context.Context, cfg Config) (err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("googlecloud.Init: %w", err)
+		}
+	}()
```
apriil15, line 63: Good approach to wrap up error using defer, learn it.

Just curious about using named return, is it mainly to avoid declaring an error variable for readability?

go/plugins/googlecloud/googlecloud.go
```diff
@@ -35,7 +37,10 @@ import (
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 )
 
-type Options struct {
+// Config is configuration for this plugin.
```
randall77, line 40: Config provides configuration options for the Init function.

randall77 (approved):

go/plugins/googlecloud/googlecloud.go
jba: You have to name the error return, or assigning to it on line 61 won't change the error returned by the function.

go/plugins/googlecloud/googlecloud.go
jba: Done.

PR #359, opened 2024-06-07
[Go] rewrite ai package Prompt to use PromptAction and Register

### LINKED ISSUE #356: [Go] rewrite ai/prompt.go to match JS more closely
Currently, we distinguish between a Prompt interface and the prompt action.
In the JS, and now in the rest of the Go ai package, the former interfaces are now type aliases for actions.
Also, in the JS, the action of a prompt is to render it, not to generate text from it.

Rewrite the general prompt API to conform more closely to what the JS does (js/ai/src/prompt.ts).
The JS PromptAction is generic in its input, so we can't use a type alias; we will need a new generic type.
We should also change the action to render.

ianlancetaylor:
Also provide ai.Render to render a prompt and generate a response.

Fixes #356

### DIFF
--- a/go/ai/prompt.go
+++ b/go/ai/prompt.go
@@ -37,18 +37,26 @@ type PromptRequest struct {
 	Model string `json:"model,omitempty"`
 }
 
-// Prompt is the interface used to execute a prompt template and
-// pass the result to a [ModelAction].
-type Prompt interface {
-	Generate(context.Context, *PromptRequest, func(context.Context, *Candidate) error) (*GenerateResponse, error)
-}
+// A PromptAction is used to generate content from an AI model using a prompt.
+type PromptAction = core.Action[*PromptRequest, *GenerateResponse, *Candidate]
 
-// RegisterPrompt registers a prompt in the global registry.
-func RegisterPrompt(provider, name string, prompt Prompt) {
+// DefinePrompt register the given function as a prompt action,
+// and returns a [PromptAction] that runs it.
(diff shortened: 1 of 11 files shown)


go/ai/prompt.go
```diff
@@ -37,18 +37,26 @@ type PromptRequest struct {
 	Model string `json:"model,omitempty"`
 }
 
-// Prompt is the interface used to execute a prompt template and
-// pass the result to a [ModelAction].
-type Prompt interface {
-	Generate(context.Context, *PromptRequest, func(context.Context, *Candidate) error) (*GenerateResponse, error)
-}
+// A PromptAction is used to generate content from an AI model using a prompt.
+type PromptAction = core.Action[*PromptRequest, *GenerateResponse, *Candidate]
```
jba, line 41: Maybe I'm misreading the JS, but it seems to me that the prompt action doesn't do generation, it just produces the input:
```
type PromptAction<I extends z.ZodTypeAny = z.ZodTypeAny> = Action<
   I,
    typeof GenerateRequestSchema>
```
which seems to correspond to the (currently invalid) Go
```
type PromptAction[I any] = Action[I, *GenerateRequest, struct{}]
```

ianlancetaylor: Hmmm, I expect you are right.  Now I'm not sure what `PromptAction` is for on the TypeScript side.

jba: Presumably you call the prompt to get the request, then call ai.Generate with the request, but I haven't checked.

ianlancetaylor:
Taking a different approach in #364.

PR #365, opened 2024-06-07
test that streaming and final answer contain the same text.

randall77:
This should already be the case, just testing it.

### DIFF
--- a/go/plugins/googleai/googleai_test.go
+++ b/go/plugins/googleai/googleai_test.go
@@ -150,10 +150,8 @@ func TestLive(t *testing.T) {
 			t.Fatal(err)
 		}
 		out2 := ""
-		for _, c := range final.Candidates {
-			for _, p := range c.Message.Content {
-				out2 += p.Text
-			}
+		for _, p := range final.Candidates[0].Message.Content {
+			out2 += p.Text
 		}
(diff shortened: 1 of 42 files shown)


go/plugins/googleai/googleai_test.go
```diff
@@ -141,14 +141,23 @@ func TestLive(t *testing.T) {
...
+		final, err := ai.Generate(ctx, g, req, func(ctx context.Context, c *ai.GenerateResponseChunk) error {
 			parts++
 			out += c.Content[0].Text
 			return nil
 		})
 		if err != nil {
 			t.Fatal(err)
 		}
+		out2 := ""
+		for _, c := range final.Candidates {
```
apascal07, line 153: I'm not sure this is right. IIRC, candidate chunks are streamed in parallel and can be streamed in any order so it's not necessarily going to be like aaabbbccc.

randall77: There's only one candidate in this test, so I think it doesn't matter.
I'll just use Candidates[0] instead of looping on the Candidates slice.

randall77:
Abandoning, managed to get in a bad github state. Remailed in khr-work26.

PR #367, opened 2024-06-08
[Go] proposal: vector DB Inits return actions

jba:
Vector DBs only register two actions: an indexer and a retriever.
Continue to allow users to look up these actions by name.
But also return them from Init.

### DIFF
--- a/go/plugins/localvec/localvec.go
+++ b/go/plugins/localvec/localvec.go
@@ -38,33 +38,45 @@ const provider = "devLocalVectorStore"
 
 // Config provides configuration options for the Init function.
 type Config struct {
+	Stores []StoreConfig
+}
+
+type StoreConfig struct {
 	// Where to store the data. Defaults to os.TempDir.
-	Dir             string
+	Dir string
+	// A name that uniquely identifies the store.
 	Name            string
 	Embedder        *ai.EmbedderAction
 	EmbedderOptions any
 }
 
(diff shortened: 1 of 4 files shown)


pavelgj:
LGTM

One use case I'd like to point out, and I think it works fine with this proposal -- initializing multiple vector stores from the plugin:

```
	cfi, cfr, err := localvec.Init(context.Background(), localvec.Config{
		Name:     "cat-facts",
		Embedder: googleai.Embedder("embedding-001"),
	})
	dfi, dfr, err := localvec.Init(context.Background(), localvec.Config{
		Name:     "dog-facts",
		Embedder: googleai.Embedder("embedding-001"),
	})
```

`Init` conveys the semantics well -- you are only allowed to initialize it once per given `Name`.

pavelgj (approved):

ianlancetaylor (approved):

jba:
Actually, I think you've provided a counterexample. I didn't realize you might want to instantiate a vector DB with two indexes in the same program. That makes the vector DB case look more like the generative model case, where there could be multiple named entities of the same kind. Now that I took a better look at your rag test app, I see that you actually use that ability.

So I think for consistency, a single Init should completely initialize the plugin.

Once we have multiple retrievers and indexers, I think `pinecone.Retriever("my-index")` is the best we can do.

PR #369, opened 2024-06-09
[Go] suggestion: associate type params with Action Types

jba:
Define a type called Assoc that links the Go type of an action
to a value of the ActionType enum. That is, a line like

   const fooAssoc = atype.Assoc[Action[int, bool, string]](atype.Foo)

encodes the relationship between the ActionType enum value Foo and
actions with type parameters [int, bool, string].

The main benefit is that the relationship needs to be written only once
in the program. Previously it appeared twice: in the Define function and
in the Lookup function.

### DIFF
--- a/go/ai/embedder.go
+++ b/go/ai/embedder.go
@@ -25,6 +25,8 @@ import (
 // multidimensional vector.
 type EmbedderAction = core.Action[*EmbedRequest, []float32, struct{}]
 
+const embedderAssoc = atype.Assoc[EmbedderAction](atype.Embedder)
+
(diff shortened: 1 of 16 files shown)


jba:
Reviewer questions:
- Does this add more complexity than it's worth?
- Can you think of a better name than Assoc?

ianlancetaylor:
This does seem a bit complicated to me.  It's clever but I'm concerned that future readers will find it difficult to puzzle out what is going on here.

jba:
OK, I sort of agree.

PR #372, opened 2024-06-10
[Go] googleai.Init: return actions

jba:
Init returns the models and embedders that it defines.

Users can access defined actions in one of three ways: by slice index,
by calling Action.Name on a slice element, or by calling
the Model or Embedder functions.


go/plugins/googleai/googleai.go
```diff
@@ -42,20 +42,25 @@ type Config struct {
 	Embedders []string
 }
 
-func Init(ctx context.Context, cfg Config) (err error) {
+// Init initializes the plugin.
```
randall77, line 45: s/the/this/

go/samples/coffee-shop/main.go
```diff
@@ -104,7 +104,7 @@ func main() {
 		fmt.Fprintln(os.Stderr, "You can get an API key at https://ai.google.dev.")
 		os.Exit(1)
 	}
-	err := googleai.Init(context.Background(), googleai.Config{APIKey: apiKey})
+	_, _, err := googleai.Init(context.Background(), googleai.Config{APIKey: apiKey})
```
randall77, line 107: Just to understand, we could do instead here:

gs, _, err := googleai.Init(context.Background(), googleai.Config{APIKey: apiKey, Models: []string{"gemini-1.5-pro"}})
g := gs[0]

?

It seems weird to me that the empty list means "initialize them all". In the call above, the plugin will initialize all the embedders because we didn't specify, and didn't need, any embedders. Maybe there is a more explicit way to say "all"? config.Models == []string{"*"}?
(Fine to tackle in a separate CL.)

randall77 (approved):

jba:
Abandoning this in favor of #379.

PR #375, opened 2024-06-10
[Go] pgvector sample

jba:
A sample demonstrating how to use Postgres's vector extension
to build an indexer and retriever.


go/samples/pgvector/main.go
```diff
@@ -0,0 +1,215 @@
...
+			Document: &ai.Document{Content: []*ai.Part{ai.NewTextPart(in.Question)}},
+			Options:  in.Show,
+		})
+		if err != nil {
+			return "", err
+		}
+		for _, doc := range res.Documents {
+			fmt.Printf("%+v %q\n", doc.Metadata, doc.Content[0].Text)
+		}
+		// Use documents in RAG prompts.
```
randall77, line 105: Is this a TODO? I'm not sure what the point of retrieving here is, as you don't actually answer the question asked.

ianlancetaylor:
Are we not going to add this as a plugin?

go/samples/pgvector/main.go
jba: It's deliberately a sketch. So yeah it's a TODO, but not for us, for whoever copies the sample.

I based it off of https://github.com/firebase/genkit/blob/main/docs/templates/pgvector.md.
I decided to write actual code rather than a markdown file so I could be sure it actually worked.

ianlancetaylor (approved):

PR #378, opened 2024-06-10
add ollama plugin

huangjeff5:


jba (changes requested):

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/firebase/genkit/go/ai"
+)
+
+const provider = "ollama"
+
+func defineModel(name string, serverAddress string) {
+	meta := &ai.ModelMetadata{
+		Label: "Ollama - " + name,
+		Supports: ai.ModelCapabilities{
+			Multiturn: true,
```
pavelgj, line 38: also need to support system prompt, ollama supports it (follow up ok)

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,226 @@
...
+func Model(name string) *ai.ModelAction {
+	return ai.LookupModel(provider, name)
+}
+
+// Config provides configuration options for the Init function.
+type Config struct {
+	// API key. Required.
+	ServerAddress string
+	// Generative models to provide.
+	Model string
```
pavelgj, line 56: this needs to be an array

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,226 @@
...
+format: the format to return a response in. Currently the only accepted value is json
+options: additional model parameters listed in the documentation for the Modelfile such as temperature
+system: system message to (overrides what is defined in the Modelfile)
+template: the prompt template to use (overrides what is defined in the Modelfile)
+context: the context parameter returned from a previous request to /generate, this can be used to keep a short conversational memory
+stream: if false the response will be returned as a single response object, rather than a stream of objects
+raw: if true no formatting will be applied to the prompt. You may choose to use the raw parameter if you are specifying a full templated prompt in your request to the API
+keep_alive: controls how long the model will stay loaded into memory following the request (default: 5m)
+*/
+type OllamaRequest struct {
```
pavelgj, line 81: can be private?

pavelgj (approved):

go/plugins/ollama/ollama.go
huangjeff5: Sounds good, will follow up on this after hack week.

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+const provider = "ollama"
+
+func defineModel(name string, serverAddress string) {
+	meta := &ai.ModelMetadata{
+		Label: "Ollama - " + name,
+		Supports: ai.ModelCapabilities{
+			Multiturn: true,
+		},
+	}
+	g := generator{model: name, serverAddress: serverAddress}
```
randall77, line 41: This should probably be g := &generator{...}, just so it matches the pointer methods we define.
(The other model plugins do it the way you did, so probably not a huge deal.)

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+
+// Model returns the [ai.ModelAction] with the given name.
+// It returns nil if the model was not configured.
+func Model(name string) *ai.ModelAction {
+	return ai.LookupModel(provider, name)
+}
+
+// Config provides configuration options for the Init function.
+type Config struct {
+	// API key. Required.
```
randall77, line 53: This comment is presumably not correct?

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+		if err := scanner.Err(); err != nil {
+			return nil, fmt.Errorf("error reading stream: %v", err)
+		}
+		// Handle end of stream (optional: send a final candidate to signal completion)
+	}
+	//Return an empty generate response, since we use callback for streaming
+	return &ai.GenerateResponse{}, nil
+}
+
+// convertParts serializes a slice of *ai.Part into a a map (represent Ollama message type)
```
randall77, line 161: remove duplicate "a".

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+		ai.RoleSystem: "system",
+		// Add more mappings as needed
+	}
+
+	partMap := map[string]string{}
+	for _, part := range parts {
+		partMap["role"] = roleMapping[role]
+		switch {
+		case part.IsText():
+			partMap["content"] = part.Text
```
randall77, line 175: What happens if there are multiple parts? Shouldn't we concatenate these, or make a list, or something?

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+options: additional model parameters listed in the documentation for the Modelfile such as temperature
+system: system message to (overrides what is defined in the Modelfile)
+template: the prompt template to use (overrides what is defined in the Modelfile)
+context: the context parameter returned from a previous request to /generate, this can be used to keep a short conversational memory
+stream: if false the response will be returned as a single response object, rather than a stream of objects
+raw: if true no formatting will be applied to the prompt. You may choose to use the raw parameter if you are specifying a full templated prompt in your request to the API
+keep_alive: controls how long the model will stay loaded into memory following the request (default: 5m)
+*/
+type ollamaRequest struct {
+	Messages []map[string]string `json:"messages"`
```
randall77, line 84: I'm not a fan of this being just a map[string]string. We should probably define internal types here with the correct fields/subfields and have json labels. Just taking a stab at it:

type ollamaMessage struct {
    role string // `json:"role"`
    content string // `json:"content"`
}

Then ollamaRequest.Messages would have type []*ollamaMessage

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+	Model    string              `json:"model"`
+	Stream   bool                `json:"stream"`
+}
+
+// Generate makes a request to the Ollama API and processes the response.
+func (g *generator) generate(ctx context.Context, input *ai.GenerateRequest, cb func(context.Context, *ai.GenerateResponseChunk) error) (*ai.GenerateResponse, error) {
+	// Step 1: Combine parts from all messages into a single payload slice
+	var messages []map[string]string
+
+	// Add all messages to history field.
```
randall77, line 94: There's no history field.

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+	for _, m := range input.Messages {
+		message, err := convertParts(m.Role, m.Content)
+		if err != nil {
+			return nil, fmt.Errorf("error converting message parts: %v", err)
+		}
+		messages = append(messages, message)
+	}
+	fmt.Println("should stream", cb != nil)
```
randall77, line 102: Remove this logging?

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+			}
+			cb(ctx, chunk)
+		}
+
+		if err := scanner.Err(); err != nil {
+			return nil, fmt.Errorf("error reading stream: %v", err)
+		}
+		// Handle end of stream (optional: send a final candidate to signal completion)
+	}
+	//Return an empty generate response, since we use callback for streaming
```
randall77, line 157: In the other plugins, we wound up having the final return value be the merging of all the streaming responses we already returned (as if it were a non-streaming request). We should probably do the same here.

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+	return &ai.GenerateResponse{}, nil
+}
+
+// convertParts serializes a slice of *ai.Part into a a map (represent Ollama message type)
+func convertParts(role ai.Role, parts []*ai.Part) (map[string]string, error) {
+	roleMapping := map[ai.Role]string{
```
randall77, line 163: This could be an unexported global variable instead of a local one. Then it is clear it is static.

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+			partMap["content"] = part.Text
+		default:
+			return nil, errors.New("unknown content type")
+		}
+	}
+	return partMap, nil
+}
+
+func translateChunk(input string) (*ai.GenerateResponseChunk, error) {
+	log.Printf("translating chunk")
```
randall77, line 184: remove log

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+	var response GenerateResponse
+
+	if err := json.Unmarshal([]byte(input), &response); err != nil {
+		return nil, fmt.Errorf("error parsing response JSON: %v", err)
+	}
+	chunk := &ai.GenerateResponseChunk{
+		Index:   0,
+		Content: make([]*ai.Part, 0, 1),
+	}
+	return chunk, nil
```
randall77, line 194: Don't we have to set chunk.Content[0] to something?
Maybe we need a test :)

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+type GenerateResponse struct {
```
randall77, line 197: This type can be unexported. (I think? Not sure about if json can handle that. If not, maybe it needs to be in an internal package somewhere, or maybe a function-local type. We don't want this exported. Same goes for the types I was mentioning earlier instead of the map[string]string)

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,228 @@
...
+
+// translateResponse deserializes a JSON response from the Ollama API into a GenerateResponse.
+func translateResponse(responseData []byte) (*ai.GenerateResponse, error) {
+	var response GenerateResponse
+
+	if err := json.Unmarshal(responseData, &response); err != nil {
+		return nil, fmt.Errorf("error parsing response JSON: %v", err)
+	}
+	generateResponse := &ai.GenerateResponse{
+		Candidates: make([]*ai.Candidate, 0, 1),
```
randall77, line 214: You don't need make(..,0,1), just use nil. Append will allocate for you.

go/plugins/ollama/ollama.go
huangjeff5: Ah thats true, I only handled the text case here where there is only 1 part but that's a good point, I updated the logic to make it generic.

go/plugins/ollama/ollama.go
huangjeff5: Done

go/plugins/ollama/ollama.go
huangjeff5: Yeah you're right. This PR doesn't handle streaming fully. I'll have it be a follow on PR.

go/plugins/ollama/ollama.go
huangjeff5: What does supporting system prompt mean exactly? Would that be looking for the first system message in the genkit GenerateRequest.Message array?

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,340 @@
...
+format: the format to return a response in. Currently the only accepted value is json
+options: additional model parameters listed in the documentation for the Modelfile such as temperature
+system: system message to (overrides what is defined in the Modelfile)
+template: the prompt template to use (overrides what is defined in the Modelfile)
+context: the context parameter returned from a previous request to /generate, this can be used to keep a short conversational memory
+stream: if false the response will be returned as a single response object, rather than a stream of objects
+raw: if true no formatting will be applied to the prompt. You may choose to use the raw parameter if you are specifying a full templated prompt in your request to the API
+keep_alive: controls how long the model will stay loaded into memory following the request (default: 5m)
+*/
+type ollamaRequest struct {
```
randall77, line 100: It's not clear at first reading what the difference is between ollamaRequest and ollamaGenerateRequest (and similar for Response). Adding a comment somewhere about the difference would help.

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,340 @@
...
+// Generate makes a request to the Ollama API and processes the response.
+func (g *generator) generate(ctx context.Context, input *ai.GenerateRequest, cb func(context.Context, *ai.GenerateResponseChunk) error) (*ai.GenerateResponse, error) {
+
+	stream := cb != nil
+	var payload any
+	isChatModel := g.model.Type == "chat"
+	if !isChatModel {
+		payload = ollamaGenerateRequest{
+			Model:  g.model.Name,
+			Prompt: concatMessages(input, ai.Role("user")),
```
randall77, line 138: This feels strange that we're concatenating all the same-role text. It loses the interleaving of the conversation so far.
Does the JS ollama client do this?

go/plugins/ollama/ollama.go
```diff
@@ -0,0 +1,340 @@
...
+			Stream:   stream,
+		}
+	}
+	client := &http.Client{
+		Timeout: time.Second * 30,
+	}
+	payloadBytes, err := json.Marshal(payload)
+	// Determine the correct endpoint
+	endpoint := g.serverAddress + "/api/chat"
+	if g.model.Type != "chat" {
```
randall77, line 164: !isChatModel

randall77 (approved):

go/plugins/ollama/ollama.go
huangjeff5: Good point, I renamed it to ollamaChatRequest as well.

go/plugins/ollama/ollama.go
huangjeff5: True... there are some issues with the ollama JS client as well (doesn't have a system prompt), but you're right that it makes more sense to interleave all the messages. (at least the non-system messages)

go/plugins/ollama/ollama.go
```diff
@@ -246,3 +273,14 @@ func translateChunk(input string) (*ai.GenerateResponseChunk, error) {
 	chunk.Content = append(chunk.Content, aiPart)
 	return chunk, nil
 }
+
+// getPrompt extracts the text content from a GenerateRequest and concatenates it into a single string.
+func getPrompt(input *ai.GenerateRequest) string {
+	prompt := ""
```
jba, line 328: This is quadratic. Use strings.Builder.

huangjeff5: Gotcha, thanks for the pointer. Fixed

PR #379, opened 2024-06-10
[Go] proposal: revised API

jba:
In this API, configuration (defining actions) should still happen
at the beginning of the program, but can be done incrementally by
calling functions on a plugin. For example:

    err := googleai.Init(...)   // creates state common to all actions
    m1 := googleai.DefineModel("gemini-1.0-pro")
    e1 := googleai.DefineEmbedder("emb-01")

When everything has been configured, the program must call genkit.Init.
This has two effects:
- Attempts to register additional actions will fail.
- The dev server is started.

This API combines the convenience and clarity of building and storing
actions individually with the requirement that everything that the
program needs is registered early.

DO NOT SUBMIT. Implementation is incomplete. For discussion only.

### DIFF
--- a/go/core/registry.go
+++ b/go/core/registry.go
@@ -49,6 +49,7 @@ func init() {
 type registry struct {
 	tstate  *tracing.State
 	mu      sync.Mutex
+	frozen  bool // when true, no more additions
 	actions map[string]action
 	flows   []flow
 	// TraceStores, at most one for each [Environment].
@@ -98,6 +99,9 @@ func (r *registry) registerAction(a action) {
 	key := fmt.Sprintf("/%s/%s", a.actionType(), a.Name())
 	r.mu.Lock()
 	defer r.mu.Unlock()
+	if r.frozen {
+		panic(fmt.Sprintf("attempt to register action %s in a frozen registry. Register before calling genkit.Init", key))
+	}
 	if _, ok := r.actions[key]; ok {
 		panic(fmt.Sprintf("action %q is already registered", key))
 	}
@@ -106,6 +110,12 @@ func (r *registry) registerAction(a action) {
 	slog.Info("RegisterAction",
 		"type", a.actionType(),
 		"name", a.Name())
+}
+
+func (r *registry) freeze() {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.frozen = true
 }
 
 // lookupAction returns the action for the given key, or nil if there is none.
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -23,6 +23,30 @@ import (
 	"github.com/firebase/genkit/go/core"
 )
 
+// Options are options to [Init].
+type Options = core.Options
+
+// Init initializes Genkit.
+// After it is called, no further actions can be defined.
+//
+// Init starts servers depending on the value of the GENKIT_ENV
+// environment variable and the provided options.
+//
+
+// If GENKIT_ENV = "dev", a development server is started
+// in a separate goroutine at the address in opts.DevAddr, or the default
+// if empty.
+//
+// If opts.FlowAddr is a value other than "-", a flow server is started (see [StartFlowServer])
+// and the call to Init waits for the server to shut down.
+// If opts.FlowAddr == "-", no flow server is started and Init returns immediately.
+//
+// Thus Init(nil) will start a dev server in the "dev" environment, will always start
+// a flow server, and will pause execution until the flow server terminates.
+func Init(opts *Options) error {
+	return core.InternalInit(opts)
+}
+
 // DefineFlow creates a Flow that runs fn, and registers it as an action.
 //
 // fn takes an input of type In and returns an output of type Out.
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -25,11 +25,13 @@ import (
 
 // Options are options to [Init].
 type Options struct {
-	DevAddr string
 	// If "-", do not start a FlowServer.
 	// Otherwise, start a FlowServer on the given address, or the
 	// default of ":3400" if empty.
 	FlowAddr string
+	// The names of flows to serve.
+	// If empty, all registered flows are served.
+	Flows []string
 }
 
 // Init initializes Genkit.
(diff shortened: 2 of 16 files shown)


go/core/registry.go
```diff
@@ -109,6 +113,16 @@ func (r *registry) registerAction(provider string, a action) {
 		"name", a.name())
 }
 
+func (r *registry) freeze() {
+	r.mu.Lock()
+	r.mu.Unlock()
```
ianlancetaylor, line 118: Missing `defer`.

jba:
I believe this is complete and correct.

Open questions:
- For vector DB plugins, it's not clear whether the indexer and retriever should be defined separately, as I've done for pinecone, or together, as I did for localvec. A separate DefineIndexer and DefineRetriever is more in keeping with the rest of the API, but provides an opportunity for different args to be passed, which would give rise to confusing bugs. But if one function is used, users who only want one of the two actions must nevertheless get both. It's also unclear what to name the function. In localvec I called it `DefineStore`, but that would be an odd name for pinecone, since it doesn't use that term anywhere else.

-  `genkit.Init` is designed so the default (passing `nil`) provides a good experience. But its behavior with arguments is complicated. Is there a simpler design?

ianlancetaylor:
A vector DB plugin is useless if you don't have both an indexer and a retriever, so I would vote for defining them together.  Maybe `DefineVectorStore` or `DefineDataStore`?  I don't think it matters that Pinecone doesn't use those terms.

jba:
I imagine that it would be common in larger setups to have one process doing the indexing, in the background as it were, while the frontend retrieves. Much like pkgsite or indeed Google Search.

ianlancetaylor:
Fair enough.

go/plugins/googleai/googleai.go
```diff
@@ -30,79 +30,75 @@ import (
...
-type Config struct {
-	// API key. Required.
-	APIKey string
-	// Generative models to provide.
-	// If empty, a complete list will be obtained from the service.
-	Models []string
-	// Embedding models to provide.
-	// If empty, a complete list will be obtained from the service.
-	Embedders []string
+var state struct {
```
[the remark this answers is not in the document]
jba: The bigger problem is the illusion that that would give of being able to create multiple plugin instances of the same kind. Ultimately, the registry is global and it's a constraint that its keys must be unique. You couldn't write
```
p1 := pinecone.New()
p1. DefineIndexer("foo", ...)
p2 := pineceone.New()
p2.DefineIndexer("foo", ...)
```
The last line will panic, because both keys are "indexer/pinecone/foo".

We could in theory allow the user to pick the second component, like
```
p1 := pinecone.New("my-first-pinecone-plugin")
```
but I think that would diverge too much from the JS approach.

Another way to put it is that the global state in the plugins is just an extension of the global state in the registry.

jba:
@pavelgj Can you comment? I believe this is the design we discussed over VC. I want to make sure you're OK with it before proceeding.

go/genkit/genkit.go
```diff
@@ -23,6 +23,30 @@ import (
 	"github.com/firebase/genkit/go/core"
 )
 
+// Options are options to [Init].
```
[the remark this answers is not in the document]
jba: Done.

go/genkit/genkit.go
```diff
@@ -23,6 +23,30 @@ import (
...
+type Options = core.Options
+
+// Init initializes Genkit.
+// After it is called, no further actions can be defined.
+//
+// Init starts servers depending on the value of the GENKIT_ENV
+// environment variable and the provided options.
+//
+
```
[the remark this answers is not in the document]
jba: Done.

go/genkit/genkit.go
```diff
@@ -23,6 +23,30 @@ import (
...
+// If GENKIT_ENV = "dev", a development server is started
+// in a separate goroutine at the address in opts.DevAddr, or the default
```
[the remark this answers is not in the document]
jba: Done.

go/genkit/genkit.go
```diff
@@ -23,6 +23,35 @@ import (
 	"github.com/firebase/genkit/go/core"
 )
 
+// Options are options to [Init].
+type Options struct {
+	DevAddr string
```
pavelgj, line 28: Dev server address/port probably should not be allowed to be specified programmatically. They are managed by the CLI and passed in via env var.

go/genkit/genkit.go
```diff
@@ -23,6 +23,35 @@ import (
...
+	// If "-", do not start a FlowServer.
+	// Otherwise, start a FlowServer on the given address, or the
+	// default of ":3400" if empty.
+	FlowAddr string
```
pavelgj, line 32: We probably want options to specify which flows we want to expose via the flows server. Perhaps this would work better as a strict?

`FlowsServer: {Addr:..., Flows:...}`

go/plugins/googleai/googleai.go
```diff
@@ -30,79 +30,75 @@ import (
...
-		defineEmbedder(name, client)
+// DefineModel defines a model with the given name.
+func DefineModel(name string) *ai.ModelAction {
+	state.mu.Lock()
+	defer state.mu.Unlock()
+	if !state.initted {
+		panic("googleai.Init not called")
 	}
-	return nil
+	return defineModel(name)
```
pavelgj, line 73: This needs `ai.ModelCapabilities`.

```go
type ModelMetadata struct {
	Name         string
	Label        string
	Capabilities ai.ModelCapabilities
}

func DefineModel(metadata ModelMetadata) *ai.ModelAction {
  // ...
}
```

go/plugins/googleai/googleai.go
```diff
@@ -115,6 +111,66 @@ func defineEmbedder(name string, client *genai.Client) {
 	})
 }
 
+// DefineAllModels defines all models known to the service.
+func DefineAllModels(ctx context.Context) ([]*ai.ModelAction, error) {
```
pavelgj, line 115: relevant conversation https://github.com/firebase/genkit/pull/426#discussion_r1644596614

go/genkit/genkit.go
jba: Done.

go/genkit/genkit.go
jba: done

go/plugins/googleai/googleai.go
jba: Will address in a follow-up PR.

randall77 (approved):

pavelgj (approved):

PR #384, opened 2024-06-10
implement vertexai streaming

### LINKED ISSUE #344: [Go] streaming for vertexai plugin
Implement streaming generation for the vertexai plugin.

randall77:
Fixes #344


go/go.sum
```diff
@@ -5,8 +5,12 @@ cloud.google.com/go/ai v0.5.0 h1:x8s4rDn5t9OVZvBCgtr5bZTH5X0O7JdE6zYo+O+MpRw=
 cloud.google.com/go/ai v0.5.0/go.mod h1:96VBphk70e0zdXZrbtgPuKYRZsQ3UktSUXhuojwiKA8=
 cloud.google.com/go/aiplatform v1.66.0 h1:bbFYY4JInclG10czRFUYj2rjD+obhh3Gi9zVlyoMgEc=
```
ianlancetaylor, line 6: Run `go mod tidy`.

go/plugins/vertexai/vertexai.go
```diff
@@ -213,13 +211,77 @@ func (g *generator) generate(ctx context.Context, input *ai.GenerateRequest, cb
...
 
-	r := translateResponse(resp)
-	r.Request = input
+	// Streaming version.
+	iter := cs.SendMessageStream(ctx, parts...)
+	r := &ai.GenerateResponse{Request: input, Candidates: make([]*ai.Candidate, input.Candidates)}
+	for {
+		chunk, err := iter.Next()
+		if err != nil {
+			if err.Error() == "no more items in iterator" {
```
ianlancetaylor, line 231: My understanding is that you can write

    if err == iterator.Done

where iterator is google.golang.org/api/iterator/.

https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines

ianlancetaylor (approved):

go/plugins/vertexai/vertexai.go
jba: Correct.

go/plugins/vertexai/vertexai.go
```diff
@@ -213,13 +211,77 @@ func (g *generator) generate(ctx context.Context, input *ai.GenerateRequest, cb
...
+				Index:   tc.Index,
+			})
+			if err != nil {
+				return nil, err
+			}
+			// Save candidate in full response structure.
+			if old := r.Candidates[tc.Index]; old == nil {
+				r.Candidates[tc.Index] = tc
+			} else {
+				// Need to merge two "parts" of a candidate.
```
jba, line 253: Would it help if vertexai/genai.joinResponses was exported?

randall77: Yes, probably would. I think the general state of streaming documentation is pretty poor, so it's hard to know for sure.

go/plugins/vertexai/vertexai.go
randall77: I tried that. I couldn't get it to work. But of course, now it works. Maybe that was fixed somewhere between 0.7.1 and 0.10.0.
And of course, it isn't documented how you should do it.

go/go.sum
randall77: Done.

go/plugins/vertexai/vertexai.go
randall77: That's a good reference though, I can just copy what they do.

randall77: They just need to export the merged field somehow, they are already calling it. I think you did the same for googleai?
It's almost accessible, I can get the content field from the chat history. But the rest is inaccessible.

jba: See https://github.com/googleapis/google-cloud-go/pull/10364

randall77: Using the newly exported api.

PR #392, opened 2024-06-12
[VertexAI] Fix Gemini 1.5 Pro model version

Dabolus:
Originally reported [here](https://github.com/TheFireCo/genkit-plugins/issues/85) (wrong repo).
Gemini 1.5 Pro available model version was incorrectly set to Gemini 1.0 Pro instead.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


MichaelDoyle (approved):
Thank you! Nice catch.

PR #397, opened 2024-06-12
Add explicit return types for plugins that define models, etc

MichaelDoyle:
I think this will help folks that want to contribute to the plugins with better type safety. I've seen a few errors go by (or almost go by) related to this.

### DIFF
--- a/js/plugins/chroma/src/index.ts
+++ b/js/plugins/chroma/src/index.ts
@@ -20,7 +20,9 @@ import {
   defineIndexer,
   defineRetriever,
   Document,
+  IndexerAction,
   indexerRef,
+  RetrieverAction,
   retrieverRef,
 } from '@genkit-ai/ai/retriever';
 import { genkitPlugin, PluginProvider } from '@genkit-ai/core';
@@ -120,7 +122,7 @@ export function chromaRetriever<
   createCollectionIfMissing?: boolean;
   embedder: EmbedderArgument<EmbedderCustomOptions>;
   embedderOptions?: z.infer<EmbedderCustomOptions>;
-}) {
+}): RetrieverAction<z.ZodOptional<typeof ChromaRetrieverOptionsSchema>> {
   const { embedder, collectionName, embedderOptions } = params;
   return defineRetriever(
     {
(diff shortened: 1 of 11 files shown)


js/plugins/chroma/src/index.ts
```diff
@@ -120,7 +122,7 @@ export function chromaRetriever<
   createCollectionIfMissing?: boolean;
   embedder: EmbedderArgument<EmbedderCustomOptions>;
   embedderOptions?: z.infer<EmbedderCustomOptions>;
-}) {
+}): RetrieverAction<z.ZodOptional<typeof ChromaRetrieverOptionsSchema>> {
```
MichaelDoyle, line 125: This seemed a little funky to me. Not sure if this is 💯 .

It looks like we're possibly double-wrapping in an Optional? https://github.com/firebase/genkit/blob/main/js/ai/src/retriever.ts#L164-L168

js/plugins/pinecone/src/index.ts
```diff
@@ -185,7 +187,7 @@ export function configurePineconeIndexer<
   textKey?: string;
   embedder: EmbedderArgument<EmbedderCustomOptions>;
   embedderOptions?: z.infer<EmbedderCustomOptions>;
-}) {
+}): IndexerAction<z.ZodOptional<typeof PineconeIndexerOptionsSchema>> {
```
MichaelDoyle, line 190: This seemed a little funky to me. Not sure if this is 💯 .

It looks like we're possibly double-wrapping in an Optional? https://github.com/firebase/genkit/blob/main/js/ai/src/embedder.ts#L74-L78

MichaelDoyle:
@pavelgj friendly ping on this one

js/plugins/chroma/src/index.ts
ssbushi: When I first made these abstractions in `js/ai`, I wanted to make as few assumptions as possible. That meant that the options should not be `ZodOptional` since we do not know what kind of retriever they are building and what options are required vs optional. 

IMO, the `/ai` abstractions should not add a `.optional()` on the schema and the providers can add them in the custom schema they define (`ChromaRetrieverOptionsSchema` in this case). 

That being said, this code has through many changes and its evident that it no longer holds to that thought. The problem here does not seem to be double wrapping though -- there is no Optional on the `ChromaRetrieverOptionsSchema` object and it is added by `RetrieverAction`.

PR #406, opened 2024-06-13
Remove duplicate paths.

kmandrika:
This come into play when a span has several subspans.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/flow/src/telemetry.ts
+++ b/js/flow/src/telemetry.ts
@@ -79,13 +79,15 @@ export function writeFlowSuccess(flowName: string, latencyMs: number) {
 
   const paths = traceMetadataAls.getStore()?.paths || new Set<PathMetadata>();
   if (paths) {
-    const relevantPaths = Array.from(paths).filter((meta) =>
-      meta.path.includes(flowName)
+    const relevantPaths = new Map(
+      Array.from(paths)
+        .filter((meta) => meta.path.includes(flowName))
+        .map((path) => [path.path, path])
     );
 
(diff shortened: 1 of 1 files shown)


js/flow/src/telemetry.ts
```diff
@@ -79,13 +79,15 @@ export function writeFlowSuccess(flowName: string, latencyMs: number) {
...
+        .map((path) => [path.path, path])
     );
 
     logger.logStructured(`Paths[/${flowName}]`, {
       flowName: flowName,
-      paths: relevantPaths.map((p) => p.path),
+      paths: [...relevantPaths.keys()],
     });
 
     relevantPaths.forEach((p) => {
```
andrewbrook, line 93: I don't think this takes into account the new data structure

andrewbrook (changes requested):
Add unit tests?

js/flow/src/telemetry.ts
kmandrika: Could you say more here, not sure what you mean

andrewbrook: Ah, I didn't expect that first arg of forEach was the entry value. So I guess this is fine.

kmandrika:
> Add unit tests?

On it!

kmandrika:
Thinking more about this & playing around with test cases, I don't think removing duplicate paths makes sense any more.

In cases (e.g. tool calls) where a parent span can have multiple subspans (some of which have the same path), the traversal of those paths should be counted because it happened.

Going to close this for now.

PR #420, opened 2024-06-14
Go docs: models.md and flows.md

kevinthecheung:


docs-go/models.md
```diff
@@ -0,0 +1,263 @@
...
+    Tools: []*ai.ToolDefinition{myJoke},
+  }
+  response, err := ai.Generate(context.Background(), gemini15pro, &request, nil)
+  ```
+
+This will automatically call the tools in order to fulfill the user prompt.
+
+<!-- TODO: returnToolRequests: true` -->
+
+### Adding retriever context
```
kevinthecheung, line 195: Does the Go library have this? I couldn't find the context field being used anywhere

nohe427 (approved):
LGTM

nohe427 (approved):

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+  menuSuggestionFlow := genkit.DefineFlow(
+    "menuSuggestionFlow",
+    func(ctx context.Context, restaurantTheme string, _ genkit.NoStream) (string, error) {
+      suggestion := makeMenuItemSuggestion(restaurantTheme)
+      return suggestion, nil
+    },
+  )
+  ```
+
+You can define input and output schemas for a flow, which is useful when you
```
ianlancetaylor, line 24: This suggests that people will in fact define input and output schemas for a flow, but that's not what they do.  What they do is write a flow that accepts or returns a structured input value.  Maybe something like (assuming this is accurate):

The argument and result types of a flow may be structured values. Genkit will produce JSON schemas for these values using invopop/jsonschema. Those JSON schemas will be used by the Genkit CLI and UI.

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+    func(
+      ctx context.Context,
+      restaurantTheme inputType,
+      callback func(context.Context, streamType) error,
+    ) (outputType, error) {
+      var menu outputType = ""
+      menuChunks := make(chan streamType)
+      go makeFullMenuSuggestion(restaurantTheme, menuChunks)
+      for {
+        chunk, more := <-menuChunks
```
ianlancetaylor, line 87: In Go we typically call the second variable here `ok`, meaning that the value (`chunk` in this case) is valid.  The name `more` hints that there are more values to come, which isn't quite right.

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+  type streamType string
+
+  menuSuggestionFlow := genkit.DefineFlow(
+    "menuSuggestionFlow",
+    func(
+      ctx context.Context,
+      restaurantTheme inputType,
+      callback func(context.Context, streamType) error,
+    ) (outputType, error) {
+      var menu outputType = ""
```
ianlancetaylor, line 83: May as well model good Go code here by writing

```Go
    var menu strings.Builder
    ...
        menu.WriteString(string(chunk))
    ...
    return outputType(menu.String())
```

Alternatively

```Go
    var menuVals []string
    ...
        menuVals = append(menuVals, string(chunk))
    ...
    return outputType(strings.Join(menuVals, ""))
```

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
+# Flows
+
+Flows are functions with some additional characteristics: they are strongly
+typed, streamable, locally and remotely callable, and fully observable.
+Firebase Genkit provides CLI and developer UI tooling for working with flows
+(running, debugging).
+
+## Defining flows
+
+A flow wraps a function:
```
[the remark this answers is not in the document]
apascal07: In JS, this is correct: they are both a callable function and they wrap a function that does user-defined work. In Go, this is not strictly true. It's a struct that contains a function. But the concepts are the same so I wonder if we should stick with "flows = functions" because that's what we aim them to be, but because Go isn't as flexible as JS, to achieve the side effects of tracing, etc, we need to store them as structs.

apascal07 (commented):
@kevinthecheung Please be aware of slight change here: https://github.com/firebase/genkit/pull/428

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+
+- {Go}
+
+  ```go
+  menuSuggestionFlow := genkit.DefineFlow(
+    "menuSuggestionFlow",
+    func(ctx context.Context, restaurantTheme string, _ genkit.NoStream) (string, error) {
```
pavelgj, line 17:
```suggestion
    func(ctx context.Context, restaurantTheme string) (string, error) {
```

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+    ItemName    string `json:"item_name"`
+    Description string `json:"description"`
+    Calories    int    `json:"calories"`
+  }
+  ```
+
+  ```go
+  menuSuggestionFlow := genkit.DefineFlow(
+    "menuSuggestionFlow",
+    func(ctx context.Context, restaurantTheme string, _ genkit.NoStream) (MenuSuggestion, error) {
```
pavelgj, line 43:
```suggestion
    func(ctx context.Context, restaurantTheme string) (MenuSuggestion, error) {
```

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+
+- {Go}
+
+  ```go
+  // Types for illustrative purposes.
+  type inputType string
+  type outputType string
+  type streamType string
+
+  menuSuggestionFlow := genkit.DefineFlow(
```
pavelgj, line 76:
```suggestion
  menuSuggestionFlow := genkit.DefineStreamingFlow(
```

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+- {Go}
+
+  To deploy flows using Cloud Run and similar services, define your flows, and
+  then call `StartFlowServer()`:
+
+  ```go
+  func main() {
+    genkit.DefineFlow(
+      "menuSuggestionFlow",
+      func(ctx context.Context, restaurantTheme string, _ genkit.NoStream) (string, error) {
```
pavelgj, line 146:
```suggestion
      func(ctx context.Context, restaurantTheme string) (string, error) {
```

docs-go/flows.md
```diff
@@ -0,0 +1,187 @@
...
+## Flow observability
+
+Sometimes when using 3rd party SDKs that are not instrumented for observability,
+you might want to see them as a separate trace step in the Developer UI. All you
+need to do is wrap the code in the `run` function.
+
+```go
+genkit.DefineFlow(
+  "menuSuggestionFlow",
+  func(ctx context.Context, restaurantTheme string, _ genkit.NoStream) (string, error) {
```
pavelgj, line 179:
```suggestion
  func(ctx context.Context, restaurantTheme string) (string, error) {
```

docs-go/flows.md
```diff
@@ -1,33 +1,40 @@
 # Flows
 
-Flows are functions with some additional characteristics: they are strongly
-typed, streamable, locally and remotely callable, and fully observable.
-Firebase Genkit provides CLI and developer UI tooling for working with flows
-(running, debugging).
+Flows are runnable objects with some additional characteristics over simple
```
apascal07, line 3: I believe the Go folks have previously instructed me to avoid using the term "object" in Go contexts. I'm not sure what a good replacement is here though. @ianlancetaylor

apascal07: Food for thought: `Flows are wrapped functions with some additional characteristics over direct calls: ...`

In the future, we can supplement that list with "interruptable/resumable" or similar.

ianlancetaylor: "wrapped functions" SGTM

jba:
Docs! Awesome!

Can we have the code snippets inserted from actual runnable code, like samples? Otherwise they will get obsolete quickly, especially in this phase of relatively rapid change.

We have [a simple tool to do that](https://go.googlesource.com/example/+/refs/heads/master/internal/cmd/weave/weave.go) lying around.

pavelgj (approved):

jba (approved):
Let's land this. We can tweak incrementally.
I will own weaving the text with real code.

PR #421, opened 2024-06-15
Adds context and history support to Dotprompt. Fixes #418

### LINKED ISSUE #418: [Dotprompt] Support `history` and `context` when passed in.
This was partially implemented but didn't actually work. A prompt should be able to receive both `context` and `history` and render them appropriately using specified sections or with a default solution.

mbleigh:
This adds a few new features to Dotprompt:

1. History is now supported in `myPrompt.generate()`. It is inserted just before the last user message generated by the prompt by default, or wherever the `{{history}}` helper is provided if it is provided.
2. The `{{section "output|context"}}` helper now works, allowing you to control where in the flow of your prompt output instructions or document context are injected into the prompt.


maxl0rd (commented):
This approach looks good to me. 
The build error seems unrelated.
Should documentation be updated?

js/ai/src/model.ts
```diff
@@ -109,6 +109,7 @@ export type Role = z.infer<typeof RoleSchema>;
 export const MessageSchema = z.object({
   role: RoleSchema,
   content: z.array(PartSchema),
+  metadata: z.record(z.unknown()).optional(),
```
pavelgj, line 112: parts already have metadata: https://github.com/firebase/genkit/blob/83398d6e9762e8c2b21fa4d1de5258229fce499a/js/ai/src/model.ts#L46
Do we want to keep both?

mbleigh: I think so, they are different altitudes and so provide opportunities for different types of transformation.

pavelgj (approved):

PR #423, opened 2024-06-16
[Go] cross-language test: googleai embedder

jba:
Add a test designed to be run in different languages.

This uses a cross-language test framework I'm developing,
github.com/jba/xltest.

At its heart is the file testdata/googleai-embedder.yaml, which defines
the test in a language-neutral way. The test refers to a json file
with the expected trace.

The code in tests/googleai_test.go reads the test file and runs
the test. The main challenge is making two traces comparable,
which requires canonicalizing their span IDs.

To make sure we are testing the complete logic of an embedder action,
we pass a mock HTTP client to the googleai plugin, so we control
the interaction with the service.

Capturing traces cleanly also required some changes to the tracing
system, to allow installing a trace store temporarily and then
removing it.


go/plugins/googleai/googleai.go
```diff
@@ -40,6 +40,8 @@ type Config struct {
 	// Embedding models to provide.
 	// If empty, a complete list will be obtained from the service.
 	Embedders []string
+
+	ClientOptions []option.ClientOption
```
ianlancetaylor, line 44: This needs docs.  In particular that this replaces the `APIKey` field.

go/tests/googleai_test.go
```diff
@@ -0,0 +1,176 @@
...
+	t.Run("embed", func(t *testing.T) {
+		test, err := xltest.ReadFile(filepath.Join("testdata", "googleai-embedder.yaml"))
+		if err != nil {
+			t.Fatal(err)
+		}
+		valfunc := validateEmbedder
+		if *update {
+			valfunc = updateEmbedder
+		}
+		test.Run(t, func(text string) (embedderResult, error) {
```
ianlancetaylor, line 60: This will be clearer if we pull the function literal into its own variable, so that the `valfunc` variable isn't hidden.

go/tests/googleai_test.go
```diff
@@ -0,0 +1,176 @@
...
+		cmpopts.IgnoreFields(tracing.Data{}, "TraceID", "StartTime", "EndTime"),
+		cmpopts.IgnoreFields(tracing.SpanData{}, "TraceID", "StartTime", "EndTime"),
+	}
+	if diff := gocmp.Diff(want.trace, got.trace, opts...); diff != "" {
+		return fmt.Errorf("traces: %s", diff)
+	}
+	return nil
+}
+
+// renameSpans changes the keys of td.Spans to s0, s1, ... in order of the span start time,
```
ianlancetaylor, line 112: Pedantically I think it's s000, s001, ....

jba: done

go/tests/googleai_test.go
jba: done

go/plugins/googleai/googleai.go
jba: Done.

go/tests/testdata/googleai-embedder-trace.json
```diff
@@ -0,0 +1,36 @@
+{
```
ssbushi, line 1: What is the source of this trace? This does not match with what is generated by the JS SDK.

jba: YAY! That is exactly why we need these.
The source is the Go SDK.
Please replace with what _is_ generated by JS so the Go can match it.
Also, please implement the similar test for JS so if the JS changes, your test will break too.

go/tests/googleai_test.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		m[newID] = span
+		// A span references it own span ID and possibly its parent's.
+		span.SpanID = oldIDToNew[span.SpanID]
+		if span.ParentSpanID != "" {
+			span.ParentSpanID = oldIDToNew[span.ParentSpanID]
+		}
+	}
+	td.Spans = m
+}
```
jba, line 142: @pavelgj This code makes it possible to compare two traces completely, except for times and the trace ID.

go/tests/testdata/googleai-embedder-trace.json
ssbushi: Updated the trace here -- this test now fails.

jba:
I'm going to close this in favor of using the existing JS test framework for the reflection API. See #549 and #551.

This test did reveal some incompatibilities. One was that the Go embedder design didn't match the JS. #537 fixed that. 

Others are that the display name doesn't see to be set in Go, and the "genkit:path" metadata values are very different. I don't know how important those are. 
@pavelgj

PR #427, opened 2024-06-18
fix: [Go] explicitlty listen on 127.0.0.1 to avoid firewall warning p…

pavelgj:


go/core/servers.go
```diff
@@ -70,7 +70,7 @@ func startDevServer(addr string) error {
 	// Don't use "localhost" here. That only binds the IPv4 address, and the genkit tool
```
randall77, line 70: This comment needs fixing.
So "localhost" doesn't work, but "127.0.0.1" does?
(maybe take a look at the PR that added this comment)

randall77 (approved):

go/core/servers.go
pavelgj: @apascal07 do you remember?

apascal07: I imagine it'll still only bind the IPv4 since it's an IPv4 address. Have you tried testing E2E?

pavelgj: you mean with `genkit start`? yes, works fine.

apascal07: If the dev UI works then I think the original comment can be removed.

pavelgj: removed

PR #429, opened 2024-06-18
feat: [Go] added middleware framework for actions and models

pavelgj:
Middleware allows providing reusable request/response pre/post processors. On the JS side we use middleware for polyfiling various model behaviours like system prompt and fetching media from URLs.

### DIFF
--- a/go/core/action.go
+++ b/go/core/action.go
@@ -215,7 +215,7 @@ func (a *Action[In, Out, Stream]) Run(ctx context.Context, input In, cb func(con
 			}
 			var output Out
 			if err == nil {
-				dispatch := ChainMiddleware(a.middleware...)
+				dispatch := chainMiddleware(a.middleware...)
 				output, err = dispatch(func(ctx context.Context, di In) (Out, error) {
 					return a.fn(ctx, di, cb)
 				})(ctx, input)
--- a/go/core/action.go
+++ b/go/core/action.go
@@ -28,6 +28,26 @@ import (
 	"github.com/invopop/jsonschema"
 )
 
+// Middleware is a functions that take in the request object, response object,
+// and the next middleware function for the action.
+//
+// Middleware functions can:
+//   - execute arbitrary code;
+//   - change the request and response;
+//   - terminate response by returning a response (or error);
+//   - call the next middleware function.
+//
+// If the current middleware function does not terminate the middlware chain it
+// must call next middlware in the chain to pass control to the next middleware
+// function, otherwise the request will be left hanging.
+type Middleware[I, O any] func(context.Context, I, func(ctx context.Context, input I) (O, error)) (O, error)
+
+// Middlewares returns an array of middlewares that are passes in as an argument.
+// `core.Middlewares(apple, banana)` is identical to `[]core.Middleware[InputType, OutputType]{apple, banana}`
+func Middlewares[I, O any](ms ...Middleware[I, O]) []Middleware[I, O] {
+	return ms
+}
+
 // Func is the type of function that Actions and Flows execute.
 // It takes an input of type Int and returns an output of type Out, optionally
 // streaming values of type Stream incrementally by invoking a callback.
@@ -63,6 +83,7 @@ type Action[In, Out, Stream any] struct {
 	// optional
 	description string
 	metadata    map[string]any
+	middleware  []Middleware[In, Out]
 }
 
 // See js/core/src/action.ts
@@ -78,18 +99,18 @@ func defineAction[In, Out any](r *registry, provider, name string, atype atype.A
 	return a
 }
 
-func DefineStreamingAction[In, Out, Stream any](provider, name string, atype atype.ActionType, metadata map[string]any, fn Func[In, Out, Stream]) *Action[In, Out, Stream] {
-	return defineStreamingAction(globalRegistry, provider, name, atype, metadata, fn)
+func DefineStreamingAction[In, Out, Stream any](provider, name string, atype atype.ActionType, metadata map[string]any, middleware []Middleware[In, Out], fn Func[In, Out, Stream]) *Action[In, Out, Stream] {
+	return defineStreamingAction(globalRegistry, provider, name, atype, metadata, middleware, fn)
 }
 
-func defineStreamingAction[In, Out, Stream any](r *registry, provider, name string, atype atype.ActionType, metadata map[string]any, fn Func[In, Out, Stream]) *Action[In, Out, Stream] {
-	a := newStreamingAction(name, atype, metadata, fn)
+func defineStreamingAction[In, Out, Stream any](r *registry, provider, name string, atype atype.ActionType, metadata map[string]any, middleware []Middleware[In, Out], fn Func[In, Out, Stream]) *Action[In, Out, Stream] {
+	a := newStreamingAction(name, atype, metadata, middleware, fn)
 	r.registerAction(provider, a)
 	return a
 }
 
 func DefineCustomAction[In, Out, Stream any](provider, name string, metadata map[string]any, fn Func[In, Out, Stream]) *Action[In, Out, Stream] {
-	return DefineStreamingAction(provider, name, atype.Custom, metadata, fn)
+	return DefineStreamingAction(provider, name, atype.Custom, metadata, nil, fn)
 }
 
 // DefineActionWithInputSchema creates a new Action and registers it.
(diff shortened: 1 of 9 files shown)


go/core/action.go
```diff
@@ -28,6 +28,26 @@ import (
...
+// Middleware is a functions that take in the request object, response object,
+// and the next middleware function for the action.
+//
+// Middleware functions can:
+//   - execute arbitrary code;
+//   - change the request and response;
+//   - terminate response by returning a response (or error);
+//   - call the next middleware function.
+//
+// If the current middleware function does not terminate the middlware chain it
```
ianlancetaylor, line 40: s/middlware/middleware/

go/core/action.go
```diff
@@ -28,6 +28,26 @@ import (
...
+// must call next middlware in the chain to pass control to the next middleware
+// function, otherwise the request will be left hanging.
+type Middleware[I, O any] func(context.Context, I, func(ctx context.Context, input I) (O, error)) (O, error)
+
+// Middlewares returns an array of middlewares that are passes in as an argument.
+// `core.Middlewares(apple, banana)` is identical to `[]core.Middleware[InputType, OutputType]{apple, banana}`
```
ianlancetaylor, line 46: Markdown syntax doesn't work in Go doc comments.

go/core/action.go
```diff
@@ -28,6 +28,26 @@ import (
...
+// Middleware functions can:
+//   - execute arbitrary code;
+//   - change the request and response;
+//   - terminate response by returning a response (or error);
+//   - call the next middleware function.
+//
+// If the current middleware function does not terminate the middlware chain it
+// must call next middlware in the chain to pass control to the next middleware
+// function, otherwise the request will be left hanging.
+type Middleware[I, O any] func(context.Context, I, func(ctx context.Context, input I) (O, error)) (O, error)
```
ianlancetaylor, line 43: Passing in the next function to call works, but I think more typical would be to return an error like `ErrMiddlewareDone`.  If the caller sees that error, it will return the output without continuing to call middleware or the model.

go/core/action.go
```diff
@@ -28,6 +28,26 @@ import (
 	"github.com/invopop/jsonschema"
 )
 
+// Middleware is a functions that take in the request object, response object,
```
ianlancetaylor, line 31: In general I think @jba 's goal was for the core package to be for plugins and the genkit and ai packages to be for users.  This makes me wonder who is going to write `Middleware` values: the user or plugins?

pavelgj: Middleware is a feature of the core framework at the action level and ai package will contain a couple of common model middlewares (see: https://github.com/firebase/genkit/blob/main/js/ai/src/model/middleware.ts).
There are ideas to potentially implement some retriever middleware as well (for advances RAG patterns).

go/core/action.go
pavelgj: Done.

go/core/action.go
pavelgj: Sad... removed.

go/core/action.go
ianlancetaylor: What I mean is: which code is going to pass the slice of middleware functions?  Will an application that uses Genkit be writing its own middleware functions?

go/core/action.go
ianlancetaylor: There is only one caller of the middleware functions as far as I can see, in `Action.Run`.  I meant that that function could loop through the middleware slice and call each function.  If one of them returned `ErrMiddlewareDone`, it would break out of the loop.

go/core/action.go
pavelgj: No, middleware functions are not meant to be written by the application/plugin developer (it's OK if they want to -- middleware is a powerful design pattern), but meant to be used by them. Here's an example of reusable middleware used to add common features to models: https://github.com/firebase/genkit/blob/main/js/plugins/vertexai/src/gemini.ts#L450-L464

go/core/action.go
pavelgj: You might be thinking of a different pattern. I refactored the interface a bit, following an existing Go middleware pattern implementation https://cs.opensource.google/go/x/pkgsite/+/d8c29b80:internal/middleware/middleware.go

ianlancetaylor: I think the main point I was missing is that this middleware can transform the output, so it needs to run both before and after the other code.  Thanks.

go/core/action.go
```diff
@@ -28,6 +28,38 @@ import (
...
+// some way.
+//
+// Middleware functions can:
+//   - execute arbitrary code;
+//   - change the request and response;
+//   - terminate response by returning a response (or error);
+//   - call the next middleware function.
+type Middleware[I, O any] func(MiddlewareHandler[I, O]) MiddlewareHandler[I, O]
+
+type MiddlewareHandler[I, O any] func(ctx context.Context, input I) (O, error)
```
ianlancetaylor, line 42: Does this need to be exported?  If it does, it needs a doc comment.

go/core/action.go
```diff
@@ -28,6 +28,38 @@ import (
...
+
+// Middlewares returns an array of middlewares that are passes in as an argument.
+// core.Middlewares(apple, banana) is identical to []core.Middleware[InputType, OutputType]{apple, banana}
```
ianlancetaylor, line 45: I'm not sure why people wouldn't just use a slice constructor here.  Does this function pull its weight?

go/core/action.go
```diff
@@ -28,6 +28,38 @@ import (
...
+func Middlewares[I, O any](ms ...Middleware[I, O]) []Middleware[I, O] {
+	return ms
+}
+
+// ChainMiddleware creates a new Middleware that applies a sequence of
+// Middlewares, so that they execute in the given order when handling action
+// request.
+// In other words, ChainMiddleware(m1, m2)(handler) = m1(m2(handler))
+func ChainMiddleware[I, O any](middlewares ...Middleware[I, O]) Middleware[I, O] {
```
ianlancetaylor, line 54: Is there a reason to export this?

pavelgj: Made private for now.

go/core/action.go
pavelgj: So, for example here's what it will look like when trying to pass in model middleware:
```go
[]core.Middleware[*ai.GenerateRequest, *ai.GenerateResponse]{m1, m2}
```

vs
```go
core.Middlewares(m1, m2)
```

I prefer the second one...

go/core/action.go
pavelgj: Added doc.

go/core/action.go
ianlancetaylor: Ah, no type inference on compound literals.  Fair point.

go/core/action.go
```diff
@@ -178,7 +215,10 @@ func (a *Action[In, Out, Stream]) Run(ctx context.Context, input In, cb func(con
 			}
 			var output Out
 			if err == nil {
-				output, err = a.fn(ctx, input, cb)
+				dispatch := chainMiddleware(a.middleware...)
```
jba, line 218: Maybe I'm missing something, but I don't think we need to pass down the middlewares as a separate thing.

Say we changed middleware handlers so they have the same type as the action function itself, by including the callback argument.

Then chaining the handlers followed by the action function doesn't have to happen here, it could happen higher up.
So instead of
```
DefineAction(..., Middlewares(h1, h2), f)
```
we could write
```
DefineAction(..., h1(h2(f)))
```
This effectively removes the machinery of middlewares from the implementation. A middleware handler is just a function from action-functions to action-functions. We can keep the `MiddlewareHandler` type and perhaps provide a couple of common ones, but I don't think we need the rest of it.

go/core/action.go
```diff
@@ -78,18 +109,18 @@ func defineAction[In, Out any](r *registry, provider, name string, atype atype.A
 	return a
 }
 
-func DefineStreamingAction[In, Out, Stream any](provider, name string, atype atype.ActionType, metadata map[string]any, fn Func[In, Out, Stream]) *Action[In, Out, Stream] {
-	return defineStreamingAction(globalRegistry, provider, name, atype, metadata, fn)
+func DefineStreamingAction[In, Out, Stream any](provider, name string, atype atype.ActionType, metadata map[string]any, middleware []Middleware[In, Out, Stream], fn Func[In, Out, Stream]) *Action[In, Out, Stream] {
+	return defineStreamingAction(globalRegistry, provider, name, atype, metadata, middleware, fn)
```
jba, line 113: I'm still unclear about why the middlewares are getting passed all the way down. I can just apply them to the action function at the top level. I don't think anything in action.go needs to change.

PR #435, opened 2024-06-20
Update action names to match JS SDK

ssbushi:
Using `provider/name` as the action name, instead of `name`. This matches the JS SDK.

Checklist (if applicable):
- [x] Tested (manually, unit tests)


jba:
I'd be surprised if this didn't break something somewhere.

We've had a few iterations on action and flow naming, so I'd like to put this PR on hold until we have cross-language tests that verify the traces. #423 is an example. Once both Go and JS implementations have tests that use the same golden traces, we can make changes like this with confidence, and without having to do any manual verification.

ssbushi:
@jba I agree. I'm all for more robust tests and I can hold this off till your PR is merged.

I left a comment on #423 to clarify the test.

ssbushi:
FYI -- @pavelgj  has already fixed this in #438. I'm closing out this PR.

PR #441, opened 2024-06-21
fix: exit from `nonDurableExpressHandler` when authPolicy check fails

pavelgj:


sam-gc (approved):

idangozlan:
I was just about to create a PR for that, haha. Nice! Hopefully a next version will be released soon. thanks @pavelgj .

PR #443, opened 2024-06-21
[plugins/firebase] defineFirestoreRetriever: Allow supplying collection at retrieval time

mbleigh:
Checklist (if applicable):
- [x] Docs updated

### DIFF
--- a/docs/plugins/firebase.md
+++ b/docs/plugins/firebase.md
@@ -123,9 +123,9 @@ const docs = await retrieve({
 
 Available retrieval options include:
 
-* `limit`: Specify the number of matching results to return.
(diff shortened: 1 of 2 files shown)


pavelgj:
Can you elaborate on the use-case? I'm trying to check whether this impacts tracing and playground (should be fine, I believe).

mbleigh:
Wrote this real quick after seeing https://github.com/firebase/genkit/discussions/394

If you use subcollections e.g. `users/{id}/messages` then you need to be able to supply the collection name at retrieval time because it's not fixed.

pavelgj (approved):

PR #450, opened 2024-06-22
feat: allow specifying middleware on the `generate` function

pavelgj:
The idea is to be able to do things like

```ts
await generate({
  prompt: prompt,
  model: model,
  use: [
    retry({ maxRetries: 2 }),
    smoothStreaming({ cadence: 25, randomize: true })
  ],
})


pavelgj:
Nope... this is bad. Should not be allowed!!!

PR #468, opened 2024-06-24
chore: make firebase template auth policy a comment to reduce friction

pavelgj:


genkit-tools/cli/config/firebase.index.ts.template
```diff
@@ -20,9 +20,10 @@ export const menuSuggestionFlow = onFlow(
     inputSchema: z.string(),
     outputSchema: z.string(),
     authPolicy: firebaseAuth((user) => {
-      if (!user.email_verified) {
-        throw new Error("Verified email required to run flow");
-      }
+      // TODO: insert authorization logic. For example:
```
apascal07, line 23: I'm split on this. 1) Auth is important, 2) it doesn't really make sense for this sample, 3) should either have it in place or omit the comment entirely relying on docs for auth info. It's a bit distracting because it's unclear whether it's necessary for the sample to work. And if it's not necessary, there are a bunch of optional things we could be calling out but we don't for the sake of simplicity.

pavelgj: cc @chrisraygill

sam-gc: The options are `firebaseAuth()` or `noAuth()`. A comment isn't ideal but it still at least guarantees that Firebase Auth is being used. `noAuth()` would mean no auth checks at all.

sam-gc: How about making the comment:
```js
// Firebase Auth is required to call this flow using the Firebase Functions SDK.
// TODO: Write additional logic tailored to the needs of your app.
```

pavelgj, line 23:
```suggestion
      // Firebase Auth is required to call this flow using the Firebase Functions SDK.
      // TODO: Write additional logic tailored to the needs of your app.
      // For example:
```

pavelgj: PTAL

apascal07 (approved):

genkit-tools/cli/config/firebase.index.ts.template
apascal07: LGTM.

PR #471, opened 2024-06-24
Update ollama js plugin to support system message

huangjeff5:
Ollama models support the "system" role. Ensure that it is supported in our plugin too.

### DIFF
--- a/js/plugins/ollama/src/index.ts
+++ b/js/plugins/ollama/src/index.ts
@@ -282,15 +282,15 @@ function readChunks(reader) {
 
 function getPrompt(input: GenerateRequest): string {
   return input.messages
-    .filter(m => m.role !== "system") // Exclude system messages
-    .map(m => m.content.map(c => c.text).join())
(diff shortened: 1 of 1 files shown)


js/plugins/ollama/src/index.ts
```diff
@@ -279,8 +281,18 @@ function readChunks(reader) {
...
+function getPrompt(input: GenerateRequest): string {
+  return input.messages
+    .filter((m) => m.role !== 'system') // Exclude system messages
+    .map((m) => m.content.map((c) => c.text).join())
+    .join();
+}
+
+function getSystemMessage(input: GenerateRequest): string {
+  return input.messages
+    .filter((m) => m.role === 'system') // Exclude system messages
```
MichaelDoyle, line 293: Can you just update the comment here

i.e.) `// Include only the system messages`

Or alternatively, just drop it entirely. :)

pavelgj (approved):

PR #472, opened 2024-06-25
Go docs: prompts and dotprompts

kevinthecheung:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


ianlancetaylor (approved):

docs-go/dotprompt.md
```diff
@@ -0,0 +1,317 @@
...
+  ```
+
+  You can call the prompt's `Generate` method to render the template and pass it
+  to the model API in one step:
+
+  ```go
+  // The .prompt file specifies vertexai/gemini-1.5-pro, so make sure it's set
+  // up:
+  projectID := os.Getenv("GCLOUD_PROJECT")
+  vertexai.Init(context.Background(), projectID, "us-central1")
```
ianlancetaylor, line 63: Better to write

    ctx := context.Background()

and then use `ctx` here and below rather than calling `context.Background` multiple times.

docs-go/dotprompt.md
```diff
@@ -0,0 +1,317 @@
...
+  prompt, err := dotprompt.Open("greeting")
+  ```
+
+  You can call the prompt's `Generate` method to render the template and pass it
+  to the model API in one step:
+
+  ```go
+  // The .prompt file specifies vertexai/gemini-1.5-pro, so make sure it's set
+  // up:
+  projectID := os.Getenv("GCLOUD_PROJECT")
```
ianlancetaylor, line 62: Use the new weave program for this code?

docs-go/dotprompt.md
```diff
@@ -0,0 +1,317 @@
...
+  	&dotprompt.PromptRequest{
+  		Variables: GreetingPromptInput{
+  			Location: "the beach",
+  			Style:    "a fancy pirate",
+  			Name:     "Ed",
+  		},
+  	},
+  	nil,
+  )
+
```
ianlancetaylor, line 82: Always a good idea to check errors in examples, even if you just call `log.Fatal`.

docs-go/dotprompt.md
```diff
@@ -0,0 +1,317 @@
...
+    name: string
+    email?: string
+  metadata?(object): # objects are also denoted via parentheses
+    updatedAt?: string, ISO timestamp of last update
+    approvedBy?: integer, id of approver
+  extra?: any, arbitrary extra data
+  (*): string, wildcard field
+```
+
+The above schema is equivalent to the following TypeScript interface:
```
ianlancetaylor, line 126: I'm not sure it's all that helpful for the Go docs to refer to a TypeScript interface.  Can we instead show the JSON version of the corresponding JSON schema?

docs-go/dotprompt.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+The URL can be `https://` or base64-encoded `data:` URIs for "inline" image
+usage. In code, this would be:
+
+- {Go}
+
+  ```go
+  dotprompt.SetDirectory("prompts")
+  describeImagePrompt, err := dotprompt.Open("describe_image")
```
ianlancetaylor, line 262: Check errors.

docs-go/dotprompt.md
```diff
@@ -0,0 +1,317 @@
...
+
+  imageBytes, err := os.ReadFile("img.jpg")
+  encodedImage := base64.StdEncoding.EncodeToString(imageBytes)
+  dataUri := fmt.Sprintf("data:image/jpeg;base64,%s", encodedImage)
```
ianlancetaylor, line 266: Really minor, but I would write

    dataURI := "data:image/jpeg;base64," + encodedImage

docs-go/dotprompt.md
```diff
@@ -0,0 +1,367 @@
...
+
+  ```go
+  describeImagePrompt, err := dotprompt.OpenVariant("describe_image", "gemini15")
+  ```
+
+The prompt loader will attempt to load the variant of that name, and fall back
+to the baseline if none exists. This means you can use conditional loading based
+on whatever criteria makes sense for your application:
+
+```ts
```
jba, line 357: Translate to Go.

jba (changes requested):

go/internal/doc-snippets/prompts.go
```diff
@@ -0,0 +1,95 @@
...
+				{Content: []*ai.Part{ai.NewTextPart(prompt)}},
+			}}, nil
+		},
+	)
+	//!-ex03.1
+
+	//!+ex03.2
+	request, err := helloPrompt.Render(context.Background(), HelloPromptInput{UserName: "Fred"})
+	if err != nil {
+		log.Fatal(err)
```
jba, line 89: change these to `return err`. That's more common.

jba (changes requested):

jba (approved):

PR #479, opened 2024-06-25
[Go] use a Makefile to generate doc-go files

jba:


docs-go/Makefile
```diff
@@ -0,0 +1,10 @@
+WEAVE=$(HOME)/go/bin/weave 
+
+all: $(WEAVE) flows.md models.md
```
ianlancetaylor, line 3: Shouldn't there be a change to a `go:generate` line somewhere?

jba: Are you suggesting that a go:generate should call this makefile?
I was thinking that this is something that the docs person runs manually.
But I guess it could be both; we could add a go:generate to the directory where the snippets live.

Done.

ianlancetaylor (approved):

PR #483, opened 2024-06-26
[Go] Ergonomic improvements to Go API.

apascal07:


go/plugins/vertexai/vertexai.go
```diff
@@ -70,7 +71,19 @@ func Init(ctx context.Context, projectID, location string) error {
 	if state.initted {
 		panic("vertexai.Init already called")
 	}
+	if projectID == "" {
+		projectID = os.Getenv("GCLOUD_PROJECT")
```
pavelgj, line 75: for the record/followup
```go
import 	"cloud.google.com/go/auth/credentials"


	creds, err := credentials.DetectDefault(&credentials.DetectOptions{})
	if err != nil {
		return err
	}
	projectID, err := creds.QuotaProjectID(ctx)
	if err != nil {
		return err
	}
```

pavelgj: I was expecting `creds.ProjectID(ctx)` to work... but doesn't for some reason.

apascal07: I looked at this, it only handles ADC: https://pkg.go.dev/cloud.google.com/go/auth/credentials#DetectDefault

There's no client library that does everything the Node.js google-auth-library does AFAICT.

go/ai/gen.go
```diff
@@ -203,3 +203,13 @@ type ToolResponse struct {
 	// An example might be map[string]any{"name":"Thomas Jefferson", "born":1743}.
 	Output map[string]any `json:"output,omitempty"`
 }
+
+// NewTextMessages is a helper function that creates a slice of [Message] from one role and string.
+func NewTextMessages(r Role, m string) []*Message {
```
pavelgj, line 208: wee can iterate on this, but "New Text Messages" from single string reads a bit weird...

go/plugins/googleai/googleai.go
```diff
@@ -85,6 +85,9 @@ func Init(ctx context.Context, apiKey string) (err error) {
 	}
 	state.client = client
 	state.initted = true
+	for model, caps := range knownCaps {
+		DefineModel(model, &caps)
```
pavelgj, line 89: needs error handling

go/plugins/vertexai/vertexai.go
```diff
@@ -89,6 +102,9 @@ func Init(ctx context.Context, projectID, location string) error {
 	if err != nil {
 		return err
 	}
+	for model, caps := range knownCaps {
+		DefineModel(model, &caps)
```
pavelgj, line 106: needs error handling

go/ai/gen.go
apascal07: I agree, I'm open to ideas because I haven't been able to come up with anything better. :-)

pavelgj (approved):

go/plugins/vertexai/vertexai.go
pavelgj: we also only do ADC in nodejs

go/ai/gen.go
apascal07: Alternatively: `NewTextMessageAsSlice()`

pavelgj (approved):

go/ai/gen.go
pavelgj: It's fine, we can iterate later.

genkit-tools/cli/src/commands/init/init-go.ts
```diff
@@ -36,48 +36,30 @@ interface ModelOption {
...
-		if err != nil {
-			log.Fatal(err)
-		}
-	}`;
-
-const ollamaInitTmpl = `if err := ollama.Init(context.Background(), ollama.Config{ServerAddress: "http://127.0.0.1:11434", Models: []ollama.ModelDefinition{{Name: "gemma"}}}); err != nil {
-		log.Fatal(err)
-	}`;
+/** Ollama init call template. Keep indenting to for expected output. */
+const ollamaInit = `ollama.Init(context.Background(), ollama.Config{
```
jba, line 40: Do all these end up in `main`? If so, let's start main with 
```
ctx := context.Background()
```
and use `ctx` throughout.

jba (changes requested):

go/ai/gen.go
```diff
@@ -203,3 +203,13 @@ type ToolResponse struct {
 	// An example might be map[string]any{"name":"Thomas Jefferson", "born":1743}.
 	Output map[string]any `json:"output,omitempty"`
 }
+
+// NewTextMessages is a helper function that creates a slice of [Message] from one role and string.
```
jba, line 207: I think this function is too specific. I'd rather see something like
```
func NewUserGenerateRequest(parts ...*ai.Part)
```

go/plugins/googleai/googleai.go
```diff
@@ -85,6 +85,12 @@ func Init(ctx context.Context, apiKey string) (err error) {
 	}
 	state.client = client
 	state.initted = true
+	for model, caps := range knownCaps {
+		_, err := DefineModel(model, &caps)
```
jba, line 89: if _, err := XXX; err != nil {...}

go/plugins/vertexai/vertexai.go
```diff
@@ -89,6 +102,12 @@ func Init(ctx context.Context, projectID, location string) error {
 	if err != nil {
 		return err
 	}
+	for model, caps := range knownCaps {
+		_, err := DefineModel(model, &caps)
```
jba, line 106: ditto

go/plugins/vertexai/vertexai.go
```diff
@@ -70,7 +71,19 @@ func Init(ctx context.Context, projectID, location string) error {
 	if state.initted {
```
jba, line 71: Update func comment three lines up from here to clarify when you need to call DefineModel.

Also for consistency shouldn't we define all known embedding models?

jba (changes requested):

go/ai/gen.go
apascal07: None of us is exactly happy with it but the only way it's useful is if it saves the full 7-8 LOC for the simplest case, otherwise I'm not sure it's worth it. Let's discuss post-bug bash once folks have a chance to start with and extend the sample.

go/plugins/vertexai/vertexai.go
```diff
@@ -50,6 +51,8 @@ var (
 		"gemini-1.5-pro":   multimodal,
 		"gemini-1.5-flash": multimodal,
 	}
+
+	knownEmbedders = []string{"text-embedding-004", "multimodalembedding"}
```
pavelgj, line 55: Current list on the JS side:
```
  'textembedding-gecko@003'
  'textembedding-gecko@002'
  'textembedding-gecko@001'
  'text-embedding-004'
  'textembedding-gecko-multilingual@001'
  'text-multilingual-embedding-002'
```

PR #484, opened 2024-06-26
feat(js): add support for streaming json output

cabljac:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/ai/src/extract.ts
+++ b/js/ai/src/extract.ts
@@ -15,7 +15,10 @@
  */
 
 import JSON5 from 'json5';
-import untruncateJson from 'untruncate-json';
+import { Allow, parse } from 'partial-json';
+
+const parsePartialJson = (jsonString: string) =>
+  JSON.stringify(parse(jsonString, Allow.ALL));
 
 /**
  * Extracts JSON from string with lenient parsing rules to improve likelihood of successful extraction.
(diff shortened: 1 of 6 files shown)


js/ai/src/generate.ts
```diff
@@ -399,6 +406,18 @@ export class GenerateResponseChunk<T = unknown>
...
+  /**
+   * Attempts to extract the longest valid JSON substring from the accumulated chunks.
+   * @returns The longest valid JSON substring found in the accumulated chunks.
+   */
+  output(): T | null {
+    if (!this.accumulatedChunks) return null;
+    const accumulatedText = this.accumulatedChunks
+      .map((chunk) => chunk.content.map((part) => part.text || '').join(''))
+      .join('');
+    return extractJson(untruncateJson(accumulatedText));
```
cabljac, line 418: I think this won't work exactly, as the text probably comes back with a ```json prefix

js/ai/src/extract.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import JSON5 from 'json5';
+import untruncateJson from 'untruncate-json';
```
pavelgj, line 18: We have a sample here using the `partial-json` package:
https://github.com/firebase/genkit/blob/1f6534b2a586006317699345935ca8268ddc6a1d/samples/js-angular/server/src/jsonStreaming.ts#L20C31-L20C43

Any obvious differences between those two?

pavelgj:
I like this approach.

js/ai/src/extract.ts
```diff
@@ -56,3 +60,46 @@ export function extractJson<T = unknown>(text: string): T | null {
   }
   throw new Error(`No JSON object or array found in model output: ${text}`);
 }
+
+export function extractAndUntruncateJson<T = unknown>(text: string): T | null {
```
pavelgj, line 64: I mean, I totally understand understand this by reading the code, just asking for a friend, can you explain a bit what this function does any why it's necessary? maybe a comment (with examples) would be helpful...

cabljac: I was mostly following the logic of the existing `extractJson` function above it, but maybe I should have just added an optional parseFn argument to `extractJson` instead of repeating it haha

But I can also add comments no problem

cabljac: I've updated the PR so that i'm just modifying the existing method, i've added some comments too, wdyt?

pavelgj (approved):

PR #487, opened 2024-06-26
Requiring manual Firebase app initialization to use admin sdk features

maxl0rd:
- Removing opportunistic, automatic application initialization where the admin SDK is used in the Firebase plugin.
- Customer code will _always_ be required to call initializeApp() before configuring features that require the admin SDK.
- Updating test apps that use flow auth or functions

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


MichaelDoyle:
Friendly ping @sam-gc

maxl0rd:
This is a breaking change, so we discussed some alternatives to this proposal.

An alternative approach we came up with was to allow passing an app object into the initialization of the Firebase plugin, so that the developer can manage initialization. By making this optional, we'd avoid any breaking api or behavior changes.

sam-gc:
Sorry for missing this

PR #491, opened 2024-06-26
Add a status label to metrics so that they can be filtered by success state

bryanatkinson:
Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/flow/src/telemetry.ts
+++ b/js/flow/src/telemetry.ts
@@ -163,6 +163,13 @@ export function writeFlowFailure(
       error: err.name,
       path: failPath,
     });
+
+    pathLatencies.record(latencyMs, {
+      ...pathDimensions,
+      status: 'failure',
+      error: err.name,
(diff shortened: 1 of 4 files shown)


js/flow/src/telemetry.ts
```diff
@@ -144,19 +146,20 @@ export function writeFlowFailure(
...
       });
 
       pathLatencies.record(p.latency, {
         ...pathDimensions,
+        status: 'success',
         path: p.path,
       });
     });
 
     pathCounter.add(1, {
```
andrewbrook, line 160: Should we be tracking failed path latency here?

andrewbrook (approved):

kmandrika (approved):

js/flow/src/telemetry.ts
bryanatkinson: Added.

PR #501, opened 2024-06-27
[Go] Implemented more robust server startup/shutdown.

### LINKED ISSUE #395: [Go] `genkit start` does not properly shut down previous runner process before starting new one
**Describe the bug**
```
Detected a change in main.go. Waiting for other changes before reloading.
Reloading app code.
App process exited with code null
Starting app at `.`...
Warn: Port 3100 not available, using 3101 instead.
time=2024-06-12T10:17:19.460-04:00 level=INFO msg="starting dev server"
time=2024-06-12T10:17:19.461-04:00 level=INFO msg="starting flow server"
time=2024-06-12T10:17:19.461-04:00 level=INFO msg=listening addr=:3400
time=2024-06-12T10:17:19.461-04:00 level=INFO msg=listening addr=:3101
time=2024-06-12T10:17:19.461-04:00 level=INFO msg="listen tcp :3400: bind: address already in use"
exit status 1
App process exited with code 1
```

**To Reproduce**
reproducible on `coffee-shop`

apascal07:
Fixes #395.

This change spawns app code on a detached process group so that any subsequent sub-processes are always `kill()`ed from the runner (e.g. `go run` spawns another process).

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated


jba (approved):

go/core/servers.go
```diff
@@ -43,36 +44,149 @@ import (
...
+	if opts.FlowAddr != "-" {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			s, err := startFlowServer(opts.FlowAddr, opts.Flows, errCh)
+			if err != nil {
+				errCh <- fmt.Errorf("failed to start flow server: %w", err)
+				return
+			}
+			servers = append(servers, s)
```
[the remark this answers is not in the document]
apascal07: This reminded me that one of the error handlers won't ever do anything because we sent a permanent `nil` from `startServer`. Errors will only be sent on the channel now. PTAL.

genkit-tools/common/src/runner/runner.ts
```diff
@@ -258,7 +269,16 @@ export class Runner {
           this.appProcess = null;
           resolve();
         });
-        this.appProcess.kill();
+        if (process.platform === 'win32') {
```
pavelgj, line 272: have you looked at https://www.npmjs.com/package/terminate?

apascal07: No, good suggestion, will try this!

go/genkit/genkit.go
```diff
@@ -93,18 +93,6 @@ func Run[Out any](ctx context.Context, name string, f func() (Out, error)) (Out,
...
 
-// StartFlowServer starts a server serving the routes described in [NewFlowServeMux].
-// It listens on addr, or if empty, the value of the PORT environment variable,
-// or if that is empty, ":3400".
-//
-// In development mode (if the environment variable GENKIT_ENV=dev), it also starts
-// a dev server.
-//
-// StartFlowServer always returns a non-nil error, the one returned by http.ListenAndServe.
-func StartFlowServer(addr string, flows []string) error {
```
pavelgj, line 104: docs need updating

docs-go/flows.md
```diff
@@ -148,7 +148,7 @@ first.
 - {Go}
 
   To deploy flows using Cloud Run and similar services, define your flows, and
-  then call `StartFlowServer()`:
+  then call `Init()`:
```
kevinthecheung, line 151: why this change? i thought StartFlowServer was clearer but maybe it's not accurate. Init() sounds like you always need to call it -- do you?

apascal07: `Init` is not just the equivalent of `StartFlowServer`. It also starts the Reflection API server (previously a side effect of `StartFlowServer` which is not technically correct, it freezes the registry, and in the future will house all other Genkit initialization work that we might want to do.

So yes, you always need to call it. The equivalent in JS is `configureGenkit()` which must always be called and the flow server and Reflection API server will soon be started from there implicitly as well.

pavelgj (approved):

go/core/servers.go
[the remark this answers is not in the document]
jba: Sorry, I forgot about errgroup: https://pkg.go.dev/golang.org/x/sync@v0.7.0/errgroup
That will deal with errors and context cancellation.
If you want to move on, I'll approve this PR and use errgroup in my own PR.

jba (approved):

go/core/servers.go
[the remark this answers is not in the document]
apascal07: I gave it a shot refactoring with errgroup. The main err handling/waiting is cleaner, but it doesn't work without significant refactor of how `Server` is created/returned to be kept track of for shutdown since `ListenAndServe()` is a blocking function. I'll keep it as is for now and we can revisit in the future in needed.

PR #503, opened 2024-06-27
Adds `defineSchema` to support registered schema use in Dotprompt.

### LINKED ISSUE #417: [Dotprompt] Schema registry
This feature would actually be usable throughout Genkit but would be most useful for Dotprompt. If Genkit supported a schema registry, it would be much easier to share schema definitions between text prompt files and code.

```ts
export const MySchema = defineSchema("MySchema", z.object({
  field1: z.string(),
  field2: z.string().optional(),
}));
```

Once a schema has been registered, it can be used inside a Dotprompt file:

```handlebars
---
model: vertexai/gemini-1.5-flash-preview
input:
  schema:
    name: string
output:
  schema: MySchema
---

Hello, {{name}}! This prompt outputs MySchema.
```

Then when using the prompt, we can pass the schema to a generic argument:

```ts
import { MySchema } from "./schemas";

const myPrompt = prompt<{name: string}, typeof MySchema>("myPrompt");
```

This would be a big win for prompts that end up with complex output schemas that then have to essentially be defined twice.

mbleigh:
This adds a new `defineSchema` method exported from `core` and corresponding support in Dotprompt, allowing schema reuse between code and prompts. Fixes #417

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated


js/core/src/registry.ts
```diff
@@ -53,6 +55,15 @@ function pluginsByName(): Record<string, PluginProvider> {
...
 }
+function schemasByName(): Record<
+  string,
+  { schema?: z.ZodTypeAny; jsonSchema?: JSONSchema }
+> {
+  if (global[SCHEMAS_BY_NAME] === undefined) {
+    global[SCHEMAS_BY_NAME] = {};
+  }
+  return global[SCHEMAS_BY_NAME];
+}
```
pavelgj, line 66: Do we ever want/need to access the schema from the dev ui (to display/inspect)?  or is the idea that the schema refs are always going to be fully resolved in dotprompts, etc. before the action is exposed to the dev ui?

mbleigh: Good question, I _think_ they'll be resolved in the prompt loading process but there's probably some confirming of that needs to happen.

pavelgj (approved):

js/core/src/registry.ts
pavelgj: Looking at the unit test it looks like it gets resolved... should be fine I think.

PR #506, opened 2024-06-27
Check auth on streamFlow() calls

sam-gc:
Still need to figure out longer term how to enforce flows with auth policies....


i2amsam (approved):
For long term robustness it seems like maybe runflow should be special case of streamFlow, or they should share a preamble setup processes, but this looks like a good immediate fix

PR #523, opened 2024-07-02
[Go] improve API

jba:
This PR consists solely of refactoring. There are no behavior changes.

Address the following warts on the API:

- Although flows are a key concept in Genkit, and the genkit package
  includes a DefineFlow function, the Flow type itself lived in the core
  package.

- The core package contained several exported methods that were intended
  only to be called by the genkit package, a clear sign that something
  is off in the API design.

The actual visible changes to achieve this are small, but a lot of
code was moved. Most of the changes involve moving unexported symbols
into internal packages so they could be used by other packages.

The main API changes are:

- Flow is now in the genkit package.

- There are no more InternalXXX symbols in core.

There are some other minor changes, like the removal of the NoStream
type.


go/internal/registry/registry.go
```diff
@@ -35,44 +35,46 @@ import (
 
 // The global registry, used in non-test code.
 // A test may create their own registries to avoid conflicting with other tests.
-var globalRegistry *registry
+var Global *Registry
```
apascal07, line 38: Given this is a var, does it make sense to keep it unexported and export a `GlobalReg()` function instead so that the global registry can't be overwritten/nilled out?

apascal07 (commented):
Looks good overall!

go/internal/registry/registry.go
jba: Good question.

I would do that if it was user-visible. But it's in an internal package, meaning only code in or under the `go` directory can use it. It's part of the culture to allow this level of "unsafety"; we wouldn't do that in Java but in Go I feel it's okay. It is a judgment call though. Maybe Ian or Keith will disagree.

randall77: Exported variables in internal packages seems fine.
The only thing I worry about is if there is a protocol involved in accessing it, in which case it should be unexported and have accessor functions. Typically that would be if we need to hold a lock around the access, or use atomic read/write, or whatever.
If it is ok to just read this variable whenever (which I think it is, given it is set in init?), then exported is fine.

go/ai/generate.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/firebase/genkit/go/core"
 	"github.com/firebase/genkit/go/core/logger"
+	"github.com/firebase/genkit/go/internal"
```
randall77, line 28: internal is kind of a weird package name, usually it is internal/something. Perhaps internal/schema?
Maybe ok though?

go/genkit/flow.go
```diff
@@ -81,48 +85,70 @@ import (
...
+// DefineStreamingFlow creates a streaming Flow that runs fn, and registers it as an action.
+//
+// fn takes an input of type In and returns an output of type Out, optionally
+// streaming values of type Stream incrementally by invoking a callback.
+//
+// If the function supports streaming and the callback is non-nil, it should
+// stream the results by invoking the callback periodically, ultimately returning
+// with a final return value. Otherwise, it should ignore the callback and
+// just return a result.
+func DefineStreamingFlow[In, Out, Stream any](
```
randall77, line 131: Might want to mention that we expect all the data to be available from the final result. Streaming results are just a preview; the final result object contains (at least) all the streaming responses.

go/genkit/flow.go
```diff
@@ -285,28 +303,22 @@ func (f *Flow[In, Out, Stream]) runInstruction(ctx context.Context, inst *flowIn
...
 
-// flow is the type that all Flow[In, Out, Stream] have in common.
-type flow interface {
-	Name() string
-
-	// runJSON uses encoding/json to unmarshal the input,
-	// calls Flow.start, then returns the marshaled result.
-	runJSON(ctx context.Context, input json.RawMessage, cb streamingCallback[json.RawMessage]) (json.RawMessage, error)
-}
+// Flow[I, O, S] implements the flow interface.
```
randall77, line 306: Should this comment live up at the definition of Flow?
Or do you mean here "The following methods ensure that ... "

go/genkit/servers.go
```diff
@@ -41,84 +41,22 @@ import (
 
 	"github.com/firebase/genkit/go/core/logger"
 	"github.com/firebase/genkit/go/core/tracing"
+	"github.com/firebase/genkit/go/internal"
+	"github.com/firebase/genkit/go/internal/common"
+	"github.com/firebase/genkit/go/internal/registry"
 	"go.opentelemetry.io/otel/trace"
 )
 
 // InternalInit is for use by the genkit package only.
```
randall77, line 50: This comment can be deleted.

randall77 (approved):

go/ai/generate.go
jba: How about internal/base, like cmd/compile/internal/base?

go/genkit/servers.go
jba: Done.

go/genkit/flow.go
jba: The latter. Done.

go/genkit/flow.go
jba: Done.

go/ai/generate.go
jba: will do in a followup.

randall77: That seems fine.

PR #526, opened 2024-07-02
rename internal package to base

jba:
The symbols defined directly in the `internal` directory are
now in the `base` package under `internal`.


go/core/action.go
```diff
@@ -22,8 +22,8 @@ import (
 
 	"github.com/firebase/genkit/go/core/logger"
 	"github.com/firebase/genkit/go/core/tracing"
-	"github.com/firebase/genkit/go/internal"
 	"github.com/firebase/genkit/go/internal/atype"
+	"github.com/firebase/genkit/go/internal/base"
```
randall77, line 26: "base" and "common" seem equally catch-all. Could we just use a single package?

randall77 (approved):

go/core/action.go
jba: Unfortunately core/tracing imports internal/base, and common imports core/tracing.
Maybe we can come up with a better name for common.

PR #528, opened 2024-07-03
[Go] test picoschema with yaml test cases

jba:
Use the yaml file of test cases to test the Go picoschema implementation.

This uncovered a few problems. Some new features were implemented in
JS but not in Go; they are implemented here.

More seriously, it was discovered that the JSON schema package we
are using, github.com/invopop/jsonschema, does not support
arrays for the "type" field, and probably never will (see
https://github.com/invopop/jsonschema/issues/134). Tests that
require that are skipped.

### DIFF
--- a/go/plugins/dotprompt/picoschema_test.go
+++ b/go/plugins/dotprompt/picoschema_test.go
@@ -15,158 +15,87 @@
 package dotprompt
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/invopop/jsonschema"
 	"gopkg.in/yaml.v3"
 )
 
 // TestPicoschema tests the same cases as picoschema_test.ts.
 func TestPicoschema(t *testing.T) {
-	tests := []struct {
-		description string
-		yaml        string
-		want        string
-	}{
-		{
-			description: "simple scalar, no description",
-			yaml:        `schema: string`,
-			want:        `{ "type": "string" }`,
-		},
-		{
-			description: "simple scalar, with description",
-			yaml:        `schema: number, the description`,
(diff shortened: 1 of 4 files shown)


go/plugins/dotprompt/picoschema_test.go
```diff
@@ -15,158 +15,87 @@
...
-    foo: {type: string}`,
-			want: `{ "type": "object", "properties": { "foo": { "type": "string" } } }`,
-		},
+	type test struct {
+		Description string
+		YAML        string
+		Want        map[string]any
+	}
+
+	convertSchema(&jsonschema.Schema{})
```
ianlancetaylor, line 35: Why call `convertSchema` here and ignore the result?

go/plugins/dotprompt/picoschema_test.go
```diff
@@ -15,158 +15,87 @@
...
+				t.Skip("no support for type as an array")
+			}
+			var val any
+			if err := yaml.Unmarshal([]byte(test.YAML), &val); err != nil {
+				t.Fatalf("YAML unmarshal failure: %v", err)
+			}
 
-		// The tests, copied from TypeScript, use a schema field.
-		val = val.(map[string]any)["schema"]
+			// The tests, copied from TypeScript, use a schema field.
```
ianlancetaylor, line 61: You can drop "copied from TypeScript" now.

ianlancetaylor (approved):
Thanks.

go/plugins/dotprompt/picoschema_test.go
jba: Done.

go/plugins/dotprompt/picoschema_test.go
jba: Oops, leftover debugging line.

PR #533, opened 2024-07-03
Move logging to ensure correct context for span

andrewbrook:
Move logging for generate action to ensure correct context for span

Checklist (if applicable):

### DIFF
--- a/js/ai/src/model.ts
+++ b/js/ai/src/model.ts
@@ -304,12 +304,15 @@ export function defineModel<
     },
     (input) => {
       const startTimeMs = performance.now();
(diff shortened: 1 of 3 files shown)


js/ai/src/model.ts
```diff
@@ -304,12 +304,15 @@ export function defineModel<
     },
     (input) => {
       const startTimeMs = performance.now();
+      telemetry.recordGenerateActionInputLogs(options.name, input);
```
kmandrika, line 307: We should swap these two lines

kmandrika (approved):

PR #540, opened 2024-07-04
[Go] plugins/googleai: fix locking

jba:
Ensure that the global state is protected during
Init, DefineModel and DefineEmbedder.


go/plugins/googleai/googleai.go
```diff
@@ -140,7 +138,6 @@ func KnownModels() []string {
 	return keys
 }
 
-// requires state.mu
```
randall77, line 143: This is still true, right? The lowercase methods require the lock to be held.

randall77 (approved):

go/plugins/googleai/googleai.go
jba: Right. Done.

PR #542, opened 2024-07-04
Fix bugs related to combining tools and output conformance

mbleigh:
Before, combining a tool call with output conformance could result in two different errors:

1. `Vertex response generation failed: ClientError: [VertexAI.ClientError]: Within a single message, FunctionResponse cannot be mixed with other type of part in the request for sending chat message.`
2. A schema validation error.

The first was caused by the output conformance middleware appending instructions to the last message instead of the last *user* role message.

The second was caused by the output schema validation trying to validate all candidates, even if the message returned contained no text or data part (e.g. a tool request).

Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -641,7 +641,8 @@ export async function generate<
   if (resolvedOptions.output?.schema || resolvedOptions.output?.jsonSchema) {
     // find a candidate with valid output schema
     const candidateErrors = response.candidates.map((c) => {
-      if (c.toolRequests()?.length) return null;
+      // only validate candidates that have output
+      if (!c.output()) return null;
 
       try {
         parseSchema(c.output(), {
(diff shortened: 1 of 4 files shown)


huangjeff5 (approved):

js/ai/src/generate.ts
```diff
@@ -635,6 +635,9 @@ export async function generate<
   if (resolvedOptions.output?.schema || resolvedOptions.output?.jsonSchema) {
     // find a candidate with valid output schema
     const candidateErrors = response.candidates.map((c) => {
+      // only validate candidates that have output
```
huangjeff5, line 638: Wait... If output schema is set, doesn't that mean we should always expect the candidate to have an output matching the described schema? 

Is there a case where empty output & output=schema is considered valid?

mbleigh: The scenario is when there is no `text` or `data` in the message at all, but only e.g. a `toolRequest`. It's a fair point though that an empty text or data response should still be validated, so I've changed the logic to explicitly check for the existence of at least one text or data part.

PR #543, opened 2024-07-04
[Go] Add snippet for pgvector retriever in Golang

ssbushi:
Add snippet for pgvector retriever in Golang

- [x] Tested (manually)
- [x] Docs updated


jba:
Looks like something went wrong here. I had a PR for this, #375, but I messed up and it got merged into the wrong branch, not main.

I recommend doing this as in that PR. Then you have a piece of code that compiles. This will inevitably become stale.

jba:
I re-sent #375 as #583.

I think whose code we end up using is less important than checking in actual code. You can use the weave tool to insert parts of the code into the markdown doc.

ssbushi:
I don't care whose code we use. Your snippet looks cleaner too!

I will branch off your PR and update the docs file using `weave`.

PR #549, opened 2024-07-05
add e2e test for reflection API

jba:
Add a test that checks the responses from the reflection API.

Describe the test in a YAML file so it can be used by other languages.


jba:
ping

pavelgj (approved):

Issue #545, opened 2024-07-04
[Tooling] Evaluation exits early if a prompt output is blocked


chrisraygill:
**Describe the bug**
Evaluation flows exit early if one of the prompts outputs gets blocked entirely When running an evaluation where one of my responses is blocked by gemini, so the flow never completes and errors early, the evaluation framework shuts down and doesn’t provide any output at all.

Output:
```
…
Error: Some flows failed with the following errors
Error: FAILED_PRECONDITION: All candidates returned finishReason issues: ["blocked"]
Received quitquitquit
OpenTelemetry SDK shut down.
Reflection API has succesfully shut down.
Reflection API has succesfully shut down.
App process exited with code 0
```

As a developer, having a prompt never finish for whatever reason is valuable feedback so I can understand where and how to better handle those situations.

[Internal bug link](https://buganizer.corp.google.com/issues/348407693)

PR #559, opened 2024-07-08
Gracefully handle inference errors when running `eval:flow`

ssbushi:
1/2 fixes for #545 

### DIFF
--- a/js/testapps/cat-eval/src/pdf_rag.ts
+++ b/js/testapps/cat-eval/src/pdf_rag.ts
@@ -24,6 +24,7 @@ import { defineFlow, run } from '@genkit-ai/flow';
 import { geminiPro } from '@genkit-ai/googleai';
 import { chunk } from 'llm-chunk';
 import path from 'path';
+import { getDocument } from 'pdfjs-dist-legacy';
 import * as z from 'zod';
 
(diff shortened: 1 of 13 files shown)


pavelgj (approved):

js/testapps/cat-eval/src/pdf_rag.ts
```diff
@@ -67,26 +68,6 @@ export const pdfQA = defineFlow(
...
-          {
-            category: 'HARM_CATEGORY_HARASSMENT',
-            threshold: 'BLOCK_NONE',
-          },
-          {
-            category: 'HARM_CATEGORY_SEXUALLY_EXPLICIT',
-            threshold: 'BLOCK_NONE',
-          },
-        ],
-      },
```
MichaelDoyle, line 89: Calling it out *just* in case - you might still want this (evaluators should probably never be allowed to get blocked for content).

MichaelDoyle (approved):

js/testapps/cat-eval/src/pdf_rag.ts
ssbushi: This setting is for the flow under test, not the evaluators.

PR #560, opened 2024-07-09
fix(js): split tool response and other parts in vertexai gemini

cabljac:
Fixes https://github.com/firebase/genkit/issues/452 

Only manually tested so far in a couple of specific cases.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/plugins/vertexai/src/gemini.ts
+++ b/js/plugins/vertexai/src/gemini.ts
@@ -279,6 +279,61 @@ export function toGeminiMessage(
   };
 }
 
+export function toGeminiMessages(
+  messages: MessageData[],
+  model?: ModelReference<z.ZodTypeAny>
+): Content[] {
+  const geminiMessages: Content[] = [];
+
+  for (let i = 0; i < messages.length; i++) {
+    if (messages[i].content.some((part) => part.toolResponse)) {
+      const toolResponseMessage = messages[i];
+
+      // get the most recent user message in the history
+      const previousUserMessageIndex = messages
+        .slice(0, i)
+        .reverse()
+        .findIndex((m) => m.role === toGeminiRole('user', model));
+
+      if (previousUserMessageIndex === -1) {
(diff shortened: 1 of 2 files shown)


js/plugins/vertexai/src/gemini.ts
```diff
@@ -532,8 +532,19 @@ export function geminiModel(name: string, vertex: VertexAI): ModelAction {
...
+          const toolResponses = msg.parts.filter(
+            (part) => part.functionResponse
+          );
+          const nonToolParts = msg.parts.filter(
+            (part) => !part.functionResponse
+          );
+          // The Gemini API does not support sending tool responses in the same message as other parts
+          for (const part of toolResponses) {
+            chat.sendMessage([part]);
+          }
```
pavelgj, line 546: what's going on here? it looks like it makes the chat call for each tool response separately, but not doing anything with the response?

cabljac: I wasn't 100% on this, but the thinking was the chat instance in the sdk will keep hold of the responses, but I need to look at this again. For instance surely it needs to be awaited if it works how I thought it did

pavelgj: `sendMessage` makes a call to the API... and returns the response. This code will result in multiple LLM calls instead of 1, which sounds.... expensive. :)

cabljac: Ah we should definitely group them into one call haha. The issue in question is that a FunctionResponse isn't allowed to be in the same message as other parts, I believe

pavelgj: I think it's just a matter of splitting them into separate messages... and moving the non tool parts into history.

cabljac:
closing in favour of https://github.com/firebase/genkit/pull/560

PR #578, opened 2024-07-10
RAG docs for Go

kevinthecheung:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


go/internal/doc-snippets/rag/main.go
```diff
@@ -0,0 +1,227 @@
...
+			// Add chunks to the index.
+			err = menuPdfIndexer.Index(ctx, &ai.IndexerRequest{
+				Documents: docs, Options: nil,
+			})
+			return nil, err
+		},
+	)
+	//!-indexflow
+	
+	genkit.Init(ctx, nil)
```
jba, line 95: Check for returned error.

go/internal/doc-snippets/rag/main.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return "", err
+	}
+	
+	buf := bytes.Buffer{}
+    b, err := r.GetPlainText()
+    if err != nil {
+        return "", err
+    }
+    buf.ReadFrom(b)
```
jba, line 113: Not sure what's going on here. if `b` is a []byte, then you can just do `string(b)`.

go/internal/doc-snippets/rag/main.go
```diff
@@ -0,0 +1,227 @@
...
+		textsplitter.WithChunkSize(200),
+		textsplitter.WithChunkOverlap(20),
+	)
+	//!-splitcfg
+	//!+indexflow
+	genkit.DefineFlow(
+		"indexMenu",
+		func(ctx context.Context, path string) (any, error) {
+			// Extract plain text from the PDF.
+			pdfText, err := genkit.Run(ctx, "extract", func() (string, error) {
```
jba, line 62: Explain why you're using genkit.Run here instead of doing the work directly.

go/internal/doc-snippets/rag/main.go
```diff
@@ -0,0 +1,227 @@
...
+
+			// Split the text into chunks.
+			docs, err := genkit.Run(ctx, "chunk", func() ([]*ai.Document, error) {
+				chunks, err := splitter.SplitText(pdfText)
+				if err != nil {
+					return nil, err
+				}
+
+				var docs []*ai.Document
+				for i := range len(chunks) {
```
jba, line 77: for _,c := range chunks

go/internal/doc-snippets/rag/main.go
```diff
@@ -0,0 +1,227 @@
...
+				}
+				return docs, nil
+			})
+			if err != nil {
+				return nil, err
+			}
+
+			// Add chunks to the index.
+			err = menuPdfIndexer.Index(ctx, &ai.IndexerRequest{
+				Documents: docs, Options: nil,
```
jba, line 88: You can omit `Options: nil`

jba (changes requested):

go/internal/doc-snippets/rag/main.go
```diff
@@ -0,0 +1,227 @@
...
+	
+	genkit.Init(ctx, nil)
+}
+
+//!+readpdf
+// Helper function to extract plain text from a PDF. Excerpted from 
+// https://github.com/ledongthuc/pdf
+func readPdf(path string) (string, error) {
+	f, r, err := pdf.Open(path)
+    defer f.Close()
```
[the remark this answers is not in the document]
kevinthecheung: i don't know, so i included the check

go/internal/doc-snippets/rag/main.go
```diff
@@ -198,13 +205,14 @@ func customret() {
 			k := 3
 			preRerankK := 10
 			if opts, ok := req.Options.(CustomMenuRetrieverOptions); ok {
+				k = opts.K
```
jba, line 208: OK, so the bug here (sorry, just saw it) is that someone may do

     CustomMenuRetrieverOptions{K: 5}

leaving the other option at 0, but you set both.
So a better way is to assume 0 means unset (which I think is OK for both of these options, zero doesn't make sense, right?).
Then you'd write 
   if opts.K != 0 { k = opts.K}
and similarly for the other one.

But I think here is a nicer way:
```
opts, _ := req.Options.(CustomMenuRetrieverOptions)
// If req.Options is not the right type, then opts is a zero CustomMenuRetrieverOptions.
if opts.K == 0 { opts.K = 3 }
if ... // the other one
```
Then use opts.K, etc. in the following code.

jba (changes requested):

docs-go/rag.md
```diff
@@ -0,0 +1,392 @@
...
+
+To create a retriever, you can use one of the provided implementations or
+create your own.
+
+## Supported indexers, retrievers, and embedders
+
+Genkit provides indexer and retriever support through its plugin system. The
+following plugins are officially supported:
+
+- [Cloud Firestore vector store](plugins/firebase.md)
```
ssbushi, line 104: The go SDK only has pinecone support for RAG.

ssbushi (approved):

jba (approved):

PR #582, opened 2024-07-10
fix: fixed/updated go init template

pavelgj:


jba (approved):
OK, but this should be an actual .go file that's inserted into the ts, as we did for the doc snippets. It's going to be hard to keep this up to date.

pavelgj:
> OK, but this should be an actual .go file that's inserted into the ts, as we did for the doc snippets. It's going to be hard to keep this up to date.

A similar thought crossed my mind as I was making these changes. Something we definitely should follow up on.

PR #591, opened 2024-07-11
Fixes context middleware to append to user message only.

mbleigh:
Applies the fix from #542 to the context augmentation middleware.

Checklist (if applicable):


MichaelDoyle (approved):
Thanks! This change itself looks great. 

On a bigger picture level, I am still curious as to whether or not the middleware should be reapplied during a tool-calling loop, or if we can/should maintain the original request that went to the LLM and simple append to that. I am not so familiar with how this is al plumbed together.

PR #594, opened 2024-07-11
Cloud Run docs for Go

kevinthecheung:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


docs-go/cloud-run.md
```diff
@@ -0,0 +1,174 @@
...
+    Accept the defaults for the remaining prompts. The `genkit` tool will create
+    a sample source file to get you started developing your own AI flows.
+    For the rest of this tutorial, however, you'll just deploy the sample flow.
+
+1.  Edit the sample file (`main.go` or `genkit.go`) to explicitly specify the
+    port the flow server should listen on:
+
+    ```go
+    if err := genkit.Init(ctx,
+    	&genkit.Options{FlowAddr: ":3400"}, // Add this parameter.
```
jba, line 58: I think this happens by default.

jba (approved):

docs-go/cloud-run.md
kevinthecheung: The default is `127.0.0.1:3400`, which doesn't work when deployed as far as I can tell.

pavelgj: It should first try to get the port number from PORT env var: https://github.com/firebase/genkit/blob/4b36ece6177cbd61ffa319d8fbde987c21d27e6e/go/genkit/servers.go#L68

kevinthecheung: Yeah,. but it still prepends 127.0.0.1:

https://github.com/firebase/genkit/blob/4b36ece6177cbd61ffa319d8fbde987c21d27e6e/go/genkit/servers.go#L362-L364

pavelgj: Oh.... that's a bug.

pavelgj (approved):

PR #597, opened 2024-07-11
Go docs for Vertex and Google AI

kevinthecheung:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


go/internal/doc-snippets/googleai.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	//!-retrieve
+
+	_ = retrieveRes
+
+	var myIndexer *ai.Indexer
+	var docsToIndex []*ai.Document
+
+	//!+index
+	err = myIndexer.Index(ctx, &ai.IndexerRequest{Documents: docsToIndex})
```
jba, line 90: if err := ...; err != nil

go/internal/doc-snippets/vertexai.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/firebase/genkit/go/ai"
+	"github.com/firebase/genkit/go/plugins/vertexai"
+)
+
+func vertexaiEx(ctx context.Context) error {
+	var err error
+
+	//!+init
+	err = vertexai.Init(ctx, nil)
```
jba, line 28: ditto
Whenever there is only an error return value and the line is not too long, use the `if err :=` form.

jba (changes requested):

jba (approved):

PR #606, opened 2024-07-12
[Go] add version to attribution headers

jba:
Add a version file that contains the current tagged version.

Also add a test for googleai.
It's harder to test vertexai because it uses gRPC.

Also ran `go mod tidy`.


jba:
@pavelgj Where can I document that the version in internal/version.go must be updated before a new Go release is tagged?

go/internal/version.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+// Version is the current tagged release of this module.
+// That is, it should match the value of the latest `go/v*` git tag.
+const Version = "0.0.1"
```
pavelgj, line 19:
```suggestion
const Version = "0.0.2"
```

pavelgj (approved):

PR #608, opened 2024-07-12
[Go] ollama.Init takes Config

jba:
Pass a Config struct to Init instead of just the server address,
to allow for additional configuration options.

By passing a struct, adding options is not a breaking change.

Also make other minor stylistic changes.

### DIFF
--- a/go/plugins/ollama/ollama.go
+++ b/go/plugins/ollama/ollama.go
@@ -84,14 +84,6 @@ type ModelDefinition struct {
 	Type string
 }
 
-// Config provides configuration options for the Init function.
-type Config struct {
-	// Server Address of oLLama.
-	ServerAddress string
-	// Generative models to provide.
-	Models []ModelDefinition
-}
-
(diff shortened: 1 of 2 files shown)


go/plugins/ollama/ollama.go
```diff
@@ -194,16 +196,17 @@ func (g *generator) generate(ctx context.Context, input *ai.GenerateRequest, cb
 			Stream:   stream,
 		}
 	}
-	client := &http.Client{
-		Timeout: time.Second * 30,
-	}
+	client := &http.Client{Timeout: time.Second * 30}
```
ianlancetaylor, line 199: Really minor but I encourage people to write `30 * time.Second` so that it reads as the English "30 seconds".

ianlancetaylor (approved):

go/plugins/ollama/ollama.go
jba: Done.

PR #630, opened 2024-07-16
Feat: Graphs

debkanchan:
Graphs let us work in a cyclic fashion easily to implement workflows like reflexion. Graphs would help us build agentic workflows easily

Right now i'm just prototyping it. I need some feedback on how to best make it production grade.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/graph/LICENSE
+++ b/js/graph/LICENSE
@@ -1,0 +1,203 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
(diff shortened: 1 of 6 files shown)


cabljac:
I have a plugin i was building which does something similar, if you like I will put it in a public repo to review?

debkanchan:
Sure @cabljac

pavelgj (commented):
do you mind adding a testapp (`js/testapps/graph`) that demonstrates the graph library in action?

cabljac:
> do you mind adding a testapp (`js/testapps/graph`) that demonstrates the graph library in action?

Is this directed to me or @debkanchan ? (also happy to add a test app with the plugin)

debkanchan:
> do you mind adding a testapp (`js/testapps/graph`) that demonstrates the graph library in action?

Sure. I'll do that once I figure out 1. tracing 2. how to make it run like `run(graph)`

cabljac:
Hey @pavelgj @debkanchan this morning I opened a draft, with the code from the plugin i made before, included a test app:

https://github.com/firebase/genkit/pull/649

debkanchan:
@cabljac I read your PR. Tbh it's the most complete one so far. But I have one problem with this approach that I have also previously stated in discord.

The concept of edges is bit confusing imo. To me, each node should be able to dictate which node it wants to go next based on it's internal conditions. We should not be needing separate edges or even worse conditional edges. This only confuses new-comers and imo more verbose.

cabljac:
Ah interesting, I didn't consider this. I come from a math background where an edge is quite intuitive, but perhaps you're right and it could be confusing.

I guess any way you define a graph it will have implicit edges, and most graph related packages will deal with edges to run various algorithms. 

Perhaps the main way to define an edge could be implicitly at the node.

One thing I like about having a graph as a layer above the actual flows is that they can just point at the flows, so one flow could serve several node instances.

I should share the UI demo I made for it on discord with you. I'm away at the moment but will do when I get a chance. We should also move this to a separate repo probably. I could create one if you want, and invite you?

debkanchan:
This pr has been moved to https://github.com/TheFireCo/genkit-plugins/pull/117

PR #645, opened 2024-07-17
[Go] Make dotprompt take in `jsonschema.Schema` instead of `map[string]any` for better ergonomics.

apascal07:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/go/plugins/dotprompt/dotprompt.go
+++ b/go/plugins/dotprompt/dotprompt.go
@@ -98,14 +98,17 @@ type Config struct {
 	// Details for the model.
 	GenerationConfig *ai.GenerationCommonConfig
 
-	InputSchema      *jsonschema.Schema // schema for input variables
-	VariableDefaults map[string]any     // default input variable values
+	// Schema for input variables.
+	InputSchema *jsonschema.Schema
+
+	// Default input variable values
+	VariableDefaults map[string]any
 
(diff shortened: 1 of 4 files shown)


jba:
My main concern here is making `github.com/invopop/jsonschema.Schema` part of the API. I don't know if we do that anywhere else, but we shouldn't.

The problem is that that package is broken. It doesn't handle all JSONSchemas, and it [probably never will](invopop/jsonschema/134). It's bad enough that we use it internally, but at least we can change that. If it's part of the API, replacing it will be a breaking change.

apascal07:
> My main concern here is making `github.com/invopop/jsonschema.Schema` part of the API. I don't know if we do that anywhere else, but we shouldn't.
> 
> The problem is that that package is broken. It doesn't handle all JSONSchemas, and it [probably never will](invopop/jsonschema/134). It's bad enough that we use it internally, but at least we can change that. If it's part of the API, replacing it will be a breaking change.

Valid concern but we do have `jsonschema.Schema` part of the API already. Right above, we use it for `InputSchema`.

jba:
Oh yeah. Well that's too bad.

I think replacing that package should block going to v1.0. Unfortunately it won't be easy.

jba (approved):

PR #664, opened 2024-07-22
fix: [Go] fixed tool schema and simplifying tool definition

pavelgj:
Made it easier to define tools

```go
myJokeTool := ai.DefineTool(
	"myJoke",
	"useful when you need a joke to tell",
	func(ctx context.Context, input *any) (string, error) {
		return "haha Just kidding no joke! got you", nil
	},
)

request := ai.GenerateRequest{
	Messages: []*ai.Message{
		{Content: []*ai.Part{ai.NewTextPart("Tell me a joke.")},
			Role: ai.RoleUser},
	},
	Tools: []*ai.ToolDefinition{myJokeTool.Definition()},
}
```


go/ai/tools.go
```diff
@@ -30,29 +34,58 @@ const provider = "local"
...
 // and something in the general code should handle the JSON conversion.
 type Tool struct {
-	Definition *ToolDefinition
-	Fn         func(context.Context, map[string]any) (map[string]any, error)
+	Action action.Action
 }
 
 // DefineTool defines a tool function.
-func DefineTool(definition *ToolDefinition, metadata map[string]any, fn func(ctx context.Context, input map[string]any) (map[string]any, error)) {
+func DefineTool[In, Out any](name, description string, metadata map[string]any, fn func(ctx context.Context, input In) (Out, error)) *Tool {
```
jba, line 41: Why not carry the type safety through as far as possible, by having a `Tool[In, Out]` that contains a `core.Action[In, Out, struct{}]`.

jba: At some point you might need a `Tool` type that wasn't generic, as we do for actions, and it would have a RunJSON method. Or maybe you could just use actions for that, I'm not sure.

pavelgj: I spent a lot of time trying to make Tool generic... in most Actions (except custom actions) generic i/o types are fixed (ex. model action has GenerateRequest/Response types), so in, for example. LookupModel you can fill in the Action generics (in `LookupActionFor`) with hardcoded types. With Tools the i/o types are defined by the user per tool. So I need a non-generic type to be able to construct and return it from `LookupTool`.

pavelgj: actually nevermind... figured it out.

jba (approved):

go/ai/tools.go
```diff
@@ -29,30 +33,91 @@ const provider = "local"
...
-	Definition *ToolDefinition
-	Fn         func(context.Context, map[string]any) (map[string]any, error)
+type Tool[In, Out any] struct {
+	action *core.Action[In, Out, struct{}]
+}
+
+type toolAction struct {
+	action action.Action
+}
+
```
apascal07, line 43: Let's maintain comments.

go/ai/tools.go
```diff
@@ -29,30 +33,91 @@ const provider = "local"
 // The ToolDefinition has JSON schemas that describe the types.
 // TODO: This should be generic over the function input and output types,
 // and something in the general code should handle the JSON conversion.
-type Tool struct {
-	Definition *ToolDefinition
-	Fn         func(context.Context, map[string]any) (map[string]any, error)
+type Tool[In, Out any] struct {
```
apascal07, line 36: It's not immediately clear to me why we need both `Tool` and `toolAction`. Can you explain?

pavelgj: The way generics in Go work. I need a type that doesn't have generics to construct and return from `LookupTools`. `toolAction` is it. Otherwise when holding on to a reference to a `Tool[In, Out]` you have strongly typed (with generics) reference.

go/ai/tools.go
pavelgj: Done

apascal07 (approved):

PR #671, opened 2024-07-23
fix: point to correct binary path for global install

cabljac:
This should fix https://github.com/firebase/genkit/issues/625 

The issue is that the `node_modules` originally pointed to doesn't exist.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


genkit-tools/common/src/runner/runner.ts
```diff
@@ -177,7 +177,7 @@ export class Runner {
       case 'nodejs':
         command =
           config?.runner?.mode === 'harness'
-            ? path.join(__dirname, '../../../node_modules/.bin/tsx')
+            ? path.join(__dirname,"../../../../../.bin/tsx")
```
cabljac, line 180:
```suggestion
            ? path.join(__dirname,'../../../../../.bin/tsx')
```

MichaelDoyle:
Looping in @apascal07 who worked on this originally.

pavelgj:
> Looping in @apascal07 who worked on this originally.

yeah, I'm going to leave this one to Alex! ;)

apascal07:
Ugh. The problem here is that the path to the node_modules dir is different depending on whether you’re linking locally vs. installing from NPM registry. I don’t think there’s a single path that will work for both. Can you compare the directory structure between local linked and NPM installed Genkit to confirm?

cabljac:
@apascal07 you're right. I've added a check to see if the file is linked, maybe it seems a bit hacky, what do you think?

apascal07:
> @apascal07 you're right. I've added a check to see if the file is linked, maybe it seems a bit hacky, what do you think?

Could we instead check whether `tsx` exists at path 1  otherwise check path 2? Let's try the "installed from NPM registry" path first.

cabljac:
updated the logic to check if the file exists :)

apascal07 (approved):

PR #681, opened 2024-07-24
refactor: [Go] updated the ai.Generate API

pavelgj:
```go
model := googleai.Model("gemini-1.5-flash")

// simple text
text, err := model.GenerateText(ctx,
	ai.WithConfig(&ai.GenerationCommonConfig{Temperature: 1}),
	ai.WithTextPrompt("Tell me a joke"))

// with tools
gablorkenTool := ai.DefineTool("gablorken", "use when need to calculate a gablorken",
		Value float64
		Over  float64
	}) (float64, error) {
		return math.Pow(input.Value, input.Over), nil
	},
)

text, err := model.GenerateText(ctx,
	ai.WithCandidates(1),
	ai.WithTextPrompt("what is a gablorken of 2 over 3.5?"),
	ai.WithTools(gablorkenTool))

// with mesages
response, err = model.Generate(ctx,
	ai.WithMessages(history...),
	ai.WithTextPrompt("What were we talking about?"))

// with history
resp1, err := m.Generate(ctx, ai.WithTextPrompt("hi, my name is Heisenberg!"))

resp2, err := m.Generate(ctx,
    ai.WithMessages(resp1.History()...),
    ai.WithTextPrompt("Say my name!"))

// multimodal
resp, err := model.Generate(ctx, ai.WithMessages(
	ai.NewUserMessage(
		ai.NewTextPart("Describe the following image."),
		ai.NewMediaPart("", "data:image/jpeg;base64,"+encodedImage))))

// structured output
type GameCharacter struct {
	Name      string
	Backstory string
}

character := &GameCharacter{}
_, err = m.GenerateData(ctx, character,
	ai.WithTextPrompt("generate a game character"))

// streaming
resp, err := model.Generate(ctx,
	ai.WithTextPrompt("tell me a VERY long joke"),
	ai.WithStreaming(func(ctx context.Context, c *ai.GenerateResponseChunk) error {
		fmt.Printf("got stream chunk %v\n", c)
		return nil
	}))
```

variants to work with "raw" request are still available

```go
	resp, err := model.GenerateRaw(ctx, &ai.GenerateRequest{
		Messages: []*ai.Message{
			// ....
		},
	}, streamingCallback)
```

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -86,8 +86,102 @@ func LookupModel(provider, name string) *Model {
 	return (*Model)(core.LookupActionFor[*GenerateRequest, *GenerateResponse, *GenerateResponseChunk](atype.Model, provider, name))
 }
 
-// Generate applies the [Model] to some input, handling tool requests.
-func (m *Model) Generate(ctx context.Context, req *GenerateRequest, cb ModelStreamingCallback) (*GenerateResponse, error) {
+type GenerateRequestEditor func(req *GenerateRequest)
+
+func WithSimpleTextPrompt(prompt string) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		req.Messages = append(req.Messages, NewUserTextMessage(prompt))
+	}
+}
+
+func WithMessages(messages ...*Message) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		req.Messages = append(req.Messages, messages...)
+	}
+}
+
+func WithConfig(config any) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		req.Config = config
+	}
+}
+
+func WithCandidates(c int) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		req.Candidates = c
+	}
+}
+
+func WithContext(c ...any) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		req.Context = c
+	}
+}
+
+func WithTools(tools ...Tool) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		var toolDefs []*ToolDefinition
+		for _, t := range tools {
+			toolDefs = append(toolDefs, t.Definition())
+		}
+		req.Tools = toolDefs
+	}
+}
+
+func WithOutputSchema(schema struct{}) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		req.Output.Schema = base.SchemaAsMap(base.InferJSONSchema(schema))
+	}
+}
+
+func WithOutputFormat(format OutputFormat) GenerateRequestEditor {
+	return func(req *GenerateRequest) {
+		req.Output.Format = format
+	}
+}
+
+func (m *Model) Generate(ctx context.Context, withs ...GenerateRequestEditor) (*GenerateResponse, error) {
+	return m.StreamGenerate(ctx, nil, withs...)
+}
+
+func (m *Model) GenerateText(ctx context.Context, withs ...GenerateRequestEditor) (string, error) {
+	return m.StreamGenerateText(ctx, nil, withs...)
+}
+
+func (m *Model) StreamGenerate(ctx context.Context, cb ModelStreamingCallback, withs ...GenerateRequestEditor) (*GenerateResponse, error) {
+	req := &GenerateRequest{}
+	for _, with := range withs {
+		with(req)
+	}
+
+	return m.StreamGenerateRaw(ctx, req, cb)
+}
+
+func (m *Model) StreamGenerateText(ctx context.Context, cb ModelStreamingCallback, withs ...GenerateRequestEditor) (string, error) {
+	req := &GenerateRequest{}
+	for _, with := range withs {
+		with(req)
+	}
+
+	if req.Candidates > 1 {
+		return "", fmt.Errorf("StreamGenerateText can only be called with 1 candidate")
+	}
+
+	res, err := m.StreamGenerateRaw(ctx, req, cb)
+	if err != nil {
+		return "", err
+	}
+
+	return res.Text()
+}
+
+// GenerateRaw applies the [Model] to some input, handling tool requests.
+func (m *Model) GenerateRaw(ctx context.Context, req *GenerateRequest) (*GenerateResponse, error) {
+	return m.StreamGenerateRaw(ctx, req, nil)
+}
+
+// StreamGenerateRaw applies the [Model] to some input, handling tool requests and handles streaming.
+func (m *Model) StreamGenerateRaw(ctx context.Context, req *GenerateRequest, cb ModelStreamingCallback) (*GenerateResponse, error) {
 	if m == nil {
 		return nil, errors.New("Generate called on a nil Model; check that all models are defined")
 	}
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -94,90 +94,107 @@ type GenerateParams struct {
 }
 
 // GenerateParamsBuilder adds.
-type GenerateParamsBuilder func(req *GenerateParams)
+type GenerateParamsBuilder func(req *GenerateParams) error
 
 // WithTextPrompt adds a simple text user prompt to GenerateRequest.
 func WithTextPrompt(prompt string) GenerateParamsBuilder {
-	return func(req *GenerateParams) {
+	return func(req *GenerateParams) error {
 		req.Request.Messages = append(req.Request.Messages, NewUserTextMessage(prompt))
+		return nil
 	}
 }
 
 // WithSystemPrompt adds a simple text system prompt as the first message in GenerateRequest.
 func WithSystemPrompt(prompt string) GenerateParamsBuilder {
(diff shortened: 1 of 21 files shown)


mbleigh:
A few high-level questions/thoughts:

1. This retains the `model.Generate` approach instead of following Node.js where `generate` is a sugaring abstraction on top of the more utilitarian `Model` interface. Is this intentional / do we think it's a good idea in the long term? vs. something like `ai.GenerateText(ctx, ai.WithModel(model))`
2. This approach has no required arguments, but is going to end up erroring if you don't supply any options. Should there be a mandatory `Message prompt` argument that is the bare minimum (since config, and everything else could have defaults but there *must* be a prompt).
3. For things like `WithMessages` and `WithTools`, what happens if they're supplied multiple times? Is it replace or append?
4. (random nit) Why `SimpleTextPrompt` instead of just `TextPrompt`?

pavelgj:
> A few high-level questions/thoughts:
> 
> 1. This retains the `model.Generate` approach instead of following Node.js where `generate` is a sugaring abstraction on top of the more utilitarian `Model` interface. Is this intentional / do we think it's a good idea in the long term? vs. something like `ai.GenerateText(ctx, ai.WithModel(model))`. To put a stake in the ground I pretty strongly think Generate needs to be an abstraction to make it easier to author models and allow for things like default models.
> 2. This approach has no required arguments, but is going to end up erroring if you don't supply any options. Should there be a mandatory `Message prompt` argument that is the bare minimum (since config, and everything else could have defaults but there _must_ be a prompt).
> 3. For things like `WithMessages` and `WithTools`, what happens if they're supplied multiple times? Is it replace or append?
> 4. (random nit) Why `SimpleTextPrompt` instead of just `TextPrompt`?

1. I'm 100% with you, but wanted to avoid this and tackle next. That said, I think variadic args can be a good substitute for overloading... `ai.GenerateText(ctx, ai.WithModel(model))` or `ai.GenerateText(ctx, ai.WithModelStr("vertexai/gemini-1.5-flash"))`
2.  I thought about that and thought it's OK... we can return errors if nothing is provided. I mean, you could just pass `&GenerateRequest{}` before and the language doesn't really help making sure at least some messages are set.
3. It appends. Specifying it multiple times is actually OK:
    `model.GenerateText(ctx, ai.WithMessages(history...), ai.WithMessages(ai.NewUserTextMessage("Hi")))`
4. Agreed. Fixed.

go/ai/generate.go
```diff
@@ -86,8 +86,122 @@ func LookupModel(provider, name string) *Model {
...
+	}
+}
+
+// Generate run generate request for this model. Returns GenerateResponse struct.
+func (m *Model) Generate(ctx context.Context, withs ...GenerateRequestBuilder) (*GenerateResponse, error) {
+	return m.StreamGenerate(ctx, nil, withs...)
+}
+
+// GenerateText run generate request for this model. Returns generated text only.
+func (m *Model) GenerateText(ctx context.Context, withs ...GenerateRequestBuilder) (string, error) {
```
mbleigh, line 165: Should this return `(string, *GenerateResponse, error)`? You can always ignore the full response if you want with a `_`, but that way you still have access to the response if you want to get metadata out of it.

apascal07: Muddies the water with plain `Generate` a bit...

mbleigh: Yeah, I think I'd be okay with `resp.Text()` being a non-erroring method (it should be - nothing can go wrong at that point) and not returning the response from GenerateText

pavelgj: Done -- `resp.Text()` doesn't error anymore

go/ai/generate.go
```diff
@@ -86,8 +86,131 @@ func LookupModel(provider, name string) *Model {
...
+		var toolDefs []*ToolDefinition
+		for _, t := range tools {
+			toolDefs = append(toolDefs, t.Definition())
+		}
+		req.Request.Tools = toolDefs
+	}
+}
+
+// WithOutputSchema adds provided output schema to GenerateRequest.
+func WithOutputSchema[T any](schema T) GenerateParamsBuilder {
```
apascal07, line 151: Does `[T any]` do anything here? Please explain in the comment that the arg is the struct/value, not the schema map or anything else.

go/ai/generate.go
```diff
@@ -86,8 +86,131 @@ func LookupModel(provider, name string) *Model {
...
+		return "", err
+	}
+
+	return res.Text(), nil
+}
+
+// Generate run generate request for this model. Returns GenerateResponse struct.
+// TODO: Stream GenerateData with partial JSON
+func (m *Model) GenerateData(ctx context.Context, value any, withs ...GenerateParamsBuilder) (*GenerateResponse, error) {
+	withs = append(withs, WithOutputSchema(value))
```
apascal07, line 200: This should probably throw an error if there are any other `WithOutputSchema`. At the very least, overwrite it but that's less deterministic behavior.

go/ai/generate.go
```diff
@@ -86,8 +86,131 @@ func LookupModel(provider, name string) *Model {
...
+			req.Request.Output.Format = OutputFormatJSON
+		}
+		req.Request.Output.Schema = base.SchemaAsMap(base.InferJSONSchemaNonReferencing(schema))
+	}
+}
+
+// WithOutputFormat adds provided output format to GenerateRequest.
+func WithOutputFormat(format OutputFormat) GenerateParamsBuilder {
+	return func(req *GenerateParams) {
+		req.Request.Output.Format = format
```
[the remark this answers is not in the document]
pavelgj: done

go/ai/generate.go
pavelgj: hmmm... no... vestige of some previous iteration. Removed.

go/ai/generate.go
pavelgj: added an error if WithOutputSchema  called more than once

go/ai/generate.go
```diff
@@ -86,8 +86,170 @@ func LookupModel(provider, name string) *Model {
...
-// Generate applies the [Model] to some input, handling tool requests.
-func (m *Model) Generate(ctx context.Context, req *GenerateRequest, cb ModelStreamingCallback) (*GenerateResponse, error) {
+type GenerateParams struct {
+	Request      *GenerateRequest
+	Stream       ModelStreamingCallback
+	History      []*Message
+	SystemPrompt *Message
+}
+
+// GenerateParamsBuilder adds.
```
jba, line 96: Finish doc.

Also, we would typically call this type `GenerateOption`.
Standard pattern: https://pkg.go.dev/google.golang.org/api/option#ClientOption

go/ai/generate.go
```diff
@@ -86,8 +86,170 @@ func LookupModel(provider, name string) *Model {
...
+// WithStreaming adds a streaming callback to the generate request.
+func WithStreaming(cb ModelStreamingCallback) GenerateParamsBuilder {
+	return func(req *GenerateParams) error {
+		req.Stream = cb
+		return nil
+	}
+}
+
+// Generate run generate request for this model. Returns GenerateResponse struct.
+func (m *Model) Generate(ctx context.Context, withs ...GenerateParamsBuilder) (*GenerateResponse, error) {
```
jba, line 202: typically `opts ...GenerateOption`

jba: But I am confused: I thought on the discord we were going to make this more low-level, taking a GenerateRequest, and have functions like `ai.GenerateXXX`.

go/ai/generate.go
pavelgj: Done.

go/ai/generate.go
pavelgj: I like `GenerateOption`! Renamed.
re: `ai.GenerateXXX` that will be in the next PR. Want to keep these changes manageable/reviewable. In retrospect I should've started with that, but was super excited about this first.

go/ai/generate.go
```diff
@@ -86,8 +86,183 @@ func LookupModel(provider, name string) *Model {
...
+// GenerateParams represents various params of the Generate call.
+type GenerateParams struct {
+	Request      *GenerateRequest
+	Stream       ModelStreamingCallback
+	History      []*Message
+	SystemPrompt *Message
+}
+
+// GenerateOption configures params of the Generate call.
+type GenerateOption func(req *GenerateParams) error
```
jba, line 98: Great. 
Next question: Do you want this to be exported?

If yes, then anyone can write their own GenerateOption and use it. But we have one more exported symbol (actually several, because the fields are exported.) A bigger API surface means more for people to understand. Even though 99% (or maybe 100%) of people won't care about this type and its fields and will never use it directly.

If no, then you control the options, no one can write their own, and no one sees this type (except as an arg to the GenerateOption function, which they will happily ignore).

I would always pick no: make this unexported. You can always change your mind and export it later (although it would technically be breaking to change the signature of the GenerateOption function, it's extremely unlikely that anyone will be affected).

pavelgj: agreed. unexported.

apascal07 (approved):

PR #696, opened 2024-07-26
feat: allow for sync references to prompt files

cabljac:
This PR adds a DotpromptRef class to allow sync references to prompts in dotprompt.

resolves https://github.com/firebase/genkit/issues/513

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


mbleigh (approved):
I'm not 100% sure about the `promptRef` naming, but after thinking over it for a bit I can't come up with anything better.

Can you also update documentation with this change to change all async `prompt` loads to sync `promptRef` loads?

cabljac:
Yup will do!

I agree, it almost seems nicer to call it just `prompt` even if it returns a promptRef..

PR #705, opened 2024-07-28
Remove Google Auth Access Token Caching in PredictModel

dongyangli1226:
The changes in this PR remove caching of the google auth access token to fix a token expiration issue mentioned in: https://github.com/firebase/genkit/issues/662

Example issue scenario: Google auth access token has a default 1 hour TTL before it expires. When the auth token is generated and the Predict Model is called, the PredictModel caches the token for 50 mins and then requests to get a new token after 50 mins. However, the auth token cached in Google Auth Library hasn't expired yet and it will expire in 10 mins in this case. The PredictModel now caches a token that's about to be expired in 10 mins for another 50 mins.

Checklist (if applicable):
- [ X ] Tested (manually, unit tested, etc.)
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/plugins/vertexai/src/predict.ts
+++ b/js/plugins/vertexai/src/predict.ts
@@ -31,26 +31,18 @@ interface PredictionResponse<R> {
   predictions: R[];
 }
 
-const ACCESS_TOKEN_TTL = 50 * 60 * 1000; // cache access token for 50 minutes
-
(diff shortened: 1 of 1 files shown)


dongyangli1226:
Fixing the CLA. Closing this one

PR #706, opened 2024-07-28
Remove Google Auth Access Token Caching in PredictModel

dongyangli1226:
The changes in this PR remove caching of the google auth access token to fix a token expiration issue mentioned in: https://github.com/firebase/genkit/issues/662

Example issue scenario: Google auth access token has a default 1 hour TTL before it expires. When the auth token is generated and the Predict Model is called, the PredictModel caches the token for 50 mins and then requests to get a new token after 50 mins. However, the auth token cached in Google Auth Library hasn't expired yet and it will expire in 10 mins in this case. The PredictModel now caches a token that's about to be expired in 10 mins for another 50 mins.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/plugins/vertexai/src/predict.ts
+++ b/js/plugins/vertexai/src/predict.ts
@@ -31,26 +31,18 @@ interface PredictionResponse<R> {
   predictions: R[];
 }
 
-const ACCESS_TOKEN_TTL = 50 * 60 * 1000; // cache access token for 50 minutes
-
 export function predictModel<I = unknown, R = unknown, P = unknown>(
   auth: GoogleAuth,
   { location, projectId }: PluginOptions,
   model: string
 ) {
-  let accessToken: string | null | undefined;
-  let accessTokenFetchTime = 0;
-
   return async (
     instances: I[],
     parameters?: P
   ): Promise<PredictionResponse<R>> => {
     const fetch = (await import('node-fetch')).default;
 
-    if (!accessToken || accessTokenFetchTime + ACCESS_TOKEN_TTL < Date.now()) {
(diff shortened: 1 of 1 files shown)


dongyangli1226:
@pavelgj Friendly ping to help take a look. Thank you!

pavelgj:
Thanks for the PR. I merged some changes to this code earlier (unrelated change) as part of which I also removed the token refresh logic. Could you please check the current implementation at head and see if that addresses the issue?

dongyangli1226:
> Thanks for the PR. I merged some changes to this code earlier (unrelated change) as part of which I also removed the token refresh logic. Could you please check the current implementation at head and see if that addresses the issue?

Yeah I think that should resolve it. When do we plan to release the change? @pavelgj Thanks for the quick response.

pavelgj:
> > Thanks for the PR. I merged some changes to this code earlier (unrelated change) as part of which I also removed the token refresh logic. Could you please check the current implementation at head and see if that addresses the issue?
> 
> Yeah I think that should resolve it. When do we plan to release the change? @pavelgj Thanks for the quick response.

Great! We usually aim to release on Wednesdays +- a day.

dongyangli1226:
> > > Thanks for the PR. I merged some changes to this code earlier (unrelated change) as part of which I also removed the token refresh logic. Could you please check the current implementation at head and see if that addresses the issue?
> > 
> > 
> > Yeah I think that should resolve it. When do we plan to release the change? @pavelgj Thanks for the quick response.
> 
> Great! We usually aim to release on Wednesdays +- a day.

Awesome thank you!!

PR #716, opened 2024-07-30
feat: allow overriding location per request in vertex ai plugin

pavelgj:
```ts

configureGenkit({
  plugins: [
    vertextAI({
      location: 'us-central1',
      modelGarden: {
        models: [claude35Sonnet]
      }
    })
  ]
})

generate({
  model: claude35Sonnet,
  prompt: 'banana',
  config: {
    location: 'us-east5',
  }
})
```

### DIFF
--- a/js/plugins/vertexai/src/anthropic.ts
+++ b/js/plugins/vertexai/src/anthropic.ts
@@ -46,7 +46,7 @@ import z from 'zod';
 
 export const AnthropicConfigSchema = GenerationCommonConfigSchema.extend({
   locationOverride: z.string().optional(),
-})
+});
 
 export const claude35Sonnet = modelRef({
   name: 'vertexai/claude-3-5-sonnet',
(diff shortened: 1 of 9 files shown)


js/plugins/vertexai/src/anthropic.ts
```diff
@@ -42,6 +42,11 @@ import {
   modelRef,
 } from '@genkit-ai/ai/model';
 import { GENKIT_CLIENT_HEADER } from '@genkit-ai/core';
+import z from 'zod';
+
+export const AnthropicConfigSchema = GenerationCommonConfigSchema.extend({
+  locationOverride: z.string().optional(),
```
mbleigh, line 48: I'd just call it `location`

pavelgj: I thought about that... had a mild preference to making it explicit... but `location` is shorter... therefore better! :)
Done.

apascal07 (approved):

PR #720, opened 2024-07-31
test(go): add live tests to go ollama plugin

cabljac:
This PR resolves https://github.com/firebase/genkit/issues/609


go/plugins/ollama/ollama_live_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
+	"github.com/firebase/genkit/go/ai"
+	ollamaPlugin "github.com/firebase/genkit/go/plugins/ollama"
+	ollamaTestContainers "github.com/testcontainers/testcontainers-go/modules/ollama"
+)
+
+// setupOllamaContainer sets up the Ollama container and returns the server address.
+func setupOllamaContainer(ctx context.Context) (string, func(), error) {
```
pavelgj, line 31: This is cool... unfortunately relying on docker is problematic -- for security policy reasons it's problematic for majority of the genkit team to use.
I was thinking it's OK for the test to assume that you have ollama already installed and running (and the necessary model pulled)... maybe a comment with basic setup instructions is sufficient. 
Like other live test it should be not be enabled by default (it's OK if doesn't run as part of CI). See ex: https://github.com/firebase/genkit/blob/main/go/plugins/googleai/googleai_test.go
https://github.com/firebase/genkit/blob/main/go/plugins/vertexai/vertexai.go

they are not well documented, but to run those live tests you have to run then with flags that explicitly enable them.

pavelgj (changes requested):

go/plugins/ollama/ollama_live_test.go
cabljac: There's a module called ovai I mentioned in the discord which is a cool concept, but I'm not sure it works atm 

Idea is wrapping vertex ai with ollama API

pavelgj: I looked at it... I'm sceptical -- Ollama API (especially the multimodal cases) is slightly less expressive than Vertex Gemini APIs.

cabljac: Ah I just meant to mock llama anyway, but it's kinda irrelevant as I think it's out of date and not very active

pavelgj (approved):

PR #722, opened 2024-07-31
[Go] Added support for flow auth and Firebase auth plugin.

### LINKED ISSUE #259: [Go] Flow server API payload does not match the JS API
**Describe the bug**
The way to trigger flows in Go vs JS is not consistent at the moment.

Go: `curl -d '{"customerName": "Stimpy"}' http://localhost:3400/testAllCoffeeFlows`
JS: `curl -d '{"data": {"modelName": "vertexai/gemini-1.0-pro", "subject": "banana"}}' http://localhost:3400/jokeFlow -H "Content-Type: application/json"`

Notice that the payload has to be wrapped in `data` field for JS.

**To Reproduce**
Go: Run the `coffee-shop` sample (`go run .`) and trigger a flow using the above command.
JS: Add `startFlowServer()` to the `flow-simple-ai` sample, start the genkit server (`genkit start`) and trigger a flow using the above command. 

**Expected behavior**
Both Go and JS have similar APIs (require the "data" wrapper field).

 - OS: Linux Debian

**Go version**
go version go1.23-20240419-RC02 cl/626470163 +7f76c00fc5 X:fieldtrack,boringcrypto linux/amd64

apascal07:
Fixes #259.

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated


jba (approved):
Nicer without generics.

go/samples/firebase-auth/main.go
```diff
@@ -0,0 +1,72 @@
...
+			return errors.New("user ID does not match")
+		}
+		return nil
+	}
+	firebaseAuth, err := firebase.NewAuth(ctx, policy, true)
+	if err != nil {
+		log.Fatalf("failed to set up Firebase auth: %v", err)
+	}
+
+	flowWithRequiredAuth := genkit.DefineFlow("flow-with-required-auth", func(ctx context.Context, user string) (string, error) {
```
pavelgj, line 42: how does one get access to the auth context inside the flow?

pavelgj: I believe it's a good candidate to be put inside the `context.Context`...

```go
authContext := GetAuthContext(ctx)
```

go/genkit/flow.go
```diff
@@ -430,21 +439,21 @@ func (f *Flow[In, Out, Stream]) runJSON(ctx context.Context, authHeader string,
...
 
 // provideAuthContext provides auth context for the given auth header if flow auth is configured.
-func (f *Flow[In, Out, Stream]) provideAuthContext(ctx context.Context, authHeader string) (map[string]any, error) {
+func (f *Flow[In, Out, Stream]) provideAuthContext(ctx context.Context, authHeader string) (context.Context, error) {
 	if f.auth != nil {
-		authContext, err := f.auth.ProvideAuthContext(ctx, authHeader)
+		newCtx, err := f.auth.ProvideAuthContext(ctx, authHeader)
 		if err != nil {
-			return nil, fmt.Errorf("unauthorized: %w", err)
+			return ctx, fmt.Errorf("unauthorized: %w", err)
```
jba, line 446: It's very rare to return a valid value with an error. You should return nil as before.

go/genkit/flow.go
```diff
@@ -117,26 +117,32 @@ type flowOptions struct {
 
 type noStream = func(context.Context, struct{}) error
 
-// FlowAuth configures a auth context provider and a auth policy check for a flow.
+// FlowAuth configures an auth context provider and a auth policy check for a flow.
 type FlowAuth interface {
-	// ProvideAuthContext provides auth context from an auth header.
-	ProvideAuthContext(ctx context.Context, authHeader string) (map[string]any, error)
+	// ProvideAuthContext provides auth context from an auth header and sets it on the context.
```
jba, line 122: There are standard-ish names for these functions; see https://pkg.go.dev/context#Context.Value.

So I would call this NewContext, since it makes a new context containing auth information.

go/genkit/flow.go
```diff
@@ -117,26 +117,32 @@ type flowOptions struct {
...
+	ProvideAuthContext(ctx context.Context, authHeader string) (context.Context, error)
+
+	// SetAuthContext sets the auth context on the given context.
+	SetAuthContext(ctx context.Context, authContext map[string]any) context.Context
```
jba, line 126: This I don't really understand. Contexts are supposed to behave like functional values; you shouldn't modify them or anything inside them. If this method is supposed to do that, then can we drop it in favor of the above? If not, why not?

go/genkit/flow.go
```diff
@@ -117,26 +117,32 @@ type flowOptions struct {
...
+
+	// AuthContext retrieves the auth context from the given context.
+	AuthContext(ctx context.Context) map[string]any
```
jba, line 129: Standard name: FromContext.

jba (changes requested):

go/genkit/flow.go
apascal07: Oops, agree, didn't meant to change it.

go/genkit/flow.go
apascal07: There are two cases we're trying to cover:

1) The prod case: this is where we're given an auth token in the header and we need to make a real API call to decode it into an auth context. This is what `ProvideAuthContext` does and it returns a `context.Context` for use both by the plugin to check policy and also to propagate into flow code so that the details may be used for other calls.
2) The dev/direct run case: this is where we're not making a real API call to auth but we need to provide the auth context either from the HTTP request or directly in flow code (direct run). This auth context needs to similarly be set on `context.Context` so that it may propagate through to flow code for later use if necessary.

That's why we have `SetAuthContext` because only the plugin knows what the key it's expecting is for both policy check and for `GetAuthContext`.

go/genkit/flow.go
apascal07: Full explanation in the next comment but summarize here: it doesn't just return a new context with auth information, it actually does API calls so I'm worried that using standard names like `NewContext` may be misleading in what's happening.

go/genkit/flow.go
jba: What if SetAuthContext was called NewContext (since IIUC it is just using something it has locally) and we leave ProvideAuthContext?

jba: In either case, more doc please!

pavelgj (approved):
LGTM! Great job Alex!

go/genkit/flow.go
apascal07: Done, please take a final look to make sure it makes sense.

jba: LGTM.

jba (approved):

kevinthecheung (approved):

PR #725, opened 2024-07-31
feat: do not allow defining new actions from within other actions/flows

pavelgj:


js/core/src/registry.ts
```diff
@@ -126,10 +126,19 @@ type ActionsRecord = Record<string, Action<z.ZodTypeAny, z.ZodTypeAny>>;
  * Returns all actions in the registry.
  */
 export async function listActions(): Promise<ActionsRecord> {
+  await maybeInitializeAllPlugins();
+  return Object.assign({}, actionsById());
+}
+
+let initializedAllPlugins = false;
+export async function maybeInitializeAllPlugins() {
```
apascal07, line 134: This reads weird. It definitely *does* initialize all plugins, it's just that it may have already been done. I would suggest either `initializeAllPlugins` or `allPluginsInitialized`.

pavelgj: renamed

apascal07 (approved):

PR #755, opened 2024-08-06
[Eval] Breaking change -- Modify EvalRunKey

ssbushi:
Changes `actionId` to `actionRef`, stores the full action path. Must be submitted along with https://github.com/FirebasePrivate/genkit-ui/pull/1001

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/genkit-tools/common/src/types/eval.ts
+++ b/genkit-tools/common/src/types/eval.ts
@@ -89,7 +89,6 @@ export type EvalResult = z.infer<typeof EvalResultSchema>;
  */
 export const EvalRunKeySchema = z.object({
   actionRef: z.string().optional(),
-  datasetRef: z.string().optional(),
   evalRunId: z.string(),
   createdAt: z.string(),
 });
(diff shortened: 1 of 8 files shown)


genkit-tools/common/src/types/eval.ts
```diff
@@ -88,7 +88,7 @@ export type EvalResult = z.infer<typeof EvalResultSchema>;
  * A unique identifier for an Evaluation Run.
  */
 export const EvalRunKeySchema = z.object({
-  actionId: z.string().optional(),
+  actionRef: z.string().optional(),
```
pavelgj, line 91: there are existing/saved eval runs, this change will break those right? Can you add `actionRef` without removing `actionId`?

ssbushi: Yeah it will break. 

Yes, that's a good idea. That should be possible.

ssbushi: Given that main is only for critical fixes, I think we can merge this newer version into `next` without complications.

shrutip90 (approved):

PR #785, opened 2024-08-13
Rendering dotprompts in a new trace span

maxl0rd:
Rendering dotprompts in a new trace span for better observability.

Checklist (if applicable):

image

### DIFF
--- a/js/plugins/dotprompt/src/prompt.ts
+++ b/js/plugins/dotprompt/src/prompt.ts
@@ -28,6 +28,11 @@ import { MessageData, ModelArgument } from '@genkit-ai/ai/model';
 import { DocumentData } from '@genkit-ai/ai/retriever';
 import { GenkitError } from '@genkit-ai/core';
 import { parseSchema } from '@genkit-ai/core/schema';
+import {
+  runInNewSpan,
+  setCustomMetadataAttribute,
+  SPAN_TYPE_ATTR,
+} from '@genkit-ai/core/tracing';
 import { createHash } from 'crypto';
 import fm, { FrontMatterResult } from 'front-matter';
 import z from 'zod';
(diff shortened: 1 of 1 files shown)


js/plugins/dotprompt/src/prompt.ts
```diff
@@ -197,16 +198,41 @@ export class Dotprompt<Variables = unknown> implements PromptMetadata {
     return this._generateOptions<O>(opt);
   }
 
+  async renderInNewSpan<O extends z.ZodTypeAny = z.ZodTypeAny>(
+    opt: PromptGenerateOptions<Variables>
+  ): Promise<GenerateOptions<z.ZodTypeAny, O>> {
+    return runInNewSpan(
+      {
+        metadata: {
```
tonybaroneee, line 206: Should we include variant info in here, or will that be bundled in the name already?

tonybaroneee (approved):
Nice, thanks!

js/plugins/dotprompt/src/prompt.ts
maxl0rd: Good point. Variant isn't included in the name by default. I will add that.

js/plugins/dotprompt/src/prompt.ts
```diff
@@ -197,16 +198,41 @@ export class Dotprompt<Variables = unknown> implements PromptMetadata {
...
+          name: this.name,
```
kmandrika: I think we need like a hash of the template or something here, to serve as a proxy for version

maxl0rd:
Added an additional metadata field for the hash of the dotprompt

image

js/plugins/dotprompt/src/prompt.ts
maxl0rd: Done

PR #787, opened 2024-08-14
Fixing bugs in prompt rendering

maxl0rd:
This fixes the bug reported in https://github.com/firebase/genkit/issues/422 where `renderPrompt` does not preserve the output schema that the prompt function returns.

This also fixes the bug reported in https://github.com/firebase/genkit/issues/757 where `Dotprompt.render()` only accepts default options. With this change, the `render` and `generate` methods will accept any options, unless a model is also passed in the method, in which case it requires the custom options defined by that model.

### DIFF
--- a/js/plugins/dotprompt/src/prompt.ts
+++ b/js/plugins/dotprompt/src/prompt.ts
@@ -24,7 +24,7 @@ import {
   PromptAction,
   toGenerateRequest,
 } from '@genkit-ai/ai';
-import { GenerationCommonConfigSchema, MessageData } from '@genkit-ai/ai/model';
+import { MessageData, ModelArgument } from '@genkit-ai/ai/model';
 import { DocumentData } from '@genkit-ai/ai/retriever';
 import { GenkitError } from '@genkit-ai/core';
 import { parseSchema } from '@genkit-ai/core/schema';
@@ -42,11 +42,11 @@ import { compile } from './template.js';
 
 export type PromptData = PromptFrontmatter & { template: string };
 
-export type PromptGenerateOptions<V = unknown> = Omit<
-  GenerateOptions<z.ZodTypeAny, typeof GenerationCommonConfigSchema>,
-  'prompt' | 'model'
-> & {
-  model?: string;
+export type PromptGenerateOptions<
+  V = unknown,
+  CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
+> = Omit<GenerateOptions<z.ZodTypeAny, CustomOptions>, 'prompt' | 'model'> & {
+  model?: ModelArgument<CustomOptions>;
   input?: V;
 };
 
@@ -55,7 +55,9 @@ interface RenderMetadata {
   history?: MessageData[];
 }
 
-export class Dotprompt<Variables = unknown> implements PromptMetadata {
+export class Dotprompt<Variables = unknown>
+  implements PromptMetadata<z.ZodTypeAny>
+{
   name: string;
   variant?: string;
   hash: string;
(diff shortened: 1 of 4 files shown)


js/plugins/dotprompt/src/prompt.ts
```diff
@@ -166,9 +168,12 @@ export class Dotprompt<Variables = unknown> implements PromptMetadata {
...
   }
 
-  private _generateOptions<O extends z.ZodTypeAny = z.ZodTypeAny>(
+  private _generateOptions<
+    O extends z.ZodTypeAny = z.ZodTypeAny,
+    CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
+  >(
     options: PromptGenerateOptions<Variables>
-  ): GenerateOptions<z.ZodTypeAny, O> {
+  ): GenerateOptions<O, CustomOptions> {
```
MichaelDoyle, line 176: Why is the ordering inconsistent? Sometimes its `<O, CustomOptions> `and other times its `<CustomOptions, O>`...

maxl0rd: Unfortunately, the signature for `GenerateOptions` is `GenerateOptions<O, CustomOptions>` and the signature for `generate` is `generate<CustomOptions, O>`. Since the dotprompt `generate()` method parallels the core generate function, I kept that signature consistent.

maxl0rd: Also, this line was a bug in the previous version. It worked most of the time but only for Anys.

js/plugins/dotprompt/src/prompt.ts
```diff
@@ -42,11 +42,11 @@ import { compile } from './template.js';
 
 export type PromptData = PromptFrontmatter & { template: string };
 
-export type PromptGenerateOptions<V = unknown> = Omit<
-  GenerateOptions<z.ZodTypeAny, typeof GenerationCommonConfigSchema>,
-  'prompt' | 'model'
-> & {
-  model?: string;
+export type PromptGenerateOptions<
+  V = unknown,
```
apascal07, line 46: Any reason not to change `V` to `I` like we have everywhere else? `V` doesn't really make me think "input".

maxl0rd: That works.

apascal07:
Please add descriptive comments to all user-facing exported functions (e.g. `renderPrompt`, etc) explaining how they are to be used.

js/plugins/dotprompt/src/prompt.ts
maxl0rd: Changed `V` and `Variables` to `I` to be consistent with the rest of the codebase.

maxl0rd:
Added some better docs for public methods.

js/plugins/dotprompt/src/prompt.ts
```diff
@@ -188,25 +202,49 @@ export class Dotprompt<Variables = unknown> implements PromptMetadata {
...
 
-  render<O extends z.ZodTypeAny = z.ZodTypeAny>(
-    opt: PromptGenerateOptions<Variables>
-  ): GenerateOptions<z.ZodTypeAny, O> {
-    return this._generateOptions<O>(opt);
+  /**
+   * Renders the prompt template based on user input.
+   *
+   * @param opt Options for the prompt template, including user input variables and custom model configuration options.
+   * @returns a `GenerateOptions` object to be used with the `generate()` function.
```
apascal07, line 212: Hmm, is this true? It looks like both `render()` and `generate()` take in the same parameters.

maxl0rd: Oops. I mean `generate` from `@genkit-ai/ai` here. Not sure how to disambiguate.

maxl0rd: Done

apascal07 (approved):

PR #789, opened 2024-08-15
docs: new models.md page

kevinthecheung:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


docs/models.md
```diff
@@ -1,198 +1,645 @@
...
+- Claude 3 family of LLMs, through the [Anthropic plugin](https://thefireco.github.io/genkit-plugins/docs/plugins/genkitx-anthropic){:.external}
+- GPT family of LLMs through the [OpenAI plugin](https://thefireco.github.io/genkit-plugins/docs/plugins/genkitx-openai){:.external}
+- GPT family of LLMs through the [Azure OpenAI plugin](https://thefireco.github.io/genkit-plugins/docs/plugins/genkitx-azure-openai){:.external}
+- Command R family of LLMs through the [Cohere plugin](https://thefireco.github.io/genkit-plugins/docs/plugins/genkitx-cohere){:.external}
+- Mistral family of LLMs through the [Mistral plugin](https://thefireco.github.io/genkit-plugins/docs/plugins/genkitx-mistral){:.external}
+- Gemma 2, Llama 3, and many more open models hosted on Groq, through the
+  [Groq plugin](https://thefireco.github.io/genkit-plugins/docs/plugins/genkitx-groq){:.external}
+
+You can discover more by searching for packages starting with [`genkitx-` on
+npmjs.org](https://www.npmjs.com/search?q=genkitx){:.external}.
```
mbleigh, line 87:
```suggestion
You can discover more by searching for packages with the [`genkit-model` keyword](https://www.npmjs.com/search?q=keywords%3Agenkit-model){:.external} on NPM.
```

docs/models.md
```diff
@@ -1,198 +1,645 @@
...
+make use of these advanced parameters, you will want to also take a look at the
+documentation on dotprompt, which provides a way to specify a model, a prompt,
+and a model configuration together as one unit.
+
+You can experiment with the effect of these parameters on the output generated
+by different model and prompt combinations by using the Developer UI. Start the
+developer UI with the genkit start command and it will automatically load all of
+the models defined by the plugins configured in your project. You can quickly
+try different prompts and configuration values without having to repeatedly make
+these changes in code.
```
MichaelDoyle, line 295: This is a nice blurb about the Dev UI, but it's somewhat buried/hidden under the model parameter definitions. Consider bumping this out into a section that briefly describes using the Dev UI for interacting and experimenting with models.

docs/models.md
```diff
@@ -1,198 +1,645 @@
...
+want the model to consider, but this time by explicitly specifying the maximum
+number of tokens. Specifying a value of 1 means that the model should behave
+deterministically.
+
+Note: The astute reader might notice that the fact that these parameters are
+specific to individual models undermines the promise of being able to swap out
+models simply by passing a different value to generate. You're correct! If you
+make use of these advanced parameters, you will want to also take a look at the
+documentation on dotprompt, which provides a way to specify a model, a prompt,
+and a model configuration together as one unit.
```
MichaelDoyle, line 288: From the text - it's not clear what problem dotprompt is solving here.

MichaelDoyle (commented):
Overall really nice. Thanks!

I do worry a bit about how verbose it is, for folks who are looking to get into the code right away and/or see "what can these models _actually do_ for me".

As one example, we dedicate a lot of sentences to explaining "why" models are hosted behind API end points. We probably don't need to do this. A quick sentence about how this is a rapidly evolving field with many model providers with different (frequently changing) APIs would probably be good enough to explain why a common abstraction is needed.

docs/models.md
```diff
@@ -1,198 +1,645 @@
...
+    allergens: z.array(z.string()),
+  });
+  ```
+
+- **Retry the generate() call**. If the model you've chosen only rarely fails to
+  generate conformant output, you can treat the error as you would treat a
+  network error, and simply retry the request using some kind of incremental
+  back-off strategy.
+
+- **Generate more candidates**. Another strategy for dealing with intermittent
```
apascal07, line 409: We're going to get rid of `candidates` in the 0.6 release in favor of calling `generate()` multiple times in parallel as the end result is the same and not all models support `candidates`.

apascal07:
Overall looks good! Agree with @MichaelDoyle that it's a bit verbose but I think that we should keep it as is and try to have separate "getting started" versions of concept docs.

MichaelDoyle (approved):

PR #790, opened 2024-08-16
feat(go/plugins/firebase): add Init method to firebase go plugin

cabljac:
This PR adds an Init method to the Go firebase plugin

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/go/plugins/firebase/firebase.go
+++ b/go/plugins/firebase/firebase.go
@@ -16,31 +16,69 @@ package firebase
 
 import (
 	"context"
+	"fmt"
+	"log"
 	"sync"
 
 	firebase "firebase.google.com/go/v4"
-	"firebase.google.com/go/v4/auth"
 )
 
-type FirebaseApp interface {
-	Auth(ctx context.Context) (*auth.Client, error)
+var state struct {
+	mu      sync.Mutex
+	initted bool
+	app     *firebase.App
 }
 
-var (
-	app   *firebase.App
-	mutex sync.Mutex
-)
+// FirebasePluginConfig is the configuration for the Firebase plugin.
+type FirebasePluginConfig struct {
+	AuthOverride     *map[string]interface{} `json:"databaseAuthVariableOverride"`
+	DatabaseURL      string                  `json:"databaseURL"`
+	ProjectID        string                  `json:"projectId"`
+	ServiceAccountID string                  `json:"serviceAccountId"`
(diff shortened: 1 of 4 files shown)


go/plugins/firebase/firebase.go
```diff
@@ -16,31 +16,69 @@ package firebase
...
+}
+
+// Init initializes the Firebase app with the provided configuration.
+// If called more than once, it logs a message and returns nil.
+func Init(ctx context.Context, cfg *FirebasePluginConfig) error {
+	state.mu.Lock()
+	defer state.mu.Unlock()
+
+	if state.initted {
+		log.Println("firebase.Init: already called, returning without reinitializing")
```
apascal07, line 48: I wonder if this is the right behavior or if it should error on subsequent calls. Generally this no-op behavior is only deterministic if you are not passing a configuration that you are expecting to take effect. Here, a no-op means you don't know if the initted config matches what you're expecting.

cabljac: Good spot, I'll update it :)

go/plugins/firebase/firebase.go
```diff
@@ -16,31 +16,78 @@ package firebase
...
+
+// Init initializes the Firebase app with the provided configuration.
+// If called more than once, it logs a message and returns nil.
+func Init(ctx context.Context, cfg *FirebasePluginConfig) error {
+	state.mu.Lock()
+	defer state.mu.Unlock()
+
+	if state.initted {
+		// throw an error if the app is already initialized
+		return fmt.Errorf("firebase.Init: Firebase app already initialized")
```
apascal07, line 48:
```suggestion
		return errors.New("firebase.Init: Firebase app already initialized")
```

go/plugins/firebase/firebase.go
```diff
@@ -16,31 +16,78 @@ package firebase
...
-		app = newApp
+	state.app = app
+	state.initted = true
+
+	return nil
+}
+
+// UnInit uninitializes the Firebase app.
+
+func UnInit() {
```
apascal07, line 74: Test is in the same package, better to drop this and just clean up state directly. We don't want to expose a function like this.

apascal07 (changes requested):

cabljac:
closing in favor of https://github.com/firebase/genkit/pull/791

PR #791, opened 2024-08-16
feat(go/plugins/firebase): add firestore retriever to go firebase plugin

cabljac:
This PR adds a firestore retriever helper to the Go plugin

(Merge https://github.com/firebase/genkit/pull/790 first)

Resolves https://github.com/firebase/genkit/issues/711

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


go/plugins/firebase/firebase.go
```diff
@@ -16,31 +16,77 @@ package firebase
...
+		return fmt.Errorf("firebase.Init: %w", err)
+	}
+
+	state.app = app
+	state.initted = true
+
+	return nil
+}
+
+func UnInit() {
```
jba, line 73: Don't export this if you can avoid it.

jba: (And try hard to avoid it!)

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+			return nil, fmt.Errorf("retriever request is nil")
+		}
+
+		options := RetrieverRequestOptions{Limit: 10, DistanceMeasure: cfg.DistanceMeasure}
+
+		if req.Options != nil {
+			// Ensure that the options are of the correct type
+			parsedOptions, ok := req.Options.(RetrieverRequestOptions)
+			if !ok {
+				return nil, fmt.Errorf("firebase.Retrieve options have type %T, want %T", req.Options, &RetrieverRequestOptions{})
```
jba, line 59: Do you want a struct (line 57) or a pointer to a struct (this line)?

jba: (My college roommate pulled an all-nighter writing C code. In the morning he said, "I'm so tired, I don't know whether I want a shower or a pointer to a shower.")

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+			return nil, fmt.Errorf("%s index embedding failed: %v", provider, err)
+		}
+
+		if eres == nil || len(eres.Embeddings) == 0 {
+			return nil, fmt.Errorf("embedding result is nil or empty")
+		}
+
+		coll := cfg.Client.Collection(cfg.Collection)
+		if coll == nil {
+			return nil, fmt.Errorf("collection is nil")
```
jba, line 84: The error here is that the collection path is invalid. That will have more meaning to the user.

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		embedding := eres.Embeddings[0].Embedding
+
+		distanceMeasure := options.DistanceMeasure | cfg.DistanceMeasure
```
jba, line 89: I don't think you want a bitwise "or" here. I guess that works if one is zero and one is non-zero, but what if both are non-zero? I think you'll have to write an if statement.

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+
+		query := coll.FindNearest(cfg.VectorField, embedding, options.Limit, distanceMeasure, nil)
+
+		// Execute the query
+		iter := query.Documents(ctx)
+		if iter == nil {
+			return nil, fmt.Errorf("document iterator is nil")
```
jba, line 96: This can't be nil (https://github.com/googleapis/google-cloud-go/blob/firestore/v1.16.0/firestore/query.go#L1089).
But if you really want to be defensive I guess that's OK.

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		gotDocs, err := iter.GetAll()
+
```
jba, line 100: remove blank line

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get documents: %v", err)
```
jba, line 102: nit: we don't typically write "failed to" or "error during" because it's obvious.  "getting documents" is fine here.

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		genkitDocs := make([]*ai.Document, len(gotDocs))
+
```
jba, line 106: remove blank line

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,138 @@
...
+				continue
+			}
+
+			out := make(map[string]any)
+			out["content"] = content
+
+			metadata := make(map[string]any)
+			if len(cfg.MetadataFields) > 0 {
+				for _, field := range cfg.MetadataFields {
+					metadata[field] = doc.Data()[field]
```
jba, line 120: Repeatedly calling doc.Data is expensive.

go/plugins/firebase/retriever_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	testDocument := ai.DocumentFromText("Test document", map[string]any{"metadata": "test"})
+
+	req := &ai.RetrieverRequest{
+		Document: testDocument,
+		Options:  RetrieverRequestOptions{Limit: 2, DistanceMeasure: firestore.DistanceMeasureEuclidean},
+	}
+
+	fmt.Print(&retriever, &req)
```
jba, line 104: Remove this and other debugging code.
If you want to leave in some printing, use t.Log/Logf.

go/plugins/firebase/retriever_test.go
```diff
@@ -0,0 +1,129 @@
...
+	if len(resp.Documents) != 2 {
+		t.Errorf("expected 2 documents, got %d", len(resp.Documents))
+	}
+
+	for _, doc := range resp.Documents {
+		if doc == nil {
+			t.Error("retrieved document is nil")
+		}
+	}
+	fmt.Printf("Doc with content \n\n %s \n\n retrieved \n\n", resp.Documents[0].Content[0].Text)
```
jba, line 126: I'm confused: where does this test add documents to the collection?

Anyway, assuming you know the documents, you can do a little better here by checking their content, or at least part of their content (like a substring).

jba (changes requested):

go/plugins/firebase/auth_test.go
```diff
@@ -93,6 +93,7 @@ func TestProvideAuthContext(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+
```
apriil15, line 96: maybe we can remove this blank line

```suggestion
```

go/plugins/firebase/retriever.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, fmt.Errorf("%s index embedding failed: %v", provider, err)
+		}
+
+		if eres == nil || len(eres.Embeddings) == 0 {
+			return nil, fmt.Errorf("embedding result is nil or empty")
+		}
+
+		coll := cfg.Client.Collection(cfg.Collection)
+		if coll == nil {
+			return nil, fmt.Errorf("collection path is invalid")
```
jba, line 84: `fmt.Errorf("DefineFirestoreRetriever: collection path %q is invalid", cfg.Collection)`

Also I wonder if we could call this outside the closure to check the error earlier.

go/plugins/firebase/retriever_test.go
jba: I still don't see where the documents come from. If they have to be present in the DB already, add a comment explaining that.

jba (changes requested):

go/plugins/firebase/retriever_test.go
cabljac: Yup! will sort this out, applying these changes when I can in between some js work

PR #793, opened 2024-08-16
Update to suggest defineDotPrompt rather than definePrompt

i2amsam:
Update prompts.md to suggest defineDotPrompt rather than definePrompt 

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated


docs/prompts.md
```diff
@@ -17,18 +17,18 @@ approaches are also supported by working with prompts directly.
 ## Defining prompts
```
MichaelDoyle, line 17: Let's do something with the paragraph above 

> (Most developers will...)

I propose we scratch it entirely, and then in the prior paragraph, mention that Genkit provides Dotprompt as a way to maintain prompts in source files, etc.

docs/prompts.md
```diff
@@ -17,18 +17,18 @@ approaches are also supported by working with prompts directly.
...
 generate({
   model: 'googleai/gemini-1.5-flash-latest',
-  prompt: 'You are a helpful AI assistant named Walt.',
+  prompt: 'You are a helpful AI assistant named Walt, say hello',
 });
 ```
 
-In most cases, you will need to include some customer provided inputs in your prompt.
+In some cases you will need to include some customer provided inputs in your prompt.
 You could define a function to render them like this.
```
MichaelDoyle, line 32: Scratch the function business, and simplify this to be:

> In some cases you will need to include customer provided inputs in your prompt.
You might use Javascript's Template Literals to do this.
>
>  ```ts
> generate({
>   model: 'googleai/gemini-1.5-flash-latest',
>   prompt: `You are a helpful AI assistant named Walt. Say hello to ${name}.`,
> });
> ```

docs/prompts.md
```diff
@@ -42,65 +42,81 @@ generate({
 });
 ```
 
-One shortcoming of defining prompts in your code is that testing requires executing
-them as part of a flow. To facilitate more rapid iteration, Genkit provides a facility
-to define your prompts and run them in the Developer UI.
+However, to facilitate more rapid iteration, Genkit provides a way to define your prompts in a standard format which enables this prompt templating as well as rapid testing in the Developer UI.
```
MichaelDoyle, line 45: Suggestion:

> However, Genkit provides a way to define your prompts in a standardized format that facilitates more advanced templating and rapid testing in the Developer UI.

docs/prompts.md
```diff
@@ -42,65 +42,81 @@ generate({
...
 
-Use the `definePrompt` function to register your prompts with Genkit.
+Use the `defineDotprompt` function to define these structred prompts.
```
MichaelDoyle, line 47: s/structred/structured/

docs/prompts.md
```diff
@@ -42,65 +42,81 @@ generate({
...
+In the Genkit Developer UI, you can run any prompt you have defined in this way.  This allows you to experiment with individual prompts outside of the scope of
 the flows in which they might be used.
 
+<img src="resources/prompts-in-developer-ui.png" alt="The developer UI showing JSON response to the threeGreetingsPrompt" class="screenshot-attempt-right">
+
 ## Dotprompt
 
-Genkit includes the [Dotprompt](./dotprompt.md) library which adds additional
-functionality to prompts.
+See the [Dotprompt](./dotprompt.md) page for more features of the Dotprompt library, including
```
MichaelDoyle, line 117: Why not consolidate these pages into one? You've already started to do that by introducing input and output schema conformance above, which is also listed below as an advanced topic to be covered later!

MichaelDoyle:
I made a bunch of comments above, but I also want to make the following suggestion:

_I think we should combine this page and the dotprompt page into one._ The two pages currently overlap, and are in conflict with each other in a few different ways. 

For example, this page suggests developers should start with defining prompts in code (`defineDotPrompt`) and yet, the dotprompt documentation suggests that developers should start with the file format first, and introduces `defineDotPrompt` is an _alternative_ method if you really want to write it as code instead. 

Both are covering input/out schemas, but in varying levels of detail, etc.

docs/prompts.md
i2amsam: Ack. Updated to remove this paragram

docs/prompts.md
i2amsam: Good call, updated

docs/prompts.md
i2amsam: My thinking is that `dotprompt` is too complex for the intro doc since it covers many usecases, so it's better to have one intro page for prompting in general and a deep-dive page for dotprompt

kevinthecheung: I would like to try to consolidate the pages to see how it looks but I can take it on myself in a follow-up PR

kevinthecheung (approved):

docs/prompts.md
```diff
@@ -1,106 +1,118 @@
...
+    },    
+  },
+  `You are a helpful AI assistant named Walt. Say hello to {{name}}, write a response for each of the styles requested`
 );
 ```
 
-In the Genkit Developer UI, you can run any prompt you have defined in this way.
-This allows you to experiment with individual prompts outside of the scope of
-the flows in which they might be used.
+You can then call `generate` on that prompt and work with the structred output in the response
```
kevinthecheung, line 96:
```suggestion
You can then call `generate` on that prompt and work with the structured output in the response:
```

MichaelDoyle (approved):

PR #796, opened 2024-08-16
[JS] Support multiple imagen models in VertexAI

MichaelDoyle:
Also adds Imagen3 and Imagen3 Fast. Both are currently allowlist only - I'm still waiting access to try these specific models.

image

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated

### DIFF
--- a/js/plugins/vertexai/src/imagen.ts
+++ b/js/plugins/vertexai/src/imagen.ts
@@ -21,6 +21,7 @@ import {
   GenerationCommonConfigSchema,
   getBasicUsageStats,
   modelRef,
+  ModelReference,
 } from '@genkit-ai/ai/model';
 import { GoogleAuth } from 'google-auth-library';
 import z from 'zod';
@@ -45,6 +46,11 @@ export const imagen2 = modelRef({
   name: 'vertexai/imagen2',
   info: {
     label: 'Vertex AI - Imagen2',
+    versions: [
+      'imagegeneration@006',
+      'imagegeneration@005',
+      'imagegeneration@002',
+    ],
     supports: {
       media: false,
       multiturn: false,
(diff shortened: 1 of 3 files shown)


js/plugins/vertexai/src/imagen.ts
```diff
@@ -106,7 +153,15 @@ interface ImagenInstance {
...
+export function imagenModel(
+  name: string,
+  client: GoogleAuth,
+  options: PluginOptions
+) {
+  const modelName = `vertexai/${name}`;
+  const model: ModelReference<z.ZodTypeAny> = SUPPORTED_IMAGEN_MODELS[name];
+  if (!model) throw new Error(`Unsupported model: ${name}`);
+
   const predictClients: Record<
```
MichaelDoyle, line 165: I'm not changing this, but this code was curious to me. Does this actually provide any value? It appears to be intended as some type of cache/memoization, but is it actually providing that function?

js/plugins/vertexai/src/imagen.ts
```diff
@@ -53,9 +59,50 @@ export const imagen2 = modelRef({
...
+    supports: {
+      media: false,
+      multiturn: false,
+      tools: false,
+      systemRole: false,
+      output: ['media'],
+    },
+  },
+  version: 'imagen-3.0-generate-001',
+  configSchema: ImagenConfigSchema,
```
mbleigh, line 80: Have you confirmed that Imagen3 has the same config schema? 1 -> 2 was pretty different.

MichaelDoyle: Once I get allowlisted, I'll take another pass through. The way I interpret the docs, everything seems to be standardized across all of the models now, but they do support different criteria on the various options (e.g. different aspect ratios, languages, etc), some params are n/a to certain models, etc.

Imagen2 also got upgraded to v006, so there will probably be some changes applicable there as well.

MichaelDoyle:
Vertex AI has a single schema across all imagen models; I've updated it with the latest changes.

js/plugins/vertexai/src/imagen.ts
pavelgj: It's caching clients per location (location can be passes in the request)... why do it you think it may not be doing that?

pavelgj (approved):

js/plugins/vertexai/src/imagen.ts
MichaelDoyle: I think it just looks funny, coming from more of a pure OO background. Thanks.

PR #797, opened 2024-08-16
Local file based DatasetStore implementation and Tools API changes

shrutip90:
Local file based DatasetStore implementation and Tools API changes

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * A local, file-based DatasetStore implementation.
+ */
+export class LocalFileDatasetStore implements DatasetStore {
+  private readonly storeRoot;
+  private readonly indexFile;
+  private readonly INDEX_DELIMITER = '\n';
+  private static cachedDatasetStore: LocalFileDatasetStore | null = null;
+
+  private constructor() {
```
ssbushi, line 42: Let us make the constructor take the root directory location, as discussed earlier, to make any swaps easier (if we need to down the line).

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+    return await rm(filePath);
+  }
+
+  private generateRootPath(): string {
+    const rootHash = crypto
+      .createHash('md5')
+      .update(process.cwd() || 'unknown')
+      .digest('hex');
+    return path.resolve(os.tmpdir(), `.genkit/${rootHash}/datasets`);
```
ssbushi, line 203: We should not store under `tmpdir`. Use `process.cwd()` instead?

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+  private generateDatasetId(): string {
+    return uuidv4();
+  }
+
+  private generateFileName(datasetId: string): string {
+    return `${datasetId}.json`;
+  }
+
+  private getIndexFilePath(): string {
+    return path.resolve(this.storeRoot, 'index.txt');
```
ssbushi, line 215: IMO, why not make this a simple JSON file, that stores a mapping (datasetId to metadata)? No need to split and parse in that case.

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+    data: Dataset,
+    displayName?: string
+  ): Promise<DatasetMetadata> {
+    const datasetId = this.generateDatasetId();
+    const filePath = path.resolve(
+      this.storeRoot,
+      this.generateFileName(datasetId)
+    );
+
+    if (!fs.existsSync(filePath)) {
```
ssbushi, line 79: Should this be `if (fs.existsSync...)` ?

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+      throw new Error(
+        `Create dataset failed: file already exists at {$filePath}`
+      );
+    }
+
+    logger.info(`Saving Dataset to ` + filePath);
+    await writeFile(filePath, JSON.stringify(data));
+
+    const metadata = {
+      datasetId: datasetId,
```
ssbushi, line 90:
```suggestion
      datasetId,
```

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+      size: Array.isArray(data) ? data.length : data.samples.length,
+      uri: 'file://' + filePath,
+      version: 1,
+      displayName: displayName,
+      createTime: Date.now().toString(),
```
ssbushi, line 95: Nit: `const now = Date.now().toString(); createTime: now, updateTime: now`

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+      path.resolve(this.indexFile),
+      JSON.stringify(metadata) + this.INDEX_DELIMITER
+    );
+    return metadata;
+  }
+
+  async updateDataset(req: UpdateDatasetRequest): Promise<DatasetMetadata> {
+    if (!req.datasetId) {
+      return this.createDatasetInternal(req.patch, req.displayName);
+    }
```
ssbushi, line 113: Should we support this behavior?

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+      datasetId: datasetId,
+      size: Array.isArray(req.patch)
+        ? req.patch.length
+        : req.patch.samples.length,
+      uri: 'file://' + filePath,
+      version: prevMetadata.version++,
+      displayName: req.displayName,
+      createTime: prevMetadata.createTime,
+      updateTime: Date.now().toString(),
+    };
```
ssbushi, line 144: This looks like it is overriding _all_ rows during update. This works for now, but it had a more efficient upsert implementation in mind (add new rows, and update existing rows, identify rows by deterministicHash(row)). WDYT?

ssbushi: This also means that we ill have to slightly change the way we store the dataset files (keep a row identifier on file)

genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -0,0 +1,221 @@
...
+    return metadatas;
+  }
+
+  async deleteDataset(id: string): Promise<void> {
+    const filePath = path.resolve(this.storeRoot, this.generateFileName(id));
+    if (!fs.existsSync(filePath)) {
+      return;
+    }
+    return await rm(filePath);
+  }
```
ssbushi, line 196: This should also update the index?

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Don't remember talking about this, but updated it.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Done. This is harder to mock. So, I have used regular expression matching in tests and also got rid of the URI, since it not required currently anyway.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Thanks, ya, this looks much better now.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Done.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Done.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Done.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Left it as is for now. Asked on chat, please take another look and let me know.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Done.

genkit-tools/common/src/eval/localFileDatasetStore.ts
ssbushi: Just throw if datasetId is missing.

genkit-tools/common/src/types/apis.ts
```diff
@@ -132,3 +132,18 @@ export const GetEvalRunRequestSchema = z.object({
...
+  data: EvalFlowInputSchema,
+  displayName: z.string().optional(),
+});
+
+export type CreateDatasetRequest = z.infer<typeof CreateDatasetRequestSchema>;
+
+export const UpdateDatasetRequestSchema = z.object({
+  /** Supports upsert */
+  patch: EvalFlowInputSchema,
+  datasetId: z.string().optional(),
```
ssbushi, line 146: This can be required.

genkit-tools/common/tests/eval/localFileDatasetStore_test.ts
```diff
@@ -0,0 +1,341 @@
...
+      );
+
+      const datasetMetadata = await DatasetStore.createDataset(
+        CREATE_DATASET_REQUEST
+      );
+
+      expect(fs.promises.writeFile).toHaveBeenCalledTimes(2);
+      expect(fs.promises.writeFile).toHaveBeenNthCalledWith(
+        1,
+        expect.stringMatching(
```
ssbushi, line 153: How about `expect.stringContaining`? No need for a regex check IMO

genkit-tools/common/tests/eval/localFileDatasetStore_test.ts
```diff
@@ -0,0 +1,341 @@
...
+        ),
+        JSON.stringify(CREATE_DATASET_REQUEST.data)
+      );
+      const metadataMap = {
+        [SAMPLE_DATASET_ID_1]: SAMPLE_DATASET_METADATA_1_V1,
+      };
+      expect(fs.promises.writeFile).toHaveBeenNthCalledWith(
+        2,
+        expect.stringMatching(
+          /^(?:\/|[a-zA-Z]:\\)(?:[^\\\/:*?"<>|]+[\\\/])*datasets\/index\.json$/
```
ssbushi, line 164: Same here and below.

shrutip90: Done.

genkit-tools/common/tests/eval/localFileDatasetStore_test.ts
shrutip90: Ah, yes, that works.

genkit-tools/common/src/types/apis.ts
shrutip90: Done.

genkit-tools/common/src/eval/localFileDatasetStore.ts
shrutip90: Removed since we don't need this after making it required.

ssbushi (approved):

PR #802, opened 2024-08-19
Provide service account creds to gcloud from environment variables

schnecle:
Allow customers to pass Google Service Account credentials directly via an environment variable following https://github.com/googleapis/google-auth-library-nodejs?tab=readme-ov-file#loading-credentials-from-environment-variables

Checklist (if applicable):
   Updated the cat-eval example to validate Firestore, GCloud, and Vertex
- [ ] Changelog updated
- [x] Docs updated
   Added docs around the new option to pass in credentials directly via environment variable

### DIFF
--- a/docs/plugins/google-cloud.md
+++ b/docs/plugins/google-cloud.md
@@ -47,13 +47,48 @@ export default configureGenkit({
 
 When running in production, your telemetry gets automatically exported.
 
-The plugin requires the Google Cloud project ID and your Google Cloud project credentials. If you're running your flow from a Google Cloud environment (Cloud Functions, Cloud Run, etc), the project ID and credentials are set automatically. Running in other environments requires setting the `GCLOUD_PROJECT` environment variable to your Google Cloud project and authenticating using the `gcloud` tool:
+### Authentication
+
+The plugin requires the Google Cloud project ID and your Google Cloud project credentials. If you're running your flow from a Google Cloud environment (Cloud Functions, Cloud Run, etc), the project ID and credentials are set automatically. 
+
+#### Application Default Credentials
+
+Running in other environments requires setting the `GCLOUD_PROJECT` environment variable to your Google Cloud project and authenticating using the `gcloud` tool:
 
 ```posix-terminal
 gcloud auth application-default login
 ```
 
 For more information, see the [Application Default Credentials](https://cloud.google.com/docs/authentication/provide-credentials-adc) docs.
+
+#### Service Account Credentials
+
+If you are using a service account and running outside of a Google Cloud environment, you can set your credentials as an environment variable. Follow instructions here to [set up your Google Cloud Service Account Key](https://cloud.google.com/iam/docs/keys-create-delete#creating).
+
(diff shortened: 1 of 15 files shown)


andrewbrook (approved):

kmandrika (approved):

docs/plugins/google-cloud.md
```diff
@@ -47,14 +47,49 @@ export default configureGenkit({
...
 gcloud auth application-default login
 ```
 
 For more information, see the [Application Default Credentials](https://cloud.google.com/docs/authentication/provide-credentials-adc) docs.
 
+#### Service Account Credentials
+
+If you are using a service account and running outside of a Google Cloud environment, you can set your credentials as an environment variable. Follow instructions here to [set up your Google Cloud Service Account Key](https://cloud.google.com/iam/docs/keys-create-delete#creating).
+
+Once you have downloaded the key file, you can specify the credentials in two ways a file location using the `GOOGLE_APPLICATION_CREDENTIALS` environment variable or directly copy the contents of the json file to the environment variable `GCLOUD_SERVICE_ACCOUNT`.
```
pavelgj, line 68: Not a fan of `GCLOUD_SERVICE_ACCOUNT` env var name. It makes me thing it should be just the service account name, not the credentials JSON blob.

How about  `GOOGLE_SERVICE_ACCOUNT_CREDS_JSON` or at least `GOOGLE_SERVICE_ACCOUNT_CREDS`?

js/plugins/firebase/src/index.ts
```diff
@@ -42,19 +42,51 @@ interface FirestorePluginParams {
...
+        process.env.GCLOUD_SERVICE_ACCOUNT
+      );
+      const authOptions = { credentials: serviceAccountCreds };
+      const authClient = new GoogleAuth(authOptions);
+
+      projectId = await authClient.getProjectId();
+      credentials = await authClient.getCredentials();
+    } else {
+      const authClient = new GoogleAuth();
+      projectId = params?.projectId || (await getProjectId(authClient));
```
pavelgj, line 61: I think this line (and line 57) should be outside of the if/else statement. Explicitly specified project id (`params?.projectId`) should take precedence over the one returns from auth client.

schnecle: I waffled on this because if the projectId differs from the credentials provided, would this lead to auth errors? I'm happy to take this out of here, but it felt odd to allow developers to override the projectId to one that differs from the service account credential.

docs/plugins/google-cloud.md
schnecle: yeah, I also didn't love the name, but also didn't love how long it can quickly get. If you're ok with `GOOGLE_SERVICE_ACCOUNT_CREDS`, I'll do that for brevity.

js/plugins/firebase/src/index.ts
schnecle: I made the change, but I do think it might be worth considering.

pavelgj (approved):

PR #818, opened 2024-08-26
Added metricExportTimeoutMillis to google-cloud.md

chees:
Added `metricExportTimeoutMillis: 5_000,` to the telemetryConfig. This is needed when you define `metricExportIntervalMillis` like in the example. Without it won't run due to errors.

### DIFF
--- a/docs/plugins/google-cloud.md
+++ b/docs/plugins/google-cloud.md
@@ -88,6 +88,7 @@ googleCloud({
       '@opentelemetry/instrumentation-net': { enabled: false },
     },
     metricExportIntervalMillis: 5_000,
+    metricExportTimeoutMillis: 5_000,
   },
 });
 ```


docs/plugins/google-cloud.md
```diff
@@ -88,6 +88,7 @@ googleCloud({
       '@opentelemetry/instrumentation-net': { enabled: false },
     },
     metricExportIntervalMillis: 5_000,
+    metricExportTimeoutMillis: 5_000,
```
bryanatkinson, line 91: The default values for both of these fields is 300_000. I'd suggest having this example use those defaults explicitly. Using 5s intervals is useful during development, but could end up using a lot more metric quota than is needed if deployed to production.

My guess is that this example was failing because the export interval would be lower than the default timeout.

bryanatkinson (approved):

docs/plugins/google-cloud.md
MichaelDoyle: @chees if you can make the change above, we'll get this merged in asap. Thanks!

MichaelDoyle:
Quick ping @chees if you can make the change above, we'll get this merged in asap. Thanks!

PR #822, opened 2024-08-27
Makes inputSchema optional for tools.

### LINKED ISSUE #814: [Tooling] Tools fail if defined without arguments
**Describe the bug**

I have a simple tool that returns some data from a database. It does not need an argument. As a result of not having an argument, it fails in ways that are hard to debug. 

When invoking the tool directly in the dev ui, it works appropriately.

However, when used in a flow. the tool (with no input schema) errors with: 

`INVALID_ARGUMENT: Schema validation failed. Parse Errors: - tools.2.inputSchema: must be object `

with z.object({}), it errors in the flow and direct invocation with:
- flow:
`[GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash-latest:generateContent: [400 Bad Request] * GenerateContentRequest.tools[0].function_declarations[2].parameters.properties: should be non-empty for OBJECT type`

- tool:
`TRPCClientError: INVALID_ARGUMENT: Schema validation failed. Parse Errors:

- (root): must be object`

with z.any() the tool works, and returns a populated object, but the flow errors with 

`Cannot read properties of undefined (reading 'toUpperCase')
TypeError: Cannot read properties of undefined (reading 'toUpperCase')
    at convertSchemaProperty (/Users/connor/fora-echo/node_modules/@genkit-ai/googleai/lib/gemini.js:234:78)`

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - mac OS: 14.4 (23E214)

mbleigh:
Fixes #814 and also fixes a nasty regression with the new `generateAction` that caused `response.toHistory()` to stop working.

There's still an issue with tool calling and `toHistory`, which is that the tool request / responses are elided from history once the LLM comes back with a full response. We should fix that.

Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -16,10 +16,10 @@
 
 import {
   Action,
-  GenkitError,
-  StreamingCallback,
   config as genkitConfig,
+  GenkitError,
   runWithStreamingCallback,
+  StreamingCallback,
 } from '@genkit-ai/core';
 import { lookupAction } from '@genkit-ai/core/registry';
 import { toJsonSchema, validateSchema } from '@genkit-ai/core/schema';
@@ -27,8 +27,8 @@ import { z } from 'zod';
 import { DocumentData } from './document.js';
 import { extractJson } from './extract.js';
 import {
-  GenerateUtilParamSchema,
   generateAction,
+  GenerateUtilParamSchema,
   inferRoleFromParts,
 } from './generateAction.js';
 import {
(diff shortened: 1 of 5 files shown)


js/ai/src/generate.ts
```diff
@@ -610,7 +610,11 @@ export async function generate<
 
   return await runWithStreamingCallback(
     resolvedOptions.streamingCallback,
-    async () => new GenerateResponse<O>(await generateAction(params))
+    async () =>
+      new GenerateResponse<O>(
```
pavelgj, line 614: Do you mind adding a quick unit test for this?

pavelgj (approved):

js/ai/src/generate.ts
mbleigh: Done.

PR #823, opened 2024-08-27
fix(go): remove unnecessary nil check for lookup model

yukinagae:
Remove unnecessary nil check as lookup model (`ai.LookupModel(provider, name)`) returns `(*ai.modelActionDef)(nil)` when no model is found, which is different from `nil` in this case.

The below playground shows how it looks like:
https://go.dev/play/p/u-OvxaXc3aZ

Also, this article explains this problem.
https://vitaneri.com/posts/check-for-nil-interface-in-go 

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated

### DIFF
--- a/genkit-tools/cli/config/main.go.template
+++ b/genkit-tools/cli/config/main.go.template
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log"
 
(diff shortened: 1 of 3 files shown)


yukinagae:
I'll close this PR because removing the nil check in this case seems like bad practice. Instead of removing the nil check, I'll modify `ai.LookupModel(provider, name)` to properly return nil when no model is found.

yukinagae:
I've submitted an issue instead of this PR: https://github.com/firebase/genkit/issues/847

PR #751, opened 2024-08-06
refactor(go/plugins/ollama): use ollama go client in plugin

cabljac:
This PR refactors the existing ollama plugin to use the ollama go client (which is used by ollama itself)

This will keep us in sync with ollama better, and make adding additional features easier as ollama adds them.

The PR also adds in embedders for ollama, resolving https://github.com/firebase/genkit/issues/728

During testing I also found some logic which could use a nil pointer check (in the validateCandidate logic, e.g called by generate.go in the `ai` package)

Checklist (if applicable):
- [x] Tested (manually, unit tested, etc.)
- [ ] Changelog updated
- [ ] Docs updated

PR #841, opened 2024-08-30
feat(go): Add ollama embeddings support

huangjeff5:
Checklist (if applicable):
- [x ] Tested manually

### DIFF
--- a/go/plugins/ollama/embed.go
+++ b/go/plugins/ollama/embed.go
@@ -26,7 +26,7 @@ import (
 )
 
 type EmbedOptions struct {
-	Model string `json:"model,omitempty"`
+	Model string `json:"model"`
 }
 
 type ollamaEmbedRequest struct {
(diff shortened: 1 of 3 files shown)


cabljac:
Hey Jeff I added embeddings here - https://github.com/firebase/genkit/pull/751

huangjeff5:
> Hey Jeff I added embeddings here - #751

Discussed offline; we will continue integrating directly with REST API so that we don't add an unnecessary dependency (SDK).

go/plugins/ollama/embed.go
```diff
@@ -0,0 +1,168 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/firebase/genkit/go/ai"
+)
+
+var supportedEmbeddingModels = []string{
```
pavelgj, line 28: we should not be hardcoding any models... like we do with genai models -- model must be specified in the ollama config.

go/plugins/ollama/embed.go
```diff
@@ -0,0 +1,152 @@
...
+
+// DefineEmbedder defines an embedder with a given server address.
+func DefineEmbedder(serverAddress string, model string) ai.Embedder {
+	state.mu.Lock()
+	defer state.mu.Unlock()
+	if !state.initted {
+		panic("ollama.Init not called")
+	}
+	return ai.DefineEmbedder(provider, serverAddress, func(ctx context.Context, req *ai.EmbedRequest) (*ai.EmbedResponse, error) {
+		if req.Options == nil {
```
pavelgj, line 135: does this also need 
```go
if req.Options.Model == "" {
  req.Options.Model = model
}
```
?

Otherwise it looks like `req.Options.Model` is required field

pavelgj (approved):

PR #863, opened 2024-09-05
fix: use google gemini 1.5 models against the v1 API

pavelgj:


js/plugins/googleai/src/index.ts
```diff
@@ -71,7 +71,7 @@ export const googleAI: Plugin<[PluginOptions] | []> = genkitPlugin(
           googleAIModel(name, options?.apiKey, undefined, options?.baseUrl)
         ),
         ...Object.keys(SUPPORTED_V15_MODELS).map((name) =>
-          googleAIModel(name, options?.apiKey, 'v1beta', options?.baseUrl)
+          googleAIModel(name, options?.apiKey, undefined, options?.baseUrl)
```
apascal07, line 74: Does it default to latest when unset?

apascal07 (approved):

js/plugins/googleai/src/index.ts
pavelgj: yes. It's an override option for the googleai SDK. undefined goes to whatever is the latest considered by the SDK

PR #877, opened 2024-09-08
feat: remove trace store and send traces to telemetry server

pavelgj:
PR train... 
parent: https://github.com/firebase/genkit/pull/876


js/core/src/tracing.ts
```diff
@@ -93,10 +90,11 @@ export async function cleanUpTracing(): Promise<void> {
  * Returns `undefined` if no trace store implementation is configured.
```
apascal07, line 90: Update comment here and any other outdated one.

PR #892, opened 2024-09-11
Updates isRoot logic to no longer assume root when calling newTrace.

bryanatkinson:
This can be called multiple times when running child flows.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/core/src/tracing/instrumentation.ts
+++ b/js/core/src/tracing/instrumentation.ts
@@ -43,6 +43,7 @@ export async function newTrace<T>(
   },
   fn: (metadata: SpanMetadata, rootSpan: ApiSpan) => Promise<T>
 ) {
+  const isRoot = traceMetadataAls.getStore() ? false : true;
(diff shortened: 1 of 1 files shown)


js/core/src/tracing/instrumentation.ts
```diff
@@ -43,6 +43,7 @@ export async function newTrace<T>(
   },
   fn: (metadata: SpanMetadata, rootSpan: ApiSpan) => Promise<T>
 ) {
+  const isRoot = traceMetadataAls.getStore() ? false : true;
```
andrewbrook, line 46: opt nit: `const isRoot = !traceMetadataAls.getStore()`

andrewbrook (approved):

js/core/src/tracing/instrumentation.ts
bryanatkinson: Added

pavelgj (approved):

PR #900, opened 2024-09-12
[WIP] Implicit plugins

kmandrika:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/core/src/plugin.ts
+++ b/js/core/src/plugin.ts
@@ -26,15 +26,10 @@ export interface Provider<T> {
 }
 
 export enum PluginProvidesType {
-  UNSPECIFIED = 0x0,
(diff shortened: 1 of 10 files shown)


js/core/src/plugin.ts
```diff
@@ -25,12 +25,25 @@ export interface Provider<T> {
   value: T;
 }
 
+export enum PluginProvidesType {
```
kmandrika, line 28: FYI, this is wrong, will fix.

kmandrika: Needs to work as a bitmask

pavelgj:
There are massive changes w.r.t. @apascal07 is working on in next branch.

PR #902, opened 2024-09-13
Remove durable flows

pavelgj:


debkanchan:
why tho?

pavelgj:
> why tho?

We want to simplify the core framework, durable flows add a lot of complexity. We are looking to support stateful agentic use cases via a more focused framework rather than via general purpose workflow engine.

js/flow/src/flow.ts
```diff
@@ -790,43 +397,38 @@ export class Flow<
...
-        const state = await this.runDirectly(input, {
+        const result = await this.invoke(input, {
           streamingCallback: ((chunk: z.infer<S>) => {
             res.write(JSON.stringify(chunk) + streamDelimiter);
           }) as S extends z.ZodVoid ? undefined : StreamingCallback<z.infer<S>>,
           auth,
         });
-        res.write(JSON.stringify(state.operation));
+        res.write({
+          result: result.result, // Need more results!!!!
```
apascal07, line 407: ?

js/flow/src/index.ts
```diff
@@ -14,26 +14,18 @@
...
-export { FirestoreStateStore } from './firestoreStateStore.js';
 export {
   Flow,
   defineFlow,
   defineStreamingFlow,
   startFlowsServer,
   type CallableFlow,
   type FlowAuthPolicy,
-  type StepsFunction,
+  type FlowFn as StepsFunction,
```
apascal07, line 24: `type FlowFn,`

apascal07 (approved):
End of an era. 🥲

js/flow/src/flow.ts
pavelgj: `result: result.result`... joke?! I try...

apascal07: Oh I saw that and was like jeeze how many more are you expecting? Didn't get it was a meta-joke. :-)

PR #920, opened 2024-09-17
fix: correctly pass local auth context when doing runFlow

pavelgj:


apascal07 (approved):

js/flow/src/flow.ts
```diff
@@ -638,11 +638,15 @@ export async function runFlow<
     );
   }
 
-  const state = await flow.runEnvelope({
-    start: {
-      input,
+  const state = await flow.runEnvelope(
```
pavelgj, line 641: Just realized that also need to do this for streamFlow

pavelgj: Done.

PR #921, opened 2024-09-17
fix: fixed local auth handling for streamed flows

pavelgj:


apascal07:
Can you fix `defineFlow` -> `ai.defineFlow` please.

PR #923, opened 2024-09-17
Allow plugins to implicitly register as providers.

kmandrika:
In practice, this removes the need to specify a telemetry {} block if a plugin within the plugins list is a telemetry provider.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


js/core/src/plugin.ts
```diff
@@ -24,12 +24,20 @@ export interface Provider<T> {
   value: T;
 }
 
+export enum PluginAbilityType {
+  UNSPECIFIED = 0,
+  FLOW_STATE_STORE = 1 << 0,
```
pavelgj, line 29: this is gone

pavelgj:
since this is a plugin API change could you please bring this API proposal up for discussion on discord?

js/core/src/plugin.ts
kmandrika: Removed

PR #926, opened 2024-09-18
[JS] Refactored Genkit API to no longer use a global registry.

apascal07:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


js/testapps/dev-ui-gallery/src/main/flows-firebase-functions.ts
```diff
@@ -15,14 +15,15 @@
...
 
 import { generate } from '@genkit-ai/ai';
+import { run } from '@genkit-ai/core';
 import { firebaseAuth } from '@genkit-ai/firebase/auth';
 import { noAuth, onFlow } from '@genkit-ai/firebase/functions';
-import { defineFlow, run } from '@genkit-ai/flow';
 import { gemini15Flash } from '@genkit-ai/googleai';
 import { DecodedIdToken } from 'firebase-admin/auth';
 import * as z from 'zod';
+import { genkit } from '../index.js';
```
pavelgj, line 24: I wonder if we should go back to this pattern -- `initializeGenkit` goes into `genkit.ts` file and then you import genkit instance like this...

apascal07: It's user preference, I don't mind it either way. It just depends on the size and complexity of the codebase. I think we should stick with intro docs with everything in the same file to reduce learning curve.

genkit-tools/cli/config/firebase.index.ts.template
```diff
@@ -11,7 +11,7 @@ $GENKIT_MODEL_IMPORT
 import {firebaseAuth} from "@genkit-ai/firebase/auth";
 import {onFlow} from "@genkit-ai/firebase/functions";
  
-configureGenkit({
+const genkit = initializeGenkit({
```
pavelgj, line 14:
```suggestion
const ai = genkit({
```

genkit-tools/cli/config/firebase.index.ts.template
```diff
@@ -23,6 +23,7 @@ $GENKIT_CONFIG_PLUGINS
 
 // Define a simple flow that prompts an LLM to generate menu suggestions.
 export const menuSuggestionFlow = onFlow(
+  genkit,
```
pavelgj, line 26:
```suggestion
  ai,
```

genkit-tools/cli/config/firebase.index.ts.template
```diff
@@ -2,7 +2,7 @@ import * as z from "zod";
 
 // Import the Genkit core libraries and plugins.
 import {generate} from "@genkit-ai/ai";
-import {configureGenkit} from "@genkit-ai/core";
+import {initializeGenkit} from "@genkit-ai/core";
```
pavelgj, line 5:
```suggestion
import {genkit} from "@genkit-ai/core";
```

genkit-tools/cli/config/nextjs.genkit.ts.template
```diff
@@ -4,12 +4,11 @@ import * as z from 'zod';
 
 // Import the Genkit core libraries and plugins.
 import { generate } from '@genkit-ai/ai';
-import { configureGenkit } from '@genkit-ai/core';
-import { defineFlow } from '@genkit-ai/flow';
+import { initializeGenkit } from '@genkit-ai/core';
```
pavelgj, line 7:
```suggestion
import { genkit } from '@genkit-ai/core';
```

genkit-tools/cli/config/nextjs.genkit.ts.template
```diff
@@ -4,12 +4,11 @@ import * as z from 'zod';
...
 $GENKIT_CONFIG_IMPORTS
 $GENKIT_MODEL_IMPORT
 
-configureGenkit({
+const genkit = initializeGenkit({
```
pavelgj, line 11:
```suggestion
const ai = genkit({
```

genkit-tools/cli/config/nextjs.genkit.ts.template
```diff
@@ -20,7 +19,7 @@ $GENKIT_CONFIG_PLUGINS
 });
 
 // Define a simple flow that prompts an LLM to generate menu suggestions.
-const menuSuggestionFlow = defineFlow(
+const menuSuggestionFlow = genkit.defineFlow(
```
pavelgj, line 22:
```suggestion
const menuSuggestionFlow = ai.defineFlow(
```

genkit-tools/cli/config/nodejs.index.ts.template
```diff
@@ -2,23 +2,25 @@ import * as z from 'zod';
 
 // Import the Genkit core libraries and plugins.
 import { generate } from '@genkit-ai/ai';
-import { configureGenkit } from '@genkit-ai/core';
-import { defineFlow, startFlowsServer } from '@genkit-ai/flow';
+import { initializeGenkit } from '@genkit-ai/core';
```
pavelgj, line 5:
```suggestion
import { genkit } from '@genkit-ai/core';
```

genkit-tools/cli/config/nodejs.index.ts.template
```diff
@@ -2,23 +2,25 @@ import * as z from 'zod';
...
 $GENKIT_CONFIG_IMPORTS
 $GENKIT_MODEL_IMPORT
 
-configureGenkit({
+const genkit = initializeGenkit({
```
pavelgj, line 9:
```suggestion
const ai = genkit({
```

js/core/src/flow.ts
```diff
@@ -370,81 +333,263 @@ export class Flow<
...
+    });
+  }
+}
+
+/**
+ * Options to configure the flow server.
+ */
+export interface FlowServerOptions {
+  /** Which environment(s) to run the flow server in. Defaults to `prod`. */
+  runInEnv?: 'all' | 'prod' | 'dev';
```
pavelgj, line 420: I'm not sold on this.... let's leave this out for now.

js/core/src/flow.ts
```diff
@@ -370,81 +333,263 @@ export class Flow<
...
+  /** List of flows to expose via the flow server. If not specified, all registered flows will be exposed. */
+  flows?: Flow<any, any, any>[];
```
pavelgj, line 422: this one is problematic... we can't have the flow ref when you're creating genkit instance... we need a different way to control which flows to expose in the flow server. maybe ignore for this PR.

apascal07: Ugh, I thought about this but then "realized" that since they're module-level, it has a fancier resolving mechanism but you're right. Even using string references won't help because it will still try to start up before they're defined. Bummer. Will think about options here.

js/core/src/genkit.ts
```diff
@@ -0,0 +1,277 @@
...
+  /** Name of the flow state store to use. If not specified, a dev flow state store will be used. The flow state store must be registered in the config. */
+  flowStateStore?: string;
+  /** Whether to enable tracing and metrics. */
+  enableTracingAndMetrics?: boolean;
+  /** Level at which to log messages.*/
+  logLevel?: 'error' | 'warn' | 'info' | 'debug';
+  /** Directory where dotprompts are stored. */
+  promptDir?: string;
+  /** Telemetry configuration. */
+  telemetry?: TelemetryOptions;
```
pavelgj, line 61: can you please add a FIXME -- I want to move this outta here... telemetry should not (cannot) be scoped to genkit instance.

pavelgj (approved):

PR #931, opened 2024-09-18
Update the flow name extraction regex to support the scenario where there is a parent span.

bryanatkinson:
When running locally, there is a parent `dev-run-action-wrapper` span, and this results in an extracted flow name that looks like `dev-run-action-wrapper/{<flowName>`.

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


PolinaGo (approved):
Recent refactoring resulted in flow names being prepended with /{dev-run-action-wrapper}/ in some cases.

kmandrika (approved):

PR #933, opened 2024-09-19
fix model name lookup in dotprompt plugin and add a test

### LINKED ISSUE #932: [Go] When using the dotprompt plugin with a `ModelName` config, it will panic
**Describe the bug**
When creating a prompt with a `ModelName` Config, the provided model name is not considered and it will panic. This is due to a bug when looking up the model in:

https://github.com/firebase/genkit/blob/2940fd520ebde799c74581921546b655b2cab898/go/plugins/dotprompt/genkit.go#L207-L225

The `model` is shadowed inside the if statement and result to the lookedup model going out of context after the if statement.

**To Reproduce**

```golang
prompt, _ := dotprompt.Define("menu", `Suggest an item for the menu of a {{ type }} themed restaurant`, dotprompt.Config{ModelName: "googleai/gemini-1.5-flash", InputSchema: r.Reflect(Input{}), OutputSchema: os, OutputFormat: ai.OutputFormatJSON})

resp, _ := prompt.Generate(ctx,
			&dotprompt.PromptRequest{
				Variables: input,
				Config:    &ai.GenerationCommonConfig{Temperature: 1},
			},
			nil,
		)
```

**Expected behavior**

When a model name is provided, either use the provided model name or return an error if the model lookup fails.

Not relevant

** Go version
Not relevant

**Additional context**
The panic will happen in line:

https://github.com/firebase/genkit/blob/2940fd520ebde799c74581921546b655b2cab898/go/plugins/dotprompt/genkit.go#L227

This is similar to https://github.com/firebase/genkit/issues/847 but not quite the same issue.

dhiaayachi:
Fix #932

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/go/plugins/dotprompt/genkit_test.go
+++ b/go/plugins/dotprompt/genkit_test.go
@@ -73,3 +73,36 @@ func TestExecute(t *testing.T) {
 		t.Errorf("fake model replied with %q, want %q", got, want)
 	}
 }
+
+func TestExecuteWithModelName(t *testing.T) {
+	ai.DefineModel("test", "test", nil, testGenerate)
+	p, err := New("TestExecute", "TestExecute", Config{ModelName: "test/test"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	resp, err := p.Generate(context.Background(), &PromptRequest{}, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(resp.Candidates) != 1 {
+		t.Errorf("got %d candidates, want 1", len(resp.Candidates))
+		if len(resp.Candidates) < 1 {
+			t.FailNow()
+		}
+	}
+	msg := resp.Candidates[0].Message
+	if msg == nil {
+		t.Fatal("response has candidate with no message")
+	}
+	if len(msg.Content) != 1 {
+		t.Errorf("got %d message parts, want 1", len(msg.Content))
+		if len(msg.Content) < 1 {
+			t.FailNow()
+		}
+	}
+	got := msg.Content[0].Text
+	want := `AI reply to "TestExecute"`
+	if got != want {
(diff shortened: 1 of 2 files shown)


go/plugins/dotprompt/genkit_test.go
```diff
@@ -73,3 +73,36 @@ func TestExecute(t *testing.T) {
 		t.Errorf("fake model replied with %q, want %q", got, want)
 	}
 }
+
+func TestExecuteWithModelName(t *testing.T) {
+	ai.DefineModel("test", "test", nil, testGenerate)
+	p, err := New("TestExecute", "TestExecute", Config{ModelName: "test/test"})
```
yukinagae, line 79: @dhiaayachi 

It might be a bit hard to understand, but `ai.DefineModel()` is a function with side effects that internally registers the defined Model in a global scope registry. When doing so, if there is a duplicate Model with the same `[provider/model_name]`, an error is triggered, and the Model cannot be registered.

If you want the simplest fix, you can modify the test code as follows to ensure that the Models do not conflict:

```go
ai.DefineModel("test", "test2", nil, testGenerate)
p, err := New("TestExecute", "TestExecute", Config{ModelName: "test/test2"})
```

Alternatively, you can use something like [TestMain](https://pkg.go.dev/testing#hdr-Main) to ensure that `ai.DefineModel("test", "test", nil, testGenerate)` is called only once within the test file.

yukinagae: This is the reason why the CI test failed.

dhiaayachi: Thank you @yukinagae, I decided to use subtests to avoid having to define the model as a global var.

mbleigh (approved):
I ran into this same thing quite recently and had the same one-character fix in a local branch. LGTM.

PR #944, opened 2024-09-23
feat(js/plugins/vertexai): strict json mode

cabljac:
This PR adds strict json mode support to Vertex AI plugin in `next`

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/plugins/vertexai/src/gemini.ts
+++ b/js/plugins/vertexai/src/gemini.ts
@@ -398,7 +398,10 @@ function fromGeminiFunctionResponsePart(part: GeminiPart): Part {
 }
 
 // Converts vertex part to genkit part
-function fromGeminiPart(part: GeminiPart): Part {
+function fromGeminiPart(part: GeminiPart, jsonMode: boolean): Part {
+  if (jsonMode && part.text !== undefined) {
+    return { data: JSON.parse(part.text) };
+  }
   if (part.text !== undefined) return { text: part.text };
   if (part.functionCall) return fromGeminiFunctionCallPart(part);
   if (part.functionResponse) return fromGeminiFunctionResponsePart(part);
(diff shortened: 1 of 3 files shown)


js/plugins/vertexai/src/gemini.ts
```diff
@@ -398,7 +398,10 @@ function fromGeminiFunctionResponsePart(part: GeminiPart): Part {
 }
 
 // Converts vertex part to genkit part
-function fromGeminiPart(part: GeminiPart): Part {
+function fromGeminiPart(part: GeminiPart, jsonMode: boolean): Part {
+  if (jsonMode && part.text !== undefined) {
+    return { data: JSON.parse(part.text) };
```
pavelgj, line 403: is `JSON.parse` safe here? Do we need to guard agains markdown and extraneous text? Or does the gemini json mode guarantee to ouput clean json only?

js/plugins/vertexai/src/gemini.ts
```diff
@@ -517,11 +521,18 @@ export function geminiModel(
         }
       }
 
+      const tools = request.tools?.length
+        ? [{ functionDeclarations: request.tools?.map(toGeminiTool) }]
+        : [];
+
+      // Cannot use tools and function calling at the same time
+      const jsonMode =
+        (request.output?.format === 'json' || !!request.output?.schema) &&
```
pavelgj, line 530: this implementation does not support passing output schema as model responseSchema?
https://ai.google.dev/gemini-api/docs/structured-output?lang=node#supply-schema-in-config

js/plugins/vertexai/src/gemini.ts
cabljac: it's not safe, we need to use lenient parsing, there's an open issue for it. Working on a fix (this is also in main)

js/plugins/vertexai/src/gemini.ts
cabljac: Yeah we need to add this feature in too, do you want these changes on this PR? This PR was in theory just to sync with what went into `main`

cabljac: i'll add as commits here actually, that makes sense

js/plugins/vertexai/src/gemini.ts
cabljac: I think it will guarantee clean json but not complete JSON, but will double check

cabljac:
closing as resolved in https://github.com/firebase/genkit/pull/1149/files

responseSchema will be dealt with in a new PR

PR #948, opened 2024-09-23
[Evals] Add endpoint to start new evaluation run

ssbushi:


genkit-tools/cli/src/commands/eval-flow.ts
```diff
@@ -50,7 +52,10 @@ export const evalFlow = new Command('eval:flow')
   )
   .argument('<flowName>', 'Name of the flow to run')
   .argument('[data]', 'JSON data to use to start the flow')
-  .option('--input <filename>', 'JSON batch data to use to run the flow')
+  .option(
+    '--input <input>',
+    'Input dataset ID or JSON file to be used for evaluation'
```
shrutip90, line 57: nit: Maybe a separate --datasetid option could be more intuitive to use.

genkit-tools/common/src/eval/evaluate.ts
```diff
@@ -35,47 +36,91 @@ import { enrichResultsWithScoring, extractMetricsMetadata } from './parser';
...
 interface BulkRunResponse {
   traceId?: string;
-  response: any;
+  response?: any;
 }
 
-/** Handles the Inference part of Inference-Evaluation cycle */
-export async function runInference(
+/**
+ * Starts a new evaluation run. Intended to be used via the reflection.
```
shrutip90, line 43: nit: reflection API?

genkit-tools/common/src/eval/evaluate.ts
```diff
@@ -113,54 +158,72 @@ export async function getMatchingEvaluators(
...
   return responses;
 }
 
-async function fetchDataSet(
-  runner: Runner,
-  flowName: string,
-  states: BulkRunResponse[],
-  parsedData: EvalFlowInput
-): Promise<EvalInput[]> {
+async function fetchDataSet(params: {
```
shrutip90, line 218: nit: can we rename this to fetchEvalInput to distinguish from the stored datasets?

shrutip90 (approved):

genkit-tools/cli/src/commands/eval-flow.ts
ssbushi: I think the versatility helps. We have to rethink the args for CLI commands anyway. We can incorporate this there. Thanks!

PR #980, opened 2024-09-27
Marking genkit root features

maxl0rd:
Marking genkit root features to facilitate queries for the Genkit entry point (eg flow, a prompt, generate, etc.)


schnecle (approved):
LGTM!

PR #1004, opened 2024-10-03
fix(docs/dotprompt/schema): use correct schemas to type inference

DenisVCode:
Current example is wrong and does't correctly infer schema for the output. It sends only one schema, where 2 are expected, one for input, one for output.

I have followed the current docs, but replaced the single schema with 2 schemas, one for output and one for input, that would produce correctly working snippet. 

Related Discord chat: https://discord.com/channels/1255578482214305893/1255934063173435402/1291469593235296439

Checklist (if applicable):
- [x] Docs updated

### DIFF
--- a/docs/dotprompt.md
+++ b/docs/dotprompt.md
@@ -151,11 +151,19 @@ a schema registered with `defineSchema` by name. To register a schema:
 import { defineSchema } from '@genkit-ai/core';
 import { z } from 'zod';
 
-const MySchema = defineSchema(
-  'MySchema',
+const MyOutputSchema = defineSchema(
+  'MyOutputSchema',
   z.object({
     field1: z.string(),
     field2: z.number(),
+  })
+);
+
+const MyInputSchema = defineSchema(
+  'MyInputSchema',
+  z.object({
+    field3: z.string(),
+    field4: z.number(),
   })
 );
 ```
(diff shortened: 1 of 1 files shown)


xavidop (approved):

docs/dotprompt.md
```diff
@@ -180,9 +190,9 @@ import { promptRef } from "@genkit-ai/dotprompt";
 
 const myPrompt = promptRef("myPrompt");
 
-const result = await myPrompt.generate<typeof MySchema>({...});
+const result = await myPrompt.generate<typeof MyInputSchema, typeof MyOutputSchema>({...});
```
pavelgj, line 193: Looks like the first generic parameter of `.generate` is model custom options schema.
You don't need to define `MyInputSchema` for this example.

```suggestion
const result = await myPrompt.generate<any, typeof MyOutputSchema>({...});
```

DenisVCode: Gotcha thanks, is it possible to add strongly typed inputs too? But I guess that's more than a doc edit. What would be good example custom options to add there instead of any please?

pavelgj: I'm wondering if it worth updating these anymore. Prompts have undergone pretty significant changes in the next branch for the next release (very soon): https://github.com/firebase/genkit/blob/64cd023d7c645857fe21c4c5404d0e58a1e27f1e/js/ai/src/prompt.ts#L80

PR #1017, opened 2024-10-07
Add explicit model attribute to make filtering by model easier.

schnecle:
Checklist (if applicable):
- [ ] Docs updated

### DIFF
--- a/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
+++ b/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
@@ -341,7 +341,10 @@ class AdjustingTraceExporter implements SpanExporter {
   }
 
   private markGenkitModel(span: ReadableSpan): ReadableSpan {
-    if (span.attributes['genkit:metadata:subtype'] === 'model') {
+    if (
+      span.attributes['genkit:metadata:subtype'] === 'model' &&
+      span.attributes['genkit:name']
+    ) {
       span.attributes['genkit:model'] = span.attributes['genkit:name'];
     }
     return span;
--- a/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
+++ b/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
@@ -244,6 +244,7 @@ class AdjustingTraceExporter implements SpanExporter {
       span = this.markErrorSpanAsError(span);
       span = this.markFailedAction(span);
       span = this.markGenkitFeature(span);
+      span = this.markGenkitModel(span);
       span = this.normalizeLabels(span);
       return span;
     });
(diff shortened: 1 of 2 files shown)


js/plugins/google-cloud/tests/traces_test.ts
```diff
@@ -14,10 +14,13 @@
  * limitations under the License.
  */
 
+import { runWithRegistry } from '@genkit-ai/core/registry';
 import { ReadableSpan } from '@opentelemetry/sdk-trace-base';
 import { Genkit, genkit, run, z } from 'genkit';
 import assert from 'node:assert';
 import { after, before, beforeEach, describe, it } from 'node:test';
+import { generate } from '../../../ai/src/generate.js';
```
schnecle, line 22: Should these be different imports?

maxl0rd: Yes, they should probably be `from '@genkit-ai/ai`

js/plugins/google-cloud/src/gcpOpenTelemetry.ts
```diff
@@ -338,6 +339,13 @@ class AdjustingTraceExporter implements SpanExporter {
     }
     return span;
   }
+
+  private markGenkitModel(span: ReadableSpan): ReadableSpan {
+    if (span.attributes['genkit:metadata:subtype'] === 'model') {
```
maxl0rd, line 344: In the other examples, I added `&& attributes['genkit:name']` to the condition, just to avoid the corner case where we write a null key. Shouldn't ever happen but...

js/plugins/google-cloud/tests/traces_test.ts
```diff
@@ -132,6 +135,50 @@ describe('GoogleCloudTracing', () => {
     assert.equal(spans[1].attributes['genkit/feature'], 'niceFlow');
   });
 
+  it('adds the genkit/model label for model actions', async () => {
+    const echoModel = runWithRegistry(ai.registry, () =>
```
maxl0rd, line 139: runWithRegistry doesn't return anything

maxl0rd (commented):
Just a few nits

js/plugins/google-cloud/tests/traces_test.ts
schnecle: then how is this passing?!

js/plugins/google-cloud/src/gcpOpenTelemetry.ts
schnecle: Good call!

js/plugins/google-cloud/tests/traces_test.ts
schnecle: I think this does some magic because it looks like we are using it to return stuff all over the place. Example: https://github.com/firebase/genkit/blob/next/js/ai/tests/generate/generate_test.ts#L595

js/plugins/google-cloud/tests/traces_test.ts
schnecle: it actually looks like it's now 'genkit' and 'genkit/model'

js/plugins/google-cloud/tests/traces_test.ts
schnecle: There is no declared return type, but the function is returning whatever the internal function is doing.

maxl0rd (approved):

PR #1028, opened 2024-10-08
Set span type to prompt when rendering a dotprompt

MichaelDoyle:
This will enable the Dev UI to properly deal with prompts in a more generic way when rendering badges, loading traces into the runner, etc, versus having to check for each different type of prompt ['prompt', 'DOTprompt', 'FOOprompt'] etc.

image

Checklist (if applicable):

### DIFF
--- a/js/plugins/dotprompt/src/prompt.ts
+++ b/js/plugins/dotprompt/src/prompt.ts
@@ -234,7 +234,7 @@ export class Dotprompt<I = unknown> implements PromptMetadata<z.ZodTypeAny> {
           input: opt,
         },
         labels: {
-          [SPAN_TYPE_ATTR]: 'dotprompt',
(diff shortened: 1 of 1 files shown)


MichaelDoyle:
cc @tonybaroneee this will be relevant for the firebase console as well.

tonybaroneee (approved):

MichaelDoyle:
On second thought, this doesn't look to be _exactly_ what we want.

I think we want

```
type: 'action'
subtype: 'prompt'
```

maxl0rd:
I think this is on the right track, but I have a counter proposal since this is colliding with the feature tracing change.

tonybaroneee (changes requested):
(See Max's comment)

PR #1029, opened 2024-10-08
feat: make ai.definePrompt callable as a function

pavelgj:
```ts
const hi = ai.definePrompt({
    name: 'hi',
    model: gemini15Flash,
    input: {
      schema: z.object({
        name: z.string(),
      }),
    },
  },
  'hi {{ name }}'
);

const response = await hi({ name: 'Genkit' });
console.log(response.text())

// or stream

const { response, stream } = await hi.stream({ name: 'Genkit' });
  console.log(chunk.text());
}
const responseText = (await response).text();
console.log(response.text())
```

or

```ts
const hi = ai.definePrompt({
    name: 'hi',
    model: 'echoModel',
    input: {
      schema: z.object({
        name: z.string(),
      }),
    },
    config: {
      temperature: 11,
    },
  },
  async (input) => {
    return {
      messages: [
        { role: 'user', content: [{ text: `hi ${input.name}` }] },
      ],
    };
  }
);
      
const response = await hi({ name: 'Genkit' });
console.log(response.text())

// or stream

const { response, stream } = await hi.stream({ name: 'Genkit' });
  console.log(chunk.text());
}
const responseText = (await response).text();
console.log(response.text())
```


js/genkit/src/genkit.ts
```diff
@@ -81,19 +91,65 @@ export interface GenkitOptions {
...
+> {
+  (
+    input?: z.infer<I>,
+    opts?: z.infer<CustomOptions>
+  ): Promise<GenerateResponse>;
+
+  /**
+   * Generates a streaming response by rendering the prompt template with given user input and then calling the model.
+   *
+   * @param opt Options for the prompt template, including user input variables and custom model configuration options.
```
MichaelDoyle, line 114: Add doc for `@param input` ?

BTW, when would I use `stream` versus `generateStream`, and how would the user know? I think the comments read the same right now for both methods.

pavelgj: Updated the docs.

so `(..)`/`.stream` are "convenience" shortcuts for `generate`/`generateStream` where you can just provide the inputs (and optionally model config). If the user needs to provide/override more generate options (additional tools, output settings, etc) then they can use the more verbose `generate`/`generateStream`.

apascal07 (approved):

pavelgj:
> I can see we use this stream method: const { response, stream } = await hi.stream({ name: 'Genkit' });
> 
> Can we do this? const { response, stream } = await hi.generate({ name: 'Genkit' });

you mean 
```ts
const { response, stream } = await hi.generateStream({ name: 'Genkit' });
```
?

`hi.stream` is essentially `hi.generateStream`, it just takes input as main argument.  You can pass in all the other `generate` options as the second parameter:

```ts
const { response, stream } = await hi.stream({ name: 'Genkit' }, {model, config: { temperature: 11 }});
```

JGSolutions:
@pavelgj 
trying to use generateSteam but I get this error

Property 'generateStream' does not exist on type 'ExecutablePrompt<any, ZodTypeAny, ZodTypeAny>'.t

pavelgj:
> @pavelgj trying to use generateSteam but I get this error
> 
> Property 'generateStream' does not exist on type 'ExecutablePrompt<any, ZodTypeAny, ZodTypeAny>'.t

Try `myPrompt.stream(input, { config: { temperature: 1 }}`

JGSolutions:
@pavelgj i tried the stream but I get different response and breaks my logic after that. In version 0.5 i was using generate and so by reading this doc I though I could have done the same:

https://docs.google.com/document/d/1cr0GA5B63AQXWEiXoYC0CIbuF_sPRo6Ci96x4csjZ9k/edit?tab=t.0#heading=h.u3azbka89zhw (scroll to: Executable prompts (with first class dotprompt))

I am still using ai.generate() just that i removed my .prompt file and used it directly in the generate method with RAGand firebase

```
const docs = await ai.retrieve({
            retriever: promptSearchRetriever,
            query: gg,
            options: { k: e.limit },
        });

        const { output } = await ai.generate({
            model: gemini15Flash,
            prompt: `You are a helpful AI assistant that really loves to talk about board games. Try to answer the provided query.

                  When a user says "popular", sort by rank.
                  When a user says "play time length", sort by maxPlayTime.
                  When a user says "new", sort by yearPublished.

                  Query: {{ ${gg} }}`,
            output: { schema: GamesFlowOutput },
            docs // data from firestore
        });

        const games = await output as Games[];
```

Is there any specific model I can use to increase the speed when generate the response?

PR #1047, opened 2024-10-11
Using features in telemetry

maxl0rd:
Enabling top level actions such as prompts or generate calls to be used for telemetry as effectively as those wrapped in flows.

- Switching actions to use newTrace, which starts a trace similar to a flow if it is not currently inside something.
- Improving instrumentation to always assign a feature name
- Renaming metric dimensions from flowName -> featureName

### DIFF
--- a/js/plugins/google-cloud/src/telemetry/action.ts
+++ b/js/plugins/google-cloud/src/telemetry/action.ts
@@ -26,7 +26,7 @@ import {
   Telemetry,
   internalMetricNamespaceWrap,
 } from '../metrics.js';
-import { extractErrorName, extractOuterFlowNameFromPath } from '../utils';
+import { extractErrorName, extractOuterFeatureNameFromPath } from '../utils';
 
 class ActionTelemetry implements Telemetry {
   /**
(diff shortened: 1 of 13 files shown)


js/plugins/google-cloud/src/utils.ts
```diff
@@ -26,6 +26,15 @@ export function extractOuterFlowNameFromPath(path: string) {
   return flowName ? flowName[1] : '<unknown>';
 }
 
+export function extractOuterFeatureNameFromPath(path: string) {
+  if (!path || path === '<unknown>') {
+    return '<unknown>';
+  }
+
+  const flowName = path.match('/{(.+),t:(flow|action|prompt)}+');
```
tonybaroneee, line 34: Are these the only 3 top-level types? Is "helper" a thing now?

tonybaroneee (approved):
NOICE

js/plugins/google-cloud/src/utils.ts
maxl0rd: Yes, it would be good to include helper.

js/plugins/google-cloud/src/telemetry/action.ts
```diff
@@ -55,35 +55,37 @@ class ActionTelemetry implements Telemetry {
...
+      this.writeFailure(actionName, featureName, path, latencyMs, errorName);
     }
 
     logger.warn(`Unknown action state; ${state}`);
   }
 
   private writeSuccess(
     actionName: string,
-    flowName: string,
+    featureName: string,
```
bryanatkinson, line 82: Are you sure we should switch from flowName to featureName?

Since we're adding feature-level metrics without removing flow-level metrics, why not include featureName as an additional dimension on the action metrics instead of replacing?

tonybaroneee: I guess that's true, `flowName` can still be set if this action is part of a feature that is also a flow. But the question is, will `flowName` be used in the future, or will it be more "feature paths" (to drive things like our span filter)?

maxl0rd: Feature name is the same as how flow name worked before, if you are in a flow.

bryanatkinson: But flow and feature may be different if the top-most Genkit span isn't a flow.  If we're supporting separate metrics for these concepts, why wouldn't we keep them as separate dimensions on other metrics?

maxl0rd: DOne

tonybaroneee (approved):
🐚🐑🚢

bryanatkinson (approved):

PR #1060, opened 2024-10-16
feat: updated genkit plugin API to support Genkit instance

pavelgj:
```ts
export function myPlugin(params: {...}): GenkitPlugin {
  return genkitPlugin('myPlugin', async (ai) => {
    ai.defineModel({...});
    ai.defineRetriever({...});
    ai.defineIndexer({...});
    ai.defineEmbedder({...});
    ai.defineReranker({...});
  });
}

### CODE REVIEW

File: js/genkit/src/evaluator.ts [hunk 1]
```diff
@@ -14,4 +14,23 @@
  * limitations under the License.
  */
 
-export * from '@genkit-ai/ai/evaluator';
+export {
```
cabljac, line 17: I'm a little out of the loop on evaluators: Is there a defineEvaluator method? If so does it need to be exported

js/genkit/src/plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type PluginInit = (genkit: Genkit) => void | Promise<void>;
+
+export type GenkitPlugin = (genkit: Genkit) => PluginProvider;
+
+/**
+ * Defines a Genkit plugin.
+ */
+export function genkitPlugin<T extends PluginInit>(
```
cabljac, line 31: Even though it isn't defining an action itself, I could see this as being renamed `definePlugin` ?

The initialiser is basically going to define a bunch of actions right

js/genkit/src/evaluator.ts
pavelgj: yes, there is. it's `ai.defineEvaluator` now.

cabljac (approved):
One just suggestion on naming of genkitPlugin, otherwise lgtm!

js/genkit/src/plugin.ts
pavelgj: Interesting... the only thing is that it follows separate mechanic compared to other `ai.define*` methods. When you call `ai.define*` it registers whatever you define with the genkit instance. However `definePlugin` is not doing that -- it only creates a plugin, but then there's a separate `ai.registerPlugin`  process (not exactly that, but conceptually) which makes the plugin available to the genkit instance.
I'm not sure whether anyone would care about that distinction... but to me it feels important enough. :)

cabljac: that's fair :D

js/plugins/google-cloud/tests/metrics_test.ts
```diff
@@ -260,7 +252,8 @@ describe('GoogleCloudMetrics', () => {
     );
   });
 
-  it('writes action metrics for a failed action', async () => {
+  // SKIPPED -- we don't allow defining arbitrary actions anymore....
+  it.skip('writes action failure metrics', async () => {
```
MichaelDoyle, line 256: What's the eventual plan w/ the skipped tests?

js/plugins/vertexai/tests/vector-search/query_public_endpoint_test.ts
```diff
@@ -19,7 +19,8 @@ import { describe, it, Mock } from 'node:test';
 import { queryPublicEndpoint } from '../../src/vector-search/query_public_endpoint';
 
 describe('queryPublicEndpoint', () => {
-  it('queryPublicEndpoint sends the correct request and retrieves neighbors', async (t) => {
+  // FIXME -- t.mock.method is not supported node above 20
```
MichaelDoyle, line 22: nit: create an issue for these?

MichaelDoyle (approved):

js/plugins/google-cloud/tests/metrics_test.ts
pavelgj: Filed #1071 to track, but we need to decide whether these are even worth testing since generic actions are not possible to create anymore... maybe can be switched to use some other type of action.

PR #1086, opened 2024-10-21
fix(js/plugins/ollama): reconfigure ollama embedding

cabljac:
Should resolve https://github.com/firebase/genkit/issues/1080

Also enabled the tests for this plugin.

Checklist (if applicable):
- [x] Docs updated

### DIFF
--- a/docs/plugins/ollama.md
+++ b/docs/plugins/ollama.md
@@ -150,6 +150,10 @@ export default configureGenkit({
 
 ### Using Embeddings
 
+This example demonstrates generating a 768-dimensional embedding using the `nomic-embed-text` model.
+
+Make sure to run `ollama pull nomic-embed-text` to obtain this model, and `ollama serve` to serve the embedding model.
+
 Once configured, you can generate embeddings for your content using the `embed` function:
 
 ```js
@@ -165,5 +169,3 @@ const result = await embed({
   },
 });
 ```
-
-This example demonstrates generating a 768-dimensional embedding using the `nomic-embed-text` model.
--- a/docs/plugins/ollama.md
+++ b/docs/plugins/ollama.md
@@ -122,3 +122,48 @@ const llmResponse = await generate({
   prompt: 'Tell me a joke.',
 });
 ```
+
+
+### Embeddings
+
+The Ollama plugin supports embeddings, allowing you to obtain vector representations of text content. To use embeddings, configure the plugin with an appropriate model that supports embeddings, and specify the `embedder` in your code.
+
+Example configuration:
+
+```js
+import { ollama } from 'genkitx-ollama';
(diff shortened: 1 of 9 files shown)


docs/plugins/ollama.md
```diff
@@ -122,3 +122,48 @@ const llmResponse = await generate({
...
+
+```js
+import { ollama } from 'genkitx-ollama';
+
+export default configureGenkit({
+  plugins: [
+    ollama({
+      embedders: [
+        {
+          name: 'nomic-embed-text', // specify the embedding model
```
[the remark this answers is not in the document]
cabljac: yup, good catch

js/plugins/ollama/src/types.ts
```diff
@@ -0,0 +1,89 @@
...
+  serverAddress: string;
+
+  /**
+   * Optional request headers, which can be either static or dynamically generated.
+   */
+  requestHeaders?: RequestHeaders;
+}
+
+// Function type for generating request headers dynamically
+export interface RequestHeaderFunction {
```
pavelgj, line 75: This looks good, but this is a breaking change from:
```ts
((
  params: { serverAddress: string; model: ModelDefinition },
  request: GenerateRequest
) => Promise<Record<string, string> | void>)
```

I mean, if you look at docs/plugins/ollama.md "Authentication" chapter... then this:

```ts
const ollamaProd = {
  ...ollamaCommon,
  serverAddress: 'https://my-deployment',
  requestHeaders: async (params) => {
    const headers = await fetchWithAuthHeader(params.serverAddress);
    return { Authorization: headers['Authorization'] };
  },
} as OllamaPluginParams;
```

must be updated to 

```ts
const ollamaProd = {
  ...ollamaCommon,
  serverAddress: 'https://my-deployment',
  requestHeaders: async (options) => {
    const headers = await fetchWithAuthHeader(options.params.serverAddress);
    return { Authorization: headers['Authorization'] };
  },
} as OllamaPluginParams;
```

I would prefer to do it without breaking changes (at least to what we have documented)... can we do something like this?

```ts
export interface RequestHeaderFunction {
  (params: {
    serverAddress: string;
    model: ModelDefinition | EmbeddingModelDefinition;
    modelRequest?: GenerateRequest;
    embedRequest?: {docs: Document[], config?: any};
  },
  // @deprecated -- moved into params, here for backwards compatibility reasons.
  modelRequest?: GenerateRequest): Promise<Record<string, string> | void>;
}
```

cabljac: Sure!

cabljac:
closing in favour of PR to `next` [here](https://github.com/firebase/genkit/pull/1246)

PR #1118, opened 2024-10-24
Clear trace metadata from logs not used by GCP plugin.

bryanatkinson:
For logs written during span export in the GCP plugin, these default Otel trace fields are incorrect because the log is being written at a different time than during span execution.

Checklist (if applicable):
- [ ] Docs updated

### DIFF
--- a/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
+++ b/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
@@ -97,6 +97,15 @@ export class GcpOpenTelemetry {
       `projects/${projectId}/traces/${spanContext.traceId}`;
     record['logging.googleapis.com/trace_sampled'] ??= isSampled ? '1' : '0';
     record['logging.googleapis.com/spanId'] ??= spanContext.spanId;
+
+    // Clear out the duplicate trace and span information in the log metadata.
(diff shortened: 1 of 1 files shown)


MichaelDoyle (approved):

js/plugins/google-cloud/src/gcpOpenTelemetry.ts
```diff
@@ -97,6 +97,15 @@ export class GcpOpenTelemetry {
...
     record['logging.googleapis.com/spanId'] ??= spanContext.spanId;
+
+    // Clear out the duplicate trace and span information in the log metadata.
+    // These will be incorrect for logs written during span export time since
+    // the logs are written after the span has fully executed. Those logs are
+    // explicitly tied to the correct span in createCommonLogAttributes in
+    // utils.ts.
+    delete record['span_id'];
+    delete record['trace_id'];
+    delete record['trace_flags'];
```
MichaelDoyle, line 108: lgtm - would it be easy to add a quick test for this?

bryanatkinson: Not that I can tell. This is an existing gap in our logs_test; we can only verify the log message itself, and not the structured metadata.

PR #1120, opened 2024-10-24
Refactors Message into its own file

mbleigh:
Also standardizes on Message.parse and Message.parseContent for coercing lenient options into proper form.

Description here... 

Checklist (if applicable):
- [ ] Docs updated

### DIFF
--- a/js/ai/src/message.ts
+++ b/js/ai/src/message.ts
@@ -1,0 +1,115 @@
+import { extractJson } from './extract';
+import { MessageData, Part, ToolRequestPart, ToolResponsePart } from './model';
+
+/**
+ * Message represents a single role's contribution to a generation. Each message
+ * can contain multiple parts (for example text and an image), and each generation
+ * can contain multiple messages.
+ */
+export class Message<T = unknown> implements MessageData {
+  role: MessageData['role'];
+  content: Part[];
+  metadata?: Record<string, any>;
+
+  static parseData(
+    lenientMessage:
+      | string
+      | (MessageData & { content: string | Part | Part[]; role: string })
+      | MessageData,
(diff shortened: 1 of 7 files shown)


js/ai/src/message.ts
```diff
@@ -0,0 +1,131 @@
...
+ * Message represents a single role's contribution to a generation. Each message
+ * can contain multiple parts (for example text and an image), and each generation
+ * can contain multiple messages.
+ */
+export class Message<T = unknown> implements MessageData {
+  role: MessageData['role'];
+  content: Part[];
+  metadata?: Record<string, any>;
+
+  static parseData(
```
pavelgj, line 30: oh, can you also add
```ts
Message.fromText('foo')
Message.fromMedia('uri')
```

pavelgj: hmm... we also need the role... maybe 

```ts
Message.fromText('system', 'foo')
```

Basically I'd like to improve:
```ts
await ai.generate({
  messages: [ Message.fromText('system', 'foo') ]
})
```

pavelgj:
```ts
await ai.generate({
  messages: [ Message.fromText('system', 'foo') ]
})
```

vs 

```ts
await ai.generate({
  messages: [ { role: 'system'}, content: [{ text: 'foo' }] ]
})
```

not sure anymore. :D

pavelgj: feel free to ignore. :)

mbleigh: I'm going to ignore just because we actually are already doing some friendly coercing here. This works:

```ts
messages: [{role: "system", content: "talk like a pirate"}]
```

apascal07 (approved):

PR #1124, opened 2024-10-25
Fix action schema validation

MichaelDoyle:
Prune the session state from the input just prior to validation (at the action level), similar to what we're doing for rendering prompts.

See also:
https://github.com/firebase/genkit/blob/next/js/plugins/dotprompt/src/prompt.ts#L182-L198

Checklist (if applicable):
- [ ] Docs updated

### DIFF
--- a/js/core/src/action.ts
+++ b/js/core/src/action.ts
@@ -28,6 +28,8 @@ import {
 export { Status, StatusCodes, StatusSchema } from './statusTypes.js';
 export { JSONSchema7 };
 
+export const GENKIT_SESSION_STATE_INPUT_KEY = '__genkit__sessionState';
+
 export interface ActionMetadata<
   I extends z.ZodTypeAny,
   O extends z.ZodTypeAny,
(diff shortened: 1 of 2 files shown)


maxl0rd (approved):

apascal07:
Curious, what's this? Never seen it before.

MichaelDoyle:
> Curious, what's this? Never seen it before.

First time encountering as well. I believe this session state is loaded into the "input" as a means to have the dev-ui be able to paly it back when re-running traces.

js/core/src/action.ts
```diff
@@ -120,6 +122,10 @@ export function action<
       ? config.name
       : `${config.name.pluginId}/${config.name.actionId}`;
   const actionFn = async (input: I) => {
+    if (input?.hasOwnProperty(GENKIT_SESSION_STATE_INPUT_KEY)) {
+      input = { ...input };
```
pavelgj, line 126: Hmm... Why doesn't this break: https://github.com/firebase/genkit/blob/64cd023d7c645857fe21c4c5404d0e58a1e27f1e/js/genkit/tests/session_test.ts#L342

I think you're only supposed to strip it for parsing, but add it back in for action fn call on line 144......

Issue #708, opened 2024-07-29
Custom Format Parsers


mbleigh:
Currently the `output()` method on Genkit results attempts to leniently parse JSON based on whether `format` is supplied or an output schema is present. There are a variety of reasons, however, that someone might want to have greater control over output parsing than this.

Proposed is to extend the `format` option to instead be a registry of formatters with some defaults (`text` and `json` at a minimum), provided. To define a custom format parser, developers would specify it something like so:

```ts
defineFormat('csv', (req) => {
  return {
    parseResponse: (res) => {
      const toParse = extractCodeFence(res.text);
      return parseCSV(toParse);
    },
    // optional, if omitted streaming `.output()` is not supported for this format type
    parseChunk: (chunk) => {
      // here `chunk` contains only the most recent data, `partialResponse` contains
      // a Message with all chunks received so far
      const toParse = extractPartialCodeFence(chunk.accumulatedText);
      return parseCSV(toParse);
    },
    instructions: `Output should be in CSV format with the following columns:\n\n${schemaToCSVSpec(req.output.schema)}.`
  }
});
```

The parser definition semantics should be flexible enough to handle many different scenarios, including:

1. **Built-up response:** where each chunk in a stream returns an incrementally more complete response
2. **Buffered chunking:** where e.g. JSONL is streamed but only on complete objects (so `parseChunk` must have the ability to return `null` and not emit a chunk to the end user)
3. **Options/Schema:** allow a format parser to access the full request including output schema and control how output validation occurs.

To use a custom format is simple: just use the `format` option already in output.

```ts
generate({
  prompt: "Generate a contact list with 10 people.",
  output: {
    format: 'csv',
    schema: z.array(z.object({firstName: z.string(), lastName: z.string()})
  },
});
```

### Work Plan

- [x] Create an interface for formatters that can handle streaming as well as custom instructions.
- [x] Implement a base set of default formatters - `text`, `json`, `array`, `jsonl`, `enum`.
- [x] Refactor `generate` and `generateStream` to play nicely with new streaming semantics.
- [ ] Figure out how to reconcile new streaming semantics with tool loop.
- [ ] Figure out how media and multi-modal output should work.
- [x] Add `instructions?: boolean | string | (req: GenerateRequest) => string` for custom instructions control.
- [x] Add `constrained?: boolean` option to output for schema-constrained generation, and implement it for Gemini models.
- [ ] Add enum mode to Gemini models.
- [ ] Add support for `ai.defineFormat` to define custom formats.

PR #1125, opened 2024-10-25
[JS] Adds custom format implementations. #708 part 1

mbleigh:
This is the first PR in a series that will work toward #708, which has now been updated with a work plan for the various pieces that need to go into it..

@galihlprakoso FYI -- this feature is pretty big and consequential and has been on my list to work through for a long time. I appreciate the contribution (and would love if you can find more things you want to add!) but I'm going to implement custom formats directly since there's a lot of complex interaction between various parts of the library.

In this PR I've added a new interface for custom formats and implemented several default formats but haven't yet integrated it with `generate` or the broader ecosystem. It is well-tested though.

### DIFF
--- a/js/ai/src/formats/enum.ts
+++ b/js/ai/src/formats/enum.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import { GenkitError } from '@genkit-ai/core';
 import type { Formatter } from './types';
 
--- a/js/ai/src/formats/enum.ts
+++ b/js/ai/src/formats/enum.ts
@@ -1,0 +1,24 @@
+import { GenkitError } from '@genkit-ai/core';
+import type { Formatter } from './types';
+
+export const enumParser: Formatter = (request) => {
+  const schemaType = request.output?.schema?.type;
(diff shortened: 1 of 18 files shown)


galihlprakoso:
Thank you, @mbleigh! I understand now. It was a good exercise to start getting familiar with the codebase, as it allowed me to explore the core, AI, and other modules. I'll look for another task to work on.

js/ai/src/formats/enum.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (schemaType && schemaType !== 'string' && schemaType !== 'enum') {
+    throw new GenkitError({
+      status: 'INVALID_ARGUMENT',
+      message: `Must supply a 'string' or 'enum' schema type when using the enum parser format.`,
+    });
+  }
+
+  let instructions: boolean | string = false;
+  if (request.output?.schema?.enum) {
+    instructions = `Output should be ONLY one of the following enum values. Do not output any additional information or add quotes.\n\n${request.output?.schema?.enum.map((v) => v.toString()).join('\n')}`;
```
apascal07, line 31: Have you given any thought to how we might integrate this with the plan to allow devs to specify their own schema instruction message?

mbleigh: Yes, the idea is that the format provides a "base layer" of instructions that are set to be augmented into the prompt; however, the `instructions` field will be added to `output` and you can either override it with your own custom instructions or set it to `false` to opt-out of automatic instructions altogether if you included it otherwise in your prompt. For example:

```ts
ai.generate({
  prompt: "Output some JSON in the format `{foo: 'bar'}`."
  output: {format: 'json', instructions: false}
});
```

apascal07 (approved):

js/ai/src/formats/enum.ts
apascal07: Do we have a placeholder to inject the schema into the prompt along with the custom instructions? Sorry if I missed it.

mbleigh: No placeholder (since if you just wanted to insert the json schema raw you'd probably do that). For custom formatters, they have access to the whole request when returning their instructions. For users, I'm thinking we probably allow something like:

```ts
ai.generate({
  format: {
    schema: MySchema,
    instructions: (req) => `Use this format:\n\n${formatSchemaForInstructions(req.output.schema)}`,
  }
});
```

PR #1130, opened 2024-10-29
Fix: allow use more memory for some pnpm cmds

alonsopec89:
Open to discussion, but trying to do builds in a different os, I was getting "Worker terminated due to reaching memory limit: JS heap out of memory" This PR try to help people that got something like this

Checklist (if applicable):
- [X] Docs updated


alonsopec89:
closing due will be tackled here https://github.com/egoist/tsup/issues/920#issuecomment-2159611615

PR #1136, opened 2024-10-29
Moves GenerateResponse and GenerateResponseChunk to their own files

mbleigh:
Refactoring to better organize the `ai` package, also moves `generateAction` into the new `generate` dir.

Checklist (if applicable):
- [ ] Docs updated


js/ai/src/formats/types.d.ts
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import { GenerateResponse, GenerateResponseChunk } from '../generate';
+import { GenerateResponse } from '../generate';
+import { GenerateResponseChunk } from '../generate/chunk';
```
MichaelDoyle, line 18: Should this line have changed? I see both `GenerateResponse` and `GenerateResponseChunk` are still exported from `generate`.

MichaelDoyle (approved):

js/ai/src/formats/types.d.ts
mbleigh: Good catch.

mbleigh:
> Assuming it's just a move, hard to tell if anything has changed.

It is, but yes it's hard to tell.

PR #1169, opened 2024-11-04
fix: allow passing promise of TelemetryConfig to enableTelemetry

pavelgj:


js/core/src/tracing.ts
```diff
@@ -35,18 +35,26 @@ let nodeOtelConfig: TelemetryConfig | null = null;
...
 }
 
 /**
  * Enables tracing and metrics open telemetry configuration.
  */
-export function enableTelemetry(telemetryConfig: TelemetryConfig) {
+export async function enableTelemetry(
+  telemetryConfig: TelemetryConfig | Promise<TelemetryConfig>
+) {
+  global[instrumentationKey] = true;
```
andrewbrook, line 51: Could there still be a race condition if telemetryConfig takes a long time to resolve from here?

andrewbrook (approved):

js/core/src/tracing.ts
pavelgj: what might happen is that some telemetry might not get recorded until it competes initialization... but we can actually block requests on telemetry initialization... let me do that.

PR #1171, opened 2024-11-04
Formatters working E2E, including chunk and response parsers. #708 continued

mbleigh:
This will need to be refined over time, but I've gotten it working across both Google AI and Vertex now with a test of each format type against both Gemini 1.5 Flash and Pro.

Checklist (if applicable):
- [ ] Docs updated


js/testapps/format-tester/src/index.ts
```diff
@@ -0,0 +1,155 @@
...
+          description: z.string(),
+          price: z.number(),
+          stock: z.number(),
+          color: z.string(),
+          tags: z.array(z.string()),
+        })
+      ),
+    },
+  },
+  enum: {
```
pavelgj, line 117: what's the streaming expectation for `enum`? I'm seeing

```json
{"text":"HIGH \n","output":null}
```

in non-streaming mode I see `response.output` as `HIGH`.

so, should expect in streaming mode to see

```json
{"text":"HIGH \n","output":"HIGH"}
```

right?

mbleigh: I didn't write a chunk parser for enum mode since it doesn't make sense to "partially" get an enum...maybe it should just throw an error instead of nulling?

js/ai/src/formats/array.ts
```diff
@@ -24,21 +24,21 @@ export const arrayFormatter: Formatter<unknown[], unknown[]> = {
...
-    if (request.output?.schema) {
+    if (schema) {
       instructions = `Output should be a JSON array conforming to the following schema:
     
-    \`\`\`
-    ${JSON.stringify(request.output!.schema!)}
-    \`\`\`
+\`\`\`
+${JSON.stringify(schema)}
+\`\`\`
```
pavelgj, line 41: what's the plan for making these instructions configurable by the user? These formatters are not too complicated, are we saying "just write your own"?

pavelgj (approved):

js/ai/src/formats/array.ts
mbleigh: Users can do:

```ts
ai.generate({
  output: {
    format: 'array',
    instructions: false, // BYO instructions
    // or
    instructions: 'Output should be a JSON array.'
  }
});
```

PR #1188, opened 2024-11-06
Fix: display error in inspect page

alonsopec89:
Currently we are not displaying the error, with this PR we'll be able to see them
Screenshot 2024-11-05 6 06 52 PM

Checklist (if applicable):
- [ X] Tested (manually, unit tested, etc.)
- [ ] Docs updated


MichaelDoyle:
How can we change this such that the error messaging shows up in the error callout, instead of showing up as "output"?

Potentially `span.RecordError(err)`? I believe that's how it works in TS.

alonsopec89:
> How can we change this such that the error messaging shows up in the error callout, instead of showing up as "output"?
> 
> Potentially `span.RecordError(err)`? I believe that's how it works in TS.
> 
> See also:
> 
> * https://github.com/firebase/genkit/blob/main/js/core/src/tracing/instrumentation.ts#L110-L116
> * https://opentelemetry.io/docs/languages/go/instrumentation/#record-errors

@MichaelDoyle `span.RecordError(err)` is not enough, we need the `status.code == 2`, but since the enum from opentelemetry the 2 is "Ok" , should we use our own enum?

MichaelDoyle:
Ah - thanks. That's really unfortunate. :) I think using the codes.Error is still the right thing to do, even though it doesn't match.

I found some more context: https://github.com/open-telemetry/opentelemetry-go/issues/2098

We might need to set things up such that the traces page in the dev ui is aware of which runtime the trace/spans came from and does the appropriate conversion.

MichaelDoyle:
I just submitted https://github.com/FirebasePrivate/genkit-ui/pull/1096 which I think will help - can you try w/ the latest dev ui code (`next` branch)  and see how that looks?

MichaelDoyle:
Did this end up working out? Got a more recent screenshot?

alonsopec89:
> Did this end up working out? Got a more recent screenshot?

@MichaelDoyle It is still working in the newest version
Screenshot_2024-11-18_at_8 21 06_a m

MichaelDoyle (approved):

PR #1195, opened 2024-11-06
fix: made ai.prompt sync

### LINKED ISSUE #1152: [JS] ai.prompt should be sync
In `0.5.x` `promptRef` allowed sync references to prompts that could then be executed. This is how `ai.prompt` should behave, as async references are obnoxious.

```ts
// right now in `next`
const {output} = await (await ai.prompt('myPrompt'))({input: 'here'});

// it should be...
const {output} = await ai.prompt('myPrompt')({input: 'here'});
```

This is even more important to be able to set up prompt refs outside of async function bodies since top-level await is a footgun.

pavelgj:
Fixes #1152

### DIFF
--- a/js/genkit/src/genkit.ts
+++ b/js/genkit/src/genkit.ts
@@ -271,29 +271,33 @@ export class Genkit {
    *
    * @todo TODO: Show an example of a name and variant.
    */
-  async prompt<
+  prompt<
     I extends z.ZodTypeAny = z.ZodTypeAny,
     O extends z.ZodTypeAny = z.ZodTypeAny,
     CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
   >(
     name: string,
     options?: { variant?: string }
-  ): Promise<ExecutablePrompt<z.infer<I>, O, CustomOptions>> {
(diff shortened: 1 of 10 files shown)


js/genkit/src/genkit.ts
```diff
@@ -271,29 +271,33 @@ export class Genkit {
...
+      // check the registry first as not all prompt types can be
+      // loaded by dotprompt (e.g. functional)
+      let action = (await this.registry.lookupAction(
+        `/prompt/${name}`
+      )) as PromptAction<I>;
+      // nothing in registry - check for dotprompt file.
+      if (!action) {
+        action = (await prompt(this.registry, name, options))
+          .promptAction as PromptAction<I>;
+      }
```
mbleigh, line 292: You might want to look at `promptRef` and its implementation...I believe it also accounted for reloading prompts when the files changed but don't remember for sure.

pavelgj: I think it's conceptually similar...

mbleigh (approved):
We can work on reloading prompts in dev mode as a followup if needed.

apascal07 (approved):

PR #1219, opened 2024-11-08
hacks i did to make demo work

mbleigh:


MichaelDoyle:
This is targeting the "next" branch which has long since shipped (Genkit 1.0).

Issue #1116, opened 2024-10-24
[JS] premble prompt should re-render on every chat turn


pavelgj:
```ts
const greeting = ai.definePrompt(
  { name: 'greeter', tools: [changeUserName] },
  `{{ role "system"}} user's name is {{ @state.userName}}, greet them`
);

(async () => {
  const session = ai.createSession<MyState>({
    initialState: {
      userName: 'Pavel',
    },
  });

  const chat = session.chat({ prompt: greeting });

  let response = await chat.send('hi');
  console.log(response.text); // hi Pavel

  await session.updateState({
    userName: 'Chris',
  })

  response = await chat.send('hi');
  console.log(response.text); // this should say `hi Chris`, but right now says `hi Pavel`
})();
```

pavelgj:
I got this to work: 
```
const greeting = ai.definePrompt(
  { name: 'greeter' },
  `{{ role "system"}} user's name is {{ @state.userName}}. Important: greet them by this name.`
);

(async () => {
  const session = ai.createSession({
    initialState: {
      userName: 'Pavel',
    },
  });

  let chat = session.chat(greeting);

  let response = await chat.send('hi');
  console.log(response.text);

  await session.updateState({
    userName: 'Chris',
  })
  chat = session.chat(greeting); // <----- important: causes the preamble to refresh
  response = await chat.send('hi');
  console.log(response.text);
})();
```

PR #1225, opened 2024-11-09
fix: reload the preamble every time a 'fresh' chat session is created

pavelgj:
Parent PR: https://github.com/firebase/genkit/pull/1215

```ts
const greeting = ai.definePrompt(
  { name: 'greeter' },
  `{{ role "system"}} user's name is {{ @state.userName}}. Important: greet them by this name.`
);

(async () => {
  const session = ai.createSession({
    initialState: {
      userName: 'Pavel',
    },
  });

  let chat = session.chat(greeting);

  let response = await chat.send('hi');
  console.log(response.text); // hi Pavel

  await session.updateState({
    userName: 'Chris',
  })
  chat = session.chat(greeting); // <----- important: causes the preamble to refresh
  response = await chat.send('hi');
  console.log(response.text); // hi Chris
})();
```

Fixes: #1116

### DIFF
--- a/js/genkit/src/genkit.ts
+++ b/js/genkit/src/genkit.ts
@@ -308,35 +308,6 @@ export class Genkit {
   }
 
   /**
-   * Defines and registers a dotprompt.
-   *
-   * This is an alternative to defining and importing a .prompt file.
-   *
-   * ```ts
-   * const hi = ai.definePrompt(
-   *   {
-   *     name: 'hi',
-   *     input: {
-   *       schema: z.object({
-   *         name: z.string(),
-   *       }),
-   *     },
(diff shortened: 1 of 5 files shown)


js/genkit/src/genkit.ts
```diff
@@ -308,9 +308,7 @@ export class Genkit {
   }
 
   /**
-   * Defines and registers a dotprompt.
-   *
-   * This is an alternative to defining and importing a .prompt file.
+   * Defines and registers a function-based prompt.
```
pavelgj, line 311: I just swapped the order of these signatures... helps VSCode.... not sure why....

PR #1227, opened 2024-11-11
feat: added `functionCallingConfig` config option for gemini models

pavelgj:
```ts
  model: gemini15Flash,
  prompt: 'hi',
  tools: [toolA, toolB],
  config: {
    functionCallingConfig: {
      mode: 'ANY',
    }
  }
});
```

### DIFF
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -16,6 +16,7 @@
 
 import {
   FileDataPart,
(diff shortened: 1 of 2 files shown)


apascal07 (approved):
Do we expect users to consult the Gemini docs to know what these options are?

pavelgj:
> Do we expect users to consult the Gemini docs to know what these options are?

Not sure...

PR #1229, opened 2024-11-11
chore(js/testapps): update flow-sample1 and README

Gustolandia:
This PR includes updates to flow-sample1 and an enhanced README.md.


.genkit/servers/tools.json
pavelgj, line 1: should should not be checked in...

pavelgj (approved):

PR #1236, opened 2024-11-11
1174 set log level to debug in golang

alonsopec89:
Tested manually and seems like works

Checklist (if applicable):
- [ ] Docs updated


apascal07:
We want to go through all of the log statements we have today and change non-essential stuff to DEBUG instead of INFO like it is today. We still want to log all of the levels if the right argument is passed.

go/core/logger/logger.go
apascal07, line 1: Do we need the changes in this file?

alonsopec89: I just moved to a new file

go/core/logger/logger.go
```diff
@@ -26,10 +26,12 @@ import (
 func init() {
 	// TODO: Remove this. The main program should be responsible for configuring logging.
```
apascal07, line 27: Has this TODO been resolved? How does the log level get configured by the user?

alonsopec89: yes, just has been resolved I think

alonsopec89: yes, It is resolved, the user can set the log level like:

```go
if err := genkit.Init(ctx, &genkit.Options{LogLevel: slog.LevelWarn}); err != nil {
        log.Fatal(err)
}
```

apascal07: Hmm, this is a bit tangential to the logging. In theory, flows/etc can be used without calling `genkit.Init` which is mostly responsible for starting the servers, which can be sidestepped by starting up your own server and exposing the flows directly. We still want to configure the logging level appropriately in those cases. In JS, we pulled it out of Genkit instance config and it's a top-level/global function.

alonsopec89: I refactored it, now we can set the log level like:

```go
logger.SetLevel(slog.LevelError)
```

and the default log level now is: LevelDebug

apascal07 (approved):

PR #1239, opened 2024-11-11
chore: removed generate/generateStream from executable prompts

pavelgj:
unnecessary/redundant as everything can be done with ()/.stream()

### DIFF
--- a/js/ai/src/prompt.ts
+++ b/js/ai/src/prompt.ts
@@ -83,10 +83,10 @@ export interface ExecutablePrompt<
    * @param opt Options for the prompt template, including user input variables and custom model configuration options.
    * @returns the model response as a promise of `GenerateStreamResponse`.
    */
-  <In extends I, Out extends O, Opts extends CustomOptions>(
-    input?: In,
-    opts?: PromptGenerateOptions<Out, Opts>
-  ): Promise<GenerateResponse<z.infer<Out>>>;
+  (
+    input?: I,
+    opts?: PromptGenerateOptions<O, CustomOptions>
+  ): Promise<GenerateResponse<z.infer<O>>>;
 
   /**
    * Generates a response by rendering the prompt template with given user input and then calling the model.
@@ -94,10 +94,10 @@ export interface ExecutablePrompt<
    * @param opt Options for the prompt template, including user input variables and custom model configuration options.
    * @returns the model response as a promise of `GenerateStreamResponse`.
    */
-  stream<In extends I, Out extends O, Opts extends CustomOptions>(
-    input?: In,
-    opts?: PromptGenerateOptions<Out, Opts>
-  ): Promise<GenerateStreamResponse<z.infer<Out>>>;
+  stream(
+    input?: I,
+    opts?: PromptGenerateOptions<O, CustomOptions>
+  ): Promise<GenerateStreamResponse<z.infer<O>>>;
 
   /**
    * Renders the prompt template based on user input.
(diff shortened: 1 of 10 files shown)


js/ai/src/prompt.ts
```diff
@@ -66,12 +66,8 @@ export function isPrompt(arg: any): boolean {
 export type PromptGenerateOptions<
   I = undefined,
   CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
-> = Omit<
-  GenerateOptions<z.ZodTypeAny, CustomOptions>,
-  'prompt' | 'input' | 'model'
-> & {
+> = Omit<GenerateOptions<z.ZodTypeAny, CustomOptions>, 'prompt' | 'model'> & {
```
apascal07, line 69: `model` is already optional in `GenerateOptions`, why omit it just to put it back? Also, I think we need to get rid of `I` and replace it with `O` like this and carry it through in the functions:

```
export type PromptGenerateOptions<
  CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
> = Omit<GenerateOptions<O, CustomOptions>, 'prompt'>;
```

js/ai/src/prompt.ts
```diff
@@ -102,27 +98,7 @@ export interface ExecutablePrompt<
    */
   stream<Out extends O>(
```
apascal07: Also why do we have `Out extends O` here?

js/ai/src/prompt.ts
pavelgj: updated

js/ai/src/prompt.ts
pavelgj: I swear there was a reason... but don't remember... some weird inference issue. Will remove for now....

js/ai/src/prompt.ts
```diff
@@ -89,51 +85,33 @@ export interface ExecutablePrompt<
...
    * @returns a `GenerateOptions` object to be used with the `generate()` function from @genkit-ai/ai.
    */
-  render<Out extends O>(
-    opt: PromptGenerateOptions<I, CustomOptions>
-  ): Promise<GenerateOptions<CustomOptions, Out>>;
+  render(
+    opt: PromptGenerateOptions<I, CustomOptions> & {
+      input?: I;
+    }
+  ): Promise<GenerateOptions<CustomOptions, O>>;
```
apascal07, line 114: Should be `GenerateOptions<O, CustomOptions>`, it's flipped.

pavelgj: fixed

js/ai/src/prompt.ts
apascal07: ?

js/ai/src/prompt.ts
pavelgj: yes, remembered the reason -- it's to allow overriding it at the call site... ex:
```ts
await ai.prompt('recipe')<typeof RecipeSchema>(input)
```

js/ai/src/prompt.ts
pavelgj: I mean updated now...

js/ai/src/prompt.ts
```diff
@@ -89,9 +83,9 @@ export interface ExecutablePrompt<
    * @param opt Options for the prompt template, including user input variables and custom model configuration options.
    * @returns the model response as a promise of `GenerateStreamResponse`.
    */
-  <Out extends O>(
-    input?: I,
-    opts?: PromptGenerateOptions<I, CustomOptions>
+  <In extends I, Out extends O, Opts extends CustomOptions>(
```
apascal07, line 86: I understand why we have `Out` to make the return type strongly typed but not why `In` or `Opts`. The latter two can be derived from the arguments, no?

pavelgj: OK, I think doing it this way is sufficient... removed these overridable types for now... can reintroduce later if users ask for it.

```ts
await ai.prompt<typeof InputSchema, typeof OutputSchema>('recipe')(input)
```

apascal07 (approved):
Perfect! Thanks!

PR #1242, opened 2024-11-12
Add new endpoint to trpc for polling runtime status from dev ui

MichaelDoyle:
IDX does not currently support SSE (text/stream) via its proxy. We will poll for runtime(s) via dev ui for now.


tonybaroneee (approved):
Super sad face

pavelgj (approved):

PR #1254, opened 2024-11-12
fix: retry runtime file reads -- on slow machines the delay causes errors

pavelgj:


apascal07 (approved):
Is `fs.readFile` a network call in IDX?

pavelgj:
> Is `fs.readFile` a network call in IDX?

error is from `JSON.parse` because the file is empty. I suspect high initial latency on disk write ops.

PR #1293, opened 2024-11-13
docs: revised nextjs guide

kevinthecheung:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/docs/nextjs.md
+++ b/docs/nextjs.md
@@ -1,0 +1,190 @@
+# Use Genkit in a Next.js app 
+
+This page shows how you can use Genkit flows as server actions in your Next.js
+apps.
+
+## Before you begin
+
+You should be familiar with Genkit's concept of [flows](flows), and how to write
+them.
+
+## Create a Next.js project
+
+If you don't already have a Next.js project that you want to add generative AI
+features to, you can create one for the purpose of following along with this
+page:
+
+```posix-terminal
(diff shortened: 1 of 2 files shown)


docs/nextjs.md
```diff
@@ -0,0 +1,190 @@
...
+
+import { menuSuggestionFlow } from '@/app/genkit';
+import { useState } from 'react';
+
+export default function Home() {
+  const [menuItem, setMenuItem] = useState<string>('');
+
+  async function getMenuItem(formData: FormData) {
+    const theme = formData.get('theme')?.toString() ?? '';
+    const suggestion = await menuSuggestionFlow(theme);
```
pavelgj, line 109: This actually works???

pavelgj (approved):

docs/nextjs.md
kevinthecheung: it works locally. should i be more skeptical?

pavelgj: No, all good, I was hoping it would, just haven't tested it. 

I'm not sure if it's considered a good practice (re security implications), but still neat.
As a follow up we should also document how to use genkit with API routes....

kevinthecheung: fyi i just successfully deployed it with app hosting

Issue #1184, opened 2024-11-05
[JS] missing uuid import on 0.9 RC


cabljac:
**Describe the bug**

Running a simple flow in Next.js (in a test) it complained about missing `uuid`.

I'm not 100% this is a genkit issue, but my flow was just this:

```typescript
import { z } from "genkit";
import { ai } from "../genkit";

export const mainFlow = ai.defineFlow(
  {
    name: "mainFlow",
    inputSchema: z.string(),
    outputSchema: z.string(),
  },
  async (input) => input
);
```

and my test:

```typescript
import { describe, it, expect } from "vitest";
import { mainFlow } from "./mainFlow";

describe("mainFlow", () => {
  it("should run", async () => {
    const result = await mainFlow("hello");
    expect(result).toBe("hello");
  });
});
```

**Expected behavior**
Not to need to install `uuid` as a dependency, probably

**Screenshots**
<img width="441" alt="Screenshot 2024-11-05 at 20 17 41" src="https://github.com/user-attachments/assets/d3553b39-3e9c-49e0-a064-642d6b707120">

**Runtime (please complete the following information):**
 - OS: [MacOS]

** Node version
❯ node --version
v20.11.1

**Additional context**
This is on the 0.9 RC

cabljac:
We found we needed to add `uuid` during testing in [this update](https://github.com/TheFireCo/genkit-plugins/pull/155) as well

Gustolandia:
Adressed in https://github.com/firebase/genkit/pull/1296

PR #1296, opened 2024-11-14
@Invertase Fix missing uuid import for 0.9 RC

Gustolandia:
This PR addresses the missing uuid dependency issue reported in 0.9 RC. It ensures that uuid is properly installed in the required packages. The changes include adding uuid to the necessary package.json files and updating related imports.

Changes:

Added uuid to missing dependencies.
Updated relevant code to ensure uuid is correctly imported and used.

### DIFF
--- a/js/pnpm-lock.yaml
+++ b/js/pnpm-lock.yaml
@@ -718,6 +718,9 @@ importers:
       genkit:
         specifier: workspace:*
         version: link:../../genkit
+      uuid:
+        specifier: ^10.0.0
+        version: 10.0.0
     devDependencies:
       typescript:
         specifier: ^5.6.2
@@ -798,6 +801,9 @@ importers:
       pdfjs-dist-legacy:
         specifier: ^1.0.1
         version: 1.0.1
+      uuid:
+        specifier: ^10.0.0
+        version: 10.0.0
     devDependencies:
       '@types/pdf-parse':
         specifier: ^1.1.4
@@ -1197,7 +1203,7 @@ importers:
         version: link:../../plugins/ollama
       genkitx-openai:
         specifier: ^0.10.1
-        version: 0.10.1(@genkit-ai/ai@0.9.0-rc.0)(@genkit-ai/core@0.9.0-rc.0)
+        version: 0.10.1(@genkit-ai/ai@0.9.0-rc.2)(@genkit-ai/core@0.9.0-rc.2)
     devDependencies:
       rimraf:
         specifier: ^6.0.1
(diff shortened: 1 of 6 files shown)


tonybaroneee (changes requested):
Sorry, can you point me to where we use `uuid` in these respective packages? I'm not seeing it.

Gustolandia:
> Sorry, can you point me to where we use `uuid` in these respective packages? I'm not seeing it.

Hi. 

grep -rl "uuid" --exclude-dir={node_modules,dist,lib}

I installed to the nearest package.json in order to make sure it is not missing as it happened in more than 1 situation.

The full script is here, if you want to give it a look:

const fs = require('fs');
const path = require('path');
const { execSync } = require('child_process');

let uuidReferences;
try {
  uuidReferences = execSync('grep -rl "uuid" --exclude-dir={node_modules,dist,lib} .', {
    encoding: 'utf-8',
  })
    .trim()
    .split('\n')
    .filter((file) => file);
} catch (error) {
  console.error('No references to "uuid" found.');
  process.exit(0);
}

const missingDepsByDir = new Map();

uuidReferences.forEach((filePath) => {
  const packageDir = findNearestPackageJsonDir(filePath);
  const packageJsonPath = path.join(packageDir, 'package.json');

  if (!isUuidInDependencies(packageJsonPath)) {
    if (!missingDepsByDir.has(packageDir)) {
      missingDepsByDir.set(packageDir, new Set());
    }
    missingDepsByDir.get(packageDir).add('uuid');
  }
});

missingDepsByDir.forEach((deps, dir) => {
  console.log(`Installing in ${dir}: ${Array.from(deps).join(', ')}`);
  try {
    const relativeDir = path.relative(process.cwd(), dir);
    execSync(
      `pnpm add ${Array.from(deps).join(' ')} --filter ${relativeDir || '.'}`,
      {
        stdio: 'inherit',
      }
    );
  } catch (error) {
    console.error(`Failed to install dependencies in ${dir}:`, error.message);
  }
});

console.log('All missing uuid dependencies installed.');

function findNearestPackageJsonDir(filePath) {
  let dir = path.dirname(filePath);
  while (dir !== path.parse(dir).root) {
    if (fs.existsSync(path.join(dir, 'package.json'))) {
      return dir;
    }
    dir = path.dirname(dir);
  }
  throw new Error(`No package.json found for ${filePath}`);
}

function isUuidInDependencies(packageJsonPath) {
  const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
  return (
    (packageJson.dependencies && packageJson.dependencies['uuid']) ||
    (packageJson.devDependencies && packageJson.devDependencies['uuid'])
  );
}

Gustolandia:
This is addressing https://github.com/firebase/genkit/issues/1184

tonybaroneee:
These testapps and samples do not use `uuid` therefore do not need the import. I think there's a different root cause of https://github.com/firebase/genkit/issues/1184 and this is not a fix for it.

Gustolandia:
> These testapps and samples do not use `uuid` therefore do not need the import. I believe the root cause of #1184 is from something else, and this is not a fix for it.

You still had the issue with this update? I did not. My bad.

tonybaroneee:
> > These testapps and samples do not use `uuid` therefore do not need the import. I believe the root cause of #1184 is from something else, and this is not a fix for it.
> 
> You still had the issue with this update? I did not. My bad.

Nope, was just reviewing the PR :) Glad that it's not an issue anymore (in that case, feel free to close the PR!).

PR #1305, opened 2024-11-15
fix: firebase documentation after 0.9 upgrade

xavidop:
This PR updates the docs after upgrading to 0.9.0:
1. Since the firebase emulator and the genkit UI uses by default the same port (4000), it is better to start first the genkit UI and then the firebase emulator. The firebase emulator autodetect that the port 4000 is in use and uses another one.
2. Added the npm command provided by the firebase template `npm run genkit:start`
3. fixes this visualization problem: https://firebase.google.com/docs/genkit/monitoring#trace_storage_and_developer_ui

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated

### DIFF
--- a/docs/firebase.md
+++ b/docs/firebase.md
@@ -416,27 +416,11 @@ Firebase offers a
 [suite of emulators for local development](/docs/emulator-suite), which you can
 use with Genkit.
 
-First, You will need to launch the Genkit Dev UI:
-
-```posix-terminal
-cd $PROJECT_ROOT/functions
-
-npx genkit start -- npx tsx --watch src/index.ts
-```
-
-or
-  
-```posix-terminal
-cd $PROJECT_ROOT/functions
-
-npm run genkit:start
-```
-
-To use Genkit with the Firebase Emulator Suite, start the Firebase emulators
+To use the Genkit Dev UI with the Firebase Emulator Suite, start the Firebase emulators
 like this:
 
 ```posix-terminal
-firebase emulators:start --inspect-functions
+npx genkit start -- firebase emulators:start --inspect-functions
 ```
 
 This will run your code in the emulator and run the Genkit framework in
--- a/docs/firebase.md
+++ b/docs/firebase.md
@@ -400,29 +400,48 @@ cd $PROJECT_ROOT/functions
 npx genkit start -- npx tsx --watch src/index.ts
 ```
 
+or
+  
+```posix-terminal
+cd $PROJECT_ROOT/functions
+
+npm run genkit:start
+```
+
+You can now navigate to the URL printed by the `genkit start` command to access.
+
 ## Optional: Developing using Firebase Local Emulator Suite
 
 Firebase offers a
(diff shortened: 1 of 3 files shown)


js/plugins/vertexai/src/gemini.ts
```diff
@@ -507,7 +507,8 @@ export function defineGeminiModel(
       }
       // Cannot use tools and function calling at the same time
       const jsonMode =
-        (request.output?.format === 'json' || !!request.output?.schema) &&
+        (request.output?.format === 'json' || !!request.output?.schema ||
```
pavelgj, line 510: I'm not sure the issue should be solved at the plugin level. 
I'm thinking perhaps somewhere here: https://github.com/firebase/genkit/blob/4e53051b0e18aad185d7b5709262242bf50f0a3f/js/ai/src/generate.ts#L249 
I'm looking into solving this at genkit `generate` level.

xavidop: that works too, I created a PR in here too: https://github.com/TheFireCo/genkit-plugins/pull/161

xavidop:
> @xavidop this is awesome. would you mind splitting the code changes from the docs changes (i.e. 2 PRs), since they are not related? I think we'll have an easier time getting the right folks to review that way.
> 
> Mike D

@MichaelDoyle done!

docs/firebase.md
```diff
@@ -400,30 +400,49 @@ cd $PROJECT_ROOT/functions
...
+
+npm run genkit:start
+```
+
 To use Genkit with the Firebase Emulator Suite, start the Firebase emulators
 like this:
 
 ```posix-terminal
-GENKIT_ENV=dev firebase emulators:start --inspect-functions
+firebase emulators:start --inspect-functions
```
pavelgj, line 439: The intent here is that we want to show how to "attach" the dev ui to the code running inside the emulator. With 0.9 it's actually even easier:

```posix-terminal
npx genkit start -- firebase emulators:start --inspect-functions
```

pavelgj: just tested to make sure it works -- the command must be run from the `functions` folder

xavidop: Sure! I just wanted to add the npm command in the new template to avoid some friction just in case people do not want to use `npx`. I also wanted to share how to run the firebase emulator with the genkit UI just in case someone wants to use it because it is very useful when you developing firebase apps.

pavelgj: `npx genkit start -- firebase emulators:start --inspect-functions` is how you run emulator with dev ui attached. The way you have written makes to dev ui connect to a separate instance of the code, so dev ui is not connected to the emulator.

xavidop: let me modify the docs to add that option, I do not think that is in the docs

xavidop: @pavelgj what about now?
https://github.com/firebase/genkit/pull/1305/commits/93feb2947a7c3cad8924e750d12086d986585562

PR #1308, opened 2024-11-15
docs(go): basic 0.9 updates

kevinthecheung:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/docs-go/cloud-run.md
+++ b/docs-go/cloud-run.md
@@ -1,8 +1,6 @@
 # Firebase Genkit with Cloud Run
 
-You can deploy Firebase Genkit flows as web services using Cloud Run. This page,
-as an example, walks you through the process of deploying the default sample
-flow.
+You can deploy Firebase Genkit flows as web services using Cloud Run.
 
 1.  Install the [Google Cloud CLI](https://cloud.google.com/sdk/docs/install) if
     you haven't already.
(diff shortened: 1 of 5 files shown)


docs-go/cloud-run.md
```diff
@@ -118,7 +119,7 @@ flow.
     1.  Start the UI:
 
         ```posix-terminal
-        genkit start
+        genkit start -- go run main.go
```
apascal07, line 122: I think the recommended way is `go run .` instead of `go run main.go`.

docs-go/cloud-run.md
```diff
@@ -139,14 +140,14 @@ flow.
     - {Gemini (Google AI)}
 
       ```posix-terminal
-      gcloud run deploy --port 3400 \
+      gcloud run deploy \
```
apascal07, line 143: Any reason you removed this? I'm not sure whether it's needed or not, I just don't think anything has changed in terms of specifying a port.

docs-go/deploy.md
```diff
@@ -120,7 +120,7 @@ actions.
     1.  Start the UI:
 
         ```posix-terminal
-        genkit start
+        genkit start -- go run main.go
```
apascal07, line 123: Same here.

go/internal/doc-snippets/init/main.go
```diff
@@ -38,38 +35,26 @@ func main() {
...
+	m := googleai.Model("gemini-1.5-flash")
+	if m == nil {
+		log.Fatal("Failed to find model")
+	}
 
-	// Initialize Genkit and start a flow server. This call must come last,
-	// after all of your plug-in configuration and flow definitions. When you
-	// pass a nil configuration to Init, Genkit starts a local flow server,
-	// which you can interact with using the developer UI.
-	if err := genkit.Init(ctx, nil); err != nil {
```
apascal07, line 70: I don't know if this is the right place but we still want to show that `genkit.Init(...)` is still required. Do we have it elsewhere?

kevinthecheung: It's in the deployment docs but it doesn't seem like I needed it here? I mean it runs, anyway. I guess I'm still not really clear on what it's for.

docs-go/cloud-run.md
kevinthecheung: I thought it's preferable to just use the port that cloud run sets via PORT when you deploy

go/internal/doc-snippets/init/main.go
apascal07: `genkit.Init` starts the reflection server and the flow server. So it'll run without it but dev UI won't work and in Go, deploying this code won't expose it in any way via HTTP. It's a required part of the setup, whether dev or prod.

go/internal/doc-snippets/init/main.go
apascal07, line 1: Ok wait, if you're going to remove the flow from this sample then there's no purpose to `genkit.Init`. It's both or none. :)

PR #1309, opened 2024-11-15
Correct JS Flows example #3

inlined:
Example is for structured output but the schema was string.

Checklist (if applicable):
- [ ] Changelog updated
- [x] Docs updated

### DIFF
--- a/js/doc-snippets/src/flows/index.ts
+++ b/js/doc-snippets/src/flows/index.ts
@@ -67,7 +67,10 @@ export const menuSuggestionFlowMarkdown = ai.defineFlow(
   {
     name: 'menuSuggestionFlow',
     inputSchema: z.string(),
-    outputSchema: z.string(),
+    outputSchema: z.object({
+      dishname: z.string(),
+      description: z.string(),
+    }),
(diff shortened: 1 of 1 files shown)


kevinthecheung:
This one really does return a string though

kevinthecheung:
In the documentation, this example is used to show that the output schema of a generate call might not be the same as the output schema of the flow it’s in

js/doc-snippets/src/flows/index.ts
```diff
@@ -67,7 +67,10 @@ export const menuSuggestionFlowMarkdown = ai.defineFlow(
   {
     name: 'menuSuggestionFlow',
     inputSchema: z.string(),
-    outputSchema: z.string(),
+    outputSchema: z.object({
```
pavelgj, line 70: No... the flow output type is `z.string()`... see line 84:
```ts
return `**${output.dishname}**: ${output.description}`
```

inlined: Oh whoops. Dunno how I got confused when reading docs.

PR #1316, opened 2024-11-15
breaking(go): update flow streaming protocol to SSE

hugoaguirre:
OK
```shell
curl -X POST -H "Accept: text/event-stream" http://localhost:3400/streamy -d '{"data":5}'
data: {"count":0}\n\n
data: {"count":1}\n\n
data: {"count":2}\n\n
data: {"count":3}\n\n
data: {"count":4}\n\n
data: {"result": "done: 5, streamed: 5 times"}\n\n
```

ERROR
```shell
curl -X POST -H "Accept: text/event-stream" http://localhost:3400/streamy -d '{"data":5}'                                                                       
data: {"message": {"count":0}}\n\n
data: {"message": {"count":1}}\n\n
data: {"message": {"count":2}}\n\n
data: {"message": {"count":3}}\n\n
data: {"message": {"count":4}}\n\n
data: {"error": {"status": "INTERNAL", "message": "stream flow error", "details": "this is a dummy error"}}\n\n         
```

### DIFF
--- a/go/genkit/servers.go
+++ b/go/genkit/servers.go
@@ -418,7 +418,7 @@ func nonDurableFlowHandler(f flow) func(http.ResponseWriter, *http.Request) erro
 		out, err := f.runJSON(r.Context(), r.Header.Get("Authorization"), body.Data, callback)
 		if err != nil {
 			if r.Header.Get("Accept") == "text/event-stream" || stream {
-				_, err = fmt.Fprintf(w, `data: {"error": {"status": "INTERNAL", "message": "stream flow error", "details": %s}}\n\n`, err)
+				_, err = fmt.Fprintf(w, `data: {"error": {"status": "INTERNAL", "message": "stream flow error", "details": "%s"}}\n\n`, err)
 				return err
 			}
 			return err
--- a/go/genkit/servers.go
+++ b/go/genkit/servers.go
@@ -404,7 +404,7 @@ func nonDurableFlowHandler(f flow) func(http.ResponseWriter, *http.Request) erro
 			w.Header().Set("Transfer-Encoding", "chunked")
 			// Stream results are newline-separated JSON.
 			callback = func(ctx context.Context, msg json.RawMessage) error {
-				_, err := fmt.Fprintf(w, "data: %s\n\n", msg)
+				_, err := fmt.Fprintf(w, `data: {"message": %s}\n\n`, msg)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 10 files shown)


go/genkit/servers.go
```diff
@@ -380,12 +380,12 @@ func nonDurableFlowHandler(f flow) func(http.ResponseWriter, *http.Request) erro
...
 		}
 		var callback streamingCallback[json.RawMessage]
-		if stream {
+		if r.Header.Get("Accept") == "text/event-stream" || stream {
 			w.Header().Set("Content-Type", "text/plain")
 			w.Header().Set("Transfer-Encoding", "chunked")
 			// Stream results are newline-separated JSON.
 			callback = func(ctx context.Context, msg json.RawMessage) error {
-				_, err := fmt.Fprintf(w, "%s\n", msg)
+				_, err := fmt.Fprintf(w, "data: %s\n\n", msg)
```
pavelgj, line 388: this needs to be wrapped in a `{"message":  ...}` object.

pavelgj:
```suggestion
				_, err := fmt.Fprintf(w, "data: {"message": %s}\n\n", msg)
```

hugoaguirre: Thanks Pavel, adding the changes now

pavelgj: I would recommend testing this using the genkit client library:
https://github.com/firebase/genkit/blob/main/js/genkit/src/client/client.ts


```ts
import { streamFlow } from 'genkit/client';

(async () => {
  const response = await streamFlow({
    url: 'http://localhost:3400/myTestFlow',
    input: {foo: 'bar'},
  });

  for await (const chunk of response.stream()) {
    console.log(chunk)
  }
  console.log(await response.output())
})()
```

go/tests/api_test.go
```diff
@@ -45,7 +45,7 @@ type test struct {
 const hostPort = "http://localhost:3100"
 
 func TestReflectionAPI(t *testing.T) {
-	filenames, err := filepath.Glob(filepath.FromSlash("../../tests/*.yaml"))
+	filenames, err := filepath.Glob(filepath.FromSlash("../../tests/reflection_api_tests.yaml"))
```
hugoaguirre, line 48: @apascal07 is it okay if I remove the wildcard from here?

apascal07: Yes

tests/src/flow_server_test.ts
```diff
@@ -0,0 +1,104 @@
...
+    let fetchopts = {
+      method: 'POST',
+      headers: {
+        Accept: 'text/event-stream',
+        Connection: 'keep-alive',
+      },
+      body: JSON.stringify(test.post),
+    } as RequestInit;
+
+    await fetchData(`${url}/${test.path}`, fetchopts);
```
pavelgj, line 61: I wonder if it would be better to use `genkit/client` library here for the test?

tests/flow_server_tests.yaml
```diff
@@ -0,0 +1,13 @@
+# This file describes the responses to HTTP requests made
+# to the flow server
+
+# TODO: add more test cases
+
+app: flow_server
```
pavelgj, line 6: you're not testing the streamed data?

tests/src/flow_server_test.ts
hugoaguirre: Ah, good idea. Could you illuminate me on why using `genkit/client` would be better for this? Is it for compatibility reasons?

Using the raw request, we can inspect headers and response payloads to ensure the proper formatting.

tests/flow_server_tests.yaml
hugoaguirre: Yes. My understanding is that this file should be used for test cases for the flow server, unless this label is for something else that I'm not aware of.

Should I change it to something more specific? e.g. `app: stream_flows`

apascal07: I think Pavel means that you're not consuming the whole stream, just one chunk. Now that I think about it, you do need to consume the stream so you can check the final response.

hugoaguirre: Got it. I'll add the expected stream values and test against them 😄

pavelgj (approved):

hugoaguirre:
Go test cases failing due to unrelated errors, fix can be found here: https://github.com/firebase/genkit/pull/1595

PR #1357, opened 2024-11-21
feat: added a way to run action with telemetry, no more 'dev-run-action-wrapper'

pavelgj:
Checklist (if applicable):
- [x] Tested (manually, unit tested)

### DIFF
--- a/js/core/src/reflection.ts
+++ b/js/core/src/reflection.ts
@@ -25,12 +25,7 @@ import { GENKIT_VERSION } from './index.js';
 import { logger } from './logging.js';
 import { Registry } from './registry.js';
 import { toJsonSchema } from './schema.js';
-import {
-  flushTracing,
-  newTrace,
-  setCustomMetadataAttribute,
-  setTelemetryServerUrl,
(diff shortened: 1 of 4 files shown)


js/core/src/reflection.ts
```diff
@@ -169,48 +164,30 @@ export class ReflectionServer {
           return;
         }
         if (stream === 'true') {
-          const result = await newTrace(
-            { name: 'dev-run-action-wrapper' },
```
apascal07, line 173: Why did we have this to begin with? Just trying to consider if it's consequence free to remove.

apascal07 (approved):

MichaelDoyle (approved):

js/core/src/reflection.ts
pavelgj: we had it because we didn't have a clean way to get trace id from an action (what this PR adds). We had to wrap it in a separate span from which we extracted the trace id, then we marked that span as 'genkit-dev-internal' and dev ui has code to ignore those spans when it sees them.

PR #1362, opened 2024-11-21
Remove duplicated call to authPolicy

MichaelDoyle:
Not sure how I ran across this, but it seems to be erroneous that we call this.authPolicy twice. Let me know if I am missing something, though. :)

### DIFF
--- a/js/core/src/flow.ts
+++ b/js/core/src/flow.ts
@@ -349,19 +349,6 @@ export class Flow<
       return;
     }
 
-    try {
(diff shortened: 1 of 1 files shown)


js/core/src/flow.ts
```diff
@@ -349,19 +349,6 @@ export class Flow<
       return;
     }
 
-    try {
```
pavelgj, line 352: lol... what happened here?! Bad merge conflict maybe....
Or maybe guarding against non-deterministic auth policies? j/k

PR #1365, opened 2024-11-21
feat: expose defineFormat on Genkit instance, allowing to define custom formats

pavelgj:
```ts
    ai.defineFormat(
      {
        name: 'banana',
        format: 'banana',
      },
      (schema) => {
        let instructions: string | undefined;

        if (schema) {
          instructions = `Output should be in banana format`;
        }

        return {
          parseChunk: (chunk) => {
            return `banana: ${chunk.content[0].text}`;
          },

          parseMessage: (message) => {
            return `banana: ${message.content[0].text}`;
          },

          instructions,
        };
      }
    );
```

### DIFF
--- a/js/genkit/src/genkit.ts
+++ b/js/genkit/src/genkit.ts
@@ -253,6 +253,37 @@ export class Genkit {
 
   /**
    * Defines and registers a custom model output formatter.
+   * 
+   * Here's an example of a custom JSON output formatter:
+   * 
+   * ```ts
+   * import { extractJson } from 'genkit/extract';
+   *
+   * ai.defineFormat(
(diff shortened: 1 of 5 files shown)


js/genkit/src/genkit.ts
```diff
@@ -247,6 +251,18 @@ export class Genkit {
     return defineSchema(this.registry, name, schema);
   }
 
+  /**
+   * Defines and registers a custom model output formatter.
```
apascal07, line 255: Can you add a little bit more here explaining how it's used like some of the other define methods do?

apascal07 (approved):

PR #1370, opened 2024-11-21
feat(js/plugins/checks): checks evaluator plugin returns multiple scores

HunterHeston:
Update checks evaluators to return multiple results, one per requested policy. 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated

### DIFF
--- a/js/plugins/checks/README.md
+++ b/js/plugins/checks/README.md
@@ -78,41 +78,13 @@ Create a JSON file with the data you want to test. Add as many test cases as you
 
 ```
 
-### Run the evaluators
+### Run the evaluator
 
 ```bash
 # Run all configured classifiers.
-genkit eval:run test-dataset.json --evaluators=checks/all_metrics
-
-# Run just the DANGEROUS_CONTENT classifier.
-genkit eval:run test-dataset.json --evaluators=checks/dangerous_content
-
-# Run just the HARASSMENT classifier.
-genkit eval:run test-dataset.json --evaluators=checks/harassment
-
-# Run just the HATE_SPEECH classifier.
-genkit eval:run test-dataset.json --evaluators=checks/hate_speech
-
-# Run just the MEDICAL_INFO classifier.
-genkit eval:run test-dataset.json --evaluators=checks/medical_info
-
-# Run just the OBSCENITY_AND_PROFANITY classifier.
-genkit eval:run test-dataset.json --evaluators=checks/obscenity_and_profanity
-
-# Run just the PII_SOLICITING_RECITING classifier.
-genkit eval:run test-dataset.json --evaluators=checks/pii_soliciting_reciting
-
-# Run just the SEXUALLY_EXPLICIT classifier.
-genkit eval:run test-dataset.json --evaluators=checks/sexually_explicit
-
-# Run just the VIOLENCE_AND_GORE classifier.
-genkit eval:run test-dataset.json --evaluators=checks/violence_and_gore
+genkit eval:run test-dataset.json --evaluators=checks
 ```
 
-```bash
-# Run all classifiers.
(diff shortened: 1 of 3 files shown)


js/plugins/checks/src/evaluation.ts
```diff
@@ -134,15 +140,20 @@ function createPolicyEvaluator(
...
-        },
-        testCaseId: datapoint.testCaseId,
-      };
+        }
+      });
+
+      return {
+        evaluation: evaluationResults,
+        testCaseId: datapoint.testCaseId
+      }
```
HunterHeston, line 156: @ssbushi Please take a look.

pavelgj:
please follow https://www.conventionalcommits.org/en/v1.0.0/ for PRs

js/plugins/checks/src/evaluation.ts
```diff
@@ -104,15 +116,14 @@ function createPolicyEvaluator(
   projectId: string,
   auth: GoogleAuth,
   ai: Genkit,
-  policy_config: ChecksEvaluationMetricConfig
+  policy_config: ChecksEvaluationMetricConfig[],
```
ssbushi, line 119: This still looks like it is defining one evaluator per policy (which will make one API call per policy). 

You'd want to define a single genkit evaluator that accepts an array of policies and evaluates for all of them. The response can include multiple `ScoreSchema` objects.

ssbushi (changes requested):
See above comment.

js/plugins/checks/src/evaluation.ts
HunterHeston: I left the original logic of creating an evaluator instance for each policy and an evaluator for all the policies configured. The thought was that some users might want to just run one of the policies individually.

That happens here in the constructor:
```ts
  // Individual evaluators, one per configured metric.
  const evaluators = policy_configs.map((policy_config) => {
    return createPolicyEvaluator(
      projectId,
      auth,
      ai,
      [policy_config],
      policy_config.type as string
    );
  });

  // Single evaluator instnace with all configured policies.
  evaluators.push(
    createPolicyEvaluator(projectId, auth, ai, policy_configs, 'all_metrics')
  );
```ts
I'm not sure if that's the right approach though. I'm leaning towards just removing all the individual evaluators and having a single evaluator named checks that runs all configured policies every time. If a user wants to run a different set of policies or one in isolation then they would have to update their plugin config. What do you think?

Comment (HunterHeston) [reply js/plugins/checks/src/evaluation.ts]: @ssbushi removed the extra evaluators. Users can now just run `genkit eval:run test-dataset.json --evaluators=checks`

File: js/plugins/checks/src/evaluation.ts [hunk 3]
```diff
@@ -77,11 +77,7 @@ export function checksEvaluators(
     }
   );
 
-  const evaluators = policy_configs.map((policy_config) => {
-    return createPolicyEvaluator(projectId, auth, ai, policy_config);
-  });
-
-  return evaluators;
+  return [createPolicyEvaluator(projectId, auth, ai, policy_configs)];
```
ssbushi, line 80: Why not just make this return an `EvaluatorAction` instead of an array?

ssbushi (approved):

js/plugins/checks/README.md
```diff
@@ -78,16 +78,11 @@ Create a JSON file with the data you want to test. Add as many test cases as you
...
 ```bash
-# Run just the DANGEROUS_CONTENT classifier.
-genkit eval:run test-dataset.json --evaluators=checks/dangerous_content
-```
-
-```bash
-# Run all classifiers.
-genkit eval:run test-dataset.json --evaluators=checks/dangerous_content,checks/pii_soliciting_reciting,checks/harassment,checks/sexually_explicit,checks/hate_speech,checks/medical_info,checks/violence_and_gore,checks/obscenity_and_profanity
+# Run all configured classifiers.
+genkit eval:run test-dataset.json --evaluators=checks
```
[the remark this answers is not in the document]
HunterHeston: Updated!

js/plugins/checks/src/evaluation.ts
HunterHeston: done!

PR #1375, opened 2024-11-22
feat(js/core): added a way to run actions with side-channel data and streaming

pavelgj:
```ts
    const act = action(
      {
        name: 'foo',
        inputSchema: z.string(),
        outputSchema: z.number(),
      },
      async (input, { sendChunk, context }) => {
        checkAuth(context);
        sendChunk(1);
        sendChunk(2);
        sendChunk(3);
        return input.length;
      }
    );

    const chunks: any[] = [];
    await act('1234', {
      context: { user: 'pavel' },
      onChunk: (c) => chunks.push(c),
    });
```

Checklist (if applicable):

### DIFF
--- a/js/core/src/action.ts
+++ b/js/core/src/action.ts
@@ -34,7 +34,7 @@ export { JSONSchema7 };
 export interface ActionMetadata<
   I extends z.ZodTypeAny,
   O extends z.ZodTypeAny,
-  M extends Record<string, any> = Record<string, any>,
+  S extends z.ZodTypeAny,
 > {
   actionType?: ActionType;
   name: string;
@@ -43,7 +43,8 @@ export interface ActionMetadata<
   inputJsonSchema?: JSONSchema7;
   outputSchema?: O;
   outputJsonSchema?: JSONSchema7;
-  metadata?: M;
+  streamSchema?: S;
+  metadata?: Record<string, any>;
 }
 
 /**
@@ -93,16 +94,15 @@ export interface ActionFnArg<S> {
 export type Action<
   I extends z.ZodTypeAny = z.ZodTypeAny,
   O extends z.ZodTypeAny = z.ZodTypeAny,
-  M extends Record<string, any> = Record<string, any>,
-  S = any,
+  S extends z.ZodTypeAny = z.ZodTypeAny,
 > = ((
   input: z.infer<I>,
   options?: ActionRunOptions<S>
 ) => Promise<z.infer<O>>) & {
-  __action: ActionMetadata<I, O, M>;
+  __action: ActionMetadata<I, O, S>;
   run(
     input: z.infer<I>,
-    options?: ActionRunOptions<S>
+    options?: ActionRunOptions<z.infer<S>>
   ): Promise<ActionResult<z.infer<O>>>;
 };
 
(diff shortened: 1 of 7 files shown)


js/core/src/action.ts
```diff
@@ -57,16 +57,53 @@ export interface ActionResult<O> {
...
+
 /**
  * Self-describing, validating, observable, locally and remotely callable function.
  */
 export type Action<
   I extends z.ZodTypeAny = z.ZodTypeAny,
   O extends z.ZodTypeAny = z.ZodTypeAny,
   M extends Record<string, any> = Record<string, any>,
-> = ((input: z.infer<I>) => Promise<z.infer<O>>) & {
+  S = any,
```
apascal07, line 97: Why mismatch between `Action` and `ActionParams`?

js/core/src/action.ts
```diff
@@ -142,16 +212,20 @@ export function action<
   I extends z.ZodTypeAny,
   O extends z.ZodTypeAny,
   M extends Record<string, any> = Record<string, any>,
```
apascal07, line 214: I never know what this is... but maybe it should go last?

js/core/src/action.ts
pavelgj: trying to avoid passing schema around when not necessary... arguably nothing will save this type (first two type args are zod schemas already)... but generally a good practice.

js/core/src/action.ts
```diff
@@ -89,15 +127,27 @@ type ActionParams<
   outputSchema?: O;
   outputJsonSchema?: JSONSchema7;
   metadata?: M;
-  use?: Middleware<z.infer<I>, z.infer<O>>[];
+  use?: Middleware<z.infer<I>, z.infer<O>, z.infer<S>>[];
+  streamingSchema?: S;
 };
 
+export type SimpleMiddleware<I = any, O = any> = (
```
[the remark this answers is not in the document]
pavelgj: action middleware can optionally transform side channel data, but having to pass it around when you don't need  to is cumbersome. So, there are two ways to define middleware:

```ts
    use: [
      async (input, next) => transformData(await next(input + 'middle1')),
      async (input, opts, next) =>
        transformData(await next(input + 'middle2', transformSideChannel(opts))),
    ],
```

first one is simple variant -- doesn't take opts (side channel) and `next` callback also only takes input (no opts), so side channel data is passed through untouched.
The second one takes both input and opts and `next` takes an optionally transformed side channel.

This is to support middleware that can modify stream data.

js/core/src/action.ts
pavelgj: actually... this needs to be schema... updating...

js/core/src/action.ts
```diff
@@ -106,31 +156,51 @@ export function actionWithMiddleware<
...
   }) as Action<I, O, M>;
   wrapped.__action = action.__action;
-  wrapped.run = async (req: z.infer<I>): Promise<ActionResult<z.infer<O>>> => {
+  wrapped.run = async (
+    req: z.infer<I>,
+    options?: ActionRunOptions<z.infer<S>>
+  ): Promise<ActionResult<z.infer<O>>> => {
     let telemetry;
-    const dispatch = async (index: number, req: z.infer<I>) => {
+    const dispatch = async (
```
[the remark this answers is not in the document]
pavelgj: what do you mean?

js/core/src/action.ts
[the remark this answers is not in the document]
apascal07: Gotcha, makes sense.

apascal07 (approved):

PR #1376, opened 2024-11-22
fix(js/genkit): properly look up prompt variants

MichaelDoyle:
Description here... 

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


js/genkit/src/genkit.ts
pavelgj, line 1: If it's not too much trouble, could please add a unit test for this in js/genkit/tests/prompts_test.ts

pavelgj (approved):

PR #1378, opened 2024-11-22
feat(js/core): add genit version to runtime files

MichaelDoyle:
Description here... 

Checklist (if applicable):
- [x] Tested (*manually*, unit tested, etc.)
- [ ] Changelog updated
- [ ] Docs updated

As seen in Dev UI

image

cc @tonybaroneee

### DIFF
--- a/genkit-tools/common/src/manager/types.ts
+++ b/genkit-tools/common/src/manager/types.ts
@@ -38,6 +38,8 @@ export interface RuntimeInfo {
   timestamp: string;
   /** Display name for the project, typically basename of the root folder */
   projectName?: string;
+  /** Genkit framework version */
+  genkitVersion?: string;
 }
 
 export enum RuntimeEvent {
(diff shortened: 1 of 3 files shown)


tonybaroneee (approved):

genkit-tools/common/src/manager/types.ts
```diff
@@ -38,6 +38,8 @@ export interface RuntimeInfo {
   timestamp: string;
   /** Display name for the project, typically basename of the root folder */
   projectName?: string;
+  /** Genkit framework version */
+  genkitVersion?: string;
```
pavelgj, line 42: questionL should `genkitVersion` include the runtime type, ex: `node/1.2.3` or should be have a separate field for the type?

MichaelDoyle: Very good callout.

Updated the code to pull from the same place as the header, so we can align on how this should be done. No need for me to reinvent the wheel. :)

apascal07 (approved):

pavelgj:
we should do this change for Go at the same time

genkit-tools/common/src/manager/types.ts
pavelgj: AS we discussed, let's add a `reflectionApiSpecVersion` here as well. `genkitVersion` will be only used for FYI purposes, and the "version mismatch callout" should be triggered on `reflectionApiSpecVersion`

@apascal07

MichaelDoyle:
Abandoned. Pavel ended up implementing on a different PR.

PR #1380, opened 2024-11-22
feat: dotprompt dev mode reload behaviour

cabljac:
Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


pavelgj:
We actually still want something like this, but I'm going to close this PR because it's hopelessly outdated.

PR #1400, opened 2024-11-26
fix(go): allow multiple schemas using picoschema

falonso81:
Allow multiple type in picoschema

Checklist (if applicable):
- [ ] Changelog updated
- [ ] Docs updated


pavelgj:
please follow https://www.conventionalcommits.org/en/v1.0.0/ for PRs

go/plugins/dotprompt/picoschema_test.go
```diff
@@ -97,3 +100,54 @@ func replaceEmptySchemas(m map[string]any) any {
 	}
 	return m
 }
+
+func replaceAnyOfWithTypeArray(schema map[string]any) {
```
apascal07, line 104: Can you help me understand what this is for? Is it only needed for the test or everywhere?

alonsopec89: that it is a helper function, to match exactly the test cases with the string expected, without it we have error because the serializate func outputs something like `string("red", "blue")` and we want `AnyOf("red", "blue")`, but the important refactor I think it is about: picoschema.go https://github.com/firebase/genkit/pull/1400/files#diff-d643807a9a75f3e4805fd792bbd253ad99819a506fcfb5bdf9ffa20660d87a4eR109

apascal07: Can you add a sample with an actual dotprompt file that uses these features that weren't supported before? The tests are good but we're doing quite a bit of extra logic here so I want to make sure that it actually works in practice.

alonsopec89: seems like dotpromot have now his own project,

PR #1411, opened 2024-11-26
fix: fixed gemini model handling of nullish optional fields in response schema

pavelgj:
Zod schema like this:

```ts
schema: z.object({
  title: z.string(),
  subtitle: z.string().nulish(),
}),
```

and  picoschema like this:

```yaml
output:
  schema:
    title: string
    subtitle?: string
```

produce schema:
```json
{
  "type": "object",
  "properties": {
    "title": {
      "type": "string"
    },
    "subtitle": {
      "type": [
        "string",
        "null"
      ]
    }
  },
  "required": [
    "title"
  ],
  "additionalProperties": false
}
```

which is not a valid schema in Gemini API request for constrained generation.

```
  status: 400,
  statusText: 'Bad Request',
  errorDetails: [
    {
      '@type': 'type.googleapis.com/google.rpc.BadRequest',
      fieldViolations: [
        {
          field: 'generation_config.response_schema.properties[1].value',
          description: `Invalid JSON payload received. Unknown name "type" at 'generation_config.response_schema.properties[1].value': Proto field is not repeating, cannot start list.`
        }
      ]
    }
  ]
```

convert
```
      "type": [
        "string",
        "null"
      ]
```

to

```
      "type": "string",
```

Checklist (if applicable):

### DIFF
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -459,9 +459,11 @@ export function cleanSchema(schema: JSONSchema): JSONSchema {
     if (typeof out[key] === 'object') {
       out[key] = cleanSchema(out[key]);
     }
+    // Zod nullish() and picoschema optional fields will produce type `["string", "null"]`
+    // which is not supported by the model API. Convert them to just `"string"`.
     if (key === 'type' && Array.isArray(out[key])) {
       // find the first that's not `null`.
-      out[key] = out[key].find(t => t !== 'null');
+      out[key] = out[key].find((t) => t !== 'null');
     }
   }
   return out;
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -449,7 +449,7 @@ export function fromGeminiCandidate(
   };
 }
 
-function cleanSchema(schema: JSONSchema): JSONSchema {
+export function cleanSchema(schema: JSONSchema): JSONSchema {
   const out = structuredClone(schema);
   for (const key in out) {
     if (key === '$schema' || key === 'additionalProperties') {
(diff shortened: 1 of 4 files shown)


mbleigh:
This was intentional and `["string", "null"]` is a valid JSON schema type. I added it as the default behavior because Gemini would frequently add `null` values to optional keys instead of omitting them, causing schema validation to fail when it should have succeeded.

Why is this becoming a problem?

pavelgj:
> This was intentional and `["string", "null"]` [is a valid JSON schema type](https://stackoverflow.com/questions/16241333/specify-a-value-can-be-a-string-or-null-with-json-schema). I added it as the default behavior because Gemini would frequently add `null` values to optional keys instead of omitting them, causing schema validation to fail when it should have succeeded.
> 
> Why is this becoming a problem?

Gemini response schema does not accept this as a valid schema.
```
  status: 400,
  statusText: 'Bad Request',
  errorDetails: [
    {
      '@type': 'type.googleapis.com/google.rpc.BadRequest',
      fieldViolations: [
        {
          field: 'generation_config.response_schema.properties[1].value',
          description: `Invalid JSON payload received. Unknown name "type" at 'generation_config.response_schema.properties[1].value': Proto field is not repeating, cannot start list.`
        }
      ]
    }
  ]
  ```

mbleigh:
> > This was intentional and `["string", "null"]` [is a valid JSON schema type](https://stackoverflow.com/questions/16241333/specify-a-value-can-be-a-string-or-null-with-json-schema). I added it as the default behavior because Gemini would frequently add `null` values to optional keys instead of omitting them, causing schema validation to fail when it should have succeeded.
> > Why is this becoming a problem?
> 
> Gemini response schema does not accept this as a valid schema.
> 
> ```
>   status: 400,
>   statusText: 'Bad Request',
>   errorDetails: [
>     {
>       '@type': 'type.googleapis.com/google.rpc.BadRequest',
>       fieldViolations: [
>         {
>           field: 'generation_config.response_schema.properties[1].value',
>           description: `Invalid JSON payload received. Unknown name "type" at 'generation_config.response_schema.properties[1].value': Proto field is not repeating, cannot start list.`
>         }
>       ]
>     }
>   ]
> ```

Then let's fix it at the Gemini plugin level, not the Picoschema level. In the Gemini plugins, can we recursively look for array `type` options and replace them with the first non-`"null"` value? There's already a `cleanSchema` function because I had to do other cleanup to make Gemini happy.

mbleigh (approved):

PR #1426, opened 2024-11-27
Add reference to genkit object to firebase retriever docs

arbrown:
defineFirestoreRetriever requires a reference to a genkit object.

I didn't add the import / initializiation, since that is covered in a snippet above.


MichaelDoyle:
Docs now live in their own repo -- this typo was long since corrected.

PR #1429, opened 2024-11-27
feat(go): add image support to vertexai plugin

falonso81:
Add support to vertexai plugin
 - [1364](https://github.com/firebase/genkit/issues/1364)

Checklist (if applicable):


go/plugins/vertexai/vertexai.go
```diff
@@ -169,6 +172,9 @@ func defineModel(name string, caps ai.ModelCapabilities) ai.Model {
 		input *ai.ModelRequest,
 		cb func(context.Context, *ai.ModelResponseChunk) error,
 	) (*ai.ModelResponse, error) {
+		if strings.HasPrefix(name, "imagen") {
```
apascal07, line 175: This code block is auto-generated/copied from googleai.go so it's going to get overwritten. Can you figure out a way around it?

apascal07 (commented):
Can you add tests and/or a sample that uses this? I see there are many more parameters in JS like watermark, aspect ratio, etc. How come these were omitted?

go/plugins/vertexai/vertexai.go
alonsopec89: I just put outside of the code copied

MichaelDoyle:
vertex plugin was deprecated in favor of the genai pluginl

PR #1433, opened 2024-11-30
refactor(js/core): refactored flow to be an action

pavelgj:


go/core/action.go
```diff
@@ -264,3 +264,20 @@ func LookupActionFor[In, Out, Stream any](typ atype.ActionType, provider, name s
 	}
 	return a.(*Action[In, Out, Stream])
 }
+
+var actionContextKey = base.NewContextKey[int]()
+
+// WithActionContext returns a new context with action runtime context (side channel data)
+// value set.
+func WithActionContext(ctx context.Context, actionContext map[string]any) context.Context {
```
pavelgj, line 272: I'm thinking this is temporary...  we need to do something like this in Go: https://github.com/firebase/genkit/pull/1375

apascal07 (approved):

PR #1436, opened 2024-12-02
feat(js/plugins/vertexai): add Mistral MaaS integration

cabljac:
This PR will resolve https://github.com/firebase/genkit/issues/1386

TODO:
- [x] switch out the custom HTTPServer and hooks etc for the standard SDK client approach (now that they're fixing it)
- [x] handle streaming responses

Checklist (if applicable):


cabljac:
Probably want to wait for https://github.com/mistralai/client-ts/issues/52 to be resolved (in their next release) so we can support 2411?

js/testapps/mistral-models/package.json
pavelgj, line 1: Could you please consolidate model garden plugin tests apps into a single test app?

pavelgj (approved):

PR #1447, opened 2024-12-03
breaking(go): upgrade ergonomics of prompt definition and calling

dysrama:


go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+
+// Adds the struct for output to the prompt.
+func WithOutputType(output any) PromptOption {
+	return func(p *Prompt) error {
+		r := &jsonschema.Reflector{
+			AllowAdditionalProperties: false,
+			DoNotReference:            true,
+		}
+		p.OutputSchema = r.Reflect(output)
+		return nil
```
apascal07, line 379: Set `OutputFormat` to `JSON` automatically here.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+// Adds default input variable values
+func WithDefaults(variableDefaults map[string]any) PromptOption {
+	return func(p *Prompt) error {
+		p.Config.VariableDefaults = variableDefaults
+		return nil
+	}
+}
+
+// Adds arbitrary metadata.
+func WithMetaData(metaData map[string]any) PromptOption {
```
apascal07, line 400: s/MetaData/Metadata, s/metaData/metadata

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+}
+
+// Adds Generation details for the model to the prompt.
+func WithGenerationConfig(config *ai.GenerationCommonConfig) PromptOption {
+	return func(p *Prompt) error {
+		p.Config.GenerationConfig = config
+		return nil
+	}
+}
+
```
apascal07, line 358: Also add `WithInputSchema` and `WithOutputSchema` that will take in a schema directly.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
 		sortSchemaSlices(s.Items)
 	}
 }
+
+// Adds a variant name to the prompt
+func WithVariant(variant string) PromptOption {
```
apascal07, line 334: This probably doesn't make much sense in code vs. files. I would leave it out for now.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+	return func(p *Prompt) error {
+		p.Config.Variant = variant
+		return nil
+	}
+}
+
+// Adds tools to the prompt
```
apascal07, line 341: I believe proper comment format would be "WithTools adds tools to the prompt." Same for others.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+
+// Adds Generation details for the model to the prompt.
+func WithGenerationConfig(config *ai.GenerationCommonConfig) PromptOption {
+	return func(p *Prompt) error {
+		p.Config.GenerationConfig = config
+		return nil
+	}
+}
+
+// Adds the struct for input to the prompt.
```
apascal07, line 359: It's not necessarily a struct, it could be a primitive too I think. Maybe "WithInputType uses the type provided to derive the input schema." Same for output.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+}
+
+// Adds arbitrary metadata.
+func WithMetaData(metaData map[string]any) PromptOption {
+	return func(p *Prompt) error {
+		p.Config.Metadata = metaData
+		return nil
+	}
+}
+
```
apascal07, line 406: Add `WithModelName` that takes in a name.

go/plugins/dotprompt/genkit.go
```diff
@@ -174,9 +171,16 @@ func (p *Prompt) Register() error {
 // the prompt.
 //
 // This implements the [ai.Prompt] interface.
-func (p *Prompt) Generate(ctx context.Context, pr *PromptRequest, cb func(context.Context, *ai.ModelResponseChunk) error) (*ai.ModelResponse, error) {
+func (p *Prompt) Generate(ctx context.Context, pr *PromptRequest, opts ...PromptOption) (*ai.ModelResponse, error) {
```
apascal07, line 174: We want to remove the `PromptRequest` arg and replace it with options only. Also, we want to have a different set of options for `Generate()` since not everything that makes sense at definition makes sense at generation. For example, the input/output schema is relative to the prompt and what it expects so it makes sense for it to be set at definition only.

At generation, we only want what's in `PromptRequest' + streaming. Streaming option should be removed from `PromptOption`.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+	return func(p *Prompt) error {
+		var toolSlice []ai.Tool
+		toolSlice = append(toolSlice, tools...)
+		p.Tools = toolSlice
+		return nil
+	}
+}
+
+// Adds Generation details for the model to the prompt.
+func WithGenerationConfig(config *ai.GenerationCommonConfig) PromptOption {
```
apascal07, line 352: Maybe just `WithConfig` or `WithCommonConfig`?

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +329,99 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+		if p.Config.ModelName != "" {
+			return errors.New("dotprompt.WithModel: config must specify exactly one of ModelName and Model")
+		}
+		p.Config.Model = model
+		return nil
+	}
+}
+
+// WithStreaming adds a streaming callback to the generate request.
+func WithStreaming(cb ai.ModelStreamingCallback) PromptOption {
```
apascal07, line 419: This probably belongs only in the `Generate()` options.

apascal07 (changes requested):

go/plugins/dotprompt/dotprompt.go
```diff
@@ -114,6 +110,13 @@ type Config struct {
 	Metadata map[string]any
 }
 
+type Options interface {
```
apascal07, line 113: Revert?

apascal07 (changes requested):

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +330,125 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+		p.Config.OutputSchema = r.Reflect(output)
+		p.Config.OutputFormat = ai.OutputFormatJSON
+		return nil
+	}
+}
+
+// WithOutputSchema adds the schema for output to the prompt.
+func WithOutputSchema(schema *jsonschema.Schema) PromptOption {
+	return func(p *Prompt) error {
+		if p.Config.OutputSchema != nil {
```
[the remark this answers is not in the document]
dysrama: There's both a WithOutputType and WithOutputSchema function, you asked me to add WithOutputSchema in the last review. Not sure what to fix here

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +330,125 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+}
+
+// WithInputType uses the type provided to derive the input schema.
+func WithInputType(input any) PromptOption {
+	return func(p *Prompt) error {
+		if p.Config.InputSchema != nil {
+			return errors.New("dotprompt.WithInputType: cannot set both input schema and input type")
+		}
+
+		r := &jsonschema.Reflector{
```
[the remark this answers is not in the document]
dysrama: I'm assuming you mean the InferJSONSchemaNonReferencing call, and not also changing from a json schema to a map?

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +330,125 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+// WithOutputFormat adds the desired output format to the prompt
+func WithOutputFormat(format ai.OutputFormat) PromptOption {
+	return func(p *Prompt) error {
+		p.Config.OutputFormat = format
+		return nil
+	}
+}
+
+// WithDefaults adds default input variable values
+func WithDefaults(variableDefaults map[string]any) PromptOption {
```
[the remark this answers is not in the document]
dysrama: If we combine WithInputType and WithDefaults, then if you choose to set inputSchema, you would have no way to pass in defaults at the prompt level? I'm assuming defaults work right now also when setting a schema directly over a type?

We could rename WithDefaults to WithDefaultVariables, then it would match the WithVariables on generate, and the WithDefaultModel/WithModel and WithDefaultModelName/WithModelName naming we went with yesterday.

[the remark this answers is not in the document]
apascal07: I'm not sure that `WithInputSchema` is all that important to have, for example, we don't offer that option for model generation -- it's input type only (despite the names there saying "with schema", which I think we should rename). JSON schema is kind of an implementation detail of Genkit, Go users will be using Go types.

go/plugins/dotprompt/dotprompt.go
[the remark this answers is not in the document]
apascal07: Ah sorry, yeah we need to check `OutputSchema` for both type and schema -- the old error message was misleading.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +327,134 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+		p.Config.OutputFormat = ai.OutputFormatJSON
+		return nil
+	}
+}
+
+// WithOutputFormat adds the desired output format to the prompt
+func WithOutputFormat(format ai.OutputFormat) PromptOption {
+	return func(p *Prompt) error {
+		if p.Config.OutputSchema == nil {
+			return errors.New("dotprompt.WithOutputFormat: outputSchema needs to be set, to set outputFormat")
```
apascal07, line 410: This isn't quite right. If it's `OutputFormatText`, schema does not need to be set. If it's `OutputFormatJSON`, schema needs to be set. I think we should also be flexible here and if the user has already implicitly set the format by setting the type/schema, we shouldn't error if they also do `WithOutputFormat(ai.OutputFormatJSON) after` since it's valid. Does that make sense?

go/plugins/dotprompt/genkit.go
```diff
@@ -32,17 +32,21 @@ type PromptRequest struct {
...
 	Config *ai.GenerationCommonConfig `json:"config,omitempty"`
 	// Context to pass to model, if any.
 	Context []any `json:"context,omitempty"`
 	// The model to use. This overrides any model specified by the prompt.
-	Model string `json:"model,omitempty"`
+	Model ai.Model `json:"model,omitempty"`
+	// The name of the model to use. This overrides any model specified by the prompt.
+	ModelName string `json:"model,omitempty"`
+	// Streaming callback function
+	Stream ai.ModelStreamingCallback
```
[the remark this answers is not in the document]
apascal07: Still missing it looks like?

go/plugins/dotprompt/genkit.go
```diff
@@ -216,10 +236,76 @@ func (p *Prompt) Generate(ctx context.Context, pr *PromptRequest, cb func(contex
...
+		p.Model = &model
+		return nil
+	}
+}
+
+// WithModelName adds the name of the Model to use. This overrides any model specified by the prompt.
+func WithModelName(model string) GenerateOption {
+	return func(p *PromptRequest) error {
+		if p.ModelName != "" || p.Model != nil {
+			return errors.New("dotprompt.WithModelName: config must specify exactly once, either ModelName or Model")
```
apascal07, line 295: s/config/model? Here and above. "model must be specified exactly once, either ModelName or Model"

go/plugins/dotprompt/dotprompt.go
dysrama: OK, I get it. Will make some changes

go/plugins/dotprompt/dotprompt.go
```diff
@@ -85,24 +86,20 @@ type Config struct {
...
-
 	// Details for the model.
 	GenerationConfig *ai.GenerationCommonConfig
 
 	// Schema for input variables.
 	InputSchema *jsonschema.Schema
 
 	// Default input variable values
-	VariableDefaults map[string]any
+	InputDefault map[string]any
```
apascal07, line 102: Nit: `DefaultInput`

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +327,117 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+
+		return nil
+	}
+}
+
+// WithOutputFormat adds the desired output format to the prompt
+func WithOutputFormat(format ai.OutputFormat) PromptOption {
+	return func(p *Prompt) error {
+		if p.Config.OutputFormat != "" && p.Config.OutputFormat != format {
+			return errors.New("dotprompt.WithOutputFormat: outputFormat does not match set outputSchema")
```
apascal07, line 401: Nit: s/outputFormat/OutputFormat + the other ones to match the actual capitalization of the field in case someone goes looking for it.

go/plugins/dotprompt/genkit.go
```diff
@@ -216,10 +236,106 @@ func (p *Prompt) Generate(ctx context.Context, pr *PromptRequest, cb func(contex
...
+		return "", err
+	}
+
+	return res.Text(), nil
+}
+
+// GenerateData runs generate request for this prompt. Returns ModelResponse struct.
+// TODO: Stream GenerateData with partial JSON
+func (p *Prompt) GenerateData(ctx context.Context, value any, opts ...GenerateOption) (*ai.ModelResponse, error) {
+	with := WithOutputType(value)
```
apascal07, line 260: Nit: Can just append `WithOutputType` to opts and let it be handled later so this isn't done in multiple places? Would be nice to throw a custom error if doing `GenerateData` + manually passing `WithOutputType` (user might be confused because they didn't set output type twice) but I don't think there's a succinct way to do that...

go/plugins/dotprompt/dotprompt.go
```diff
@@ -321,3 +327,117 @@ func sortSchemaSlices(s *jsonschema.Schema) {
...
+		p.Config.InputDefault = inputDefault
+		return nil
+	}
+}
+
+// WithOutputType uses the type provided to derive the output schema.
+func WithOutputType(output any) PromptOption {
+	return func(p *Prompt) error {
+		if p.Config.OutputSchema != nil {
+			return errors.New("dotprompt.WithOutputType: cannot set outputType more than once. Set either outputType or outputSchema")
```
apascal07, line 387: Stale error message

go/samples/menu/s02.go
```diff
@@ -51,12 +51,10 @@ func setup02(_ context.Context, m ai.Model) error {
...
 		 {{question}} ?`,
-		dotprompt.Config{
-			Model:        m,
-			InputSchema:  menuQuestionInputSchema,
-			OutputFormat: ai.OutputFormatText,
-			Tools:        []ai.Tool{menuTool},
-		},
+		dotprompt.WithDefaultModel(m),
+		dotprompt.WithInputType(menuQuestionInput{}),
+		dotprompt.WithOutputFormat(ai.OutputFormatText),
```
apascal07, line 56: Let's omit this from examples and automatically set output format to text if unset.

go/samples/coffee-shop/main.go
```diff
@@ -159,12 +146,10 @@ func main() {
...
 	simpleStructuredGreetingPrompt, err := dotprompt.Define("simpleStructuredGreeting", simpleStructuredGreetingPromptTemplate,
-		dotprompt.Config{
-			Model:        g,
-			InputSchema:  r.Reflect(simpleGreetingInput{}),
-			OutputFormat: ai.OutputFormatJSON,
-			OutputSchema: r.Reflect(simpleGreetingOutput{}),
-		},
+		dotprompt.WithDefaultModel(g),
+		dotprompt.WithInputType(simpleGreetingInput{}),
+		dotprompt.WithOutputFormat(ai.OutputFormatJSON),
```
apascal07, line 151: Let's omit the setting of output format here since we set type. We want to show the simplest most canonical ways of doing things.

go/plugins/dotprompt/dotprompt.go
```diff
@@ -85,24 +86,20 @@ type Config struct {
 	ModelName string
 
 	// The Model to use.
-	// If this is non-nil, Model should be the empty string.
-	Model ai.Model
+	// If this is non-nil, ModelName should be the empty string.
+	Model *ai.Model
```
apascal07, line 91: I'm sorry, I misled you here. I didn't realize that ai.Model is an interface so we actually don't want it to be a pointer. Interfaces can be nil.

go/plugins/dotprompt/dotprompt_test.go
```diff
@@ -133,6 +151,204 @@ func TestPrompts(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if dotPrompt.Config.InputSchema == nil {
+			t.Error("inputschema not set")
+		}
+		if dotPrompt.Config.InputDefault == nil {
+			t.Error("Input default not set")
+		}
+		if dotPrompt.Config.InputDefault["text"] != "test" {
```
apascal07: Does this pass? Shouldn't it be "test"?

apascal07 (approved):

apascal07:
Can you add code snippets to the PR description? Can copy most of what I put in the issue, it's mostly updated -- just so people can see at a glance what changed.

PR #1470, opened 2024-12-09
feat(go/plugins/firebase): add retriever, init methods

cabljac:
This PR adds retrievers to the Go firebase plugin.

Should resolve https://github.com/firebase/genkit/issues/711

- Consolidated state management into a single state struct, similar to the Google AI plugin
- Added thread-safety with mutex
- Added support for retrievers in the state management

Checklist (if applicable):


go/plugins/firebase/firebase.go
```diff
@@ -16,31 +16,90 @@ package firebase
...
+	defer state.mu.Unlock()
 
-// app returns a cached Firebase app.
-func App(ctx context.Context) (FirebaseApp, error) {
-	mutex.Lock()
-	defer mutex.Unlock()
-	if app == nil {
-		newApp, err := firebase.NewApp(ctx, nil)
+	if state.initted {
+		log.Println("firebase.Init: plugin already initialized, returning without reinitializing")
```
apascal07, line 52: We discussed this before, but I don't remember the outcome of the discussion -- are we sure we want to no-op when it's possible there is a different config being passed in?

go/plugins/firebase/firebase.go
```diff
@@ -16,31 +16,90 @@ package firebase
...
+	defer state.mu.Unlock()
+
+	if !state.initted {
+		return nil, fmt.Errorf("firebase.App: Firebase app not initialized. Call Init first")
+	}
+	return state.app, nil
+}
+
+// Retrievers returns the cached list of retrievers.
+func Retrievers(ctx context.Context) ([]ai.Retriever, error) {
```
apascal07, line 97: Don't need `ctx` here or above?

apascal07 (approved):

go/plugins/firebase/firebase.go
cabljac: > I wonder if this is the right behavior or if it should error on subsequent calls. Generally this no-op behavior is only deterministic if you are not passing a configuration that you are expecting to take effect. Here, a no-op means you don't know if the initted config matches what you're expecting.

yup you're right, sorry it got lost across versions of this PR, will look to change

PR #1477, opened 2024-12-10
fix(js/core): correctly handle errors when streaming

pavelgj:
Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/manager/manager.ts
+++ b/genkit-tools/common/src/manager/manager.ts
@@ -25,6 +25,7 @@ import {
   RunActionResponseSchema,
 } from '../types/action';
 import * as apis from '../types/apis';
+import { GenkitErrorData } from '../types/error';
 import { TraceData } from '../types/trace';
 import { logger } from '../utils/logger';
 import {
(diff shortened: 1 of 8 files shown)


genkit-tools/common/src/manager/manager.ts
```diff
@@ -198,9 +198,25 @@ export class RuntimeManager {
...
       stream.on('end', () => {
-        const actionResponse = RunActionResponseSchema.parse(
-          JSON.parse(buffer)
-        );
+        const parsedBuffer = JSON.parse(buffer);
+        if (parsedBuffer.error) {
+          const err = new GenkitToolsError(
+            `Error running action key='${input.key}'.`
+          );
+          // massage the error into a shape dev ui expects
```
apascal07, line 206: Should we/do we have a common type for this? Feels like we should given we share types between CLI/Dev UI.

js/core/src/reflection.ts
```diff
@@ -156,31 +156,50 @@ export class ReflectionServer {
...
+              message,
+              details: {
+                stack,
               },
-            } as RunActionResponse)
-          );
-          response.end();
+            };
+            if ((err as any).traceId) {
+              errorResponse.details.traceId = (err as any).traceId;
```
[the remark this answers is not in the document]
pavelgj: js/core/src/action.ts line 284

genkit-tools/common/src/manager/manager.ts
pavelgj: Done.

apascal07 (approved):

PR #1492, opened 2024-12-11
chore: updated README logo

pavelgj:


peterfriese (changes requested):
Hi @pavelgj - is the gray background for the logo intentional? I think the previous version (with light and dark mode assets) looked perfectly fine both in light and dark mode. Plus, it also uses the official logo from the Firebase brand toolkit (https://firebase.google.com/brand-guidelines)

pavelgj:
> Hi @pavelgj - is the gray background for the logo intentional? I think the previous version (with light and dark mode assets) looked perfectly fine both in light and dark mode. Plus, it also uses the official logo from the Firebase brand toolkit (https://firebase.google.com/brand-guidelines)

yeah, this readme needs to be renderable in places other than GH (ex. we're working on API reference, and it uses that readme as the landing page). Current logo is using github specific markdown render features to swap out images depending on the GH theme which doesn't work in other places.

peterfriese:
> > Hi @pavelgj - is the gray background for the logo intentional? I think the previous version (with light and dark mode assets) looked perfectly fine both in light and dark mode. Plus, it also uses the official logo from the Firebase brand toolkit (https://firebase.google.com/brand-guidelines)
> 
> yeah, this readme needs to be renderable in places other than GH (ex. we're working on API reference, and it uses that readme as the landing page). Current logo is using github specific markdown render features to swap out images depending on the GH theme which doesn't work in other places.

Understood. The brand guidelines explicitly say "do not alter the logo", and using a different background color falls in this category.

Can we either

1. Use a different file as the landing page for the ref docs
2. Cherry pick the GitHub CSS and use the relevant parts in the ref docs?

PR #1529, opened 2024-12-16
feat(js/plugins): added custom dimensions parameter to text-embedding

river0g:
How to use
```typescript
ai.embed({
  embedder: textEmbedding005,
  content: 'String to be embedded',
  options: {
    outputDimensionality: 500 // Between 1 and 768, inclusive
  }
})
```

Checklist (if applicable):

### DIFF
--- a/js/plugins/googleai/src/embedder.ts
+++ b/js/plugins/googleai/src/embedder.ts
@@ -37,13 +37,13 @@ export const GeminiEmbeddingConfigSchema = z.object({
   title: z.string().optional(),
   version: z.string().optional(),
   /**
-   * The `dimensions` parameter allows you to specify the dimensionality of the embedding output.
+   * The `outputDimensionality` parameter allows you to specify the dimensionality of the embedding output.
    * By default, the model generates embeddings with 768 dimensions. Models such as
    * `text-embedding-004`, `text-embedding-005`, and `text-multilingual-embedding-002`
    * allow the output dimensionality to be adjusted between 1 and 768.
    * By selecting a smaller output dimensionality, users can save memory and storage space, leading to more efficient computations.
    **/
-  dimensions: z.number().min(1).max(768).optional(),
+  outputDimensionality: z.number().min(1).max(768).optional(),
 });
 
 export type GeminiEmbeddingConfig = z.infer<typeof GeminiEmbeddingConfigSchema>;
@@ -76,8 +76,6 @@ export const SUPPORTED_MODELS = {
   'embedding-001': textEmbeddingGecko001,
   'text-embedding-004': textEmbedding004,
 };
-
-const customDimensionsSupportedModelNames = [textEmbedding004.name];
 
 export function defineGoogleAIEmbedder(
   ai: Genkit,
@@ -116,11 +114,6 @@ export function defineGoogleAIEmbedder(
       info: embedder.info!,
     },
     async (input, options) => {
-      const dimensions =
-        customDimensionsSupportedModelNames.includes(embedder.name) &&
-        options?.dimensions
-          ? options.dimensions
-          : undefined;
       const client = new GoogleGenerativeAI(apiKey!).getGenerativeModel({
         model:
           options?.version ||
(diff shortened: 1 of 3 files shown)


js/plugins/googleai/src/embedder.ts
```diff
@@ -36,6 +36,14 @@ export const GeminiEmbeddingConfigSchema = z.object({
...
   title: z.string().optional(),
   version: z.string().optional(),
+  /**
+   * The `dimensions` parameter allows you to specify the dimensionality of the embedding output.
+   * By default, the model generates embeddings with 768 dimensions. Models such as
+   * `text-embedding-004`, `text-embedding-005`, and `text-multilingual-embedding-002`
+   * allow the output dimensionality to be adjusted between 1 and 768.
+   * By selecting a smaller output dimensionality, users can save memory and storage space, leading to more efficient computations.
+   **/
+  dimensions: z.number().min(1).max(768).optional(),
```
pavelgj, line 46: the underlying option is called `outputDimensionality`... I like `dimensions` better, but I still think we should stick as close to to API option naming as possible. It's easier to cross-reference these options with API documentation...

js/plugins/googleai/src/embedder.ts
```diff
@@ -69,6 +77,8 @@ export const SUPPORTED_MODELS = {
   'text-embedding-004': textEmbedding004,
 };
 
+const customDimensionsSupportedModelNames = [textEmbedding004.name];
```
pavelgj, line 80: I don't think this is necessary. I don't think we want to maintain this mapping. Let the API return errors if the option is not supported.

js/plugins/vertexai/src/embedder.ts
```diff
@@ -39,6 +39,14 @@ export const VertexEmbeddingConfigSchema = z.object({
...
   location: z.string().optional(),
   version: z.string().optional(),
+  /**
+   * The `dimensions` parameter allows you to specify the dimensionality of the embedding output.
+   * By default, the model generates embeddings with 768 dimensions. Models such as
+   * `text-embedding-004`, `text-embedding-005`, and `text-multilingual-embedding-002`
+   * allow the output dimensionality to be adjusted between 1 and 768.
+   * By selecting a smaller output dimensionality, users can save memory and storage space, leading to more efficient computations.
+   **/
+  dimensions: z.number().min(1).max(768).optional(),
```
pavelgj, line 49: same thing on the naming -- let's call it `outputDimensionality` for consistency

js/plugins/vertexai/src/embedder.ts
```diff
@@ -83,6 +91,12 @@ export const SUPPORTED_EMBEDDER_MODELS: Record<string, EmbedderReference> = {
   // ]),
 };
 
+const customDimensionsSupportedModelNames = [
```
pavelgj, line 94: same thing here, let's drop this mapping.

js/plugins/googleai/src/embedder.ts
river0g: Indeed, based on your comment, I also feel that an API error would be the more natural approach.
I thought it was important to ensure that the issue is not hidden when the option is applied to an embedding model that does not support it.
Thank you for your comment.

js/plugins/vertexai/src/embedder.ts
```diff
@@ -145,7 +153,8 @@ export function defineVertexAIEmbedder(
             task_type: options?.taskType,
             title: options?.title,
           };
-        })
+        }),
+        { outputDimensionality: options?.outputDimensionality }
```
pavelgj, line 157: why is not along with other options (lines 152-154)?

river0g: Thank you for your question.
Based on the API interface, it seems that `outputDimensionality` cannot be included under `instances`, and must be placed under `parameters`.
(documentation here: [Get text embeddings](https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings#get_text_embeddings_for_a_snippet_of_text))

After verifying the actual behavior, I found that when `outputDimensionality` is placed under `instances` (lines 152-154), it is ignored, and the default dimensionality is returned.

pavelgj: Perfect! Just making sure.

pavelgj (approved):

PR #1545, opened 2024-12-19
feat: Write OTel GenAI semantic convention client-side metrics for token usage

bryanatkinson:
Adds support for writing the `gen_ai.client.token.usage` metric as described by the OpenTelemetry GenAI semantic conventions: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage

I added this directly to the framework instead of in the google-cloud plugin since they seem generally useful. So far I have only added support to the Gemini and Imagen models to see if this approach looks good. If it does, I will carry it through to the other models in the vertex model garden and add the `gen_ai.client.operation.duration` metric.

Checklist (if applicable):

### DIFF
--- a/js/ai/src/metrics.ts
+++ b/js/ai/src/metrics.ts
@@ -74,13 +74,16 @@ const tokenUsage = new MetricHistogram('gen_ai.client.token.usage', {
   unit: 'token',
 });
 
-const operationDuration = new MetricHistogram('gen_ai.client.operation.duration', {
-  description: 'Time taken for GenAI operations',
-  valueType: ValueType.DOUBLE,
-  unit: 'token',
-});
+const operationDuration = new MetricHistogram(
+  'gen_ai.client.operation.duration',
+  {
+    description: 'Time taken for GenAI operations',
+    valueType: ValueType.DOUBLE,
+    unit: 'token',
+  }
+);
 
-export function writeMetrics(resp: GenerateResponse, durationMs: double): void {
+export function writeMetrics(resp: GenerateResponse, durationMs: number): void {
   const commonDimensions = {
     'gen_ai.client.framework': 'genkit',
     'gen_ai.operation.name': resp.clientTelemetry?.operationName,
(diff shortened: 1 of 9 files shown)


js/plugins/google-cloud/tests/metrics_test.ts
```diff
@@ -422,7 +422,7 @@ describe('GoogleCloudMetrics', () => {
     );
   });
 
-  it('writes feature label to generate and action metrics when running inside an action', async () => {
+  it.only('writes feature label to generate and action metrics when running inside an action', async () => {
```
kmandrika, line 425: I think this needs to be undone

js/ai/src/metrics.ts
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Wrapper for OpenTelemetry metrics.
+ *
+ * The OpenTelemetry {MeterProvider} can only be accessed through the metrics
+ * API after the NodeSDK library has been initialized. To prevent race
+ * conditions we defer the instantiation of the metric to when it is first
+ * ticked.
+ */
+class Metric<T> {
```
kmandrika, line 31: Not a huge fan of having this be in core - are we positive that this can't live in the plugin?

js/plugins/google-cloud/tests/metrics_test.ts
bryanatkinson: oops. Removed.

js/ai/src/metrics.ts
bryanatkinson: There's no reason it couldn't be done in the plugin, but these semantic conventions are intended so that different tools can understand the telemetry data, so only exporting them when you're using GCP somewhat defeats the purpose.

bryanatkinson:
I've also got a follow-up that includes writing the span attributes and I think cleans things up a little bit: https://github.com/firebase/genkit/compare/bryanatkinson-semconv-metrics...bryanatkinson-semconv-spans?expand=1

I can switch this PR to using that if you're OK with the general approach.

PR #1568, opened 2024-12-21
chore(js/plugins/vertexai): remove unused variable req

river0g:
The req object is not being used.
Since `const response = await predictClient([instance], toParameters(request));` utilizes the contents of the req object, I believe the req object is not necessary at this point.

Checklist (if applicable):


MichaelDoyle:
the vertexai plugin was deprecated in favor of the google genai plugin.

PR #1569, opened 2024-12-21
refactor(js/flows): made flow be actually actions, actions can stream, moved flow server to express plugin

pavelgj:
Removed all express stuff from flows core. Flows are just simple actions now. Actions have the flow interface (including streaming).

Removed auth policy from flows. The policy is now a plugin feature. `startFlowServer` also moved to the express plugin.

```ts
import { startFlowServer, withAuth } from '@genkit-ai/express';

startFlowServer({
  flows: [
    simpleFlow,
    withAuth(flowThatNeedsAuth, authProvider, ({ auth, action, input, request }) => {
      if (!auth) {
        throw new Error('Authorization required.');
      }
    })
  ],
});
```

`run` is now `ai.run`.

Checklist (if applicable):

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -250,11 +250,12 @@ alongside the native flows. You have two options:
 1.  Use whatever server framework you like, and pass the auth context through via
     the flow call as noted above.
 
-1.  Use the built-in `startFlowsServer()` and provide Express middleware in the
-    flow config:
+1.  Use  `startFlowsServer()` available via `@genkit-ai/express` plugin and provide
+    Express auth middleware in the flow server config:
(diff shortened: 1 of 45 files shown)


chrisraygill:
Remind me, what's the use case for startFlowServer since we introduced the Express handler?

pavelgj:
> Remind me, what's the use case for startFlowServer since we introduced the Express handler?

It much less boilerplate for simple cases. And we have existing users.... I don't want to break them. So I'm moving it rather than removing.

MichaelDoyle (approved):

PR #1570, opened 2024-12-22
docs: update README.md about samples

river0g:
I updated the sample repo section in the README.md.
Reference: https://github.com/firebase/genkit/pull/1540

Since the same content is included in the README.md in the root directory, I felt the sample/README.md was unnecessary and decided to remove it.

Checklist (if applicable):


MichaelDoyle:
We've since changed the samples around quite a bit (and added content to the README file deleted here)

PR #1575, opened 2024-12-26
feature(go): add support for multiple model versions

hugoaguirre:
Adding the support for multiple model versions and grouping the versions in the model metadata.

### Gemini models being rendered in Dev UI frontpage
image

### Model versions being grouped in the Dev UI
image

### Playing around with `gemini-1.5-flash-8b-latest`
image

### Checking raw request
image

Checklist (if applicable):


hugoaguirre:
Hi @pavelgj, @apascal07 

Let me know your opinion about this approach.
I wanted to keep the model definition as it previously was because I think it was very simple and verbose.

The only changes added were:
- Rewrote how the known models are defined in the plugins (adding a new struct named: `ai.ModelDetails` that encapsulates the model capabilities and its versions)
- `ai.ModelMetadata` was updated to add model versions so they can be grouped together

go/plugins/googleai/googleai.go
```diff
@@ -47,10 +47,26 @@ var state struct {
...
+
+		"gemini-1.5-pro": {
+			Versions: []string{"gemini-1.5-pro-latest", "gemini-1.5-pro-001", "gemini-1.5-pro-002"},
+			Caps:     gemini.Multimodal,
+		},
+
+		"gemini-1.5-flash-8b": {
+			Versions: []string{"gemini-1.5-flash-8b-latest", "gemini-1.5-flash-8b-001"},
+			Caps:     gemini.Multimodal,
+		},
```
alonsopec89, line 69: this hardcoded strings of the version management can become a bit complicated to maintain, and since we are using the same ones in googleai and vertexai, I think it would be good to have in some constants file, maybe in the genkit package or another new one, so that it is used here and in vertexai

hugoaguirre: Good observation @alonsopec89. 

I was looking at how things were implemented in Genkit JS and plugins are in charge of defining the models they use regardless if more plugins are using the same models. e.g. [js vertex](https://github.com/firebase/genkit/blob/main/js/plugins/vertexai/src/gemini.ts#L218) and [js googleai](https://github.com/firebase/genkit/blob/main/js/plugins/googleai/src/gemini.ts#L112).

About placing these model definitions in `genkit` package... my understanding is that Genkit should be model-agnostic so I believe that's the reason behind why plugins are defining their models on their packages.

hugoaguirre:
> How is this to be used in code? I don't see any way to reference the exact version I want to use in a generate call for example.
Feedback addressed in the last commit

apascal07:
Overall, looks good. There's an outstanding issue that you are indirectly (but not fully) touching upon here: https://github.com/firebase/genkit/issues/1744

It's unrelated to this PR but it might make sense to just combine it, or at least do it just for the `versions` field and save the rest for another PR.

go/ai/generate.go
```diff
@@ -62,18 +54,24 @@ func DefineModel(
...
 	}
 	if metadata.Label != "" {
 		metadataMap["label"] = metadata.Label
 	}
 	supports := map[string]bool{
-		"media":      metadata.Supports.Media,
-		"multiturn":  metadata.Supports.Multiturn,
-		"systemRole": metadata.Supports.SystemRole,
-		"tools":      metadata.Supports.Tools,
+		"media":      metadata.Info.Supports.Media,
```
apascal07, line 68: Unmarshal `ModelInfoSupports` into the map programatically so we have full coverage of fields going forward.

go/ai/generate.go
```diff
@@ -256,6 +267,39 @@ func Generate(ctx context.Context, r *registry.Registry, opts ...GenerateOption)
...
+		return false, errors.New("wrong model name")
+	}
+
+	a := core.LookupActionFor[*ModelRequest, *ModelResponse, *ModelResponseChunk](r, atype.Model, parts[0], parts[1])
+	if a == nil {
+		return false, errors.New("model action not defined")
+	}
+
+	if !modelVersionSupported(v, a.Desc().Metadata) {
+		return false, fmt.Errorf("version not supported: %s", v)
```
apascal07, line 284: Add a test that covers this.

go/ai/generate.go
```diff
@@ -256,6 +267,39 @@ func Generate(ctx context.Context, r *registry.Registry, opts ...GenerateOption)
...
 
+// validateModelVersion checks in the registry the action of the
+// given model version and determines whether its supported or not.
+func validateModelVersion(r *registry.Registry, v string, req *generateParams) (bool, error) {
+	parts := strings.Split(req.Model.Name(), "/")
+	if len(parts) != 2 {
+		return false, errors.New("wrong model name")
+	}
+
+	a := core.LookupActionFor[*ModelRequest, *ModelResponse, *ModelResponseChunk](r, atype.Model, parts[0], parts[1])
```
apascal07, line 278: Use `ai.LookupModel` here.

go/ai/generate.go
```diff
@@ -256,6 +267,39 @@ func Generate(ctx context.Context, r *registry.Registry, opts ...GenerateOption)
...
+	if a == nil {
+		return false, errors.New("model action not defined")
```
apascal07, line 280: `"model %s not found", model`

apascal07 (changes requested):

go/ai/generate.go
```diff
@@ -35,18 +35,11 @@ type modelAction = core.Action[*ModelRequest, *ModelResponse, *ModelResponseChun
...
-	Media      bool // the model supports media as well as text input
-	Tools      bool // the model supports tools
-	SystemRole bool // the model supports a system prompt or role
-}
-
 // ModelMetadata is the metadata of the model, specifying things like nice user-visible label, capabilities, etc.
 type ModelMetadata struct {
 	Label    string
-	Supports ModelCapabilities
+	Versions []string
```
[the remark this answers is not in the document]
hugoaguirre: Fair enough, I thought we wanted to keep `ModelMetadata` as well. 
Changing it.

go/ai/generate.go
hugoaguirre: For the record: After internal conversation with Alex, it was decided not to follow this suggestion since these changes will be breaking current functionality: `DefineModel` would have to check for errors during Marshal/Unmarshal process and return value must change to `(Model, error)` which will break the current state of the API.

apascal07 (approved):

PR #1583, opened 2025-01-04
feat(js/ai): added a simple way to interrupt tool execution

pavelgj:
```ts
ai.defineTool(..., (input, {interrupt}) => {
  interrupt();
});
```

### DIFF
--- a/js/ai/src/generate/action.ts
+++ b/js/ai/src/generate/action.ts
@@ -329,7 +329,8 @@ async function generate(
         previousChunks: accumulatedChunks,
         parser: resolvedFormat?.handler(request.output?.schema).parseChunk,
       }
-    ));
+    )
+  );
   if (interruptedParts.length > 0) {
     return {
       ...response.toJSON(),
--- a/js/ai/src/generate/action.ts
+++ b/js/ai/src/generate/action.ts
@@ -244,6 +244,7 @@ async function generate(
   let newTools = options.rawRequest.tools;
   let newToolChoice = options.rawRequest.toolChoice;
   let interruptedParts: Part[] = [];
+  let pendingToolRequests: Part[] = [];
   for (const part of toolCalls) {
     if (!part.toolRequest) {
       throw Error(
@@ -282,11 +283,22 @@ async function generate(
       }
     } else {
       try {
+        const toolOutput = await tool(part.toolRequest?.input);
+        const requestPart = { ...part } as Part;
+        requestPart.metadata = {
+          ...part.metadata,
+          pendingToolResponse: {
+            name: part.toolRequest.name,
+            ref: part.toolRequest.ref,
+            output: toolOutput,
+          },
+        };
+        pendingToolRequests.push(requestPart);
         toolResponses.push({
           toolResponse: {
             name: part.toolRequest.name,
             ref: part.toolRequest.ref,
-            output: await tool(part.toolRequest?.input),
+            output: toolOutput,
           },
         });
       } catch (e) {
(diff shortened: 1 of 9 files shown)


js/ai/src/tool.ts
```diff
@@ -187,3 +187,16 @@ export function defineTool<I extends z.ZodTypeAny, O extends z.ZodTypeAny>(
...
+ */
+export class ToolInterruptError extends Error {}
+
+/**
+ * Interrupts current tool execution causing tool request to be returned in the generation response.
+ * Should only be called within a tool.
+ */
+export function interruptTool() {
+  throw new ToolInterruptError();
+}
```
mbleigh, line 202: I'd like this to be able to take an argument of `Record<string, any>` and whatever gets passed to that gets annotated onto the toolRequest it's associated with.

```ts
const transferMoney = ai.defineTool({
  inputSchema: z.object({amount: z.number(), fromAccount: z.string(), toAccount: z.string()}),
  outputSchema: z.object({balance: z.number()}),
}, ({amount, fromAccount, toAccount}, {interrupt, context}) {
  if (amount > 1000) interrupt({confirm: "The requested amount is large, please confirm all the details carefully."});
  if (!getUserAccounts(context.auth.uid).includes(toAccount)) interrupt({confirm: `The account '${toAccount}' does not belong to you. Please confirm you wish to transfer money to an account owned by someone else.`;
  return doMoneyTransfer(fromAccount,toAccount,amount);
});
```

This will allow the developer to customize and control how interrupts happen and also lay foundation for higher-level interrupt primitives down the road.

js/ai/src/tool.ts
```diff
@@ -187,3 +187,16 @@ export function defineTool<I extends z.ZodTypeAny, O extends z.ZodTypeAny>(
...
+/**
+ * Thrown when tools execution is interrupted. It's meant to be caugh by the framework, not public API.
+ */
+export class ToolInterruptError extends Error {}
+
+/**
+ * Interrupts current tool execution causing tool request to be returned in the generation response.
+ * Should only be called within a tool.
+ */
+export function interruptTool() {
```
mbleigh, line 200: What do you think about adding `interrupt` as an item in the second argument of the tool instead?

```ts
ai.defineTool(..., (input, {context, interrupt}) => {
  interrupt();
});
```

js/ai/src/tool.ts
pavelgj: this implies this interrupt is resumable/rerunnable where the interrupt won't be triggered? Like we we had with durable flows? How do you imagine the user facing api for this?

mbleigh:
It's up to the dev to figure it what to do with the metadata, for now it's
just an annotation and the dev can use it to drive UI etc.

I want there to be a mechanism/API to explicitly signal (maybe via
context?) that it's a resumed interruption, but I think we can figure that
out later.

js/ai/src/tool.ts
pavelgj: done

js/ai/src/generate/action.ts
```diff
@@ -249,29 +255,53 @@ async function generate(
...
+            ref: part.toolRequest.ref,
+            output: await tool(part.toolRequest?.input),
+          },
+        });
+      } catch (e) {
+        if (e instanceof ToolInterruptError) {
+          logger.debug(`interrupted tool ${part.toolRequest?.name}`);
+          if (e.metadata) {
+            part.metadata = { ...part.metadata, interrupt: e.metadata };
+          }
```
mbleigh, line 299:
```suggestion
          part.metadata = { ...part.metadata, interrupt: e.metadata || true };
```

js/ai/src/generate/action.ts
```diff
@@ -287,6 +317,16 @@ async function generate(
...
+  if (interruptedParts.length > 0) {
+    return {
+      ...response.toJSON(),
+      finishReason: 'interrupted',
+      message: {
+        role: 'model',
+        content: interruptedParts,
+      },
+    };
+  }
```
mbleigh, line 329: Will this work properly if the model had already generated some text before it made the tool request? It looks like it won't, but maybe I'm not understanding the codepath. The existing parts should be examined and the request swapped instead of only returning the interrupt I think.

mbleigh (approved):
I'm actually pretty happy with this API sitting with it overnight, and I think we could do `defineInterrupt` as a thin sugaring on top of it.

mbleigh (approved):

js/ai/src/generate/action.ts
pavelgj: done

js/ai/src/generate/action.ts
pavelgj: done

PR #1590, opened 2025-01-07
[chore]: Standardize package exports in a way that silences pnpm warnings

### LINKED ISSUE #1589: [JS] pnpm hints that package exports are in the incorrect order
**Describe the bug**
pnpm build warns that 

**To Reproduce**
Run `pnpm build`

**Expected behavior**
Clean build

**Screenshots**
Build produces warnings such as:

```
plugins/pinecone build$ npm-run-all build:clean check compile
[8 lines collapsed]
[1 lines collapsed]
│          ╵       ~~~~~~~
│   The "import" condition comes earlier and will be used for all "import" statements:
│     package.json:51:6:
│       51 │       "import": "./lib/index.mjs",
│          ╵       ~~~~~~~~
│   The "require" condition comes earlier and will be used for all "require" calls:
│     package.json:50:6:
│       50 │       "require": "./lib/index.js",
│          ╵       ~~~~~~~~~
│   The "import" condition comes earlier and will be used for all "import" statements:
```

 - OS: MacOS
 - Version 14.7.2

** Node version
 -  20.11.1

inlined:
Fixes #1589 to clean up warnings about our package ordering (which is inconsistent within the repository). Thank you ChatGPT for the suggested ordering, which silences pnpm. I worry but have not verified that the old way might confuse node or typescript on import style.

Checklist (if applicable):
- [x] Tested (builds without warnings)
- [x] Docs updated (N/A)

### DIFF
--- a/js/ai/package.json
+++ b/js/ai/package.json
@@ -46,80 +46,80 @@
   "exports": {
     ".": {
       "types": "./lib/index.d.ts",
-      "require": "./lib/index.js",
       "import": "./lib/index.mjs",
+      "require": "./lib/index.js",
       "default": "./lib/index.js"
     },
     "./retriever": {
       "types": "./lib/retriever.d.ts",
-      "require": "./lib/retriever.js",
       "import": "./lib/retriever.mjs",
+      "require": "./lib/retriever.js",
       "default": "./lib/retriever.js"
     },
     "./embedder": {
       "types": "./lib/embedder.d.ts",
-      "require": "./lib/embedder.js",
       "import": "./lib/embedder.mjs",
+      "require": "./lib/embedder.js",
       "default": "./lib/embedder.js"
     },
     "./evaluator": {
       "types": "./lib/evaluator.d.ts",
-      "require": "./lib/evaluator.js",
       "import": "./lib/evaluator.mjs",
+      "require": "./lib/evaluator.js",
       "default": "./lib/evaluator.js"
     },
     "./model": {
       "types": "./lib/model.d.ts",
-      "require": "./lib/model.js",
       "import": "./lib/model.mjs",
+      "require": "./lib/model.js",
       "default": "./lib/model.js"
     },
     "./model/middleware": {
       "types": "./lib/model/middleware.d.ts",
-      "require": "./lib/model/middleware.js",
       "import": "./lib/model/middleware.mjs",
+      "require": "./lib/model/middleware.js",
       "default": "./lib/model/middleware.js"
     },
     "./extract": {
       "types": "./lib/extract.d.ts",
-      "require": "./lib/extract.js",
       "import": "./lib/extract.mjs",
+      "require": "./lib/extract.js",
       "default": "./lib/extract.js"
     },
     "./tool": {
       "types": "./lib/tool.d.ts",
-      "require": "./lib/tool.js",
       "import": "./lib/tool.mjs",
+      "require": "./lib/tool.js",
       "default": "./lib/tool.js"
     },
     "./testing": {
       "types": "./lib/testing/index.d.ts",
-      "require": "./lib/testing/index.js",
       "import": "./lib/testing/index.mjs",
+      "require": "./lib/testing/index.js",
       "default": "./lib/testing/index.js"
     },
     "./reranker": {
       "types": "./lib/reranker.d.ts",
-      "require": "./lib/reranker.js",
       "import": "./lib/reranker.mjs",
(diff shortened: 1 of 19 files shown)


inlined:
Note, I ran into issues in multiple PRs that ts-node was not able to use "import" statements because the scripts code was not a module. "tsx" is already installed and is the more modern way to execute a TypeScript script, so I modified the format and lint scripts to use it.

inlined:
Looks like jest uses ts-node. Going to look into fixes.

mbleigh:
IIRC we have explicitly not done this before because it's actually correct and pnpm is wrong to be sending warnings. The ordering of `exports` is really bizarre (for instance `default` **must** be the last one) but certain toolchains we ran into had issues if we didn't have exactly this setup...I think.

inlined:
If that's the case, why are our orerings not consistent throughout the codebase?

mbleigh:
I don't remember for sure, I just remember trying to fix this same thing a few months back and going down a rabbit hole where I decided not touching it was the better course...maybe it's fine? I just know that this stuff has been very fragile to mess with previously.

mbleigh (approved):

pavelgj:
> I don't remember for sure, I just remember trying to fix this same thing a few months back and going down a rabbit hole where I decided not touching it was the better course...maybe it's fine? I just know that this stuff has been very fragile to mess with previously.

yeah, very fragile... I wouldn't trust our tests to cover these things.

inlined:
Trying to do some research because I agree this can both be fragile and important. The [node docs](https://nodejs.org/api/packages.html#conditional-exports) make it clear that default should be last but only give guidance that you should go “more to less” specific with import vs require (wtf does that even mean?!)

I checked the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html) and it definitely says “types” must come first. I ran across a bug that noted it works anyway and the team openly considered a breaking change to prevent the proliferation of malformed packages.

The best I can guess from this is that ESM modules should put import first and CSM modules should put require first because that is the naive syntax for the package

inlined:
There’s an in progress [Node docs repo](https://github.com/nodejs/package-examples/tree/main/guide) that goes into some of the pitfalls. FWIW, both of its examples have import before require. It also raises another question: given our very high minimum node version, why are we producing CSM instead of the more performant ESM?

pavelgj:
> There’s an in progress [Node docs repo](https://github.com/nodejs/package-examples/tree/main/guide) that goes into some of the pitfalls. FWIW, both of its examples have import before require. It also raises another question: given our very high minimum node version, why are we producing CSM instead of the more performant ESM?

"high minimum node version" happened organically as we discovered that some things weren't working... and I think originally we just started by copy-pasting the boilerplate from some firebase SDK, and tweaked things until it worked for us... then tweaked a lot more things as we noticed things breaking in various environments.

inlined:
After more research, I’m starting to think that we should not only make this change, but also change from CSM to ESM (which is a breaking change, so no time like the present!)

ESM is supported in all LTS node versions, loads faster (async I/O), supports tree shaking for smaller deployment sizes, and can be more easily shared in browsers (I know that’s not supported yet, but I’d like to get there). This would require zero code changes on our side because TS syntax is already ESM. I also don’t think it requires any customers to change code because we’re actually creating UMD with our exports annotations.

My somewhat informed assumption is that the change in module type would change what typescript transpiles to. I can verify if you want (once I get my test app working for no-skipLibCheck PR

The only downside is ESM is that exports are no longer dynamic and therefore you can’t replace them with mocks like Simon.

thoughts @mbleigh @pavelgj ?

mbleigh:
We can't switch to ESM because we are a library and we want maximally
compatible code. Even if people are using a version of Node.js that
supports ESM having a default ESM package is very hard to use from CJS and
lots of people (e.g. Next.js projects) are still CJS packages.

We are already shipping both CJS and ESM versions, and ESM imports will use
our ESM code.

inlined:
It may not matter as long as we either have a commonJS type with .mjs import entrypoints or a module  type with .cjs require entrypoints. 

Once I finish my other PR to get Genkit working in CF3 with a strict tsconfig, I'll tinker with importing genkit from both a CJS and ESM functions package in both TypeScript and JavaScript.

inlined:
I started going down a very long rabbit hole trying to figure out the right way to build a module that works as CJS and ESM. I have not yet been successful, but once/if I am, I'll try to apply the learnings to Genkit and firebase-functions. https://github.com/inlined/dual-mode-poc

PR #1592, opened 2025-01-10
feat(go/genkit): added persistent chat session and agent support

### LINKED ISSUE #1458: [Go] Add persistent chat session and agent support
https://firebase.google.com/docs/genkit/chat
https://firebase.google.com/docs/genkit/multi-agent

dysrama:
```go
func SimpleChat() {
	ctx := context.Background()
	g, err := genkit.New(nil)
	if err != nil {
		log.Fatal(err)
	}

	if err := vertexai.Init(ctx, g, nil); err != nil {
		log.Fatal(err)
	}

	m := vertexai.Model(g, "gemini-1.5-flash")

	chat, err := genkit.NewChat(
		ctx,
		g,
		genkit.WithModel(m),
		genkit.WithSystemText("You're a pirate first mate. Address the user as Captain and assist them however you can."),
		genkit.WithConfig(ai.GenerationCommonConfig{Temperature: 1.3}),
	)
	if err != nil {
		log.Fatal(err)
	}

	resp, err := chat.Send(ctx, "Hello")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Print(resp.Text())
}

func StatefulChat() {
	ctx := context.Background()
	g, err := genkit.New(nil)
	if err != nil {
		log.Fatal(err)
	}

	if err := vertexai.Init(ctx, g, nil); err != nil {
		log.Fatal(err)
	}

	m := vertexai.Model(g, "gemini-1.5-pro")

	// To include state in a session, you need to instantiate a session explicitly
	session, err := genkit.NewSession(ctx,
		genkit.WithSessionData(genkit.SessionData{
			State: map[string]any{
				"username": "Michael",
			},
		},
		),
	)
	if err != nil {
		log.Fatal(err)
	}

	// Modify the state using tools
	chatTool := genkit.DefineTool(
		g,
		"updateName",
		"use this tool to update the name of the user",
			Name string
		}) (string, error) {
			// Set name in state
			session, err := genkit.SessionFromContext(ctx)
			if err != nil {
				return "", err
			}

			err = session.UpdateState(map[string]any{
				"username": input.Name,
			})
			if err != nil {
				return "", err
			}

			return "changed username to " + input.Name, nil
		},
	)

	chat, err := genkit.NewChat(
		ctx,
		g,
		genkit.WithModel(m),
		genkit.WithSystemText("You're Kitt from Knight Rider. Address the user as Kitt would and always introduce yourself."),
		genkit.WithConfig(ai.GenerationCommonConfig{Temperature: 1}),
		genkit.WithSession(session),
		genkit.WithTools(chatTool),
	)
	if err != nil {
		log.Fatal(err)
	}

	resp, err := chat.Send(ctx, "Hello, my name is Earl")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Print(resp.Text())
	fmt.Print(session.SessionData.State["username"])
}

func MultiThreadChat() {
	ctx := context.Background()
	g, err := genkit.New(nil)
	if err != nil {
		log.Fatal(err)
	}

	if err := vertexai.Init(ctx, g, nil); err != nil {
		log.Fatal(err)
	}

	m := vertexai.Model(g, "gemini-1.5-flash")

	pirateChat, err := genkit.NewChat(
		ctx,
		g,
		genkit.WithModel(m),
		genkit.WithSystemText("You're a pirate first mate. Address the user as Captain and assist them however you can."),
		genkit.WithConfig(ai.GenerationCommonConfig{Temperature: 1.3}),
		genkit.WithThreadName("pirate"),
	)
	if err != nil {
		log.Fatal(err)
	}

	resp, err := pirateChat.Send(ctx, "Hello")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Print(resp.Text())

	lawyerChat, err := genkit.NewChat(
		ctx,
		g,
		genkit.WithModel(m),
		genkit.WithSystemText("You're a lawyer. Give unsolicited advice no matter what is asked."),
		genkit.WithConfig(ai.GenerationCommonConfig{Temperature: 1.3}),
		genkit.WithThreadName("lawyer"),
	)
	if err != nil {
		log.Fatal(err)
	}

	resp, err = lawyerChat.Send(ctx, "Hello")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Print(resp.Text())
}

func PersistentStorageChat() {
	ctx := context.Background()
	g, err := genkit.New(nil)
	if err != nil {
		log.Fatal(err)
	}

	if err := vertexai.Init(ctx, g, nil); err != nil {
		log.Fatal(err)
	}

	m := vertexai.Model(g, "gemini-1.5-pro")

	// To override default in-mem session storage
	store := &MyOwnSessionStore{
		SessionData: make(map[string]genkit.SessionData),
	}
	session, err := genkit.NewSession(ctx,
		genkit.WithSessionStore(store),
	)
	if err != nil {
		log.Fatal(err)
	}

	chat, err := genkit.NewChat(
		ctx,
		g,
		genkit.WithModel(m),
		genkit.WithSystemText("You're a helpful chatbox. Help the user."),
		genkit.WithConfig(ai.GenerationCommonConfig{Temperature: 1}),
		genkit.WithSession(session),
	)
	if err != nil {
		log.Fatal(err)
	}

	resp, err := chat.Send(ctx, "Hello, my name is Earl")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Print(resp.Text())

	// Load and use existing session
	session2, err := genkit.LoadSession(ctx, "mySessionID", store)
	if err != nil {
		log.Fatal(err)
	}

	chat2, err := genkit.NewChat(
		ctx,
		g,
		genkit.WithModel(m),
		genkit.WithSystemText("You're a helpful chatbox. Help the user."),
		genkit.WithConfig(ai.GenerationCommonConfig{Temperature: 1}),
		genkit.WithSession(session2),
	)
	if err != nil {
		log.Fatal(err)
	}

	resp, err = chat2.Send(ctx, "What's my name?")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Print(resp.Text())
}

type MyOwnSessionStore struct {
	SessionData map[string]genkit.SessionData
}

func (s *MyOwnSessionStore) Get(sessionId string) (data genkit.SessionData, err error) {
	d, err := os.ReadFile("/tmp/" + sessionId)
	if err != nil {
		return data, err
	}
	err = json.Unmarshal(d, &data)
	if err != nil {
		return data, err
	}

	if data.Threads == nil {
		data.Threads = make(map[string][]*ai.Message)
	}

	s.SessionData[sessionId] = data
	return s.SessionData[sessionId], nil
}

func (s *MyOwnSessionStore) Save(sessionId string, data genkit.SessionData) error {
	s.SessionData[sessionId] = data
	d, err := json.Marshal(data)
	if err != nil {
		return err
	}
	err = os.WriteFile("/tmp/"+sessionId, []byte(d), 0644)
	if err != nil {
		return err
	}
	return nil
}

```


apascal07 (commented):
Louise, could you please a sample that uses this new API? We'll use it as the canonical one for playing around with this in Dev UI.

dysrama:
> Louise, could you please a sample that uses this new API? We'll use it as the canonical one for playing around with this in Dev UI.

Added in description, went with the same examples we have in node

apascal07:
Can you actually add 

> > Louise, could you please a sample that uses this new API? We'll use it as the canonical one for playing around with this in Dev UI.
> 
> Added in description, went with the same examples we have in node

Can you actually commit this to the `samples` directory please? We want to have a canonical example that we can run anytime.

dysrama:
> Can you actually commit this to the `samples` directory please? We want to have a canonical example that we can run anytime.

Yep, done

go/genkit/chat.go
```diff
@@ -0,0 +1,335 @@
...
+
+	return chat, nil
+}
+
+// Send sends a message to the chat, generating a response from the AI model.
+// It retrieves the chat history from the session store, adds the new message
+// to the history, and sends the entire conversation to the AI model for
+// generating a response. If a system message is set for the chat, it is
+// included in the conversation before the history.
+func (c *Chat) Send(ctx context.Context, msg any) (resp *ai.ModelResponse, err error) {
```
apascal07, line 98: I think we should break this out into two methods:

```go
func (c *Chat) Send(ctx context.Context, msgs ...*Message) (resp *ai.ModelResponse, err error)
func (c *Chat) SendText(ctx context.Context, msgText string) (resp *ai.ModelResponse, err error)
```

This is a very high touch method so maintaining strong typing here is essential. This should cover the majority of user journeys since even in the non-string scenarios we still convert to a text message.

go/genkit/chat.go
```diff
@@ -0,0 +1,335 @@
...
+	err = c.UpdateMessages(c.ThreadName, messages)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+// UpdateMessages updates the messages for the chat.
+func (c *Chat) UpdateMessages(ThreadName string, messages []*ai.Message) error {
```
apascal07, line 169:
```suggestion
func (c *Chat) UpdateMessages(thread string, msgs []*ai.Message) error {
```

go/genkit/chat.go
```diff
@@ -0,0 +1,335 @@
...
+	c.Request.Messages = messages
+	return c.Session.UpdateMessages(ThreadName, messages)
+}
+
+// WithModel sets the model for the chat.
+func WithModel(model ai.Model) ChatOption {
+	return func(c *Chat) error {
+		if c.Model != nil {
+			return errors.New("cannot set model (WithModel) more than once")
```
apascal07, line 178: To be consistent with the rest of our API, can you update all of the error messages to follow the format `genkit.WithModel: cannot set Model more than once`.

go/genkit/chat.go
```diff
@@ -0,0 +1,335 @@
...
+		if c.Request.Config != nil {
+			return errors.New("cannot set config (WithConfig) more than once")
+		}
+		c.Request.Config = &config
+		return nil
+	}
+}
+
+// WithContext adds provided context to chat.
+func WithContext(context ...any) ChatOption {
```
apascal07, line 243:
```suggestion
func WithContext(context ...*Document) ChatOption {
```

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+		if s.ID != "" {
+			return errors.New("cannot set session id (WithSessionID) more than once")
+		}
+		s.ID = id
+		return nil
+	}
+}
+
+// WithSessionData sets the session data.
+func WithSessionData(data SessionData) SessionOption {
```
apascal07, line 192: I wonder if it makes sense to pull session and/or chat into a separate package so that we can simplify these function names to `WithData`, etc. Feels a bit laborious having to type it out.

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+	Threads map[string][]*ai.Message `json:"threads,omitempty"`
+}
+
+type Session struct {
+	// Genkit
+	Genkit *Genkit
+	// The session id
+	ID string
+	// The data for the session
+	SessionData SessionData
```
apascal07, line 51:
```suggestion
	Data SessionData
```

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+	State map[string]any `json:"state,omitempty"`
+	// Schema for state variables
+	StateSchema *jsonschema.Schema `json:"stateschema,omitempty"`
+	// The messages for each thread
+	Threads map[string][]*ai.Message `json:"threads,omitempty"`
+}
+
+type Session struct {
+	// Genkit
+	Genkit *Genkit
```
apascal07, line 47: Not used I think?

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		s.ID = id
+		return nil
+	}
+}
+
+// WithSessionData sets the session data.
+func WithSessionData(data SessionData) SessionOption {
+	return func(s *Session) error {
+		if s.SessionData.Threads != nil {
```
apascal07, line 194: Is this the right comparison?

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+
+type Session struct {
+	// Genkit
+	Genkit *Genkit
+	// The session id
+	ID string
+	// The data for the session
+	SessionData SessionData
+	// Default state variable values
+	DefaultState map[string]any
```
apascal07, line 53: Do we need this field? Can we just set `Data` with the default when we construct the `Session`?

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+
+		defaultState = structMap
+	}
+
+	return schema, defaultState, nil
+}
+
+// Default in-memory session store.
+type InMemSessionStore struct {
+	SessionData map[string]SessionData
```
apascal07, line 275:
```suggestion
type InMemorySessionStore struct {
	Data map[string]SessionData
```

apascal07: Can you add a comment explaining what the key is?

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+}
+
+// Helper function to derive schema and defaults from state.
+func getSchemaAndDefaults(state any) (schema *jsonschema.Schema, defaults map[string]any, err error) {
```
apascal07, line 232: I think we do this elsewhere, like in prompt/generate. Can we extract this out into a utility function that works for all of them (it should I think) and write some tests so we have consistent behavior across the codebase? Although with the other changes I'm suggesting, not sure if it still makes sense to do so.

Specifically for this, I don't know if it makes sense to allow primitives as a type/value. We should force them to make it a struct/map so we don't have to provide the magic key value of "state".

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+		defaultState = map[string]any{"state": v}
+	case float32:
+	case float64:
+		defaultState = map[string]any{"state": v}
+	case string:
+		defaultState = map[string]any{"state": v}
+	// Pass map directly
+	case map[string]any:
+		defaultState = v
+	case nil:
```
apascal07, line 247: Nil should be an error. Doesn't make sense to pass that in as a state type.

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+			return errors.New("cannot set session store (WithSessionStore) more than once")
+		}
+		s.Store = store
+		return nil
+	}
+}
+
+// WithStateType uses the type provided to derive the state schema.
+// If passing eg. a struct with values, the struct definition will serve as the schema, the values will serve as defaults.
+func WithStateType(state any) SessionOption {
```
apascal07, line 215: More comments below but I'm leaning towards making this very opinionated, like structs only. Unlike other situations like this in other APIs, state is de facto a map/struct, and since we need to know how to marshal it, I'm leaning towards a struct.

go/genkit/session.go
apascal07: We should probably unexport this and `Store` since we have `UpdateState()` and `UpdateMessages()`. We wouldn't want someone misunderstanding the API and accidentally directly modifying the state which then doesn't get saved.

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+	Get(sessionId string) (data SessionData, err error)
+	Save(sessionId string, data SessionData) error
+}
+
+type SessionData struct {
+	// Any state that should be stored
+	State map[string]any `json:"state,omitempty"`
+	// Schema for state variables
+	StateSchema *jsonschema.Schema `json:"stateschema,omitempty"`
+	// The messages for each thread
```
apascal07, line 41:
```suggestion
	// Messages by thread name.
```

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+		ID:          sessionId,
+		SessionData: sessionData,
+		Store:       store,
+	}
+
+	return session, nil
+}
+
+// UpdateState takes any data and sets it as state in the store.
+func (s *Session) UpdateState(state any) error {
```
apascal07, line 121: This looks like it modifies the state schema as well which we don't want to be doing. We should validate that the type matches the expected schema and update the values only.

go/genkit/session.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	s.SessionData = sessionData
+
+	s.SessionData.Threads[threadName] = messages
+	return s.Store.Save(s.ID, s.SessionData)
+}
+
+// Set current session ID in context.
+func (s *Session) SetContext(ctx context.Context) context.Context {
+	ctx = context.WithValue(ctx, currentSessionID, s.ID)
```
apascal07, line 154: We have context helpers in https://github.com/firebase/genkit/blob/72cc19ab00eb1d01c3c4236d4217d862762293c8/go/internal/base/context_key.go that might be useful here.

apascal07 (changes requested):

go/genkit/chat.go
dysrama: This is not until https://github.com/firebase/genkit/pull/1949/files#diff-730aeaa8bffbe3697ac7f27e59542972a2a5225deba8c0463f34f854d1c8c4c1R128 is merged right?

go/genkit/chat.go
```diff
@@ -0,0 +1,335 @@
...
+		if len(c.Request.Tools) != 0 {
+			return errors.New("cannot set tools (WithTools) more than once")
+		}
+		c.Request.Tools = tools
+		return nil
+	}
+}
+
+// WithOutputSchema adds provided output schema to the chat.
+func WithOutputSchema(schema any) ChatOption {
```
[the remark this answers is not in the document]
dysrama: WithOutputSchema in generate also take any and doesn't do any type checking? The schema in chat is just passed along to the generate WithOutputSchema function

go/genkit/session.go
dysrama: I rewrote this, so it's a much simpler helper function now, and doesn't check for primitives, so doesn't match the others anymore

go/genkit/session.go
dysrama: not directly no, but `s.data != (SessionData{})` because "struct containing map[string]any cannot be compared". So this was the best option I found. Any suggestions would be appreciated, can't figure out a better way

go/genkit/chat.go
dysrama: Fixed

apascal07:
> Any update on this? I would like to try switch from Node to Go, but this is the feature I'm missing

Hi David, thanks for your interest in this! We're going to come back to this shortly and try to have it out in the next few weeks. Thanks for your patience.

fterrag:
@apascal07 Hi there! What's needed to move this PR along?

apascal07:
> @apascal07 Hi there! What's needed to move this PR along?

We will pick work back up here very soon with a redefined design and scope.

PR #1597, opened 2025-01-13
fix(js/ai): skip response schema validation on tool calls

pavelgj:
Fixes https://github.com/firebase/genkit/issues/889

### DIFF
--- a/js/ai/src/generate/action.ts
+++ b/js/ai/src/generate/action.ts
@@ -190,16 +190,14 @@ async function generate(
   );
 
   // Throw an error if the response is not usable.
-  response.assertValid(
-    request,
-    // Skip schema validation if there are tools requests
-    (response.message?.content.filter((part) => !!part.toolRequest).length ??
-      0) > 0
-  );
+  response.assertValid();
   const message = response.message!; // would have thrown if no message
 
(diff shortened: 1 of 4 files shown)


js/ai/src/generate/action.ts
```diff
@@ -190,7 +190,12 @@ async function generate(
   );
 
   // Throw an error if the response is not usable.
-  response.assertValid(request);
+  response.assertValid(
+    request,
+    // Skip schema validation if there are tools requests
+    (response.message?.content.filter((part) => !!part.toolRequest).length ??
+      0) > 0
+  );
```
mbleigh, line 198: There's already logic to check for tool requests a few lines down, couldn't you do something there?

Basically do the valid assert right before it would return out of the tool loop anyway?

pavelgj: that's how I did it originally, but then I thought that we probably don't want to skip the whole validation, just the schema part... options were to either add a param to skip the schema validation (like a I did), or split out schema validation out of `response.assertValid` into a separate method and plug it into the condition below... or could keep it simple an skip validation entirely in case of tools...

pavelgj: split up validation functions

mbleigh (approved):

PR #1602, opened 2025-01-13
[Feat][Js] Make ai.generateStream no longer require a double await

### LINKED ISSUE #1593: FR: Remove double await for ai.generateStream
Currently ai.generateStream is an asynchronous API which returns asynchronous objects. This is a minor annoyance and would be nice to fix.

inlined:
Fixes #1593 

Rewrote generateStream so you longer need to call `await` to get the return value of objects that are themselves async.

As a style change I named the struct for {resolve, reject, promise} as Task because that's how I've commonly seen this design called.

The former method for passing into an async generator was a bit fragile because it depended on the exact order of the callbacks to the run loop (in fact, in my refactoring I broke it). Instead I created a buffering utility class called `Channel<T>`, inspired by Go channels but implements `AsyncIterable` for receiving.

I've had to change the copyright script from `ts-node` to `tsx` in many of my PRs because I get import vs require errors trying to load child-process. Since tsx is more officially supported I decided to change the way the script runs rather than adding module resolution annotations.

Checklist (if applicable):
- [X] Tested (unit tested, including new test)
- [X] Docs updated (updated jsdocs)

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -427,26 +427,19 @@ export function generateStream<
     | GenerateOptions<O, CustomOptions>
     | PromiseLike<GenerateOptions<O, CustomOptions>>
 ): GenerateStreamResponse<O> {
-  let fetch = createTask<GenerateResponse<O>>();
   let channel = new Channel<GenerateResponseChunk>();
 
-  (async () => {
-    try {
-      const generated = await generate<O, CustomOptions>(registry, {
-        ...options,
-        onChunk: channel.send.bind(channel),
-      });
-
-      channel.close();
-      fetch.resolve(generated);
-    } catch (err) {
-      channel.error(err);
-      fetch.reject(err);
-    }
-  })();
+  const generated = generate<O, CustomOptions>(registry, {
+    ...options,
+    onChunk: (chunk) => channel.send(chunk),
+  });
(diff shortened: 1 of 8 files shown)


inlined:
Arg last "fix" broke something. Will tinker

js/ai/src/generate.ts
```diff
@@ -355,86 +355,93 @@ export interface GenerateStreamResponse<O extends z.ZodTypeAny = z.ZodTypeAny> {
...
   let promise = new Promise<T>((res, rej) => ([resolve, reject] = [res, rej]));
-  return { resolve, reject, promise };
+  return {
+    resolve: resolve as Task<T>['resolve'],
+    reject: reject as Task<T>['reject'],
+    promise,
+  };
+}
+
+class Channel<T> implements AsyncIterable<T> {
```
pavelgj, line 376: nit/optional: this is neat and can probably be reused in other places... maybe move it to core?

pavelgj (approved):

js/ai/src/generate.ts
inlined: Will move this and Task to Core.

pavelgj:
FYI... this PR/commit does not follow the conventional commit format. Close... but not quite.

Issue #1360, opened 2024-11-21
[JS] Upgrade to 0.9.3 is causing significant increase in hallucinations


tekkeon:
**Describe the bug**
After upgrading to 0.9.3, we're noticing that the output Gemini (we're specifically using Gemini Pro 1.5 002) has numerous problems in our system. We're getting hallucinated text and overall very strange outputs that make our application unusable. When we downgrade back to 0.9.1, everything works properly again.

We looked at the rendered prompts and noticed the only difference was that `format` was now specified as `json`.

Snippit from 0.9.1:

```json
"output": {
    "jsonSchema": {
      "type": "object",
      "properties": {
...
```

Snippet from 0.9.3

```json
"output": {
    "format": "json",
    "jsonSchema": {
      "type": "object",
      "properties": {
...
```

We suspect this change to default to JSON mode may be causing the issue.

**To Reproduce**
Upgrade from 0.9.1 to 0.9.3 and run some prompts with complex outputs.

**Expected behavior**
We expected AI output to remain consistent to what our system had been producing.

**Runtime (please complete the following information):**
 - OS: MacOS 14.6.1

** Node version
 -  v18.17.1

pavelgj:
Thanks for the report. I'm investigating.

pavelgj:
There's a diff of model request between 0.5 and 0.9

![image](https://github.com/user-attachments/assets/3077d8df-e7d9-4f2f-b356-648ee73a262e)

the diff in instructions is negligible, however the new (0.9) json format does specify `contentType` and `constrained: true` option, which would make a difference in generation. In theory the difference should be for the better, but not necessarily.

pavelgj:
cc @mbleigh

pavelgj:
I have a PR: https://github.com/firebase/genkit/pull/1365
it would allow you to define custom formats and override the default behaviour changes in the json format. Ex:

```ts
ai.defineFormat(
  {
    name: 'myJson',
    format: 'json',
  },
  (schema) => {
    let instructions: string | undefined;

    if (schema) {
      instructions = `Output should be in JSON format and conform to the following schema:

\`\`\`
${JSON.stringify(schema)}
\`\`\`
`;
    }

    return {
      parseChunk: (chunk) => {
        return extractJson(chunk.accumulatedText);
      },

      parseMessage: (message) => {
        return extractJson(message.text);
      },

      instructions,
    };
  }
);

const MenuItemSchema = z.object({
  name: z.string(),
[...]

PR #1612, opened 2025-01-15
refactor(js/ai): refactored constrained generation into middleware, simplified json format

pavelgj:
Constrained generation is not coupled with `json` format. JSON format declares that it depends on constrained generation. Models that natively support constrained generation will handle it automatically, and for models that don't support constrained generation we provide `simulateConstrainedGeneration` middleware (automatically injected when the model does not declare support for constrained generation).

If a developer wants to use simulated constrained generation explicitly they can specify the middleware on the generate call:

```ts
await generate(registry, {
  prompt: 'generate json',
  use: [
    simulateConstrainedGeneration(),
  ],
  output: {
    schema,
  },
});
```

instruction for the middleware are overridable:

```ts
await generate(registry, {
  prompt: 'generate json',
  use: [
    simulateConstrainedGeneration({
      instructionsRenderer: (schema) =>
        `must be json: ${schemaToTSInterface(schema)}`,
    }),
  ],
  output: {
    schema,
  },
});
```

Checklist (if applicable):


mbleigh (approved):
I think I'm generally good with this, and I'm wondering if I made a mistake in how I was thinking about formats around streaming. I'm wondering if instead of `format: 'array'` it should have been `{format: 'json', stream: 'incremental' | 'array'}`.

I'm also wondering if `format` should accept mime types (e.g. `application/json` and the existing `json`, `text` are just shorthands that get translated into `application/json` and `text/plain`.

It would be a breaking change so if you think that sounds better we should try to work it in before 1.0

pavelgj:
> I think I'm generally good with this, and I'm wondering if I made a mistake in how I was thinking about formats around streaming. I'm wondering if instead of `format: 'array'` it should have been `{format: 'json', stream: 'incremental' | 'array'}`.
> 
> I'm also wondering if `format` should accept mime types (e.g. `application/json` and the existing `json`, `text` are just shorthands that get translated into `application/json` and `text/plain`.
> 
> It would be a breaking change so if you think that sounds better we should try to work it in before 1.0

Honestly, I like the simplicity of current formats. They kind of just make sense and simple to use. It's not obvious what benefits splitting out stream config would offer.

dongyangli1226:
@pavelgj @mbleigh 
Hi guys, thanks for this. We encountered similar issues mentioned in #1360 when upgrading to 1.0.5. Gemini models significantly hallucinate comparing to the behavior we see in 0.9.1. The fix you provided here does work. I wonder if you guys can help clarify this and when should we use this. It seems like we are following the standard of how Genkit and Dotprompt work based on documentation. I would expect declaring json schema in dotprompt will instruct the model properly. When the model hallucinates, it still generates json format correctly, however with very poor generated results.

Really appreciate your help. I think this might also affect others using dotprompt as it might seem like Gemini just doesn't follow instructions very well.

dongyangli1226:
The requests with and without the middleware looks exactly the same: (However, without using the middleware, the model doesn't consistently follow prompt instructions).

>  request {
>    "messages": [
>      {
>        "role": "user",
>        "content": [
>          {
>            "text": ""
>          }
>        ]
>      }
>    ],
>    "config": {
>      "version": "gemini-2.0-pro-exp-02-05",
>      "temperature": 0.2,
>      "safetySettings": [
>        {
>          "category": "HARM_CATEGORY_DANGEROUS_CONTENT",
>          "threshold": "BLOCK_ONLY_HIGH"
>        },
>        {
>          "category": "HARM_CATEGORY_HATE_SPEECH",
>          "threshold": "BLOCK_ONLY_HIGH"
>        },
>        {
>          "category": "HARM_CATEGORY_SEXUALLY_EXPLICIT",
>          "threshold": "BLOCK_ONLY_HIGH"
>        },
>        {
>          "category": "HARM_CATEGORY_HARASSMENT",
>          "threshold": "BLOCK_ONLY_HIGH"
>        }
>      ]
>    },
>    "tools": [],
>    "output": {
>      "constrained": true,
>      "contentType": "application/json",
>      "format": "json",
>      "schema": {
>        "type": "object",
>        "properties": {
>          "insights": {
>            "type": "array",
>            "items": {
>              "type": "object",
>              "properties": {
>                "title": {
>                  "type": "string"
>                },
>                "description": {
>                  "type": "string"
>                },
>                "type": {
>                  "type": "string",
>                  "enum": [
>                    "POTENTIAL_INCONSISTENCY",
>                    "INCOMPLETE_ANSWER",
>                    "ADMISSION"
>                  ]
>                },
>                "highlights": {
>                  "type": "array",
>                  "items": {
>                    "type": "string"
>                  }
>                },
>                "documentSnippets": {
>                  "type": "array",
>                  "items": {
>                    "type": "object",
>                    "properties": {
>                      "documentId": {
>                        "type": "string"
>                      },
>                      "text": {
>                        "type": "string"
>                      }
>                    },
>                    "required": [
>                      "documentId",
>                      "text"
>                    ],
>                    "additionalProperties": true
>                  }
>                },
>                "factIds": {
>                  "type": "array",
>                  "items": {
>                    "type": "string"
>                  }
>                },
>                "transcriptEntryIds": {
>                  "type": "array",
>                  "items": {
>                    "type": "string"
>                  }
>                },
>                "exampleQuestions": {
>                  "type": "array",
>                  "items": {
>                    "type": "string"
>                  }
>                }
>              },
>              "required": [
>                "title",
>                "description",
>                "type"
>              ],
>              "additionalProperties": true
>            }
>          }
>        },
>        "required": [
>          "insights"
>        ],
>        "additionalProperties": true,
>        "$schema": "http://json-schema.org/draft-07/schema#"
>      }
>    }
>  }

PR #1613, opened 2025-01-15
fix: handle for when no images are generated

KenCorbettJr:
Sometimes an image not actually returned from Imagen, especially if the image would violate some rule or policy.  This better handles those situations.


MichaelDoyle:
The vertexai plugin was deprecated in favor of the google genai pluginl

PR #1615, opened 2025-01-16
fix(js): Marking root spans that have failed

maxl0rd:
Addressing missing failedSpan and failedPath attributes when a failure happens in the root span.

### DIFF
--- a/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
+++ b/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
@@ -307,6 +307,26 @@ class AdjustingTraceExporter implements SpanExporter {
       )
     );
 
+    // Check if we have a failure in the root span that requires special handling
+    const rootSpan = spans.find((s) =>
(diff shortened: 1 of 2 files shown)


js/plugins/google-cloud/src/gcpOpenTelemetry.ts
```diff
@@ -307,6 +307,26 @@ class AdjustingTraceExporter implements SpanExporter {
       )
     );
 
+    // Check if we have a failure in the root span that requires special handling
```
schnecle, line 310: Should we throw this into a helper function like the ones below? Either that, we should remove the span = this.markFailedSpan call below.

schnecle (approved):

js/plugins/google-cloud/src/gcpOpenTelemetry.ts
maxl0rd: That helper and all the others run over all the individual child spans, so it's still relevant.

PR #1622, opened 2025-01-20
fix(evals): make testcaseId and traceId optional in input formats, refactor

ssbushi:
Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/js/tsconfig.json
+++ b/js/tsconfig.json
@@ -12,7 +12,7 @@
     "esModuleInterop": true,
     "strict": true,
     "target": "es2022",
-    "lib": ["es2022"],
+    "lib": ["es2022", "DOM"],
     "noEmit": true
   },
   "compileOnSave": true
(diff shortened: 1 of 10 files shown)


js/tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "esModuleInterop": true,
     "strict": true,
     "target": "es2022",
-    "lib": ["es2022"],
+    "lib": ["es2022", "DOM"],
```
pavelgj, line 15: I wonder if you only need to make this change for `js/genkit` package, not everything...

pavelgj (approved):

js/tsconfig.json
ssbushi: I don't know, @inlined made the change on js/clients and it doesn't compile on my machine. I just tried replicating that on the parent directory (`js`) and it worked...

shrutip90 (approved):

PR #1633, opened 2025-01-21
refactor(py): move initial implementation into its own temp directory for follow up PRs

yesudeep:
refactor(py): move initial implementation into its own temp directory for follow up PRs


pytemp/genkit/README.md
```diff
@@ -9,5 +9,6 @@ python3.11 -m pip install -e ".[dev]"
 
 Regen model objects
 
-```bash
+````bash
```
pavelgj, line 12: undo

PR #1647, opened 2025-01-22
feat: add new API for data validation

ssbushi:
Part of https://github.com/firebase/genkit/issues/1554

This change provides the required API. The Dev UI can use this API to add validation callouts.

Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/genkit-tools/common/src/eval/validate.ts
+++ b/genkit-tools/common/src/eval/validate.ts
@@ -52,7 +52,7 @@ export async function validateSchema(
     return { valid: true };
   }
 
-  let errorsMap: Record<string, any> = {};
+  const errorsMap: Record<string, any> = {};
 
   if (datasetId) {
     const datasetStore = await getDatasetStore();
@@ -69,7 +69,7 @@ export async function validateSchema(
 
     return Object.keys(errorsMap).length === 0
       ? { valid: true }
-      : { valid: false, datasetErrors: errorsMap };
+      : { valid: false, errors: errorsMap };
   } else {
     const dataset = InferenceDatasetSchema.parse(data);
     dataset.forEach((sample, index) => {
@@ -80,7 +80,7 @@ export async function validateSchema(
     });
     return Object.keys(errorsMap).length === 0
       ? { valid: true }
-      : { valid: false, dataErrors: errorsMap };
+      : { valid: false, errors: errorsMap };
   }
 }
 
--- a/genkit-tools/common/src/eval/validate.ts
+++ b/genkit-tools/common/src/eval/validate.ts
@@ -53,7 +53,7 @@ export async function validateSchema(
     return { valid: true };
   }
 
-  const errorsMap: Record<string, ErrorDetail[] | undefined> = {};
+  const errorsMap: Record<string, ErrorDetail[]> = {};
 
   if (datasetId) {
     const datasetStore = await getDatasetStore();
@@ -64,7 +64,7 @@ export async function validateSchema(
     dataset.forEach((sample, index) => {
       const response = validate(targetSchema, sample.input);
       if (!response.valid) {
-        errorsMap[sample.testCaseId] = response.errors;
+        errorsMap[sample.testCaseId] = response.errors ?? [];
       }
     });
 
@@ -76,7 +76,7 @@ export async function validateSchema(
     dataset.forEach((sample, index) => {
       const response = validate(targetSchema, sample.input);
       if (!response.valid) {
-        errorsMap[index.toString()] = response.errors;
+        errorsMap[index.toString()] = response.errors ?? [];
       }
     });
     return Object.keys(errorsMap).length === 0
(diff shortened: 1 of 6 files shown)


genkit-tools/common/src/server/router.ts
```diff
@@ -239,6 +244,17 @@ export const TOOLS_SERVER_ROUTER = (manager: RuntimeManager) =>
         return response;
       }),
 
+    /** Start new evaluation run */
+    validateSchema: loggedProcedure
+      .input(apis.ValidateDataRequestSchema)
+      .output(apis.ValidateDataResponseSchema)
+      .mutation(async ({ input }) => {
+        console.log(input);
```
pavelgj, line 252: remove

genkit-tools/common/src/server/router.ts
```diff
@@ -239,6 +244,17 @@ export const TOOLS_SERVER_ROUTER = (manager: RuntimeManager) =>
...
+        const response = await validateSchema(manager, input);
+        console.log(response);
```
pavelgj, line 254: remove

genkit-tools/common/src/eval/validate.ts
```diff
@@ -0,0 +1,110 @@
...
+    path: error.instancePath.substring(1).replace(/\//g, '.') || '(root)',
+    message: error.message!,
+  };
+}
+
+async function getAction(
+  manager: RuntimeManager,
+  actionRef: string
+): Promise<Action | undefined> {
+  const actions = await manager.listActions();
```
pavelgj, line 108: just FYI, but this is an expensive call... it's OK depending on application, but in this case seems fine.

genkit-tools/common/src/server/router.ts
```diff
@@ -239,6 +244,17 @@ export const TOOLS_SERVER_ROUTER = (manager: RuntimeManager) =>
         return response;
       }),
 
+    /** Start new evaluation run */
```
pavelgj, line 247: update please

genkit-tools/common/src/server/router.ts
```diff
@@ -239,6 +244,17 @@ export const TOOLS_SERVER_ROUTER = (manager: RuntimeManager) =>
...
+    validateSchema: loggedProcedure
+      .input(apis.ValidateDataRequestSchema)
```
pavelgj, line 249: the trpc method sounds generic, first I saw it I though "validate any arbitrary data against a schema"... but then it turned out it's eval dataset specific. Maybe call this `validateDatasetSchema`?

genkit-tools/common/src/eval/validate.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    return Object.keys(errorsMap).length === 0
+      ? { valid: true }
+      : { valid: false, datasetErrors: errorsMap };
+  } else {
+    const dataset = InferenceDatasetSchema.parse(data);
+    dataset.forEach((sample, index) => {
+      const response = validate(targetSchema, sample.input);
+      if (!response.valid) {
+        errorsMap[index] = response.errors;
```
pavelgj, line 78: is compiler not complaining here? `errorsMap` expects a string key, but `index` is a number....

genkit-tools/common/src/types/apis.ts
```diff
@@ -150,3 +150,25 @@ export const RunNewEvaluationRequestSchema = z.object({
...
+export const ValidateDataResponseSchema = z.object({
+  valid: z.boolean(),
+  datasetErrors: z
+    .record(z.string(), z.any())
+    .describe('Errors mapping when validating dataset')
+    .optional(),
+  dataErrors: z
+    .record(z.string(), z.any())
+    .describe('Errors mapping when validating raw data')
+    .optional(),
```
pavelgj, line 172: is it necessary to separate these? I mean, at the end of the day it's a map of test case identifier (or a row index, which is the same thing) to errors. Overloading inputs is OK, but overloading output should be minimized.

genkit-tools/common/src/eval/validate.ts
ssbushi: Would you rather have a `getAction` call here instead?

genkit-tools/common/src/eval/validate.ts
ssbushi: strangely, no. I have cast this to string now that you raised this.

genkit-tools/common/src/types/apis.ts
```diff
@@ -150,3 +150,23 @@ export const RunNewEvaluationRequestSchema = z.object({
...
+    data: InferenceDatasetSchema.optional(),
+  }),
+  actionRef: z.string(),
+});
+export type ValidateDataRequest = z.infer<typeof ValidateDataRequestSchema>;
+
+export const ValidateDataResponseSchema = z.object({
+  valid: z.boolean(),
+  errors: z
+    .record(z.string(), z.any())
```
pavelgj, line 166: just wondering if error should be more strongly typed than just `any`...

pavelgj (approved):

PR #1651, opened 2025-01-23
breaking(js/ai): refactored prompts API, migrated to standalone dotpr…

pavelgj:
Apologies, it's a big one... but hopefully straightforward enough to follow. Two things happening:

1. Deleted `@genkit-ai/dotprompt` package and migrated to using `dotprompt` standalone package.
2. Updated the definePrompt API as per https://discord.com/channels/1255578482214305893/1319039515624472668


js/ai/src/generate.ts
```diff
@@ -372,10 +374,12 @@ export function generateStream<
...
-  const generated = generate<O, CustomOptions>(registry, {
-    ...options,
-    onChunk: (chunk) => channel.send(chunk),
-  });
+  const generated = Promise.resolve(options).then((resolvedOptions) =>
+    generate<O, CustomOptions>(registry, {
+      ...resolvedOptions,
+      onChunk: (chunk) => channel.send(chunk),
+    })
+  );
```
mbleigh, line 382: What's this about? Probably fine just curious.

js/ai/src/generate.ts
```diff
@@ -346,6 +346,8 @@ async function resolveFullToolName(
     return `/tool/${name}`;
   } else if (await registry.lookupAction(`/prompt/${name}`)) {
     return `/prompt/${name}`;
+  } else if (await registry.lookupAction(`/prompt/dotprompt/${name}`)) {
+    return `/prompt/dotprompt/${name}`;
```
mbleigh, line 350: Do we want dotprompt segregated in the registry? Or should it just be part of `/prompt/` namespace?

js/genkit/tests/chat_test.ts
```diff
@@ -122,10 +124,11 @@ describe('chat', () => {
...
 
   it('can start chat from a prompt', async () => {
-    const preamble = ai.definePrompt(
-      { name: 'hi', config: { version: 'abc' } },
-      'hi from template'
-    );
+    const preamble = ai.definePrompt({
+      name: 'hi',
+      config: { version: 'abc' },
+      messages: 'hi from template',
```
mbleigh, line 130: 😵‍💫  So wait, what does it mean to do `messages` as a string instead of `prompt`? I think I remember we talked about this, but don't remember the outcome.

js/genkit/tests/session_test.ts
```diff
@@ -361,22 +361,20 @@ describe('session', () => {
...
-        config: { temperature: 1 },
-        description: 'Agent description',
-      },
-      '{{role "system"}} hello from template'
-    );
+    const agent = ai.definePrompt({
+      name: 'agent',
+      config: { temperature: 1 },
+      description: 'Agent description',
+      system: 'hello from template',
```
mbleigh, line 368: What would happen if someone did:

```
system: 'hello {{role "user"}}this is user now'
```

js/ai/src/prompt.ts
```diff
@@ -137,79 +174,381 @@ export interface ExecutablePrompt<
...
 }
 
+export type PartsGenerator<I, S = any> = (
+  input: I,
+  options: {
+    state?: S;
+  }
+) => Part[] | Promise<string | Part | Part[]>;
+
+export type MessageGenerator<I, S = any> = (
```
mbleigh, line 184: I think `MessagesGenerator` to be consistent in plural naming. Though I might prefer `Resolver` to `Generator` for all of these names...

js/ai/src/prompt.ts
```diff
@@ -137,79 +174,381 @@ export interface ExecutablePrompt<
...
+    );
+
+    let docs: DocumentData[] | undefined;
+    if (typeof options.docs === 'function') {
+      docs = await options.docs(input, {
+        state: session?.state,
+      });
+    } else {
+      docs = options.docs;
+    }
```
mbleigh, line 262: Not necessarily a now thing, but I feel like we should have a generic "resolver" utility that can do this kind of "if func pass it some stuff and await the response else take the literal value". We use that all over the place.

js/ai/src/generate.ts
mbleigh, line 350: Separate thing: I really want prompts to be able to have slashes in their name so that you can have a `prompts` directory with subdirectories. But the registry doesn't allow that right now even though I think it could without too much effort.

js/ai/src/prompt.ts
```diff
@@ -14,79 +14,117 @@
...
+  messages?: string | MessageData[] | MessageGenerator<z.infer<I>>;
+  docs?: DocumentData[] | DocumentGenerator<z.infer<I>>;
+  output?: OutputOptions<O>;
+  maxTurns?: number;
+  returnToolRequests?: boolean;
   metadata?: Record<string, any>;
   tools?: ToolArgument[];
   toolChoice?: ToolChoice;
   use?: ModelMiddleware[];
 }
```
mbleigh, line 119: Interfaces this big make me a little nervous...is there another canonical interface that we can derive this from to make it less likely to drift? Or should we somehow document an "if change" type thing?

js/ai/src/prompt.ts
```diff
@@ -137,79 +174,381 @@ export interface ExecutablePrompt<
...
+        },
+        ...options.metadata,
+        type: 'prompt',
+      },
     },
-    fn
+    async (input: z.infer<I>): Promise<GenerateRequest<z.ZodTypeAny>> => {
+      return toGenerateRequest(
+        registry,
+        await renderOptionsFn(input, undefined)
```
mbleigh, line 305: Why an explicit `undefined` second argument? Seems like a smell.

js/ai/src/prompt.ts
```diff
@@ -222,3 +561,158 @@ export function isExecutablePrompt(obj: any): boolean {
...
+          );
+        } else {
+          // If this prompt is in a subdirectory, we need to include that
+          // in the namespace to prevent naming conflicts.
+          let prefix = '';
+          if (promptsPath !== dirEnt.path) {
+            prefix = dirEnt.path
+              .replace(`${promptsPath}/`, '')
+              .replace(/\//g, '-');
+          }
```
mbleigh, line 598: Yeah as mentioned elsewhere, would be great if prompts could have slashes in their names before 1.0

js/ai/src/prompt.ts
```diff
@@ -222,3 +561,158 @@ export function isExecutablePrompt(obj: any): boolean {
...
+  let variant: string | null = null;
+  if (name.includes('.')) {
+    const parts = name.split('.');
+    name = parts[0];
+    variant = parts[1];
+  }
+  const source = readFileSync(join(path, filename), 'utf8');
+  const parsedPrompt = registry.dotpromptEnv.parse(source);
+  const promptMetadata =
+    await registry.dotpromptEnv.renderMetadata(parsedPrompt);
```
mbleigh, line 639: Could we keep this around as-is somewhere in the registry metadata? David East is working on "extensions" of dotprompt and needs to be able to access the `ext` field of the rendered metadata.

mbleigh (commented):
I think this mostly looks good. There's a lot of boilerplatey interface to interface stuff, but I'm not sure it can really be avoided (and hey, 4K lines removed so that's not too shabby!)

js/ai/src/generate.ts
pavelgj: This is actually a sort of unrelated bug fix -- when generateStream was made sync we broke the ability to accept options as a promise. This fixes that.

js/genkit/tests/chat_test.ts
pavelgj: you can do something like `{{role "system"}} talk like a pirate {{ role "user" }} hi` in there.

Basically, before

```ts
definePrompt(
  {...},
  `{{role "system"}} talk like a pirate {{ role "user" }} hi`
)
```

after

```ts
definePrompt(
  {
     ...,
     messages: `{{role "system"}} talk like a pirate {{ role "user" }} hi`
  },
)
```

in the `prompt` option you can only put the "user"... can't put a system prompt in there.

```ts
definePrompt(
  {
     ...,
     prompt: `hi`
  },
)
```

js/genkit/tests/session_test.ts
pavelgj: it will throw an error... `system` and `prompt` allow only one message. This would have to go under `messages`.

js/ai/src/generate.ts
pavelgj: I don't want to... but prompt file loading is async... so we need an async initialization hook that will block registry lookup, and the only thing that can do that is plugins, which require namespacing.
Can look into adding async registry providers... so that you can register an action with a promise.
Do you foresee issues with namespacing like this?

pavelgj: Yes, I believe it's a trivial validation fix. we can do it.

mbleigh: Since the actions themselves are async couldn't we push it down a layer? Have each action have a `loadAllPromptsAndStuff()` at the top that it awaits before it does other work?

js/ai/src/prompt.ts
pavelgj: Done.

js/ai/src/prompt.ts
pavelgj: yes, but those resolver functions are not all the same... 2 are (for now) but messages is different.

js/core/src/registry.ts
```diff
@@ -68,6 +71,18 @@ export class Registry {
...
+      const resolvedSchema = await this.lookupSchema(name);
+      if (!resolvedSchema) {
+        throw new GenkitError({
+          message: `Schema '${name}' not found`,
+          status: 'NOT_FOUND',
+        });
+      }
+      return toJsonSchema(resolvedSchema);
+    },
+  });
```
mbleigh, line 85: Could you switch this to `new DotpromptEnvironment` and import that directly? I'm thinking I want to get rid of `dotprompt` as a toplevel export because I think I want it to be:

```ts
const dotprompt = new DotpromptEnvironment(...);
```

or something like that instead. Otherwise it's hard to know what to name the variable.

js/ai/src/prompt.ts
pavelgj: It tries to follow GenerateOptions, but with a lot of sugar... can probably extend from it, but would have to omit and override like half the fields...

js/ai/src/prompt.ts
pavelgj: I got bitten by making it optional... when rendering from an action you actually can't override options, but in other places options are overridable. I made the arg explicit to make sure I don't forget to pass it somewhere where it is actually overridable.

js/ai/src/prompt.ts
pavelgj: see line 660, is that enough?

js/ai/src/prompt.ts
pavelgj: Noted.

js/ai/src/generate.ts
pavelgj: no, because we need action metadata to register the top action...

pavelgj: we'd need a promise action registration which only requires action type/name as input, everything else provided in a promise...

mbleigh: Ah, okay. I can't think of a reason this is a *real* problem, so I'll leave it be.

js/ai/src/prompt.ts
mbleigh: 👍

js/ai/src/prompt.ts
mbleigh: Eh, it's probably fine...

js/ai/src/generate.ts
mbleigh: To clarify: did you do it already?

js/core/src/registry.ts
mbleigh: See https://github.com/google/dotprompt/pull/13 <-- mostly I'm making a breaking change to Dotprompt so let's coordinate before we code freeze.

js/ai/src/generate.ts
pavelgj: No, I'll do it as a follow up. Probably along with async action registration.

js/core/src/registry.ts
pavelgj: Done.

mbleigh (approved):

Issue #1657, opened 2025-01-24
[Docs] Document Interrupts


mbleigh:
We're adding interrupts to 1.0 and need good documentation. Unclear if it should be its own separate guide or part of the tool calling guide.

PR #1660, opened 2025-01-25
feat(js/ai): Adds GenerateOptions#resume, ToolAction#reply, Message#interrupts

mbleigh:
```ts
const myInterrupt = ai.defineInterrupt(...);
const response = await ai.generate({tools: [myInterrupt], prompt: "Call myInterrupt"});

const resumedResponse = await ai.generate({
  messages: response.messages,
  resume: {reply: myInterrupt.reply(response.interrupts[0], "the interrupt response");
});
```

Checklist (if applicable):
- [x] Docs bug filed: #1657

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -140,17 +140,20 @@ function applyResumeOption(
   messages: MessageData[]
 ): MessageData[] {
   if (!options.resume) return [];
-  const lastModelMessage =
-    messages[messages.map((m) => m.role).lastIndexOf('model')];
   if (
-    !lastModelMessage ||
-    !lastModelMessage.content.find((p) => !!p.toolRequest)
-  )
+    messages.at(-1)?.role !== 'model' ||
+    !messages
+      .at(-1)
+      ?.content.find((p) => p.toolRequest && p.metadata?.interrupt)
+  ) {
     throw new GenkitError({
       status: 'FAILED_PRECONDITION',
-      message: `Cannot 'resume' generation when there is no prior model message with toolRequests.`,
+      message: `Cannot 'resume' generation unless the previous message is a model message with at least one interrupt.`,
     });
+  }
+  const lastModelMessage = messages.at(-1)!;
   const toolRequests = lastModelMessage.content.filter((p) => !!p.toolRequest);
+
   const pendingResponses: ToolResponsePart[] = toolRequests
     .filter((t) => !!t.metadata?.pendingToolResponse)
     .map((t) => ({
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -42,6 +42,7 @@ import {
   ModelArgument,
   ModelMiddleware,
   Part,
+  ToolResponsePart,
   resolveModel,
 } from './model.js';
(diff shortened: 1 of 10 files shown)


js/ai/src/generate.ts
```diff
@@ -97,6 +135,43 @@ export interface GenerateOptions<
...
+  const toolRequests = lastModelMessage.content.filter((p) => !!p.toolRequest);
+  const pendingResponses: ToolResponsePart[] = toolRequests
+    .filter((t) => !!t.metadata?.pendingToolResponse)
+    .map((t) => ({
+      toolResponse: t.metadata!.pendingToolResponse,
+    })) as ToolResponsePart[];
+
+  console.log('PENDING:', pendingResponses);
+
+  const reply = Array.isArray(options.resume.reply)
```
pavelgj, line 162:
```suggestion
    })) as ToolResponsePart[];

  const reply = Array.isArray(options.resume.reply)
```

js/ai/src/generate.ts
```diff
@@ -97,6 +135,43 @@ export interface GenerateOptions<
   use?: ModelMiddleware[];
 }
 
+function applyResumeOption(
+  options: GenerateOptions,
+  messages: MessageData[]
+): MessageData[] {
+  if (!options.resume) return [];
+  const lastModelMessage =
+    messages[messages.map((m) => m.role).lastIndexOf('model')];
```
pavelgj, line 144: should this be literally looking at the last message in history?

js/ai/src/tool.ts
```diff
@@ -202,6 +218,22 @@ export function defineTool<I extends z.ZodTypeAny, O extends z.ZodTypeAny>(
         interrupt: interruptTool,
       })
   );
+  (a as ToolAction<I, O>).reply = (interrupt, replyData, options) => {
+    parseSchema(replyData, {
```
pavelgj, line 222: there's an opportunity here to do what you were suggesting at one point

```ts
defineTool({...}, async (input, {interrupt}) => {
  const hitlReply = interrupt({moreInfo: 'provide foo'});
  return doToolStuff(hitlReply);
});
```

that would mean `replyData` is not tool response a whatever interrupt expects. This methods would just invoke the tool in the resumed state.

pavelgj (approved):
I like this very much!

js/ai/src/tool.ts
mbleigh: My proposal for that is [in discord](https://discord.com/channels/1255578482214305893/1332419160038178847/1332637632592871558) although the idea that `interrupt` would no-op instead of throwing when confirmed is an interesting alternative idea. Let's discuss more in chat.

js/ai/src/generate.ts
mbleigh: Gemini doesn't allow `model`, `user`, `tool` but does allow `model`, `tool`, `user` which is interesting. I made it last message since that's more restrictive and it'd be easier to loosen than tighten later.

js/ai/src/tool.ts
mbleigh: Will implement restarts in a followup PR.

PR #1661, opened 2025-01-25
docs(js): Adds a guide for interrupts.

### LINKED ISSUE #1657: [Docs] Document Interrupts

mbleigh:
Fixes #1657


docs/interrupts.md
pavelgj, line 1: should this be a section in tools docs instead?

pavelgj (approved):

PR #1662, opened 2025-01-25
refactor(js/ai): drop namespace for dotprompt, allow async action registration, allow action names with slashes

pavelgj:


js/core/src/registry.ts
```diff
@@ -133,15 +139,38 @@ export class Registry {
...
   /**
    * Returns all actions in the registry.
    * @returns All actions in the registry.
    */
   async listActions(): Promise<ActionsRecord> {
     await this.initializeAllPlugins();
+    const actions: Record<string, Action<z.ZodTypeAny, z.ZodTypeAny>> = {};
+    for (let [key, action] of Object.entries(this.actionsById)) {
+      actions[key] = await action;
+    }
```
mbleigh, line 170: Should this be:

```suggestion
    await Promise.all(Object.entries(this.actionsById).map(async ([key,action]) => {
      actions[key] = await action;
    });
```

so that it doesn't wait serially on the actions? Maybe it doesn't really matter since the promises are already in flight tho.

mbleigh (approved):

PR #1663, opened 2025-01-26
feat(js): propagate context to sub actions, expose context in prompts

pavelgj:
Allows:
```ts
definePrompt({
  prompt: 'hello {{@auth.email}}',
})
```

Also changed express plugin to put auth into `context.auth` instead of just `context`.

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -385,9 +385,11 @@ export async function generate<
           rawRequest: params,
           middleware: resolvedOptions.use,
         });
-      const response = await (resolvedOptions.context
-        ? runWithContext(registry, resolvedOptions.context, generateFn)
-        : generateFn());
+      const response = await runWithContext(
+        registry,
+        resolvedOptions.context,
+        generateFn
+      );
       const request = await toGenerateRequest(registry, {
         ...resolvedOptions,
         tools,
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -42,6 +42,7 @@ import {
   ModelArgument,
   ModelMiddleware,
   Part,
+  ToolResponsePart,
   resolveModel,
 } from './model.js';
 import { ExecutablePrompt } from './prompt.js';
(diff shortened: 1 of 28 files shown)


js/ai/src/prompt.ts
```diff
@@ -179,6 +180,7 @@ export type PartsResolver<I, S = any> = (
   input: I,
   options: {
     state?: S;
+    context?: any;
```
mbleigh, line 183: `Record<string,any>`?

mbleigh (commented):
I was thinking that `GenerateOptions` should actually have a `context` option so that context can be passed directly without needing to be embedding in a flow...WDYT?

js/ai/src/prompt.ts
pavelgj: done

pavelgj:
> I was thinking that `GenerateOptions` should actually have a `context` option so that context can be passed directly without needing to be embedding in a flow...WDYT?

Good idea. Done.

js/ai/src/generate.ts
```diff
@@ -133,6 +134,8 @@ export interface GenerateOptions<
   streamingCallback?: StreamingCallback<GenerateResponseChunk>;
   /** Middleware to be used with this model call. */
   use?: ModelMiddleware[];
+  /** Additional context (data, like e.g. auth) to be passed down to tools, prompts and other sub actions. */
+  context?: Record<string, any>;
```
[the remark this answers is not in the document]
pavelgj: @inlined and I discussed adding something similar. As good of a time as any I suppose. Done.

js/ai/src/prompt.ts
```diff
@@ -486,6 +514,11 @@ async function renderMessages<
       messages.push(
         ...(await options.messages(input, {
           state: session?.state,
+          context: {
+            ...getContext(registry),
+            ...options.context,
```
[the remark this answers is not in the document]
pavelgj: yeah, I wasn't 100% sure about merging. Last minute decision. Let's start without merging. Can tweak later if that's something we want/need.

js/ai/src/generate.ts
```diff
@@ -376,10 +380,14 @@ export async function generate<
...
-        middleware: resolvedOptions.use,
-      });
+      const generateFn = () =>
+        generateHelper(registry, {
+          rawRequest: params,
+          middleware: resolvedOptions.use,
+        });
+      const response = await (resolvedOptions.context
+        ? runWithContext(registry, resolvedOptions.context, generateFn)
+        : generateFn());
```
mbleigh, line 390: Would it be better to have `runWithContext` handle an `undefined` context so that this doesn't need to be a ternary?

mbleigh (approved):

js/ai/src/generate.ts
pavelgj: Not sure... I'm thinking we might want the ability to set undefined context (run some action in isolation)...

Another option is:
```ts
      const response = await runWithContext(registry, resolvedOptions.context || getContext(registry), generateFn); 
```

but it would mean we're unnecessarily setting context in ALS, avoidable overhead.

mbleigh: I guess I'm saying that `runWithContext` could see an `undefined` context and no-op -- not actually do anything with ALS and just fall through to calling the underlying function. So same behavior as defined here, but encapsulated in a single place.

```ts
runWithContext(registry, undefined, generateFn) // equivalent to generateFn()
runWithContext(registry, {}, generateFn) // stores empty context in ALS and runs geneateFn()
runWithContext(registry, someContext, generateFn) // stores context in ALS and runs generateFn()
```

I'm personally usually wary of these patterns because it turns `runWithContext` into `runWithContextExceptIfYouPassUndefinedContextThenItsAPassthough`, which may not be obvious to the caller. But I agree it makes call sites much cleaner.

PR #1664, opened 2025-01-26
fix(js/plugins/google-cloud): Catch errors looking for ADC

MichaelDoyle:
We have some helpful error messaging to help users apply the necessary roles to their GCP service accounts for telemetry exports. We look up the logged in service account to customize the messaging. If the lookup
fails we need to handle those errors properly.

Resolves: https://discord.com/channels/1255578482214305893/1315032928886984755

Checklist (if applicable):
- [NA] Docs updated (updated docs or a docs bug required)

### DIFF
--- a/js/plugins/google-cloud/src/auth.ts
+++ b/js/plugins/google-cloud/src/auth.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { logger } from 'genkit/logging';
-import { auth, GoogleAuth } from 'google-auth-library';
+import { auth, CredentialBody, GoogleAuth } from 'google-auth-library';
 import { GcpPrincipal, GcpTelemetryConfig } from './types.js';
 
 /**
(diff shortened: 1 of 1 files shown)


js/plugins/google-cloud/src/auth.ts
```diff
@@ -69,7 +70,12 @@ export async function credentialsFromEnvironment(): Promise<
  **/
 export async function resolveCurrentPrincipal(): Promise<GcpPrincipal> {
   const envCredentials = await credentialsFromEnvironment();
-  const adcCredentials = await auth.getCredentials();
+  let adcCredentials = {} as CredentialBody;
+  try {
+    adcCredentials = await auth.getCredentials();
+  } catch (e) {
+    // do nothing
```
pavelgj, line 77: maybe `logger.debug` something here?

pavelgj (approved):

schnecle (commented):
If we are adding in the ability to grab client_email from adc, the projectId should also come from adc. Otherwise we could allow customers to land in a state where they provide incomplete credentials via environment variable and optimistically fill them in if they just so happen to use adc. I think we kind of want to do a switch at the top level and return all ADC creds or all environment creds.

PR #1666, opened 2025-01-27
fix(js/core/async): reject channel ready promise on channel error

pavelgj:
Fixed this error case:

```ts
  const { stream } = ai.generateStream({
    prompt: 'hi',
    model: 'modelNotFound',
  });
    // something
  }
```

which was resulting in a hanging generate call.

Checklist (if applicable):


inlined (approved):
Great catch!

PR #1668, opened 2025-01-27
feat(js/plugins/vertexai): instruduced gemini model ref helper and ability to register versions

pavelgj:
`gemini` model ref helper, consistent with googleAI plugin.

```ts
import { vertexAI, gemini } from '@genkit-ai/vertexai';

const gemini15Flash042 = gemini('gemini-1.5-flash-042');
const ai = genkit({
  plugins: [
    vertexAI({
      models: [gemini15Flash042]
    })
  ],
  model: gemini15Flash042
});

await generate('hi')
```

Checklist (if applicable):
- [iou] Docs updated (updated docs or a docs bug required)

### DIFF
--- a/js/plugins/vertexai/src/gemini.ts
+++ b/js/plugins/vertexai/src/gemini.ts
@@ -721,7 +721,7 @@ export function defineGeminiModel(
 
       // Handle system instructions separately
       let systemInstruction: Content | undefined = undefined;
-      if (modelInfo?.supports?.systemRole) {
+      if (!SUPPORTED_V1_MODELS[version]) {
         const systemMessage = messages.find((m) => m.role === 'system');
         if (systemMessage) {
           messages.splice(messages.indexOf(systemMessage), 1);
--- a/js/plugins/vertexai/src/gemini.ts
+++ b/js/plugins/vertexai/src/gemini.ts
@@ -697,7 +697,7 @@ export function defineGeminiModel(
   options: PluginOptions
 ): ModelAction {
   const middlewares: ModelMiddleware[] = [];
-  if (!modelInfo?.supports?.systemRole) {
+  if (SUPPORTED_V1_MODELS[version]) {
     middlewares.push(simulateSystemPrompt());
   }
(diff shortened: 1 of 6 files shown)


js/plugins/vertexai/src/gemini.ts
```diff
@@ -619,7 +721,7 @@ export function defineGeminiModel(
 
       // Handle system instructions separately
       let systemInstruction: Content | undefined = undefined;
-      if (SUPPORTED_V15_MODELS[name]) {
+      if (modelInfo?.supports?.systemRole) {
```
MichaelDoyle, line 724: See comment above. I think we had an intentional split where system roles were supported via simulation in 1.0 era models, and via systemInstructions in 1.5+ era models.

js/plugins/vertexai/src/gemini.ts
```diff
@@ -594,19 +673,42 @@ export function defineGeminiModel(
...
+  version: string,
+  modelInfo: ModelInfo | undefined,
+  vertexClientFactory: (
+    request: GenerateRequest<typeof GeminiConfigSchema>
+  ) => VertexAI,
+  options: PluginOptions
+): ModelAction {
   const middlewares: ModelMiddleware[] = [];
-  if (SUPPORTED_V1_MODELS[name]) {
+  if (!modelInfo?.supports?.systemRole) {
```
MichaelDoyle, line 700: I seem to recall we need to simulate system prompts for 1.0 era models, but we still want to say "systemRole=true" so that the dev ui will render the system prompt input, etc.

IOW, systemRole=true applies even to cases where its simulated.

MichaelDoyle: Separately, if we don't support 1.0 anymore, we can probably remove this?

pavelgj: Good catch, we still need to support 1.0. Undone.

MichaelDoyle (approved):

PR #1670, opened 2025-01-27
feat(js): truncate input/output to align with GCP limits

kmandrika:
GCP limits log lines to 256k. This change truncates the main culprits - input and output - and attempts to truncate paths, model names, and other potential sources contributing to large log lines.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/telemetry/action.ts
+++ b/js/plugins/google-cloud/src/telemetry/action.ts
@@ -30,6 +30,8 @@ import {
   createCommonLogAttributes,
   extractErrorName,
   extractOuterFeatureNameFromPath,
+  truncate,
+  truncatePath,
 } from '../utils.js';
 
 class ActionTelemetry implements Telemetry {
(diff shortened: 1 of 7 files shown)


MichaelDoyle (approved):

MichaelDoyle:
This looks good to me overall.

I am a little confused, though, in that it seems like we likely double count metrics and double log input/output when an action is at the root? That would cause the action to be both a feature and an action? Unless I misunderstand. :)

kmandrika:
> This looks good to me overall.
> 
> FWIW, I do find the logging of input/output inside the "telemetry" classes to be a bit convoluted and confusing.
> 
> At first glance, it appears we'll double log, since an action can both be "root" (i.e. "feature") and also an "action" (child span). In practice, we're probably in the clear, because we only log input/output for tools in the action class, and nothing else. That said, this is based on the assumption that a tool call will never be root (i.e. won't be a feature). This is not a correct assumption, even though it is probably "rare" in the wild.
> 
> In general, I am not sure the behavior is correct for developers calling actions directly, outside of a flow wrt input/output logging.

Yeah, it's worth rethinking this approach at some point.

PR #1671, opened 2025-01-27
feat(multimodal)!: Added Multimodal support for Vertex AI

ifielker:
BREAKING CHANGE: Embedding type now includes metadata and ai.embed returns an array of Embedding instead of a single embedding per document.


genkit-tools/tsconfig.base.json
```diff
@@ -9,7 +9,9 @@
     "sourceMap": true,
     "noImplicitOverride": true,
     "noPropertyAccessFromIndexSignature": false,
-    "skipLibCheck": true
+    "skipLibCheck": true,
```
pavelgj, line 12: undo these? bad merge?

js/ai/src/document.ts
```diff
@@ -53,19 +57,60 @@ export type DocumentData = z.infer<typeof DocumentDataSchema>;
...
   content: Part[];
   metadata?: Record<string, any>;
+  embedMetadata?: Record<string, unknown>;
 
   constructor(data: DocumentData) {
-    this.content = data.content;
-    this.metadata = data.metadata;
+    this.content = data.content as Part[];
+    this.metadata = data.metadata as Record<string, any>;
+    this.embedMetadata = data.embedMetadata as Record<string, unknown>;
```
pavelgj, line 65: unnecessary casts

js/genkit/tests/embed_test.ts
```diff
@@ -30,10 +30,12 @@ describe('embed', () => {
...
-      const response = await ai.embed({
-        embedder: 'echoEmbedder',
-        content: 'hi',
-      });
+      const response = (
+        await ai.embed({
+          embedder: 'echoEmbedder',
+          content: 'hi',
+        })
+      )[0].embedding;
```
pavelgj, line 38: probably better to assert on the entire response... on line 45

js/plugins/chroma/package.json
```diff
@@ -47,9 +47,9 @@
   "types": "./lib/index.d.ts",
   "exports": {
     ".": {
+      "types": "./lib/index.d.ts",
```
pavelgj, line 50: undo?

js/ai/src/document.ts
```diff
@@ -40,9 +41,12 @@ export type MediaPart = z.infer<typeof MediaPartSchema>;
 export const PartSchema = z.union([TextPartSchema, MediaPartSchema]);
 export type Part = z.infer<typeof PartSchema>;
 
+// We need both metadata and embedMetadata because they can
+// contain the same fields (e.g. video start/stop) with different values.
 export const DocumentDataSchema = z.object({
   content: z.array(PartSchema),
   metadata: z.record(z.string(), z.any()).optional(),
+  embedMetadata: z.record(z.string(), z.unknown()).optional(),
```
pavelgj, line 49: please use the metadata field, separate `embedMetadata` field is not necessary.

js/ai/src/document.ts
ifielker: done

js/ai/src/document.ts
ifielker: Yes it is. As I stated in the comments above: 
    // We need both metadata and embedMetadata because they can
    // contain the same fields (e.g. video start/stop) with different values.
We would not have needed this if we went with one document per embeddings because they would have matched, but you wanted multiple. So now we have extra embed data that is different from the request metadata but has the same fields, so they can't be stored in metadata. 
e.g. 
metadata: {
someInfo: "blah",
startEmbeddingSeconds: 0,
endEmbeddingSeconds: 120,
},
embedMetadata: {
startEmbeddingSeconds: 30,
endEmbeddingSeconds: 60,
}

js/genkit/tests/embed_test.ts
ifielker: I only added the "[0].embedding" part and the formatter moved whitespace. I have no idea what this test is about. Whoever wrote it originally should fix it probably.

js/plugins/chroma/package.json
ifielker: This is what fixes the warnings about "will never be used because x is before y" that you see when you compile.

js/ai/src/document.ts
pavelgj: no, you would do it like this:

```ts
const doc = Document.fromMedia('gs://.../', 'video/mp4', {
  someInfo: "blah",
  startEmbeddingSeconds: 0,
  endEmbeddingSeconds: 120,
});
const embedding = ai.embed({
  embedder: multimodalEmbedding001,
  content: doc,
  metadata: {
    startEmbeddingSeconds: 30,
    endEmbeddingSeconds: 60,
  }
});
```

document always represents the original document.

js/plugins/chroma/package.json
pavelgj: the warning is wrong

ifielker: ok...

js/ai/src/document.ts
ifielker: Right, and then when you retrieve it from storage, you have NO IDEA which embedding it used or which part of the video to send except "the whole big chunk". So there would be no point in getting any additional embeddings past a single one for the document. We had this discussion...

ifielker: Perhaps it's not clear - the "embedMetadata that gets added to the original document is what comes back from the embedder in the embedding response metadata."

js/ai/src/embedder.ts
```diff
@@ -19,14 +19,19 @@ import { Registry } from '@genkit-ai/core/registry';
...
 
 /**
- * A batch (array) of embeddings.
+ * EmbeddingSchema includes the embedding and also metadata so you know
+ * which of multiple embeddings corresponds to which part of a document.
  */
-export type EmbeddingBatch = { embedding: Embedding }[];
+export const EmbeddingSchema = z.object({
+  embedding: z.array(z.number()),
+  embedMetadata: z.record(z.string(), z.unknown()).optional(),
```
pavelgj, line 32: just `metadata`

js/ai/src/document.ts
ifielker: So if you specify a video document from 0-120 seconds and you get super fine embeddings every 8 seconds, then you get a lot of embeddings back. But that's no good to you if they all point to the same document. Vector storage stores by embedding (number array) only, not embedding plus metadata, so if you want to know which part of the video a particular embedding is for, then you need to store the embedding metadata somewhere and the only option is in the document. It's messy, but you wanted support for multiple embeddings returned per incoming document. And documents are not unique unless they also contain the embedMetadata. There's no point in having 5 embeddings all pointing to the same document.

js/ai/src/embedder.ts
ifielker: I started that way. But it gets confusing which metadata is which, so I renamed this one to be clearer.

pavelgj: The object is embedding. The metadata field of embedding object is embedding metadata. No need to waste characters.

js/ai/src/document.ts
ifielker: Yes, but then Embedding -> Storage -> Retriever -> Document.
And now where is the embedding metadata? If you want to use it, then it needs to be in the document. Given a document with only the original metadata you have no way of knowing which part of the video to send along in your RAG query except the whole big chunk which negates the usefulness of fine grained chunking.

js/ai/src/embedder.ts
ifielker: done

js/ai/src/document.ts
ifielker: Also, for most of our vector stores, the id that we store the information under is the MD5 hash of the JSOn.stringify(document). So if 5 embeddings point to the exact same document, then we would only store either the first or the last depending on if we overwrite or skip upon finding an entry already in the store. We could change how the id is calculated, but I think that's another breaking change... which seemed unnecessary.

pavelgj: It just occurred to me that when calling `ai.embed` things like video chunking or truncation are not "metadata", but options:

```ts
const embedding = ai.embed({
  embedder: multimodalEmbedding001,
  content: Document.fromMedia('gs://.../', 'video/mp4'),
  options: {
    videoStartEmbeddingSeconds: 30,
    videoEndEmbeddingSeconds: 60,
  }
});
```

metadata should not be used for action parameters.

pavelgj: it's OK to put embedMetadata into metadata if really necessary, but it's all just metadata, it does not require a separate field.

```
metadata: {
  embedMetadata: {
      startEmbeddingSeconds: 30,
      endEmbeddingSeconds: 60,
  }
}
```

docs/plugins/vertex-ai.md
```diff
@@ -143,6 +143,24 @@ const embedding = await ai.embed({
...
+
+const embeddings = await ai.embed({
+  embedder:  multimodalEmbedding001,
+  content: [{
+    media: {
+      contentType: 'video/mp4',
+      url: 'gs://cloud-samples-data/generative-ai/video/pixel8.mp4'
+    }}]
+});
+
```
pavelgj, line 161:
```suggestion
```

js/plugins/chroma/src/index.ts
```diff
@@ -151,7 +152,7 @@ export function chromaRetriever<EmbedderCustomOptions extends z.ZodTypeAny>(
         include: getIncludes(options?.include),
         where: options?.where,
         whereDocument: options?.whereDocument,
-        queryEmbeddings: embedding,
+        queryEmbeddings: embedding[0].embedding,
```
pavelgj, line 155: I'm thinking we should throw if `embedding.length !== 1`

js/plugins/chroma/src/index.ts
```diff
@@ -312,12 +331,13 @@ export async function createChromaCollection<
...
-              content: text,
-              options: params.embedderOptions,
-            })
+          texts.map(
+            (text) =>
+              ai.embed({
+                embedder,
+                content: text,
+                options: params.embedderOptions,
+              })[0].embedding // Text only has a single embedding
```
pavelgj, line 340: also throw here if more embeddings.length !== 1

pavelgj (approved):

js/plugins/chroma/src/index.ts
ifielker: (Follow up PR)

js/plugins/chroma/src/index.ts
ifielker: Followup PR

docs/plugins/vertex-ai.md
ifielker: done

js/ai/src/document.ts
ifielker: done

PR #1673, opened 2025-01-27
refactor(js/ai): moves tool loop logic into its own file, improves readability of generate/action.ts

mbleigh:
This isn't quite ready (still two tests failing that I'm looking into) but this is a beefy refactor mainly of `ai/src/generate/action.ts` into something more manageable. I was finding it almost impossible to follow all the logic being applied so I tried to break it down into smaller pieces and rename things a bit to make it more understandable.


js/ai/src/generate/action.ts
```diff
@@ -52,15 +50,10 @@ import {
...
-import {
-  ToolAction,
-  ToolInterruptError,
-  resolveTools,
-  toToolDefinition,
-} from '../tool.js';
+import { ToolAction, resolveTools, toToolDefinition } from '../tool.js';
+import { assertValidToolNames, resolveToolRequests } from './tool-loop.js';
 
 export const GenerateUtilParamSchema = z.object({
```
pavelgj, line 58: `GenerateUtilParam` is not a very good name, is it? This is a great opportunity to rename this... before it becomes part of the cross-runtime contract (when we expose it to the dev ui).

How about GenerateActionOptions? A nod to GenerateOptions...

mbleigh: Yup, meant to do that actually and forgot. Done!

js/ai/src/generate/tool-loop.ts
pavelgj, line 1: nit: this doesn't really do any "looping", more like tool utils?

mbleigh: Done.

js/ai/src/tool.ts
```diff
@@ -46,10 +46,37 @@ export type ToolAction<
...
+   * restart constructs a tool request corresponding to the provided interrupt tool request
+   * that will then re-trigger the tool after e.g. a user confirms. The `resumedMetadata`
+   * supplied to this method will be passed to the tool to allow for custom handling of
+   * restart logic.
+   *
+   * @param interrupt The interrupt tool request you want to restart.
+   * @param resumedMetadata The metadata you want to provide to the tool to aide in reprocessing. Defaults to `true` if none is supplied.
+   * @param options Additional options for restarting the tool.
+   */
+  restart(
```
pavelgj, line 68: this is not implemented, yet?

js/ai/src/model.ts
```diff
@@ -633,7 +634,8 @@ export interface ResolvedModel<
 
 export async function resolveModel<C extends z.ZodTypeAny = z.ZodTypeAny>(
   registry: Registry,
-  model: ModelArgument<C> | undefined
+  model: ModelArgument<C> | undefined,
+  options?: { warnDeprecated?: boolean }
```
pavelgj, line 638: whence is it set?

js/ai/src/tool.ts
mbleigh: Oops! Yeah missed removing it.

js/ai/src/model.ts
mbleigh: Oops! Refactored things one too many times and forgot to add the option. It's in `generate/action.ts` now.

pavelgj (approved):

PR #1674, opened 2025-01-28
fix(js/plugins/google-cloud)!: rename io to inputAndOutput

MichaelDoyle:
IO generally refers to interfaces between a computer and the outside world. Reading/writing from disk, reading/writing from the network, etc (and also of course peripherals like mice, keyboards, monitors, etc). Renaming this removes ambiguity that we are referring to something different.

This *is* technically a breaking change. But I think its a good one, while we have a chance.

Checklist (if applicable):

### DIFF
--- a/docs/plugins/google-cloud.md
+++ b/docs/plugins/google-cloud.md
@@ -197,7 +197,7 @@ and logs.
 Provides an override that disables exporting traces while still exprting metrics
 and logs.
 
-#### disableLoggingIO
+#### disableLoggingInputAndOutput
 
 Provides an override that disables collecting input and output logs.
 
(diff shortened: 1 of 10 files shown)


docs/plugins/google-cloud.md
```diff
@@ -197,7 +197,7 @@ and logs.
 Provides an override that disables exporting traces while still exprting metrics
 and logs.
 
-#### disableLoggingIO
+#### disableLoggingInputAndOutput
```
pavelgj, line 200: Should really just link to API ref for details like this:
https://genkit-js-api.web.app/interfaces/_genkit-ai_google-cloud.GcpTelemetryConfigOptions.html#disableloggingio
Maybe someday.

js/plugins/google-cloud/tests/logs_no_input_output_test.ts
pavelgj, line 1: Might want to follow breaking change conventional commit message format for this.

pavelgj (approved):

docs/plugins/google-cloud.md
MichaelDoyle: W're going to overhaul these docs w/ David. We'll keep this in mind

js/plugins/google-cloud/tests/logs_no_input_output_test.ts
MichaelDoyle: done

PR #1678, opened 2025-01-28
fix(ai/core): correctly allow plugin initialization from runtime context

pavelgj:
Checklist (if applicable):


js/core/src/plugin.ts
pavelgj, line 1: drive by cleanup

js/core/src/action.ts
```diff
@@ -521,3 +521,13 @@ export function isInRuntimeContext(registry: Registry) {
 export function runInActionRuntimeContext<R>(registry: Registry, fn: () => R) {
   return registry.asyncStore.run(runtimeContextAslKey, 'runtime', fn);
 }
+
+/**
+ * Execute the provided function in the action runtime context.
```
apascal07, line 526: Fix?

apascal07 (approved):

js/core/src/action.ts
pavelgj:
```suggestion
 * Execute the provided function outside the action runtime context.
```

PR #1687, opened 2025-01-29
fix(js/ai/prompt): delete null descriptions returned by dotprompt

pavelgj:
it causes description getting sent to the model as:
```
  "output": {
    "format": "json",
    "contentType": "application/json",
    "constrained": true,
    "schema": {
      "type": "object",
      "properties": {
        "output": {
          "type": "string"
        }
      },
      "required": [
        "output"
      ],
      "additionalProperties": true,
      "$schema": "http://json-schema.org/draft-07/schema#",
      "description": {} // <---------
    }
  }
```

Checklist (if applicable):

### DIFF
--- a/js/ai/src/prompt.ts
+++ b/js/ai/src/prompt.ts
@@ -739,7 +739,7 @@ function loadPrompt(
 
       // dotprompt can set null description on the schema, which can confuse downstream schema consumers
       if (promptMetadata.output?.schema?.description === null) {
-        delete promptMetadata.output?.schema?.description
+        delete promptMetadata.output?.schema?.description;
       }
 
       return {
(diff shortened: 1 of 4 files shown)


js/ai/src/prompt.ts
```diff
@@ -736,6 +736,12 @@ function loadPrompt(
       if (variant) {
         promptMetadata.variant = variant;
       }
+
+      // dotprompt can set null description on the schema, which can confuse downstream schema consumers
+      if (promptMetadata.output?.schema?.description === null) {
+        delete promptMetadata.output?.schema?.description;
+      }
```
MichaelDoyle, line 743: I haven't looked that close, but why does description in particular have this problem? Do other fields also have this problem?

MichaelDoyle (approved):

PR #1690, opened 2025-01-29
fix(js)!: client streamFlow returns properties, not functions

inlined:
Reconciles the discrepancy between the client and server methods for calling an action so that they both return a Promise/AsyncIterable rather than a method returning those types.

Note for reviewer: we decided to call the firebase functions method "onCallGenkit" because we wanted to support invoking any action. Do we similarly want to rename the client SDK to runAction and streamAction?

Checklist (if applicable):
- [] Tested (compiles, but **tests fail on main**)
- [X] Docs updated (Docs already claimed this was the behavior)

### DIFF
--- a/js/genkit/src/client/client.ts
+++ b/js/genkit/src/client/client.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { Channel } from '@genkit-ai/core/async';
+
 const __flowStreamDelimiter = '\n\n';
 
 /**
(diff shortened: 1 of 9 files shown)


js/genkit/src/client/client.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { Channel } from '@genkit-ai/core/async';
```
pavelgj, line 17: this looks fine, but should be careful about pulling things in here -- `genkit/client` package is meant to be browser safe (like being used directly in angular or react client app).

pavelgj (approved):

PR #1693, opened 2025-01-29
feat(js/ai): Adds "restart" for tool interrupts with flexible resuming.

mbleigh:
Implements "restarts" on top of interrupts. Usage:

```ts
const sendText = ai.defineTool({
  name: 'sendText',
  description: 'send a text message',
  inputSchema: z.object({to: z.array(z.string()), message: z.string()}),
}, async (input, {context, interrupt, resumed}) => {
  // if not confirmed
  if (!resumed) interrupt({message: "Does this text look good to send?");
  sendTextMessage(context.auth.uid, input);
});

let response = await ai.generate({
  prompt: "Send a text to mom telling her I'll be late.",
  tools: [sendText],
});

if (interrupts.length) {
  const revisedMessage = await confirmOrReviseMessage(interrupts['0]);
  response = await ai.generate({
    resume: {restart: sendText.restart(interrupts[0], true, {replaceInput: revisedMessage});
  });
}
```

Checklist (if applicable):
- [x] Docs bug: #1657


mbleigh:
This is ready for review, the build errors are unrelated to my code.

js/ai/src/tool.ts
```diff
@@ -55,8 +61,37 @@ export type ToolAction<
...
+   * supplied to this method will be passed to the tool to allow for custom handling of
+   * restart logic.
+   *
+   * @param interrupt The interrupt tool request you want to restart.
+   * @param resumedMetadata The metadata you want to provide to the tool to aide in reprocessing. Defaults to `true` if none is supplied.
+   * @param options Additional options for restarting the tool.
+   */
+  restart(
+    interrupt: ToolRequestPart,
+    resumedMetadata?: any,
```
pavelgj, line 76: restart and resume are used interchangeably... should probably be consistent to avoid confusion. Maybe call it `restartMetadata`? and `restarted`?

mbleigh: I made the tool API `(input, {interrupt, resumed})` because `(input, {interrupt, restarted})` doesn't really feel as nice...but maybe it's fine?

Another option could be going back to the original `confirm` and make the argument `confirmed`. (Then we could have resume + complete + confirm?)

I think restart may still be the most accurate tho...

pavelgj (approved):

PR #1699, opened 2025-01-29
fix(js/ai/prompt): lazy load dotprompt files to avoid initialization race conditions

pavelgj:
Checklist (if applicable):

### DIFF
--- a/js/core/src/async.ts
+++ b/js/core/src/async.ts
@@ -89,7 +89,7 @@ export class Channel<T> implements AsyncIterable<T> {
 }
 
 /**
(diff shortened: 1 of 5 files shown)


js/core/src/async.ts
```diff
@@ -87,3 +87,29 @@ export class Channel<T> implements AsyncIterable<T> {
     };
   }
 }
+
+/**
+ * A lazy promise that does not run its executor function until then, catch or finally is called.
```
pavelgj, line 92:
```suggestion
 * A lazy promise that does not run its executor function until then is called.
```

apascal07 (approved):

PR #1724, opened 2025-01-30
fix: use server config to manage `telemetryServerUrl`

ssbushi:
- A bug causes the telemetry server started by 'genkit start' to be overridden by the ephemeral telemetry server of CLI commands like 'genkit flow:run'.
- This results in the runtime being stuck with the CLI's telemetry server URL, causing subsequent writes to fail.
- Solution in this PR is to use the `.genkit/servers` directory to track `telemetryServerUrl`

Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/docs/rag.md
+++ b/docs/rag.md
@@ -226,12 +226,12 @@ export const indexMenu = ai.defineFlow(
     filePath = path.resolve(filePath);
 
     // Read the pdf.
-    const pdfTxt = await run('extract-text', () =>
+    const pdfTxt = await ai.run('extract-text', () =>
       extractTextFromPdf(filePath)
     );
 
     // Divide the pdf text into segments.
-    const chunks = await run('chunk-it', async () =>
+    const chunks = await ai.run('chunk-it', async () =>
       chunk(pdfTxt, chunkingConfig)
     );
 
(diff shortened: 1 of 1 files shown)


ssbushi:
One thing not covered here: Should there be any cleanup of the config file? If so, where should that happen?

apascal07 (approved):

genkit-tools/cli/src/utils/manager-utils.ts
```diff
@@ -85,3 +96,61 @@ export async function runWithManager(
...
   }
 }
+
+/**
+ * Fetches an existing Telemetry Server URL from servers config. Otherwise,
+ * creates a new server and updates the server config.
+ *
+ * @returns telemetry server url
+ */
+async function getOrStartTelemetryServer(): Promise<string> {
```
pavelgj, line 106: I think this needs to be conditional,no?

pavelgj: I mean, what if I do `genkit start --` twice in the same folder? the second one won't start the telemetry server, but if you shut down the first the runtime will be without telemetry server

pavelgj: The approach we discussed: when doing `genkit start --` set GENKIT_TELEMETRY_SERVER_URL env var runtime ignores notify calls... don't like that?

ssbushi: Not that I don't like it, @apascal07 suggested using `resolveTelemetryServer` to check if there is a running server. 

Also, doesn't the env var approach also face the same problem as your second comment? Killing the first `genkit start` will kill the telemetry server but the second one is left without one.

pavelgj: No. With env var the runtime would be permanently bound to the `genkit start` cli instance that spawned it. It will always have the telemetry server.

PR #1732, opened 2025-01-31
refactor(js/ai/chat): make sendStream sync

pavelgj:
```ts
    const { response, stream } = chat.sendStream('hi');
```

Checklist (if applicable):


mbleigh (approved):
It took me a while to understand how this still works in the new formulation, but it's so much simpler! Sleek, I like it.

PR #1776, opened 2025-02-01
fix: use env var to track telemetryServerUrl

ssbushi:
- A bug causes the telemetry server started by 'genkit start' to be overridden by the ephemeral telemetry server of CLI commands like 'genkit flow:run'.
- This results in the runtime being stuck with the CLI's telemetry server URL, causing subsequent writes to fail.
- Solution in this PR is to use the `GENKIT_TELEMETRY_SERVER` env var to track `telemetryServerUrl`

Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/genkit-tools/cli/src/commands/start.ts
+++ b/genkit-tools/cli/src/commands/start.ts
@@ -35,35 +35,6 @@ export const start = new Command('start')
   .option('-p, --port <port>', 'port for the Dev UI')
   .option('-o, --open', 'Open the browser on UI start up')
   .action(async (options: RunOptions) => {
-    let runtimePromise = Promise.resolve();
-    if (start.args.length > 0) {
-      runtimePromise = new Promise((urlResolver, reject) => {
-        const appProcess = spawn(start.args[0], start.args.slice(1), {
-          env: { ...process.env, GENKIT_ENV: 'dev' },
-          shell: process.platform === 'win32',
-        });
-
-        const originalStdIn = process.stdin;
-        appProcess.stderr?.pipe(process.stderr);
-        appProcess.stdout?.pipe(process.stdout);
-        process.stdin?.pipe(appProcess.stdin);
-
-        appProcess.on('error', (error): void => {
-          console.log(`Error in app process: ${error}`);
-          reject(error);
-          process.exitCode = 1;
-        });
-        appProcess.on('exit', (code) => {
-          process.stdin?.pipe(originalStdIn);
-          if (code === 0) {
-            urlResolver(undefined);
-          } else {
-            reject(new Error(`app process exited with code ${code}`));
-          }
-        });
-      });
-    }
-
     let uiPromise = Promise.resolve();
     if (!options.noui) {
       let port: number;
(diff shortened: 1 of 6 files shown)


js/core/src/reflection.ts
```diff
@@ -226,10 +226,9 @@ export class ReflectionServer {
 
     server.post('/api/notify', async (request, response) => {
       const { telemetryServerUrl, reflectionApiSpecVersion } = request.body;
-      if (typeof telemetryServerUrl === 'string') {
-        setTelemetryServerUrl(telemetryServerUrl);
-        logger.debug(`Connected to telemetry server on ${telemetryServerUrl}`);
-      }
+      const url = process.env.GENKIT_TELEMETRY_SERVER ?? telemetryServerUrl;
```
pavelgj, line 229: Can you please also add

```ts
if (process.env.GENKIT_TELEMETRY_SERVER) {
  setTelemetryServer(process.env.GENKIT_TELEMETRY_SERVER)
}
```
to https://github.com/firebase/genkit/blob/7cb0116598b338c4e43a8c3136fadc4cf48a39da/js/core/src/tracing.ts#L51

pavelgj (approved):

genkit-tools/cli/src/commands/start.ts
```diff
@@ -76,12 +47,54 @@ export const start = new Command('start')
       } else {
         port = await getPort({ port: makeRange(4000, 4099) });
       }
-      uiPromise = startManager(true).then((manager) =>
-        startServer(manager, port)
-      );
+      uiPromise = startManager(true)
```
pavelgj, line 50: Oh right, noui case needs to be handled....

genkit-tools/cli/src/commands/start.ts
```diff
@@ -76,12 +48,53 @@ export const start = new Command('start')
       } else {
         port = await getPort({ port: makeRange(4000, 4099) });
       }
-      uiPromise = startManager(true).then((manager) =>
-        startServer(manager, port)
-      );
+      uiPromise = startManager(true).then((manager) => {
```
pavelgj, line 51: wait... you still need the manager in noui mode?

ssbushi: This is in ui mode

https://github.com/firebase/genkit/blob/bad32552acd65b344334c2dde59f7b62693c961b/genkit-tools/cli/src/commands/start.ts#L40

pavelgj: Yes. We need the manager in both modes.

js/core/src/reflection.ts
apascal07, line 229: Kinda funky logic here, we should just return from this method early if the env var is set. No reason to call `setTelemetryServerUrl` every time that notify gets hit in the env var scenario.

ssbushi: The original reason why I had this in is because the app needs at least one `setTelemetryServerUrl` to properly configure TS. In light of @pavelgj's earlier comment to add ta call to tracing.ts, the original reason no longer holds, is that correct?

apascal07: Right! Better to handle the env var scenario there.

js/core/src/reflection.ts
```diff
@@ -225,25 +225,25 @@ export class ReflectionServer {
...
-          logger.warn(
-            'WARNING: Genkit CLI version may be outdated. Please update `genkit-cli` to the latest version.'
-          );
-        } else {
-          logger.warn(
-            'Genkit CLI is newer than runtime library. Some feature may not be supported. ' +
-              'Consider upgrading your runtime library version (debug info: expected ' +
-              `${GENKIT_REFLECTION_API_SPEC_VERSION}, got ${reflectionApiSpecVersion}).`
-          );
+        if (reflectionApiSpecVersion !== GENKIT_REFLECTION_API_SPEC_VERSION) {
```
apascal07, line 232: Let's maybe keep this warning anyway?

apascal07 (approved):

js/core/src/reflection.ts
```diff
@@ -226,9 +226,13 @@ export class ReflectionServer {
 
     server.post('/api/notify', async (request, response) => {
       const { telemetryServerUrl, reflectionApiSpecVersion } = request.body;
-      if (typeof telemetryServerUrl === 'string') {
-        setTelemetryServerUrl(telemetryServerUrl);
-        logger.debug(`Connected to telemetry server on ${telemetryServerUrl}`);
+      if (!process.env.GENKIT_TELEMETRY_SERVER) {
+        if (typeof telemetryServerUrl === 'string') {
```
pavelgj, line 230: Why this check?

ssbushi: I was preserving the original check. I reckon it is a type check

pavelgj (approved):

PR #1796, opened 2025-02-03
chore: update prompts to 1.0, renames to js-prompts to match samples

i2amsam:
Updates `js-prompts` to `1.0` and adds it to the dropdown of samples in IDX.  Also renames from `prompts` to match the other samples 

Checklist (if applicable):


samples/js-prompts/src/index.ts
```diff
@@ -77,13 +78,19 @@ const threeGreetingsPrompt = ai.definePrompt(
...
 // Start a flow server, which exposes your flows as HTTP endpoints. This call
 // must come last, after all of your plug-in configuration and flow definitions.
 // You can optionally specify a subset of flows to serve, and configure some
 // HTTP server options, but by default, the flow server serves all defined flows.
-ai.startFlowServer({
+startFlowServer({
   flows: [threeGreetings, simpleTemplate, simpleDotprompt, simplePrompt],
 });
+
+// Call one of the flows just to validate everything is hooked up properly
```
pavelgj, line 92: did you mean to leave this in?

pavelgj (approved):

samples/js-prompts/src/index.ts
i2amsam: Yes, it makes the startup experience in IDX a little smoother if we output something from the main entry point

PR #1810, opened 2025-02-03
fix(js): Make flow.run's first parameter optional

inlined:
What it says on the tin. Also caught & fixed a few docs issues.

Checklist (if applicable):
- [X] Tested (unit tested)
- [X] Docs updated (updated docs)

### DIFF
--- a/js/genkit/src/genkit.ts
+++ b/js/genkit/src/genkit.ts
@@ -222,7 +222,7 @@ export class Genkit implements HasRegistry {
     options: DefineModelOptions<CustomOptionsSchema>,
     runner: (
       request: GenerateRequest<CustomOptionsSchema>,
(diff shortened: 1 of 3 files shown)


js/genkit/src/genkit.ts
```diff
@@ -721,7 +721,9 @@ export class Genkit implements HasRegistry {
    * A flow step that executes the provided function. Each run step is recorded separately in the trace.
    *
    * ```ts
-   * ai.defineFlow('hello', async() => {
```
pavelgj, line 724: Technically, this is valid....

pavelgj (approved):

js/genkit/src/genkit.ts
inlined: Reverted

PR #1818, opened 2025-02-04
feat(go): Added `/util/generate` + filled feature gaps in Generate API.

apascal07:
A lot of things happening here because there was a lot busted while trying to add `GenerateAction` and make the parity tests pass:

- [x] Updated Go schema from JS schema.
- [x] Updated parts to support the correct tool request/response schemas.
- [x] Updated `model.Generate()` to execute multiple + parallel tool requests.
- [x] Added support for interrupts in tools (didn't need to be here but... it made it easier to consider the `handleToolRequests()` refactor).
- [x] Added `GenerateAction` under `/util/generate`.
- [x] Loosened the restrictions on tool schemas (can be anything now, including `any`, instead of `map[string]any`).
- [x] Added APIs to control features like max turns, return tool requests, and tool choice (currently unused, plugins need to be updated).

Checklist (if applicable):


go/ai/tools.go
```diff
@@ -37,18 +36,51 @@ type Tool interface {
...
+// InterruptOptions provides configuration for tool interruption.
+type InterruptOptions struct {
+	Metadata map[string]any
 }
 
-// DefineTool defines a tool function.
-func DefineTool[In, Out any](r *registry.Registry, name, description string, fn func(ctx context.Context, input In) (Out, error)) *ToolDef[In, Out] {
+// ToolContext provides context and utility functions for tool execution.
+type ToolContext struct {
+	Context   context.Context
```
pavelgj, line 58: is that... "allowed"??

pavelgj: I mean, is it considered idiomatic in Go?
Reading the code below it definitely seems like a pretty clean approach, so I have no objections. Just checking, don't want to offend Go purists.

apascal07: Yes, apparently there are two idiomatic ways of doing this: 1) helper getter/setters for `context.Context` and 2) wrapping `context.Context` in another struct and passing that in place of it. It just depends. I may end up doing the former later but this is fine for now.

pavelgj (approved):

PR #1820, opened 2025-02-04
Get Python code closer to PEP 8 standard

Irillit:
I focused here on:
- Changed camelCase to snake_case in variables, fields, methods and functions;
- Renamed literals that shadowed built-in functions
- Added docstrings
- Renamed types.py to schemas.py, as types.py shadows a built-in name. I called it schemas, as it contains Pydantic models (we can call them schemas), and models might be confused in this context with ML models.
- Split __init__.py in veeneer folder
- Moved some repeated strings to class constants;
- Some other changes (context managers instead of open/close, list comprehension instead of map, Optional type instead of type | None)

Plans for the future:
- Split vertex.ai plugin from init.py to normal modules;
- Make the structure of py folder more Pythonic
- Separate the constructor of Actions into other methods


py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -0,0 +1,137 @@
...
+        if 'GENKIT_ENV' in os.environ and os.environ['GENKIT_ENV'] == 'dev':
+            cwd = os.getcwd()
+            runtimes_dir = os.path.join(cwd, '.genkit/runtimes')
+            current_datetime = datetime.datetime.now()
+            if not os.path.exists(runtimes_dir):
+                os.makedirs(runtimes_dir)
+            runtime_file_path = os.path.join(
+                runtimes_dir, f'{current_datetime.isoformat()}.json'
+            )
+            with open(runtime_file_path, 'w', encoding='utf-8') as rf:
```
Irillit, line 48: It is a context manager, that has been used instead of try, catch and close here.

py/packages/genkit/src/genkit/core/schemas.py
```diff
@@ -6,11 +6,12 @@
 #   filename:  genkit-schema.json
 #   timestamp: 2025-01-17T17:30:26+00:00
```
Irillit, line 7: This file is auto-generated. I don't have to touch it, but it should be renamed.

py/packages/genkit/src/genkit/core/action.py
```diff
@@ -20,29 +17,32 @@ class ActionResponse(BaseModel):
 
 
 class Action:
+    INPUT_KEY = 'inputSchema'
```
kirgrim, line 20: consider moving these constants to properties:

```
@property
def input_key(self) -> str:
    return 'inputSchema'
```
etc.

py/packages/genkit/src/genkit/core/reflection.py
```diff
@@ -10,9 +12,16 @@
 from genkit.core.registry import Registry
 
 
-def MakeReflectionServer(registry: Registry):
+def make_reflection_server(registry: Registry):
+    """Returns a ReflectionServer class."""
+
     class ReflectionServer(BaseHTTPRequestHandler):
```
kirgrim, line 18: move `ReflectionServer` definition outside of this function scope;
(optionally) You might consider implementing `FactoryMethod` pattern to implement this function as a class method of `ReflectionServer`

py/packages/genkit/src/genkit/core/reflection.py
```diff
@@ -10,9 +12,16 @@
...
+        """Exposes an API for local development."""
+
+        ENCODING = 'utf-8'
+
         def do_GET(self):
```
kirgrim, line 23: lets rename it to be more pythonic, something like `execute_get_request`

py/packages/genkit/src/genkit/core/reflection.py
```diff
@@ -10,9 +12,16 @@
...
+            """Handles GET requests."""
             if self.path == '/api/__health':
```
kirgrim, line 25: this probably is out of the scope for this ticket, but heartbeats are usually implemented separately

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -0,0 +1,143 @@
...
+
+    registry: Registry = Registry()
+
+    def __init__(
+        self,
+        plugins: Optional[List[Plugin]] = None,
+        model: Optional[str] = None,
+    ) -> None:
+        self.model = model
+        if 'GENKIT_ENV' in os.environ and os.environ['GENKIT_ENV'] == 'dev':
```
kirgrim, line 40: move this logic to the `@cached_property`, lets call it `dev_mode_enabled`

kirgrim: init code following dev mode logic can also be moved into separate method called e.g. `init_dev_runtime`

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -0,0 +1,143 @@
...
+                        }
+                    )
+                )
+
+            def delete_runtime_file() -> None:
+                os.remove(runtime_file_path)
+
+            atexit.register(delete_runtime_file)
+
+            threading.Thread(target=self.start_server).start()
```
kirgrim, line 66: lets name this thread and store for this object, so later we might control its lifecycle

py/samples/hello/hello.py
```diff
@@ -2,24 +2,25 @@
...
-from genkit.core.types import Message, TextPart, GenerateRequest
-from genkit.plugins.vertex_ai import vertexAI, gemini
-from genkit.veneer import Genkit
+from genkit.core.schemas import Message, TextPart, GenerateRequest
+from genkit.plugins.vertex_ai import vertex_ai, gemini
+from genkit.veneer.veneer import Genkit
 from pydantic import BaseModel, Field
 
-ai = Genkit(plugins=[vertexAI()], model=gemini('gemini-1.5-flash'))
+ai = Genkit(plugins=[vertex_ai()], model=gemini('gemini-1.5-flash'))
```
kirgrim, line 10: lets not rename plugin names for now, as we discussed that they will probably be moved to classes

kirgrim (changes requested):

py/packages/genkit/src/genkit/core/reflection.py
timere-nemo: @kirgrim 
This is a method defined in the base class BaseHTTPRequestHandler and should not be renamed.

py/packages/genkit/src/genkit/core/action.py
timere-nemo: @kirgrim 
This string is a constant used as self.metadata['inputSchema'], so I’m not sure it needs to be a function property.

py/samples/hello/hello.py
Irillit: Ok, let me remove it, so it won't interfere with my current Vertex AI ticket.

py/packages/genkit/src/genkit/core/reflection.py
Irillit: I thought Pavel wanted to implement it, but he didn't respond to me about that.
Anyway, FactoryMethod is probably required when we want to create something similar, not a one class.

py/packages/genkit/src/genkit/core/reflection.py
Irillit: @kirgrim It won't work if I rename, it is a pre-defined with httpd.

py/packages/genkit/src/genkit/core/reflection.py
Irillit: Let's refactor the structure of that file separately. (reflection.py). It requires more than just moving it out of the function.

py/packages/genkit/src/genkit/veneer/veneer.py
Irillit: I'm not sure what you mean about dev_mode_enabled. It is an env variable for now, and the only thing I might do is to use os.getenv() to avoid this long check.
In JS it is a separate function. I guess the user shouldn't set it programmatically.
```
export function getCurrentEnv(): string {
  return process.env.GENKIT_ENV || 'prod';
}

/**
 * Whether the current environment is `dev`.
 */
export function isDevEnv(): boolean {
  return getCurrentEnv() === 'dev';
}
```

Irillit: I moved a part of constructor to the _init_dev_runtime.

py/packages/genkit/src/genkit/core/action.py
kirgrim: > @kirgrim This string is a constant used as self.metadata['inputSchema'], so I’m not sure it needs to be a function property.

okay, good point

py/packages/genkit/src/genkit/core/reflection.py
kirgrim: okay, we may move this effort outside of this PR

py/packages/genkit/src/genkit/veneer/veneer.py
kirgrim: > I moved a part of constructor to the _init_dev_runtime.

Yeah, currently dev initialisation is controlled by environment variable, but this logic is injected as a part of __init__, my intention was just to logically segregate those

kirgrim (approved):

PR #1827, opened 2025-02-04
fix(js/ai): fixed GenerateStreamResponse output typing

pavelgj:


mbleigh (commented):
1. I think the stream type needs to be a separate generic (in `array` format, stream chunks are `z.infer<O>[]` for example)
2. I'd prefer that the output response be entirely unaware of Zod -- it should be perhaps `GenerateStreamResponse<O = unknown, S = Partial<O>>`

mbleigh:
> 1. I think the stream type needs to be a separate generic (in `array` format, stream chunks are `z.infer<O>[]` for example)
> 2. I'd prefer that the output response be entirely unaware of Zod -- it should be perhaps `GenerateStreamResponse<O = unknown, S = Partial<O>>`. Use `z.infer` when declaring the type from somewhere that already has Zod types.

Thinking on it more, I think it is actually always a partial of the output schema...you might be able to get away with:

```ts
O extends any[] ? Partial<O[number]>[] : Partial<O>
```

PR #1828, opened 2025-02-04
feat(python): Implementation of Ollama plugin

kirgrim:
Added support for the ollama plugin in pythonic version of Genkit based on official [library](https://github.com/ollama/ollama-python)

Usage examples:
 - Added sample in `py/samples/ollama/hello.py`for the basic interaction with the llama model via Genkit

Checklist (if applicable):


py/plugins/ollama/src/genkit/plugins/ollama/types.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from enum import StrEnum
```
Irillit, line 4: I think we shouldn't use `types.py`, as it shadows built-in name. And mypy will detect it.

py/plugins/ollama/src/genkit/plugins/ollama/types.py
```diff
@@ -0,0 +1,28 @@
...
+
+class EmbeddingModelDefinition(BaseModel):
+    name: str
+    dimensions: int
+
+
+class OllamaPluginParams(BaseModel):
+    models: List[ModelDefinition] = Field(default_factory=list)
+    embedders: List[EmbeddingModelDefinition] = Field(default_factory=list)
+    server_address: HttpUrl = Field(default=HttpUrl("http://127.0.0.1:11434"))
```
Irillit, line 27: I would move it to some constant to make it easier to read. But it is up to you, it's not a mistake to do it as you do.

py/samples/ollama/pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+  "Programming Language :: Python :: 3 :: Only",
+  "Topic :: Scientific/Engineering :: Artificial Intelligence",
+  "Topic :: Software Development :: Libraries",
+]
+dependencies = [
+  "genkit",
+  "genkit-firebase-plugin",
+  "genkit-google-ai-plugin",
+  "genkit-google-cloud-plugin",
+  "genkit-ollama-plugin",
```
Irillit, line 22: You should add Ollama dependency itself in Python, or mention how to use it in README.md.

kirgrim: > You should add Ollama dependency itself in Python, or mention how to use it in README.md.

Done.

py/plugins/ollama/src/genkit/plugins/ollama/types.py
kirgrim: > I would move it to some constant to make it easier to read. But it is up to you, it's not a mistake to do it as you do.

Good point, done

py/plugins/ollama/src/genkit/plugins/ollama/types.py
kirgrim: > I think we shouldn't use `types.py`, as it shadows built-in name. And mypy will detect it.

renamed to `models.py`

py/samples/ollama/README.md
```diff
@@ -0,0 +1,18 @@
+# Hello world
+
```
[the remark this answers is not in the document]
kirgrim: > As you use a locally installed Ollama, please, describe how to set it up.

done

py/plugins/ollama/src/genkit/plugins/ollama/plugin_api.py
```diff
@@ -0,0 +1,43 @@
...
+from genkit.plugins.ollama.utils import (
+    register_ollama_embedder,
+    register_ollama_model,
+)
+from genkit.veneer import Genkit
+
+LOG = logging.getLogger(__name__)
+
+
+def Ollama(plugin_params: OllamaPluginParams) -> Callable[[Genkit], None]:
```
Irillit, line 23: Should it be in init?

kirgrim:
@Irillit @pavelgj please take a second look, if its fine - lets merge it

py/plugins/ollama/src/genkit/plugins/ollama/plugin_api.py
pavelgj, line 23: just thinking here -- in JS we have a helper function for defining plugins
https://github.com/firebase/genkit/blob/ee4afba9ee1d7cd0d888829cfd3b312bd27fc826/js/genkit/src/plugin.ts#L31

wondering it's worth doing something similar.
The benefit of have it is that it can encapsulate genkit level plugin handling logic, like if the plugin will need to be initialized lazily, etc. things like that can be tweaked later, perhaps avoiding breaking changes to the plugins.

py/samples/ollama/hello.py
```diff
@@ -0,0 +1,57 @@
...
+    ModelDefinition,
+    OllamaAPITypes,
+)
+from genkit.plugins.ollama import Ollama
+
+from genkit.veneer import Genkit
+
+
+# model can be pulled with `ollama pull *LLM_VERSION*`
+LLM_VERSION = 'llama3.2:latest'
```
pavelgj, line 17: :caugh: gemma2 :caugh: :)

pavelgj (approved):

py/plugins/ollama/src/genkit/plugins/ollama/plugin_api.py
kirgrim: > just thinking here -- in JS we have a helper function for defining plugins
> 
> https://github.com/firebase/genkit/blob/ee4afba9ee1d7cd0d888829cfd3b312bd27fc826/js/genkit/src/plugin.ts#L31
> 
> wondering it's worth doing something similar. The benefit of have it is that it can encapsulate genkit level plugin handling logic, like if the plugin will need to be initialized lazily, etc. things like that can be tweaked later, perhaps avoiding breaking changes to the plugins.

Yeah, we also though of making a separate interface for the plugin, so gonna handle it first, and then refactor this PR

py/plugins/ollama/src/genkit/plugins/ollama/models.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class AsyncOllamaModel:
+    def __init__(
+        self, client: ollama_api.AsyncClient, model_definition: ModelDefinition
+    ):
+        self.client = client
+        self.model_definition = model_definition
+
+    async def generate(self, request: GenerateRequest) -> GenerateResponse:
```
pavelgj, line 111: FYI: you can now implement streaming here as well. See: https://github.com/firebase/genkit/pull/2068

py/plugins/ollama/src/genkit/plugins/ollama/models.py
```diff
@@ -0,0 +1,163 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+import logging
+from typing import Dict, List, Optional
```
Irillit, line 4: Let's use it as we do after we've introduced ruff. All these annotations could be done without importing typing library.

py/plugins/ollama/src/genkit/plugins/ollama/models.py
kirgrim: > #2068

I think its not only this part of code that should be refactored to support streaming - we need to define generic interface to store on_chunk callback for action in registry

py/samples/ollama/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
Irillit, line 1: I checked it here, and I wonder, if you put here a correct license, as they seem to have a MIT license.
https://github.com/ollama/ollama/blob/main/LICENSE

py/plugins/ollama/tests/test_plugin_api.py
```diff
@@ -0,0 +1,97 @@
...
+async def test_async_get_chat_model_response_from_llama_api_flow(
+    mock_ollama_api_async_client: mock.Mock, genkit_veneer_chat_model: Genkit
+):
+    mock_response_message = 'Mocked response message'
+
+    async def fake_chat_response(*args, **kwargs):
+        return ollama_api.ChatResponse(
+            message=ollama_api.Message(
+                content=mock_response_message,
+                role=Role.USER.value,
```
Irillit, line 40: If it is a StrEnum, you probably don't need a value here.

py/plugins/ollama/src/genkit/plugins/ollama/models.py
Irillit: @kirgrim I believe you can create a separate PR for that.

py/samples/ollama/LICENSE
Irillit: @yesudeep Could you also check it?

py/plugins/ollama/src/genkit/plugins/ollama/models.py
kirgrim: no worries, implemented that

py/packages/genkit/src/genkit/core/action.py
```diff
@@ -137,7 +137,7 @@ def __init__(
             on_chunk: The callback to invoke when a chunk is received.
             context: The context to pass to the action.
         """
-        self.__on_chunk = (
+        self.on_chunk = (
```
pavelgj, line 140: instead of making on_chunk public, let's add a new field `is_streaming`?

kirgrim: > instead of making `on_chunk` public, let's add a new field `is_streaming`?

good idea, done

Irillit (approved):

py/plugins/ollama/src/genkit/plugins/ollama/mixins.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+
+from genkit.core.action import ActionRunContext, noop_streaming_callback
```
Irillit, line 6: Do you use noop_streaming_callback here?

py/plugins/ollama/src/genkit/plugins/ollama/models.py
pavelgj: LGTM

pavelgj (approved):

py/plugins/ollama/src/genkit/plugins/ollama/mixins.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Common helpers extracted into a base class or module
+from genkit.core.typing import GenerateRequest, GenerationCommonConfig, TextPart
+
+import ollama as ollama_api
+
+LOG = logging.getLogger(__name__)
+
+
+class BaseOllamaModelMixin:
```
kamilkorski, line 16: Is this really a mixin or just a base class?

PR #1840, opened 2025-02-04
chore: update github workflows to use node v21

tonybaroneee:
_consistency is key_


pavelgj:
actually, I just switched them back to node 20 because we were missing some node 20 specific bugs in presubmits and 20 is out min supported version....

tonybaroneee:
> actually, I just switched them back to node 20 because we were missing some node 20 specific bugs in presubmits and 20 is our min supported version....

kk updated PR

pavelgj (approved):

PR #1845, opened 2025-02-05
fix:(go/core/tracing) doing enhancements for unit test in span data

falonso81:
 - [linking to an issue that includes more details](https://github.com/firebase/genkit/issues/225)
 - add unit test for https://github.com/firebase/genkit/issues/225

Checklist (if applicable):

### DIFF
--- a/go/core/tracing/store_test.go
+++ b/go/core/tracing/store_test.go
@@ -1,13 +1,14 @@
 // Copyright 2024 Google LLC
 // SPDX-License-Identifier: Apache-2.0
 
-
 package tracing
 
 import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


go/core/tracing/store_test.go
```diff
@@ -52,4 +53,105 @@ func TestTraceJSON(t *testing.T) {
...
+	ensureAllSpanDataFieldsArePresent(t, td, goMap)
+}
+
+// ensureAllSpanDataFieldsArePresent checks that every non-zero field in SpanData
+// (by its json tag) is present in the marshaled JSON. Fields with `,omitempty`
+// that are zero/empty are allowed to be omitted.
+func ensureAllSpanDataFieldsArePresent(t *testing.T, data Data, jsonMap map[string]any) {
+	spans, ok := jsonMap["spans"].(map[string]any)
+	if !ok {
+		t.Errorf("expected 'spans' to be a map, but got %T", jsonMap["spans"])
```
hugoaguirre, line 65: Instead of doing `t.Errorf(); return` you can instead call `t.Fatalf()`

go/core/tracing/store_test.go
```diff
@@ -52,4 +53,105 @@ func TestTraceJSON(t *testing.T) {
 	if diff := cmp.Diff(jsMap, goMap); diff != "" {
 		t.Errorf("mismatch (-want, +got):\n%s", diff)
 	}
+	ensureAllSpanDataFieldsArePresent(t, td, goMap)
+}
+
+// ensureAllSpanDataFieldsArePresent checks that every non-zero field in SpanData
+// (by its json tag) is present in the marshaled JSON. Fields with `,omitempty`
+// that are zero/empty are allowed to be omitted.
+func ensureAllSpanDataFieldsArePresent(t *testing.T, data Data, jsonMap map[string]any) {
```
hugoaguirre, line 62: I think you can get rid of ` it. *testing.T` param and just return an error if something goes wrong. This way we can check for errors in the caller function.

hugoaguirre (changes requested):

go/core/tracing/store_test.go
alonsopec89: I did the change, but we get rid of the t testing

go/core/tracing/store_test.go
alonsopec89: you right, I think now the code is better

hugoaguirre (approved):

apascal07:
I'm not sure how much value `ensureAllSpanDataFieldsArePresent` is adding here. We're essentially testing whether the JSON marshaling logic is working -- this should be well tested by Go core. The existing test reads the same `trace.json` file and compares the wire formats of both. What we want additionally (to address the linked issue) is a test of specific trace span labels -- something that we expect to be consistent between runtimes. I would look at a freshly generated trace span from JS and make sure that we are applying the same labels in Go and add a test for that (e.g. like the issue mentions, `genkit:metadata:subtype`).

PR #1849, opened 2025-02-05
feat(js/client): added defineRemoteAction remote action client

pavelgj:
Checklist (if applicable):

### DIFF
--- a/js/genkit/src/client/client.ts
+++ b/js/genkit/src/client/client.ts
@@ -15,6 +15,66 @@
  */
 
 import { Channel } from '@genkit-ai/core/async';
+import { z } from '@genkit-ai/core/zod';
+
+interface RunOptions {
+  headers?: Record<string, string>;
+}
+
+export type RemoteAction<I = any, O = any, S = any> = ((
+  input?: I,
+  options?: RunOptions
+) => Promise<O>) & {
+  stream(input?: I, opts?: RunOptions): StreamingResponse<O, S>;
+};
+
+/**
+ * Streaming response from an action.
+ */
+export interface StreamingResponse<O = any, S = any> {
+  /** Iterator over the streaming chunks. */
+  stream: AsyncIterable<S>;
+  /** Final output of the action. */
+  output: Promise<O>;
+}
+
+/**
+ * Defines a remote action which can be invoked as a function or streamed.
+ */
+export function defineRemoteAction<
+  I extends z.ZodTypeAny = z.ZodTypeAny,
+  O extends z.ZodTypeAny = z.ZodTypeAny,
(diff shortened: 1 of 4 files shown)


js/genkit/src/client/client.ts
```diff
@@ -15,6 +15,66 @@
...
+  output: Promise<O>;
+}
+
+/**
+ * Defines a remote action which can be invoked as a function or streamed.
+ */
+export function defineRemoteAction<
+  I extends z.ZodTypeAny = z.ZodTypeAny,
+  O extends z.ZodTypeAny = z.ZodTypeAny,
+  S extends z.ZodTypeAny = z.ZodTypeAny,
```
mbleigh, line 47: Ugh, not sure what to do about the generics here. We're not actually doing validation here so in some ways I wonder about omitting the schemas for now and having the generics just be the realized types. You shouldn't have to import the zod schema and all of zod just to have the correct types on your remote action.

js/genkit/src/client/client.ts
```diff
@@ -15,6 +15,66 @@
...
+  /** Iterator over the streaming chunks. */
+  stream: AsyncIterable<S>;
+  /** Final output of the action. */
+  output: Promise<O>;
+}
+
+/**
+ * Defines a remote action which can be invoked as a function or streamed.
+ */
+export function defineRemoteAction<
```
mbleigh, line 44: should it just be `remoteAction` since there's no registry we're defining it into? but on the other hand `define` feels more consistent with the server side sdk..

js/genkit/src/client/index.ts
mbleigh, line 1: We should consider exporting z from this file as well - `import {z} from "genkit"` will blow up Next.js if you do it from a file that runs in the browser

js/genkit/src/client/client.ts
pavelgj: I'm super torn on this... my thinking was that if you share zod schema between the client and server, you can reused them here... but not sure how nice it is ergonomically... ideally you should be able to just take the flow (or action) reference and infer all the typing from it...

```ts
remoteAction({
  action: myFlow,
  url: 'https://...'
});
```

mbleigh: Yeah but right now that's not very feasible because trying to import `myFlow` into a browser environment will blow up the world. Although maybe with a pure type import...

```ts
import type { someFlow } from "./flows/someFlow";

remoteAction<typeof someFlow>();
```

pavelgj: @inlined managed to make this work in https://www.npmjs.com/package/@genkit-ai/next

Issue #1852, opened 2025-02-05
[Chore] Pin the version of node used with the project


yesudeep:
**Problem statement**
Some of our systems cannot be updated to the latest versions of the operating systems
and have restricted access to updating software. 

**Recommendation**
Pinning the version of node using
PNPM appears to resolve this since PNPM can also manage node versions.

See: https://pnpm.io/npmrc#use-node-version

PR #1853, opened 2025-02-05
chore: pin node version using pnpm #1852

yesudeep:
RATIONALE: Some older systems cannot be updated to use the latest version of system software and nodejs owing to access restrictions.

Pinning the version of node used by the project and allowing PNPM to manage the version of node is a suitable workaround.

ISSUE: https://github.com/firebase/genkit/issues/1852

CHANGELOG:
- [x] Update the .npmrc file to set the version of node used by the project.
- [x] Update the CONTRIBUTING.md file to include information about managed node versions.

Checklist (if applicable):

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -14,3 +14,7 @@ package-manager-strict-version=true
 # pnpm specified in the `packageManager` field of package.json. This is the same
 # field used by Corepack.
 manage-package-manager-versions=true
+
+# Allow PNPM to manage node versions independently of the operating system
+# installed version of node.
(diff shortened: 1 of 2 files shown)


tonybaroneee:
I'm not sure that we want to be in the business of mandating the node version on the user's machine like this. There are projects like `nvm` that are better suited for this sort of thing. We support all versions of node above v20 and probably wouldn't want to stomp over whatever locally installed version someone wishes to use. In your case, I'd see if you could be granted to exception to use other versions of node from IT.

PR #1856, opened 2025-02-05
chore: update samples readme to use purple IDX button

tonybaroneee:
Consistent with main README, lets the button pop more so it's less easy to miss.

image


i2amsam (approved):
💜💜💜

PR #1861, opened 2025-02-05
update working tree output to be more verbose

i2amsam:
Help debug failures like https://github.com/firebase/genkit/actions/runs/13162235661/job/36733626823


tonybaroneee (approved):
btw, that sort of error is usually easily producible by running `pnpm i` locally and checking for any file changes

PR #1863, opened 2025-02-05
fix: include updated pnpm files

i2amsam:
Fixes e2e test failure introduced in https://github.com/firebase/genkit/pull/1839


tonybaroneee (approved):
This is weird, I'll read up more about what this link: syntax is that it thinks it should have.

PR #1869, opened 2025-02-05
fix(js/plugins/google-cloud): Make metric export time intervals mutually exclusive

schnecle:
*The short of it:*

Modify the startTime of the datapoints so that we ensure there are no overlapping export intervals.

Add a shutdown hook to our metrics wrapper ensures we are awaiting the callback function from the wrapped cloud exporter which delegates to an async function with a callback.

*The long of it:*

We were seeing occasions where exported metrics are being overwritten. This is because Cloud does not currently support the delta metric kind for custom metrics. Instead, it will convert delta metrics into "pseudo delta" metrics by calling them cumulative for the time interval provided. Where we run into trouble is that the PeriodicMetricExporterReader creates a time interval based on the last export, setting the start time of the subsequent export to the end time of the previous export. The overlap in time period means that the last in wins and is counted as the cumulative amount, resulting in what appear to be dropped metrics or sometimes, even more confusingly, metrics that increase one minute and decrease the next.

Manually tested locally using `forceDevExport = true` to make sure metrics are still being exported.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
+++ b/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
@@ -198,17 +198,6 @@ export class GcpOpenTelemetry {
           )
         )
       : new InMemoryMetricExporter(AggregationTemporality.DELTA);
-    exporter.selectAggregation = (instrumentType: InstrumentType) => {
-      if (instrumentType === InstrumentType.HISTOGRAM) {
-        return new ExponentialHistogramAggregation();
-      }
-      return new DefaultAggregation();
-    };
-    exporter.selectAggregationTemporality = (
-      instrumentType: InstrumentType
-    ) => {
-      return AggregationTemporality.DELTA;
-    };
     return exporter;
   }
 }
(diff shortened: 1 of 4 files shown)


kmandrika (approved):

js/plugins/google-cloud/src/gcpOpenTelemetry.ts
```diff
@@ -218,24 +207,75 @@ export class GcpOpenTelemetry {
...
+   * overlap with previous exports.
+   *
+   * Cloud metrics do not support delta metrics for custom metrics
+   * and will convert any DELTA aggregations to CUMULATIVE ones on
+   * export. There is implicit overlap in the start/end times that
+   * the Metric reader is sending -- the end_time of the previous
+   * export will become the start_time of the current export. This
+   * method adds a thousandth of a second to ensure discrete export
+   * timeframes.
+   */
```
MichaelDoyle, line 261: This comment could benefit from explaining "why" we need to avoid overlapping intervals (i.e. last metric wins for that interval).

schnecle: Added an additional statement about why the overlapping times are problematic.

PR #1885, opened 2025-02-06
fix: graceful handling of model input conversion

ssbushi:
Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/genkit-tools/common/src/eval/validate.ts
+++ b/genkit-tools/common/src/eval/validate.ts
@@ -92,9 +92,24 @@ function validate(
   data: unknown
 ): { valid: boolean; errors?: ErrorDetail[] } {
   const isModelAction = actionRef.startsWith('/model');
-  let input = isModelAction
-    ? getModelInput(data, /* modelConfig= */ undefined)
-    : data;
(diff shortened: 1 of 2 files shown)


genkit-tools/common/src/eval/validate.ts
```diff
@@ -92,9 +92,24 @@ function validate(
...
+  let input;
+  if (isModelAction) {
+    try {
+      input = getModelInput(data, /* modelConfig= */ undefined);
+    } catch (e) {
+      return {
+        valid: false,
+        errors: [
+          {
+            path: '(root)',
```
shrutip90, line 104: Just curious, what is this path like usually? Why are we setting it to (root) here?

shrutip90 (approved):

genkit-tools/common/src/eval/validate.ts
ssbushi: In zod schema validation, if the error is at a sub field down the heirarchy, this path specifies which field it is, (root) is when the overall object is wrong, which is suitable for this check.

PR #1887, opened 2025-02-06
fix: be more specific about Vertex project id in auth

i2amsam:
Helps prevent issues where the auth project and the configured project are different

Checklist (if applicable):


ifielker (approved):

MichaelDoyle:
i'm not sure I understand this change. Are you saying they need to be the same, and this will make the breakage more obvious? 

What was the behavior before, and what is the behavior now?

PR #1911, opened 2025-02-10
fix(go/plugins/googleai): convertTools err propagate and schema type integer

decibelcooper:
* In this CL a bug is fixed that prevented errors from being propagated from convertSchema() through convertTools().
* The above bug hid an issue where the genkit schema type switch was trying to match "int" instead of "integer".  This issue is also corrected.
* Tests updated and corrected


apascal07 (approved):

apascal07:
Thank you!

decibelcooper:
Resolves #1915

PR #1918, opened 2025-02-10
docs: Updated user-facing docs and samples to use Gemini 2.0 and `async function...` syntax

chrisraygill:
Updated user-facing docs and samples to use Gemini 2.0 and `async function...` syntax.

We should use Gemini 2.0 Flash by default instead of Gemini 1.5 Flash, now that it's GA.

Additionally, I've seen that `async function main...` is a lot more common and seemingly better understood than `(async() => ...` - so I'm switching to that by default.

Checklist (if applicable):
- [x] Tested (manually, unit tested, etc.) - It builds.

### DIFF
--- a/docs/chat.md
+++ b/docs/chat.md
@@ -28,16 +28,16 @@ Here is a minimal, console-based, chatbot application:
 
 ```ts
 import { genkit } from "genkit/beta";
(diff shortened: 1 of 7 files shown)


thedmail (approved):
Rubber-stamped, pending edits

thedmail (approved):
Rubber-stamped, pending edits

PR #1934, opened 2025-02-11
VertexAI connection parameters

Irillit:
ProjectId:

1. Passed as a parameter in the code
2. Defined as an environment variable
3.  Defined as an environment variable as a part of Firebase Config
4. Set by gcloud auth application-default login

Location:
1. Passed as a parameter in the code
2. Defined as an environment variable
3.  Default 'us-central1' region

Credentials:
1. Created from a Service Account environment variable
2. Set by gcloud auth application-default login

I don't see the cache (vertexClientFactoryCache in JS version) in initializer of [VertexAI](https://github.com/googleapis/python-aiplatform/blob/1ab43441419e6c484aee9da02a22e0b02b5e8a2b/google/cloud/aiplatform/initializer.py#L135).
I doubt if we need credentials. If we have a GCLOUD_SERVICE_ACCOUNT_CREDS, it might be more logical to provide the service account itself to the VertexAI.


py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/options.py
```diff
@@ -0,0 +1,66 @@
...
+
+@dataclasses.dataclass(slots=True)
+class PluginOptions:
+    project_id: str | None = None
+    location: str | None = None
+    google_auth: auth_credentials.Credentials | None = None
+    models: list[str] = dataclasses.field(default_factory=list)
+
+
+def get_project_from_firebase_config() -> str  | None:
```
Irillit, line 26: Should we get location from Firebase Config? As I got, it might be there as well.

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/options.py
```diff
@@ -0,0 +1,66 @@
...
+        project_id = (os.getenv(const.GCLOUD_PROJECT)
+                      or get_project_from_firebase_config())
+
+    location = (options.location
+                or os.getenv(const.GCLOUD_LOCATION)
+                or const.DEFAULT_REGION)
+
+    credentials = options.google_auth
+
+    sa_env = os.getenv(const.GCLOUD_SERVICE_ACCOUNT_CREDS)
```
Irillit, line 60: VertexAI accepts Service Account directly. Maybe we'd better provide it?
I also expect that GCLOUD_SERVICE_ACCOUNT_CREDS contains the path to the JSON key of the Service Account. Is that so?

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/constants.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+
+"""Constants for VertexAI plugin."""
+GCLOUD_PROJECT = 'GCLOUD_PROJECT'
```
Irillit, line 5: GOOGLE_CLOUD_PROJECT is also reserved by Firebase. Should we read from it?

Irillit:
@pavelgj 
I tried to repeat the order of parameter retrieval from the JS version of Genkit. Could you check if I get it right?

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/constants.py
pavelgj: let's skip GOOGLE_CLOUD_PROJECT for now. can revisit later.

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/options.py
pavelgj: this is vertex plugin, it should not be aware of Firebase config.

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/options.py
pavelgj: GCLOUD_SERVICE_ACCOUNT_CREDS contains JSON blob of the service account credentials. It was added specifically to support environments where it's hard to write the credentials file.

https://github.com/firebase/genkit/blob/90e23213e4bb0c15e453dfa6db20d32d8ad2d9ed/js/plugins/vertexai/src/common/index.ts#L64

py/packages/genkit/src/genkit/ai/model.py
```diff
@@ -2,7 +2,13 @@
...
+from genkit.core.schemas import GenerateRequest, GenerateResponse, ModelInfo
 
-from genkit.core.schemas import GenerateRequest, GenerateResponse
 
 ModelFn = Callable[[GenerateRequest], GenerateResponse]
+
+
+class ModelReference(BaseModel):
+  name: str
+  info: ModelInfo
```
pavelgj, line 14: model reference should also contain the config
https://github.com/firebase/genkit/blob/90e23213e4bb0c15e453dfa6db20d32d8ad2d9ed/js/ai/src/model.ts#L545

pavelgj: no need to add right away, could you please just add a todo?

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/gemini.py
```diff
@@ -0,0 +1,140 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+# Deprecated on 2/15/2025
+SUPPORTED_V1_MODELS = {
+  'gemini-1.0-pro': ModelReference(
+        name='vertexai/gemini-1.0-pro',
+        info=ModelInfo(
+    versions=['gemini-1.0-pro-001', 'gemini-1.0-pro-002'],
```
pavelgj, line 28: let's drop versions, we're thinking of doing that on the JS side as well

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/gemini.py
```diff
@@ -0,0 +1,140 @@
...
+    Role,
+    Supports,
+    TextPart,
+)
+
+LOG = logging.getLogger(__name__)
+
+
+# Deprecated on 2/15/2025
+SUPPORTED_V1_MODELS = {
```
pavelgj, line 24: let's drop 1.0 (already deprecated)

Irillit:
In PR #2014 I did the same with the new plugin API.

PR #1950, opened 2025-02-12
feat(py): Refactored Plugins API to follow generic `Plugin` interface.

kirgrim:
Description
 - Refactored Plugin API to use generic `genkit.core.plugin_abc.Plugin` interface
 - Updated VertexAI implementation to follow the new specification

Checklist (if applicable):

### DIFF
--- a/py/packages/genkit/src/genkit/core/plugin_abc.py
+++ b/py/packages/genkit/src/genkit/core/plugin_abc.py
@@ -1,0 +1,76 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import abc
+import typing
+
+from genkit.core.schemas import GenerateRequest, GenerateResponse
+
+if typing.TYPE_CHECKING:
+    from genkit.veneer import Genkit
+
+
+class Plugin(abc.ABC):
+    """
+    Abstract class defining common interface
+    for the Genkit Plugin implementation
+
+    NOTE: Any plugin defined for the Genkit must inherit from this class
+    """
+
+    @abc.abstractmethod
+    def attach_to_veneer(self, veneer: Genkit) -> None:
+        """
+        Entrypoint for attaching the plugin to the requested Genkit Veneer
+
+        Implementation must be provided for any inheriting plugin.
+
+        Args:
+            veneer: requested `genkit.veneer.Genkit` instance
+
+        Returns:
+            None
+        """
+        pass
+
+    def _add_model_to_veneer(
+        self, veneer: Genkit, name: str, metadata: dict | None = None
+    ) -> None:
+        """
+        Defines plugin's model in the Genkit Registry
+
+        Uses self._model_callback as a generic callback wrapper
+
+        Args:
+            veneer: requested `genkit.veneer.Genkit` instance
+            name: name of the model to attach
+            metadata: metadata information associated
+                      with the provided model (optional)
+
+        Returns:
+            None
+        """
+        if not metadata:
+            metadata = {}
+        veneer.define_model(
+            name=name, fn=self._model_callback, metadata=metadata
+        )
+
+    @abc.abstractmethod
+    def _model_callback(self, request: GenerateRequest) -> GenerateResponse:
+        """
+        Wrapper around any plugin's model callback.
+
+        Is considered an entrypoint for any model's request.
+        Implementation must be provided for any inheriting plugin.
+
+        Args:
+            request: incoming request as generic
+                     `genkit.core.schemas.GenerateRequest` instance
+
+        Returns:
+            Model response represented as generic
+            `genkit.core.schemas.GenerateResponse` instance
+        """
+        pass
--- a/py/packages/genkit/src/genkit/veneer/veneer.py
+++ b/py/packages/genkit/src/genkit/veneer/veneer.py
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 import atexit
 import datetime
 import json
+import logging
 import os
 import threading
 from collections.abc import Callable
@@ -16,11 +17,12 @@ from typing import Any
 from genkit.ai.model import ModelFn
 from genkit.ai.prompt import PromptFn
 from genkit.core.action import Action
+from genkit.core.plugin_abc import Plugin
 from genkit.core.reflection import make_reflection_server
 from genkit.core.registry import Registry
 from genkit.core.schemas import GenerateRequest, GenerateResponse, Message
 
-Plugin = Callable[['Genkit'], None]
+logger = logging.getLogger(__name__)
 
 
 class Genkit:
@@ -66,9 +68,17 @@ class Genkit:
             self.thread = threading.Thread(target=self.start_server)
             self.thread.start()
 
-        if plugins is not None:
+        if not plugins:
+            logger.warning('No plugins provided to Genkit')
+        else:
             for plugin in plugins:
-                plugin(self)
+                if isinstance(plugin, Plugin):
+                    plugin.attach_to_veneer(veneer=self)
+                else:
+                    raise ValueError(
+                        f'Invalid {plugin=} provided to Genkit: '
+                        f'must be of type `genkit.core.plugin_abc.Plugin`'
+                    )
 
     def start_server(self) -> None:
         httpd = HTTPServer(
--- a/py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/__init__.py
+++ b/py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/__init__.py
@@ -6,9 +6,10 @@
 Google Cloud Vertex AI Plugin for Genkit.
 """
 
-from collections.abc import Callable
+from typing import Any
 
 import vertexai
+from genkit.core.plugin_abc import Plugin
 from genkit.core.schemas import (
     GenerateRequest,
     GenerateResponse,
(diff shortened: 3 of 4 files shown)


py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -0,0 +1,60 @@
...
+    def attach_to_veneer(self, veneer: 'Genkit') -> None:
+        """
+        Entrypoint for attaching the plugin to the requested Genkit Veneer
+
+        :param veneer: requested `genkit.veneer.Genkit` instance
+        """
+        pass
+
+    @abc.abstractmethod
+    def add_model_to_veneer(
```
Irillit, line 29: Are you sure it should be abstract method? You have the implementation here.

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import abc
+import typing
+
+from genkit.core.schemas import GenerateRequest, GenerateResponse
+
+if typing.TYPE_CHECKING:
```
kamilkorski, line 9: Why? just import the genkit

kirgrim: > Why? just import the genkit

circular import issue: we import plugin into the genkit veneer for initializaion, here we need it only for type annotation

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class Plugin(abc.ABC):
+    """
+    Abstract class defining common interface
+    for the Genkit Plugin implementation
+    """
+
+    @abc.abstractmethod
+    def attach_to_veneer(self, veneer: 'Genkit') -> None:
```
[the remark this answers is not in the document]
kirgrim: > You are importing Genkit, use it for annotations

I do - enclosing types in quotes is acceptible for annotation, and is acceptable by both linters and IDE inspections
Using type directly will cause "not defined" issues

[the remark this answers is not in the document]
kirgrim: > > You are importing Genkit, use it for annotations
> 
> I do - enclosing types in quotes is acceptible for annotation, and is acceptable by both linters and IDE inspections Using type directly will cause "not defined" issues:

```
    def attach_to_veneer(self, veneer: Genkit) -> None:
                                       ^^^^^^
NameError: name 'Genkit' is not defined
/Users/vladuelka/Library/pnpm/global/5/.pnpm/genkit-cli@0.9.12/node_modules/genkit-cli/dist/commands/start.js:65
                    reject(new Error(`app process exited with code ${code}`));
```

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import abc
+import typing
```
Irillit, line 6: It seems 'typing' is not used anymore.

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -0,0 +1,60 @@
...
+
+        :param veneer: requested `genkit.veneer.Genkit` instance
+        """
+        pass
+
+    def add_model_to_veneer(
+        self, veneer: Genkit, name: str, metadata: dict | None = None
+    ) -> None:
+        """
+        Generic method for defining arbitrary plugin's model
```
Irillit, line 33: Please, change the description. It is no longer generic.

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -66,9 +65,9 @@ def delete_runtime_file() -> None:
             self.thread = threading.Thread(target=self.start_server)
             self.thread.start()
 
-        if plugins is not None:
+        if plugins:
             for plugin in plugins:
-                plugin(self)
+                plugin.attach_to_veneer(veneer=self)
```
Irillit, line 70: It looks like you add all the models here. I wonder, if attach_to_veneer() should accept Genkit here.You can attach it:
1. To registry inside the attach_to_veneer()
2. Just return values. We'll have the dictionary of the SUPPORTED_MODELS with ModelInfo, and, if we need, a Callable.

py/packages/genkit/src/genkit/core/plugin_abc.py
Irillit: Ok, it is still required due to circular reference.

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -66,9 +65,9 @@ def delete_runtime_file() -> None:
             self.thread = threading.Thread(target=self.start_server)
             self.thread.start()
 
-        if plugins is not None:
+        if plugins:
```
Irillit, line 68: I think, there should be an error, if there are no plugins. Genkit becomes useless.

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -16,12 +16,11 @@
...
 from genkit.core.action import Action
+from genkit.core.plugin_abc import Plugin
 from genkit.core.reflection import make_reflection_server
 from genkit.core.registry import Registry
 from genkit.core.schemas import GenerateRequest, GenerateResponse, Message
 
-Plugin = Callable[['Genkit'], None]
-
 
 class Genkit:
```
Irillit, line 25: In __init__ of that class we also have a 'model' paramter, that is added as model field and never used. At the same time attach_to_veneer registers all the models. Do we have a special reason of keeping it? Should we filter it or pass it to some method?

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -0,0 +1,60 @@
...
+                         with the provided model (optional)
+        """
+        if not metadata:
+            metadata = {}
+        veneer.define_model(
+            name=name, fn=self._model_callback, metadata=metadata
+        )
+
+    @abc.abstractmethod
+    def _model_callback(self, request: GenerateRequest) -> GenerateResponse:
```
Irillit, line 50: I wonder, if here we have only one callback. One provider has several models at once.

py/packages/genkit/src/genkit/veneer/veneer.py
kirgrim: > It looks like you add all the models here. I wonder, if attach_to_veneer() should accept Genkit here.You can attach it:
> 
> 1. To registry inside the attach_to_veneer()
> 2. Just return values. We'll have the dictionary of the SUPPORTED_MODELS with ModelInfo, and, if we need, a Callable.

1. do not really support this approach as we have a different ways of defining models/embedders/etc based on plugin
2. it is not only models, also embedders and logic is not always that trivial according to JS implementations

py/packages/genkit/src/genkit/core/plugin_abc.py
kirgrim: > I wonder, if here we have only one callback. One provider has several models at once.

afaik, as a rule, each single callback receives `GenerateRequest` and returns `GenerateResponse` I would prefer not to complicate things and handle any exceptions separately in the future

py/packages/genkit/src/genkit/core/plugin_abc.py
kirgrim: > Please use Google style docstrings: https://google.github.io/styleguide/pyguide.html#381-docstrings

updated according to the guidelines

py/packages/genkit/src/genkit/veneer/veneer.py
kirgrim: > In **init** of that class we also have a 'model' paramter, that is added as model field and never used. At the same time attach_to_veneer registers all the models. Do we have a special reason of keeping it? Should we filter it or pass it to some method?

Not sure, but would not update interface of this class as a part of this PR

Irillit: Ok, I found it later. It seems that it looks for a model.

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -16,11 +17,12 @@
...
+from genkit.core.plugin_abc import Plugin
 from genkit.core.reflection import make_reflection_server
 from genkit.core.registry import Registry
 from genkit.core.schemas import GenerateRequest, GenerateResponse, Message
 
-Plugin = Callable[['Genkit'], None]
+logger = logging.getLogger(__name__)
 
 
 class Genkit:
```
Irillit, line 28: I also heard Kamil proposed to use an interface for Genkit. If we do a separate AsyncGenkit veneer, that means that the interface might be a good idea.

kirgrim: > I also heard Kamil proposed to use an interface for Genkit. If we do a separate AsyncGenkit veneer, that means that the interface might be a good idea.

yeah, it might, but also would keep it as a separate effort from this PR

py/packages/genkit/src/genkit/core/plugin_abc.py
Irillit: Well, look. VertexAI should support the following models (from the JS code):
```
export const SUPPORTED_V1_MODELS = {
  'gemini-1.0-pro': gemini10Pro,
};

export const SUPPORTED_V15_MODELS = {
  'gemini-1.5-pro': gemini15Pro,
  'gemini-1.5-flash': gemini15Flash,
  'gemini-2.0-flash-001': gemini20Flash001,
  'gemini-2.0-flash-lite-preview-02-05': gemini20FlashLitePreview0205,
  'gemini-2.0-pro-exp-02-05': gemini20ProExp0205,
};
```

Let me check, but probably they should all be in a registry in order you want to call them.

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -0,0 +1,76 @@
...
+
+        Args:
+            veneer: requested `genkit.veneer.Genkit` instance
+
+        Returns:
+            None
+        """
+        pass
+
+    def add_model_to_veneer(
```
kamilkorski, line 37: I think it should be a protected method, it isnt used outside.

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -66,9 +68,17 @@ def delete_runtime_file() -> None:
...
+        else:
             for plugin in plugins:
-                plugin(self)
+                if isinstance(plugin, Plugin):
+                    plugin.attach_to_veneer(veneer=self)
+                else:
+                    logger.error(
+                        f'Invalid {plugin=} provided to Genkit: '
+                        f'must be of type `genkit.core.plugin_abc.Plugin`'
+                    )
```
kamilkorski, line 81: Does it need to be that defensive? I dont consider it a big problem but we already have annotations to inform user what kind of interface he should provide.

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -66,9 +68,17 @@ def delete_runtime_file() -> None:
...
+            logger.warning('No plugins provided to Genkit')
+        else:
             for plugin in plugins:
-                plugin(self)
+                if isinstance(plugin, Plugin):
+                    plugin.attach_to_veneer(veneer=self)
+                else:
+                    logger.error(
+                        f'Invalid {plugin=} provided to Genkit: '
+                        f'must be of type `genkit.core.plugin_abc.Plugin`'
```
kamilkorski, line 80: I think we should raise a RuntimeError or ValueError here. The user has passed an object with wrong interface and it is a major mistake. In my opinion we shouldnt allow the application to continue execution as if nothing has happened in this case.

kirgrim: > I think we should raise a RuntimeError or ValueError here. The user has passed an object with wrong interface and it is a major mistake. In my opinion we shouldnt allow the application to continue execution as if nothing has happened in this case.

I did not understand you correctly during discussion: yes, if type is different I have no objections with it raising an exception

py/packages/genkit/src/genkit/veneer/veneer.py
kirgrim: > Does it need to be that defensive? I dont consider it a big problem but we already have annotations to inform user what kind of interface he should provide.

I did not understand you correctly during discussion: yes, if type is different I have no objections with it raising an exception

Irillit (approved):

PR #1962, opened 2025-02-13
feat(js/ai): allow disabling native constrained generation and enabling custom instruction

pavelgj:
```ts
    const { output } = await generate(registry, {
      prompt: 'generate json',
      output: {
        constrained: false,
        instructions: true,
        schema: z.object({
          foo: z.string(),
        }),
      },
    });
```

Checklist (if applicable):

### DIFF
--- a/js/ai/src/formats/types.ts
+++ b/js/ai/src/formats/types.ts
@@ -24,7 +24,7 @@ export type OutputContentTypes = 'application/json' | 'text/plain';
 export interface Formatter<O = unknown, CO = unknown> {
   name: string;
   config: ModelRequest['output'] & {
-    defaultInstruction?: false;
+    defaultInstructions?: false;
   };
   handler: (schema?: JSONSchema) => {
     parseMessage(message: Message): O;
--- a/js/ai/src/formats/types.ts
+++ b/js/ai/src/formats/types.ts
@@ -23,7 +23,9 @@ export type OutputContentTypes = 'application/json' | 'text/plain';
 
 export interface Formatter<O = unknown, CO = unknown> {
   name: string;
-  config: ModelRequest['output'];
+  config: ModelRequest['output'] & {
+    defaultInstruction?: false;
+  };
   handler: (schema?: JSONSchema) => {
     parseMessage(message: Message): O;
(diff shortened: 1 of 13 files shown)


js/ai/src/generate.ts
```diff
@@ -343,16 +351,11 @@ export async function generate<
...
-  const instructions = resolveInstructions(
-    resolvedFormat,
-    resolvedSchema,
-    resolvedOptions?.output?.instructions
-  );
 
   const params: GenerateActionOptions = {
     model: resolvedModel.modelAction.__action.name,
     docs: resolvedOptions.docs,
-    messages: injectInstructions(messages, instructions),
```
pavelgj, line 355: @mbleigh do you remember why injection was done here? It's redundant.... Action handles it all.

js/ai/src/formats/types.ts
```diff
@@ -23,7 +23,9 @@ export type OutputContentTypes = 'application/json' | 'text/plain';
 
 export interface Formatter<O = unknown, CO = unknown> {
   name: string;
-  config: ModelRequest['output'];
+  config: ModelRequest['output'] & {
+    defaultInstruction?: false;
```
mbleigh, line 27:
```suggestion
    defaultInstructions?: false;
```

plural to match the `instructions` option

js/ai/src/generate.ts
```diff
@@ -211,7 +214,12 @@ export async function toGenerateRequest(
   );
 
   const out = {
-    messages: injectInstructions(messages, instructions),
+    messages: shouldInjectFormatInstruction(
```
mbleigh, line 217:
```suggestion
    messages: shouldInjectFormatInstructions(
```

js/ai/src/generate.ts
mbleigh: I don't remember, but I'm slightly worried it's load-bearing 😅

mbleigh (approved):

js/ai/src/generate.ts
pavelgj: I've done a lot of testing... this appears to be redundant.

PR #1981, opened 2025-02-14
docs: remove `authPolicy` section

jhuleatt:
- [`authPolicy` was deprecated in the Cloud Functions for Firebase SDK](https://github.com/firebase/firebase-functions/commit/82c4fc9bb82317161f8c7f37b912c13b27c47a11)

Checklist (if applicable):

### DIFF
--- a/docs/firebase.md
+++ b/docs/firebase.md
@@ -91,31 +91,6 @@ import { onCallGenkit } from 'firebase-functions/https';
 export generatePoem = onCallGenkit(generatePoemFlow);
 ```
 
-### Define an authorization policy
-
-All deployed flows, whether deployed to Firebase or not, should have an
-authorization policy; without one, anyone can invoke your potentially-expensive
-generative AI flows. To define an authorization policy, use the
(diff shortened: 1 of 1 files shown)


docs/firebase.md
```diff
@@ -91,31 +91,6 @@ import { onCallGenkit } from 'firebase-functions/https';
 export generatePoem = onCallGenkit(generatePoemFlow);
 ```
 
-### Define an authorization policy
```
MichaelDoyle, line 94: What should we be telling developers to do instead?

taeold: We are drafting a solution that aligns better with existing Genkit SDK (ContextProvider), and I suggest we omit this section until that solution is fully hashed and released.

taeold (approved):

MichaelDoyle:
The only problem I see is that this section serves as a “warning” not to
deploy unprotected code.

Do developers have a solution authorize their Genkit flows/actions
currently? If so, it should still be documented here?

MichaelDoyle:
Where did we land on this?

PR #1988, opened 2025-02-14
fix(js): test on node versions 20 through 22; 23 is currently broken #1995

yesudeep:
fix(js): test on node versions 20 through 22; 23 is currently broken #1995
    
ISSUE: #1995

CHANGELOG:
- [x] Update biome configuration to use arrow function arg parens always.
- [x] Update github workflows to test on node versions 20 through 22.
- [x] We will add version 23 post fixes.
- [x] Set node 22 to be the default for our eng workstations in the
  setup script for now.
REFERENCES:

The configuration for the formatting is more or less based on the Google
TypeScript formatting guidelines at:

- [x] https://google.github.io/styleguide/tsguide.html#string-literals
- [x] https://google.github.io/styleguide/jsguide.html
- [x] https://google.github.io/styleguide/tsguide.html#arrow-function-bodies
- [x] https://google.github.io/styleguide/tsguide.html#automatic-semicolon-insertion

Checklist (if applicable):


.github/workflows/e2e-tests.yml
```diff
@@ -21,15 +10,18 @@ on:
 
 jobs:
   build:
-    name: Run e2e tests
+    name: Run e2e tests (Node ${{ matrix.node-version }})
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: ['20.x', '21.x', '22.x']
```
mbleigh, line 17: we should test on 20, 22, 23. don't really care about testing on unstable versions except for the one that's about to become stable

mbleigh (commented):
One comment on Node versions to test, but I'll leave to @pavelgj to decide if we want to do this for sure.

MichaelDoyle:
Maybe you all already discussed, and I am an outlier here... but... can we separate out the toolchain changes from the style changes into separate PRs?

Some of the changes in formatting seem good. But others seem superfluous, and some even go against the style guide. Particularly, removing parens around all of the arrow functions goes against:

> Parentheses around the left-hand side of a single-argument arrow function are recommended but not required.

And is a huge part of the overall diff here.

pavelgj:
> Maybe you all already discussed, and I am an outlier here... but... can we separate out the toolchain changes from the style changes into separate PRs?
> 
> Some of the changes in formatting seem good. But others seem superfluous, and some even go against the style guide. Particularly, removing parens around all of the arrow functions goes against:
> 
> > Parentheses around the left-hand side of a single-argument arrow function are recommended but not required.
> 
> And is a huge part of the overall diff here.

I'm a little caught by surprise here....
I'm personally pretty happy with prettier and not sure we need to change it.

Issue #1705, opened 2025-01-30
[Py] Implement reflection API server using HTTP via ASGI


yesudeep:

Irillit:
During the internal discussion we decided not to change the Pavel's choice of technology, as the server is required only for local environment.

PR #1989, opened 2025-02-14
[WIP] feat(py/reflection-api): use starlette to implement the reflection api #1705

yesudeep:
Faster server response, logging, middeware, preamble, etc.

ISSUE: https://github.com/firebase/genkit/issues/1705

CHANGELOG:
- [x] Switch the reflection API server to use starlette.
- [ ] Use multiprocessing to start background jobs
- [ ] Log asynchronously in a structured format using structlog
- [ ] Add tests for the reflection API server
- [ ] Test against the genkit CLI runtime.
- [ ] Update schema to match.
- [ ] Make the server configurable.

Checklist (if applicable):

### DIFF
--- a/py/bin/sanitize_schemas.py
+++ b/py/bin/sanitize_schemas.py
@@ -82,6 +82,39 @@ class ModelConfigRemover(ast.NodeTransformer):
                         return True
         return False
 
+    def create_model_config(
+        self, extra_forbid: bool = True, populate_by_name: bool = True
+    ) -> ast.Assign:
+        """Create a model_config assignment with the specified options."""
+        keywords = []
+        if extra_forbid:
+            keywords.append(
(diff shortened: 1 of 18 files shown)


Irillit:
We had a discussion about Starlette and FastAPI, but nobody was sure. But what's the reason you've chosen Starlette?After all, FastAPI is Stalette + Pydantic.

py/bin/sanitize_schemas.py
```diff
@@ -24,11 +24,43 @@ def is_rootmodel_class(self, node: ast.ClassDef) -> bool:
                     return True
         return False
 
+    def create_model_config(
```
Irillit, line 27: Could you please check if that code is redundant with FastAPI.

PR #2015, opened 2025-02-18
feat(py): added support for action execution context

kirgrim:
Description:
 - Added support for action execution context, to make action's callback functions more stateful and flexible

Checklist (if applicable):


py/packages/genkit/src/genkit/core/action.py
```diff
@@ -66,6 +72,7 @@ def __init__(
         description: str | None = None,
         metadata: dict[ActionMetadataKey, Any] | None = None,
         span_metadata: dict[str, str] | None = None,
+        fn_context: ActionExecutionContext | None = None,
```
Irillit, line 75: Could you rename this parameter? It sounds like a Callable, some kind of decorator.

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/models/__init__.py
```diff
@@ -6,9 +6,18 @@
 Google Cloud Vertex AI Models for Genkit.
 """
 
+from genkit.core.action import ActionExecutionContext
```
Irillit, line 9: Could you avoid adding it to plugin? I'll need to merge it with very different code in my branch. It's easier to repeat.

py/packages/genkit/src/genkit/core/action.py
```diff
@@ -45,6 +45,12 @@ class ActionMetadataKey(str, Enum):
     RETURN = 'return'
 
 
+class ActionExecutionContext(BaseModel):
+    """The context to the action callback."""
+
+    model_config = ConfigDict(extra='forbid', populate_by_name=True)
```
kamilkorski, line 51: If we are adding an empty class, with another empty class that derives then it is really hard to understand why we are actually introducing that.

timere-nemo:
@kirgrim 
I have a general question about the purpose of this context and how it will be used. The ActionExecutionContext includes a general model_config field, which is present in almost every schema. Additionally, it is unclear why certain data needs to be stored within the action itself rather than within the model. If you could provide at least one use case or example, it would help clarify this

kirgrim:
> @kirgrim I have a general question about the purpose of this context and how it will be used. The ActionExecutionContext includes a general model_config field, which is present in almost every schema. Additionally, it is unclear why certain data needs to be stored within the action itself rather than within the model. If you could provide at least one use case or example, it would help clarify this

reason is that `Action` is a more generic container which could store entities of different kinds, such as Model. There is no way to persist the model itself inside the registry at the moment, without using Action as a container.

Use case is for Model callback when we need to specify, for instance, output formatting based on version of the passed model data

This will be useful for Ollama and other plugins who after refactor cannot just access the global creation context

PR #2019, opened 2025-02-18
docs: Add link to RAGAS metrics page

ssbushi:
Checklist (if applicable):

### DIFF
--- a/docs/evaluation.md
+++ b/docs/evaluation.md
@@ -252,7 +252,7 @@ given prompt
 *   Maliciousness -- Measures whether the generated output intends to deceive,
 harm, or exploit
 
-For more details on how each specific metric is computed, please refer to the 
+For more details on how each metric is computed, please refer to the 
 [RAGAS metrics documentation](https://docs.ragas.io/en/stable/concepts/metrics/available_metrics/){: .external}.
 
 ### Evaluator plugins
--- a/docs/evaluation.md
+++ b/docs/evaluation.md
@@ -252,6 +252,9 @@ given prompt
 *   Maliciousness -- Measures whether the generated output intends to deceive,
 harm, or exploit
 
+For more details on how each specific metric is computed, please refer to the 
+[RAGAS metrics documentation](https://docs.ragas.io/en/stable/concepts/metrics/available_metrics/){: .external}.
+
 ### Evaluator plugins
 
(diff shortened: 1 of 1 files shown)


docs/evaluation.md
```diff
@@ -252,6 +252,9 @@ given prompt
 *   Maliciousness -- Measures whether the generated output intends to deceive,
 harm, or exploit
 
+For more details on how each metric is computed, please refer to the 
+[RAGAS metrics documentation](https://docs.ragas.io/en/stable/concepts/metrics/available_metrics/){: .external}.
```
pavelgj, line 256: looking at the page, there are many more metrics listed there...  do we want to link to specific metrics instead?
Ex: https://docs.ragas.io/en/stable/concepts/metrics/available_metrics/faithfulness/

ssbushi: I thought of that, but RAGAS change their metrics frequently, I was afraid we might have to keep up with them so I used the parent page instead. 

I can switch to using specific metrics if you feel that's preferred.

pavelgj: if they change the metrics frequently then it means that whatever we have may not be up-to-date, right? what's the point of linking to their docs if it may or may not be accurate?

ssbushi: The reason we should link is because they have detailed explanation on how a metric is computed. Some of the feedback we recieved for evals is a scenario like:  'user expects relevancy to be 1.0 but it is only 0.5, why is that?'

The reason I am hesitant to link to individual docs for each metric is because RAGAS seem to change their metrics collection (either removing them or renaming them). E.g.: `ANSWER_RELEVANCY` in genkitEvals is now called "Response Relevancy" in RAGAS.

Catching up to RAGAS is a different problem altogether. 

You are right though... taking them to a top level directory does not guarantee that the metric details would be available. The way I see it, both approaches have their cons.

chrisraygill: Alternatively, should we just bring the explanations of the metrics into our docs so we don't need to link our to RAGAS docs?

I believe our metrics are already just "RAGAS-inspired", so it's okay if we deviate as long as we explain what they mean well.

ssbushi: That makes sense to me... but there's going to be more metrics in this plugin and we should move that into it's own documentation?

How about including a summary in the JSDoc for these metrics and linking to the JS API ref?

PR #2024, opened 2025-02-18
feat: show a link to the Dev UI on running evals

ssbushi:
Fixes: https://github.com/firebase/genkit/issues/1641

Checklist (if applicable):
- [x] Tested (manually)


genkit-tools/common/src/utils/utils.ts
```diff
@@ -215,3 +223,57 @@ export async function retriable<T>(
...
+export function isValidDevToolsInfo(data: any): data is DevToolsInfo {
+  return (
+    typeof data === 'object' &&
+    typeof data.url === 'string' &&
+    typeof data.timestamp === 'string'
+  );
+}
+
+/**
+ * Fetches the Genkit Dev UI URL if available
```
pavelgj, line 239:
```suggestion
 * Finds the URL of the currently running Genkit Dev UI instance (if available).
```

genkit-tools/common/src/utils/utils.ts
```diff
@@ -215,3 +223,57 @@ export async function retriable<T>(
...
+ */
+export async function getDevUiUrl(
+  projectRoot?: string
+): Promise<string | undefined> {
+  const serversDir = await findServersDir(projectRoot);
+  const toolsJsonPath = path.join(serversDir, 'tools.json');
```
pavelgj, line 245: you're always writing to `tools.json`? this assumes we track only the last instance of the dev ui? do we want to track all instances?

ssbushi: I did not know that we'd have multiple Dev UI instances on the same runtime. What would the difference between the instances be?

I can add support for multiple instances if we want, I just need some direction on how we do that for other state files.

apascal07: I assume like if you're working on a local version of the Dev UI.

js/testapps/evals/package.json
```diff
@@ -11,7 +11,7 @@
     "build:watch": "tsc --watch",
     "build-and-run": "pnpm build && node lib/index.js",
     "dev": "tsx --watch src/index.ts",
```
shrutip90, line 13: We already have this, do we need to change genkit:dev?

genkit-tools/common/src/utils/utils.ts
ssbushi: I dont think we can handle supporting a local Dev UI, as it is a separate process entirely and not tracked through `genkit start`

js/testapps/evals/package.json
ssbushi: I undid this change.

genkit-tools/common/src/utils/utils.ts
ssbushi: I added support for multiple Dev UI servers, PTAL.

pavelgj (approved):

PR #2029, opened 2025-02-19
feat(py): refactoring a Plugin abstraction

timere-nemo:
Description:

- Refactored a Plugin abstraction to make it more versatile in use.
- Updated the `VertexAI` plugin and a code sample

Problem:

1. Current implementation assumes that only one model is defined, whereas we may have multiple models.
2. It is unclear why `_model_callback` needs to be defined in the Plugin. Since each model (or model family) has its own execution logic, it should provide its own callback

Checklist (if applicable):

### DIFF
--- a/py/packages/genkit/src/genkit/core/plugin_abc.py
+++ b/py/packages/genkit/src/genkit/core/plugin_abc.py
@@ -8,72 +8,20 @@ from __future__ import annotations
 import abc
 import typing
 
-from genkit.core.schema_types import GenerateRequest, GenerateResponse
-
 if typing.TYPE_CHECKING:
+    # TODO: Must point to an abstraction, not an actual implementation,
+    #       which should resolve the circular import problem.
     from genkit.veneer import Genkit
 
 
 class Plugin(abc.ABC):
     """
-    Abstract class defining common interface
-    for the Genkit Plugin implementation
-
-    NOTE: Any plugin defined for the Genkit must inherit from this class
+    Abstract base class for plugins.
     """
 
     @abc.abstractmethod
-    def attach_to_veneer(self, veneer: Genkit) -> None:
+    def initialize(self, ai: Genkit) -> None:
         """
-        Entrypoint for attaching the plugin to the requested Genkit Veneer
-
-        Implementation must be provided for any inheriting plugin.
-
-        Args:
-            veneer: requested `genkit.veneer.Genkit` instance
-
-        Returns:
-            None
-        """
-        pass
-
-    def _add_model_to_veneer(
-        self, veneer: Genkit, name: str, metadata: dict | None = None
-    ) -> None:
-        """
-        Defines plugin's model in the Genkit Registry
-
-        Uses self._model_callback as a generic callback wrapper
-
-        Args:
-            veneer: requested `genkit.veneer.Genkit` instance
-            name: name of the model to attach
-            metadata: metadata information associated
-                      with the provided model (optional)
-
-        Returns:
-            None
-        """
-        if not metadata:
-            metadata = {}
-        veneer.define_model(
-            name=name, fn=self._model_callback, metadata=metadata
-        )
-
-    @abc.abstractmethod
-    def _model_callback(self, request: GenerateRequest) -> GenerateResponse:
-        """
-        Wrapper around any plugin's model callback.
-
-        Is considered an entrypoint for any model's request.
-        Implementation must be provided for any inheriting plugin.
-
-        Args:
-            request: incoming request as generic
-                     `genkit.core.schemas.GenerateRequest` instance
-
-        Returns:
-            Model response represented as generic
-            `genkit.core.schemas.GenerateResponse` instance
+        Entrypoint for attaching the plugin to the Genkit Veneer.
         """
         pass
--- a/py/packages/genkit/src/genkit/veneer/veneer.py
+++ b/py/packages/genkit/src/genkit/veneer/veneer.py
@@ -60,7 +60,7 @@ class Genkit:
         else:
             for plugin in plugins:
                 if isinstance(plugin, Plugin):
-                    plugin.attach_to_veneer(veneer=self)
+                    plugin.initialize(self)
                 else:
                     raise ValueError(
                         f'Invalid {plugin=} provided to Genkit: '
--- a/py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/plugin_api.py
+++ b/py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/plugin_api.py
@@ -37,23 +37,13 @@ class VertexAI(Plugin):
         self._gemini = Gemini(self.VERTEX_AI_GENERATIVE_MODEL_NAME)
         vertexai.init(project=project_id, location=location)
 
-    def attach_to_veneer(self, veneer: Genkit) -> None:
-        self._add_model_to_veneer(veneer=veneer)
-
-    def _add_model_to_veneer(self, veneer: Genkit, **kwargs) -> None:
-        return super()._add_model_to_veneer(
-            veneer=veneer,
+    def initialize(self, ai: Genkit) -> None:
+        ai.define_model(
             name=vertexai_name(self.VERTEX_AI_GENERATIVE_MODEL_NAME),
-            metadata=self.vertex_ai_model_metadata,
+            fn=self._gemini.handle_request,
+            metadata={
+                'model': {
+                    'supports': {'multiturn': True},
+                }
+            },
         )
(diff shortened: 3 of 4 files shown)


py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -8,72 +8,26 @@
...
-        Entrypoint for attaching the plugin to the requested Genkit Veneer
-
-        Implementation must be provided for any inheriting plugin.
-
-        Args:
-            veneer: requested `genkit.veneer.Genkit` instance
-
-        Returns:
-            None
+        Entrypoint for attaching the plugin to the Genkit Veneer.
```
kirgrim, line 25: name `register` and `ai` are ambiguous. 
I selected `attach_to_veneer` to clearly reflect the purpose of this method: to attach this plugin instance (which could actually be instantiated several times with different params)

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -8,72 +8,26 @@
 import abc
 import typing
 
-from genkit.core.schema_types import GenerateRequest, GenerateResponse
-
 if typing.TYPE_CHECKING:
+    # TODO: Must point to an abstraction, not an actual implementation,
```
kirgrim, line 12: how would you want to provide this abstraction? I believe it would not simplify solution, and would advocate for keeping this line - in future python versions it will be possible to make it without this condition

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -8,72 +8,26 @@
...
-        """
-        if not metadata:
-            metadata = {}
-        veneer.define_model(
-            name=name, fn=self._model_callback, metadata=metadata
-        )
-
-    @abc.abstractmethod
-    def _model_callback(self, request: GenerateRequest) -> GenerateResponse:
+    def __call__(self, ai: Genkit) -> None:
```
kirgrim, line 29: I am against using "magic" methods for instantiating the instance - we will not benefit much from it and it is hard to follow the idea, when we create plugin.

In addition, we also have `__init__` logic is also generic by plugin

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -8,72 +8,26 @@
...
-            None
-        """
-        if not metadata:
-            metadata = {}
-        veneer.define_model(
-            name=name, fn=self._model_callback, metadata=metadata
-        )
-
-    @abc.abstractmethod
-    def _model_callback(self, request: GenerateRequest) -> GenerateResponse:
```
kirgrim, line 64: if we remove `_model_callback` then what was the point of refactor? we will end up with completely the same spaghetti code but with the class wrapper

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -8,72 +8,26 @@
...
-        Args:
-            veneer: requested `genkit.veneer.Genkit` instance
-
-        Returns:
-            None
+        Entrypoint for attaching the plugin to the Genkit Veneer.
         """
         pass
 
-    def _add_model_to_veneer(
```
kirgrim, line 40: logic of adding model is pretty generic, why do we remove it?

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         else:
             for plugin in plugins:
                 if isinstance(plugin, Plugin):
-                    plugin.attach_to_veneer(veneer=self)
+                    plugin(self)
```
kirgrim, line 63: this line may be misleading, as calling plugin seamlessly attaches it in veneer's registry

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/__init__.py
```diff
@@ -38,13 +38,10 @@ def __init__(self, project_id: str | None = None):
...
-    def attach_to_veneer(self, veneer: Genkit) -> None:
-        self._add_model_to_veneer(veneer=veneer)
-
-    def _add_model_to_veneer(self, veneer: Genkit, **kwargs) -> None:
-        return super()._add_model_to_veneer(
-            veneer=veneer,
+    def register(self, ai: Genkit) -> None:
+        ai.define_model(
             name=self.VERTEX_AI_MODEL_NAME,
+            fn=self._handle_gemini_request,
```
kirgrim, line 44: this is trivial to add it in draft version of Vertex AI but complicated logic will require to introduce a lot of redundant code

kirgrim (changes requested):

py/packages/genkit/src/genkit/core/plugin_abc.py
Irillit: We don't even add here a Genkit class itself, we may only fill the registry. That's what we need after all.

py/packages/genkit/src/genkit/core/plugin_abc.py
kamilkorski: @kirgrim I agree with removal of this method. All it does is passing its arguments to another method with optional default.

py/packages/genkit/src/genkit/core/plugin_abc.py
kamilkorski: @kirgrim @timere-nemo Lets not transform this to function-like objects, we decided to move to oop design here so this seems to be unnecessary.

py/packages/genkit/src/genkit/core/plugin_abc.py
pavelgj: I'm thinking `initialize` might be a good name for this method. Registering in the registry (a.k.a. attaching to the veneer) is only one of the possible side-effects.

timere-nemo: @pavelgj What do you think about naming it `plugin`? Initialize sounds too similar to the `__init__` constructor

py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/__init__.py
Irillit: I thought here we can do a loop with separate models here in that case. I think we should fill the registry with all the models in a plugin anyway, otherwise I don't see the point of having it.

py/packages/genkit/src/genkit/core/plugin_abc.py
timere-nemo: My point was to have an abstraction for `Genkit` veneer, so it can be implemented using inheritance. This way, we import the abstraction instead of the current implementation, avoiding circular imports

kirgrim: > My point was to have an abstraction for `Genkit` veneer, so it can be implemented using inheritance. This way, we import the abstraction instead of the current implementation, avoiding circular imports

No reason to do that - please see #2041

py/packages/genkit/src/genkit/core/plugin_abc.py
kirgrim: yes, we do, as I mentioned previously, purpose of the model callback is to define single entry point interface for the model action handlers. This will allow developer to clearly understand what are the items that he should implement when defining new Plugin class. In addition, if we use generic "attach to model" interface, we encapsulate calling of model_callback which will give us working solution with minimal code rewriting.

py/packages/genkit/src/genkit/core/plugin_abc.py
kirgrim: > @kirgrim I agree with removal of this method. All it does is passing its arguments to another method with optional default

@kamilkorski 
Purpose of this method is to define interface for registering model. If we will have multiple defined models, we use this method "as-is". The scenario with simple overriding of this method is only for the plugins with one-to-one model relations, or when some of the arguments are deterministic (aka constants)

py/packages/genkit/src/genkit/core/plugin_abc.py
Irillit: We might have an interface later, as we may probably create an async entrypoint.

Irillit (approved):

py/packages/genkit/src/genkit/core/plugin_abc.py
Irillit: I agree too for the moment and I would add a separate interface for the model instead.

kirgrim (approved):

PR #2041, opened 2025-02-20
fix(py): Removing circular dependencies on Genkit in Plugin API

kirgrim:
Description:
 - unchained registry from veneer, making it standalone class container
 - removed Genkit Veneer references from Plugin API
 - moved implementation of registering model to `Registry`

Checklist (if applicable):


py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -23,30 +20,26 @@ class Plugin(abc.ABC):
     """
 
     @abc.abstractmethod
-    def attach_to_veneer(self, veneer: Genkit) -> None:
+    def initialize(self) -> None:
```
kirgrim, line 23: this rename is based on @pavelgj comment that this function has more general purpose

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -23,30 +20,26 @@ class Plugin(abc.ABC):
...
-        Implementation must be provided for any inheriting plugin.
-
-        Args:
-            veneer: requested `genkit.veneer.Genkit` instance
+        Entrypoint for initializing the plugin instance in Genkit
 
         Returns:
             None
         """
         pass
```
Irillit, line 30: Here I'm thinking about implementing the initialization here, but leaving _register_model() empty. I would create abstract methods for:
- register_models()
- register_indexers()
- register_retrievers()
- register_embedders()

and etc

py/packages/genkit/src/genkit/core/registry.py
```diff
@@ -12,18 +15,32 @@ class Registry:
...
+        name: str,
+        fn: ModelFn,
+        metadata: dict[str, Any] | None = None,
+    ):
+        action = Action(
+            name=name, kind=ActionKind.MODEL, fn=fn, metadata=metadata
+        )
+        cls.register_action(action=action)
+
+    @classmethod
```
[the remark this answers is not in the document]
kirgrim: > What is the rationale behind using a classmethod here?
> 
> Each Genkit instance composes its own registry instance no?

no, it is defined on the class-level

py/packages/genkit/src/genkit/core/plugin_abc.py
kirgrim: > Here I'm thinking about implementing the initialization here, but leaving _register_model() empty. I would create abstract methods for:
> 
> * register_models()
> * register_indexers()
> * register_retrievers()
> * register_embedders()
> 
> and etc

Idea of making register_model singular is in abstracting a logic of registry attachment. We attach to registry per each model, thus it is still required. But it may make sense to define register_models, although i am afraid not every plugin has multiple models

py/packages/genkit/src/genkit/core/registry.py
```diff
@@ -4,7 +4,7 @@
 """The registry is used to store and lookup resources such as actions and
```
Irillit, line 4: I support that approach, as we have the only one Registry per entire application, there shouldn't be more of them.

py/packages/genkit/src/genkit/core/registry.py
```diff
@@ -4,7 +4,7 @@
...
 flows."""
 
-from typing import Any, Callable, Dict
+from typing import Any, Callable
```
Irillit, line 7: As we discussed before, Callable should be imported from collections.

py/packages/genkit/src/genkit/core/plugin_abc.py
```diff
@@ -6,22 +6,22 @@
 from __future__ import annotations
```
Irillit, line 6: I wonder if we still need that line.

py/packages/genkit/src/genkit/core/registry.py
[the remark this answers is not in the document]
Irillit: And should we have more than one Genkit class?

py/packages/genkit/src/genkit/core/plugin_abc.py
kirgrim: > I wonder if we still need that line.

no need anymore

py/packages/genkit/src/genkit/core/registry.py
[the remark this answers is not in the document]
kirgrim: > And should we have more than one Genkit class?

makes no sense - polymorphic behaviour will be granted by the plugins api

kirgrim:
@pavelgj please take a look, it seems like we may remove this circular dependency as there is single registry per class. Is it intended/allowed/makes sense to have multiple instances of Genkit veneer instantiated during the runtime?

py/packages/genkit/src/genkit/veneer/veneer.py
```diff
@@ -29,7 +29,7 @@
 class Genkit:
     """Veneer user-facing API for application developers who use the SDK."""
 
-    registry: Registry = Registry()
```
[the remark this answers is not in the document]
kirgrim: > Per my discussion with Pavel, neither Genkit nor Registry are singletons. Please make it so that we use composition and maintain that relationship.

done

PR #2069, opened 2025-02-24
feat(go/plugin/openai): added initial openai client

kekoawong:
Adds an initial custom openai client to be used for the openai go plugin. Unit and live tests were added for this custom client, along with initial documentation for how to test the client. The main plugin functionality will be added in the next steps. 

We could also consider using the [openai go library](https://github.com/openai/openai-go) instead. A custom client was created instead since the **openai project is still under beta** and may introduce breaking changes and lack stability. Additionally, was unsure if the project's license was compatible with genkit's usage. 

Checklist (if applicable):
- [X] Tested - added unit tests and live tests
- [X] Docs updated - added an initial readme for the client


MichaelDoyle:
Thanks. We now have the `compat_oai` plugin for this purpose.

PR #2071, opened 2025-02-24
fix(js/plugins/google-cloud) don't truncate stopSequences

MichaelDoyle:
It's an array, so pass along object. 

I thought about stringify-ing it and truncating, but these should generally be small, and I think we do want the array as is.

Checklist (if applicable):


MichaelDoyle:
Log will fail to be written. Not impossible, but that's a _lot_ of stop
sequences.

Currently this is causing generate requests to fail completely if
stop sequences are specified, so this is a pretty big regression either
way. We don't inspect and truncate object-like things anywhere else, so I
opted to punt on something more robust.

schnecle (approved):

PR #2169, opened 2025-02-25
feat(evals): add `status` enum for evaluation scores

ssbushi:
Fixes https://github.com/firebase/genkit/issues/2252

Checklist (if applicable):


js/plugins/evaluators/src/types.ts
```diff
@@ -22,3 +31,66 @@ export enum GenkitMetric {
   DEEP_EQUAL = 'DEEP_EQUAL',
   JSONATA = 'JSONATA',
 }
+
+export interface BaseGenkitMetricConfig {
+  type: GenkitMetric;
+  statusOverrideFn?: (score: Score) => EvalStatusEnum;
```
ssbushi, line 37: @pavelgj I made this override method to use the score and get a new status. should we make this even more generic and just pass in the full `datapoint`?

This is part of the plugin API so should be easy to fix.... I think?

js/plugins/evaluators/src/index.ts
```diff
@@ -75,15 +72,16 @@ export function genkitEval<
...
-
-function fillScores(dataPoint: BaseEvalDataPoint, score: Score): EvalResponse {
-  return {
-    testCaseId: dataPoint.testCaseId,
-    evaluation: score,
-  };
+function fillScores(
+  dataPoint: BaseEvalDataPoint,
+  score: Score,
+  statusOverrideFn?: (s: Score) => EvalStatusEnum
```
pavelgj, line 78: does this fn need to know the metric? and maybe original status as well? so that one could do something like:

```ts
statusOverrideFn = ({score, originalStatus, metric}) => {
  if (metric === 'faithfulness') {
    return ...;
  }
  return originalStatus;
}
```

ssbushi: - originalStatus is available in the score object
- Since this is configured per metric, I think the metric field is redundant.

I like the idea of using an object for future expansion, so I added that.

pavelgj (approved):

PR #2179, opened 2025-02-27
feat(py/ai/generate): implemented basic veneer for the generate action

pavelgj:
Checklist (if applicable):


py/packages/genkit/src/genkit/ai/testing_utils.py
```diff
@@ -0,0 +1,90 @@
...
+class ProgrammableModel:
+    request_idx = 0
+    responses: list[GenerateResponse] = []
+    chunks: list[list[GenerateResponseChunk]] = None
+    last_request: GenerateRequest = None
+
+    def __init__(self):
+        self.request_idx = 0
+        self.responses = []
+        self.chunks = None
```
Irillit, line 30: You don't have to redefine them in init, as you already set them default values. They are defined on a class method, but if you like, you can move everything to init.

py/packages/genkit/src/genkit/ai/testing_utils.py
```diff
@@ -0,0 +1,90 @@
...
+        self.last_request = None
+
+    def model_fn(self, request: GenerateRequest, ctx: ActionRunContext):
+        self.last_request = request
+        response = self.responses[self.request_idx]
+        if self.chunks is not None:
```
Irillit, line 36: `is not None` is not mandatory here. `if self.chunks` would do the same.

py/packages/genkit/src/genkit/ai/testing_utils.py
pavelgj: moved everything to init... I can't figure out why, but somehow these fields behave like static fields otherwise. Literally, the tests fail because `responses` ends up with test data from multiple tests. It works fine when everything is in init. 🤷

py/packages/genkit/src/genkit/ai/testing_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    action = ai.define_model(name=name, fn=model_fn)
+
+    return (pm, action)
+
+
+class EchoModel:
+    last_request: GenerateRequest = None
+
+    def __init__(self):
+        def model_fn(request: GenerateRequest):
```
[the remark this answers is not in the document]
pavelgj: if I define it as a method, then it takes `self` as first argument and I can't use the function as the action function... but that's fine, I just wrap it like I do for programmable model. There's probably some functools magic I can do instead....

py/packages/genkit/src/genkit/ai/testing_utils.py
pavelgj: Yay! I'm learning python :)
https://stackoverflow.com/questions/207000/what-is-the-difference-between-class-and-instance-attributes

now it makes sense why I was getting this weird behaviour. So, defining them in `__init__` seems like the right thing to do.

PR #2182, opened 2025-02-27
feat(pinecone): Add support for pinecone index host url

shazy792:
Adding support for pinecone index host url to target indices as recommended by pinecone for production uses cases 
Pinecone docs ref: https://docs.pinecone.io/guides/data/target-an-index

Checklist (if applicable):


shazy792:
@pavelgj Hoping you can help review this. Should be fairly quick. Thanks!

MichaelDoyle:
@ifielker @pavelgj seems simple enough -- worth merging?

PR #2201, opened 2025-02-28
feat(js/plugins): added new gemini 2.0 flash thinking exp models

nohe427:
Added new gemini 2.0 flash thinking exp models to GoogleAI and VertexAI plugins

We added the flash thinking models recently and I want to use them in my project

Checklist (if applicable):


js/plugins/googleai/src/gemini.ts
```diff
@@ -209,6 +209,23 @@ export const gemini20ProExp0205 = modelRef({
   configSchema: GeminiConfigSchema,
 });
 
+export const gemini20FlashThinkingExp0121 = modelRef({
```
pavelgj, line 212: this also needs to be exported from the index.ts

js/plugins/vertexai/src/gemini.ts
```diff
@@ -374,6 +374,23 @@ export const gemini20ProExp0205 = modelRef({
   configSchema: GeminiConfigSchema,
 });
 
+export const gemini20FlashThinkingExp0121 = modelRef({
```
pavelgj, line 377: same here -- this needs to be exported from the index.ts

js/plugins/googleai/src/gemini.ts
nohe427: Done

PR #2234, opened 2025-03-04
fix(go): fix bug where dev UI configs aren't respected

huangjeff5:
Tested manually in Dev UI and code invocation paths.


go/plugins/googleai/googleai.go
```diff
@@ -314,26 +315,46 @@ func generate(
...
-			gm.SetTemperature(float32(c.Temperature))
-		}
-		if c.TopK != 0 {
-			gm.SetTopK(int32(c.TopK))
-		}
-		if c.TopP != 0 {
-			gm.SetTopP(float32(c.TopP))
-		}
+	var c ai.GenerationCommonConfig
+	if err := mapToStruct(input.Config.(map[string]interface{}), &c); err != nil {
```
apascal07, line 348: It's not always a map though. It only comes in as a map if the request is from Dev UI. If you run a flow with a Generate call that has a config set, it will behave correctly. So this change is now breaking that I'm pretty sure.

apascal07 (changes requested):

go/plugins/googleai/googleai.go
```diff
@@ -65,6 +66,12 @@ var (
 	knownEmbedders = []string{"text-embedding-004", "embedding-001"}
 )
 
+// GenerationGoogleAIConfig extends GenerationCommonConfig with Google AI specific settings.
+type GenerationGoogleAIConfig struct {
+	ai.GenerationCommonConfig
+	SafetySettings []*genai.SafetySetting
```
apascal07, line 72: It sucks but I think we should redefine our own so that end users don't have to depend directly on the Gemini SDK since it's an implementation detail when going through us. That's how we do it in JS too.

go/ai/gen.go
```diff
@@ -108,6 +108,7 @@ type ModelInfo struct {
 	Label    string             `json:"label,omitempty"`
 	Supports *ModelInfoSupports `json:"supports,omitempty"`
 	Versions []string           `json:"versions,omitempty"`
+	ConfigSchema map[string]any `json:"configSchema,omitempty"`
```
apascal07, line 111: This needs to come from `genkit-tools` otherwise it will get overwritten on next generation.

PR #2253, opened 2025-03-05
feat(go): Add ai > document

ssbushi:
Checklist (if applicable):

### DIFF
--- a/py/packages/genkit/src/genkit/ai/document.py
+++ b/py/packages/genkit/src/genkit/ai/document.py
@@ -1,0 +1,193 @@
+# Copyright 2025 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from copy import deepcopy
+from typing import Any
+
+from genkit.ai.embedding import EmbeddingModel
+from pydantic import BaseModel
+
+
+class EmptyPart(BaseModel):
+    """Empty part of a document."""
+
+    text: None = None
+    media: None = None
+
+
+class TextPart(BaseModel):
+    """Text part of a document.
+
+    Attributes:
+        text: The string contents of the document
+    """
+
+    text: str
+    media: None = None
+
+
+class MediaPartModel(BaseModel):
+    """Object model representing media parts.
+
+    Attributes:
+        content_type: The media content type. Inferred from data uri if not
+        provided.
+        url: A `data:` or `https:` uri containing the media content.
+    """
+
+    content_type: str | None = None
+    url: str
+
+
+class MediaPart(BaseModel):
+    """Media part of a document.
+
+    Attributes:
+        media: The media contents of the document
+    """
+
+    text: None = None
+    media: MediaPartModel
+
+
+# Part type
+type Part = TextPart | MediaPart
+
+
+class DocumentData(BaseModel):
+    """Represent complete document data.
+
+    Attributes:
+        content: The parts contained in this document
+        metadata: optional metadata
+    """
+
+    content: list[Part]
+    metadata: dict[str, Any] | None = None
+
+
+class Document(DocumentData):
+    """Represents document content along with its metadata.
+
+    This object can be embedded, indexed or retrieved. Each document can contain
+    multiple parts (for example text and an image).
+    """
+
+    def __init__(
+        self, content: list[Part], metadata: dict[str, Any] | None = None
+    ) -> None:
+        """Initialize Document object."""
+        doc_content = deepcopy(content)
+        doc_metadata = deepcopy(metadata)
+        super().__init__(content=doc_content, metadata=doc_metadata)
+
+    @staticmethod
+    def from_text(
+        text: str, metadata: dict[str, Any] | None = None
+    ) -> Document:
+        """Construct a Document from a single text part."""
+        return Document(content=[TextPart(text=text)], metadata=metadata)
+
+    @staticmethod
+    def from_media(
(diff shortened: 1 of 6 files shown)


py/packages/genkit/src/genkit/ai/embedding.py
```diff
@@ -2,9 +2,20 @@
...
+type Embedding = list[float]
+
+
+# TODO(ssbushi): Replace with generated type
+class EmbeddingModel(BaseModel):
+    """Represents an embedding with metadata."""
+
+    embedding: Embedding
+    metadata: dict[str, Any] | None = None
+
```
ssbushi, line 18: This is a placeholder for now, will update with the generated type we should get from tooling.

py/packages/genkit/src/genkit/ai/document.py
```diff
@@ -0,0 +1,193 @@
...
+    Attributes:
+        content: The parts contained in this document
+        metadata: optional metadata
+    """
+
+    content: list[Part]
+    metadata: dict[str, Any] | None = None
+
+
+class Document(DocumentData):
```
Irillit, line 71: For me it's a bit strange to inherit Pydantic model. Is it possible to use composition over inheritance here?

py/packages/genkit/src/genkit/ai/document.py
```diff
@@ -0,0 +1,193 @@
...
+
+    text: None = None
+    media: MediaPartModel
+
+
+# Part type
+type Part = TextPart | MediaPart
+
+
+class DocumentData(BaseModel):
```
pavelgj, line 59: there's already DocumentData in https://github.com/firebase/genkit/blob/15d507e5c78f1db820f7fc11b1bc6d7951a10fbe/py/packages/genkit/src/genkit/core/typing.py#L615

py/packages/genkit/src/genkit/ai/document.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if self.media():
+            return self.media()[0].url
+
+        return ''
+
+    def data_type(self) -> str | None:
+        """Gets the content_type of the data that is returned by data()."""
+        if self.text():
+            return 'text'
```
Irillit, line 143: As you've used 'text' several times, I suggest you to add it as a constant or a class field.

py/packages/genkit/src/genkit/ai/document.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class EmptyPart(BaseModel):
+    """Empty part of a document."""
+
+    text: None = None
+    media: None = None
+
+
+class TextPart(BaseModel):
```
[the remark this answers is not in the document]
ssbushi: They are now :) Thanks!

py/packages/genkit/src/genkit/ai/document.py
ssbushi: Updated.

py/packages/genkit/src/genkit/ai/document.py
ssbushi: it is possible, but why is this pattern strange?

pavelgj: we need to make a decision on how to do these wrappers in a consistent way. We already have a few here:
https://github.com/firebase/genkit/blob/main/py/packages/genkit/src/genkit/ai/model.py

I expect we will want to wrap all types from `genkit.core.typing` as the generated pydantic types are not very stable, so we will want to avoid using then directly anywhere.

py/packages/genkit/src/genkit/ai/document.py
```diff
@@ -0,0 +1,151 @@
...
+        metadata: dict[str, Any] | None = None,
+    ) -> Document:
+        """Construct a Document from a single media part."""
+        if data_type == TEXT_DATA_TYPE:
+            return Document.from_text(data, metadata)
+        return Document.from_media(data, data_type, metadata)
+
+    def text(self) -> str:
+        """Concatenates all `text` parts with no delimiter."""
+        text_parts = map(lambda part: part.root.text or '', self.content)
```
pavelgj, line 80: there's a helper here:
https://github.com/firebase/genkit/blob/238b656736b9b4832d27337867b2441776f5836c/py/packages/genkit/src/genkit/ai/model.py#L213

not sure if it's worth reusing...

py/packages/genkit/src/genkit/ai/document.py
```diff
@@ -0,0 +1,151 @@
...
+        return ''.join(text_parts)
+
+    def media(self) -> list[MediaPartModel]:
+        """Media array getter."""
+        media_parts = map(lambda part: part.root.media, self.content)
```
pavelgj, line 85: this might be a little shorter/cleaner:
```suggestion
        media_parts = part.root.media for part in self.content
```

pavelgj (approved):

py/packages/genkit/src/genkit/ai/document.py
ssbushi: Why would the generated types not be stable? They _should_ match the JS types and JS is GA so by definition it should be stable.

Irillit: Ok, adding methods /fields to the Pydantic model seems to be fine. I think, we can rely on that.
https://github.com/pydantic/pydantic/discussions/9005

PR #2257, opened 2025-03-06
fix(js): export all standard types through tooling

ssbushi:
Fixes: https://github.com/firebase/genkit/issues/2258

Checklist (if applicable):
- [x] Tested (manually, building the Dev UI)


js/ai/src/document.ts
```diff
@@ -36,8 +51,76 @@ export const MediaPartSchema = EmptyPartSchema.extend({
...
+
+/**
+ * Media part.
+ */
 export type MediaPart = z.infer<typeof MediaPartSchema>;
 
+/**
+ * Zod schema of a tool request part.
+ */
+export const ToolRequestPartSchema = EmptyPartSchema.extend({
```
pavelgj, line 63: wait.... why not import these from model.ts?

ssbushi: Model imports DocumentData from document.js which causes a cyclic dependency.

pavelgj: lets put them in a separate file them (parts.ts?)... document.ts is not a logic place for them.

ssbushi: Hmm, I thought of that but `parts.ts` has no meaningful purpose to be its own file and decided to do this instead. I guess we can move things around since they are not exposed directly anyway

pavelgj (approved):

js/ai/src/document.ts
ssbushi: Resolving, leaving as it is.

PR #2266, opened 2025-03-07
feat(go/plugins): add context-caching to VertexAI and GoogleAI plugins

hugoaguirre:
Added context-cache to VertexAI and GoogleAI plugins.
Created samples for both plugins.
Unified `go-genai` SDK used

Checklist (if applicable):


go/plugins/internal/gemini/cache.go
```diff
@@ -0,0 +1,224 @@
...
+	var nextPageToken string
+	for {
+		page, err := client.Caches.List(ctx, &genai.ListCachedContentsConfig{
+			PageSize:  CacheContentsPerPage,
+			PageToken: nextPageToken,
+		})
+		if err != nil && err != genai.ErrPageDone {
+			return nil, err
+		}
+		fmt.Printf("page found with %d elements\n\n", len(page.Items))
```
apascal07, line 185: Remove?

go/plugins/internal/gemini/cache.go
```diff
@@ -0,0 +1,224 @@
...
+
+// lookupCache retrieves a *genai.CachedContent from a given cache key
+func lookupCache(ctx context.Context, client *genai.Client, key string) (*genai.CachedContent, error) {
+	if key == "" {
+		return nil, fmt.Errorf("empty cache key detected")
+	}
+
+	var nextPageToken string
+	for {
+		page, err := client.Caches.List(ctx, &genai.ListCachedContentsConfig{
```
apascal07, line 178: Why not `client.Caches.Get(...)`?

apascal07 (changes requested):

go/plugins/internal/gemini/cache.go
```diff
@@ -0,0 +1,224 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/firebase/genkit/go/ai"
+	"google.golang.org/genai"
+)
+
+const CacheContentsPerPage = 5
+
+var ContextCacheSupportedModels = [...]string{
```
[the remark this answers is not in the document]
hugoaguirre: The compiler does not support the use of `const [...]T`, the closest we can get to a constant array is to tell the compiler to use just an array like this (with constant size) but it has to be a `var`.

go/plugins/internal/gemini/cache.go
```diff
@@ -0,0 +1,224 @@
...
+		DisplayName: key,
+		TTL:         strconv.Itoa(ttl) + "s",
+		Contents:    messagesToCache,
+	}, nil
+}
+
+// validateContextCacheRequest checks for supported models and checks if Tools
+// are being provided in the request
+func validateContextCacheRequest(request *ai.ModelRequest, modelVersion string) error {
+	if modelVersion == "" || !slices.Contains(ContextCacheSupportedModels[:], modelVersion) {
```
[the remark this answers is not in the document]
hugoaguirre: Since it is a fixed size array, it has to be converted to a slice this way

go/plugins/internal/gemini/cache.go
apascal07: This is not a practical solution. You can have many requests and many caches active. This lookup is O(n), it will be super slow for any prod uses with traffic. We need a mechanism to return `Name` back to user control so that it can be passed through to subsequent requests, both in Go and in JS.

In any case, `Caches.All()` returns an iterator over all caches available which is better to use here if we're going to brute force it. You should file an issue if you're going to check this in.

go/plugins/internal/gemini/gemini.go
```diff
@@ -265,7 +276,10 @@ func convertRequest(client *genai.Client, model string, input *ai.ModelRequest)
 	choice := convertToolChoice(input.ToolChoice, input.Tools)
 	gc.ToolConfig = choice
 
-	// TODO: configure cache
+	// configure cache
```
apascal07, line 279: Nit: this and other comments like it aren't that helpful in my opinion. Comments should describe why not what. We can see the what from the code. It's just visual noise and easy for them to become outdated. I would suggest removing them throughout these files.

go/samples/cache-gemini/main.go
```diff
@@ -0,0 +1,97 @@
...
+			return "", err
+		}
+
+		resp, err := genkit.Generate(ctx, g, ai.WithConfig(&ai.GenerationCommonConfig{
+			Temperature: 0.7,
+			Version:     "gemini-1.5-flash-001",
+		}),
+			ai.WithMessages(
+				ai.NewUserMessage(
+					ai.NewTextPart(string(textContent))),
```
apascal07, line 57: FYI: There's also `ai.NewUserTextMessage()` that makes this smoother.

go/samples/cache-gemini/main.go
```diff
@@ -0,0 +1,97 @@
...
+			ai.WithTextPrompt(prompt),
+		)
+		if err != nil {
+			return "", nil
+		}
+
+		text := resp.Text()
+
+		m := resp.Message.Metadata
+		name := ""
```
apascal07, line 70: You don't need this or `WithCacheName()` if you're going to use `resp.History()`.

go/plugins/internal/gemini/cache.go
hugoaguirre: As per the suggestion in discord, we are now using `WithCacheName()` helper to fetch in constant time the required cache

go/ai/request_helpers.go
```diff
@@ -66,3 +66,55 @@ func NewTextMessage(role Role, text string) *Message {
...
+	}
+
+	cache := make(map[string]any)
+	cache["ttlSeconds"] = ttlSeconds
+
+	if existingCache, ok := metadata["cache"].(map[string]any); ok {
+		existingCache["ttlSeconds"] = ttlSeconds
+	} else {
+		metadata["cache"] = cache
+	}
```
apascal07, line 87: I think this can be replaced with `maps.Copy()` like here: https://github.com/firebase/genkit/blob/e373b699c6ae72c01daa5e349d8b807d9ae9052c/go/ai/prompt.go#L72

go/ai/request_helpers.go
```diff
@@ -66,3 +66,55 @@ func NewTextMessage(role Role, text string) *Message {
...
+
+	cache := make(map[string]any)
+	cache["name"] = n
+
+	if existingCache, ok := metadata["cache"].(map[string]any); ok {
+		existingCache["name"] = n
+	} else {
+		metadata["cache"] = cache
+	}
```
apascal07, line 113: Same here. We want to overwrite any `name` or `ttlSeconds` and vice versa.

apascal07 (approved):

PR #2273, opened 2025-03-10
fix(go): check validation for json part on text that generated by LLM (#2265)

### LINKED ISSUE #2265: [Go] Causes an error call `ai.Generate()` with tools and output schema
**Describe the bug**
The validSchema function causes an error when I call `ai.Generate()` with `ai.WithTools()` and `ai.WithOutputSchema()`
Error message:
```plaintext
generation resulted in no candidates matching expected schema
```

**To Reproduce**
```go
type Output struct {
	GablorkenValue float64 `json:"gablorken_value"`
	Message        string  `json:"message"`
}
gablorkenTool := ai.DefineTool("gablorken", "use when need to calculate a gablorken",
		Value float64
		Over  float64
	}) (float64, error) {
		return math.Pow(input.Value, input.Over), nil
	},
)
resp, err := ai.Generate(
	ctx,                                                     //
	model,                                                   //
	ai.WithCandidates(1),                                    //
	ai.WithTextPrompt("What is a gablorken of 2 over 3.5?"), //
	ai.WithTools(gablorkenTool),                             //
	ai.WithOutputSchema(&Output{}),
)
if err != nil {
	t.Fatal(err)
}
var out Output

if err := json.Unmarshal([]byte(resp.Text()), &out); err != nil {
	t.Fatal(err)
}

const want = "12.25"
if !strings.Contains(out.Message, want) {
	t.Errorf("got %q, expecting it to contain %q", out.Message, want)
}
```

**Expected behavior**
Pass it as a tool call request without error

 - OS: MacOS
 - Version 13.2

**Go version**
go version go1.23.6 darwin/arm64

jcooky:
resolve #2265 

Change JSON schema validation to work per part instead of by content.

Checklist (if applicable):


apascal07 (approved):
Thank you for contributing!

PR #2274, opened 2025-03-10
fix(py): various fixes including missing return values; pytest watcher script

yesudeep:
CHANGELOG:
- [ ] Order the action types in registry.ts
- [ ] Sync the action kinds in registry.py
- [ ] Add script to watch python tests.
- [ ] Fix access level for Registry class members.
- [ ] Address tests that were using private members.
- [ ] Fix types and docs.

### DIFF
--- a/py/packages/genkit/src/genkit/core/registry.py
+++ b/py/packages/genkit/src/genkit/core/registry.py
@@ -29,6 +29,21 @@ type ActionName = str
 
 type ActionResolver = Callable[[ActionKind, str], None]
 
+# An action store is a nested dictionary mapping ActionKind to a dictionary of
+# action names and their corresponding Action instances.
+#
+# Structure for illustration:
+#
+# ```python
+# {
+#     ActionKind.MODEL: {
+#         'gemini-2.0-flash': Action(...),
+#         'gemini-2.0-pro': Action(...)
+#     },
+# }
+# ```
+type ActionStore = dict[ActionKind, dict[ActionName, Action]]
+
 
 class Registry:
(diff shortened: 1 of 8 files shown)


tanyasingh1902 (approved):

py/packages/genkit/src/genkit/core/registry.py
```diff
@@ -179,34 +196,38 @@ def register_value(self, type: str, name: str, value: Any):
...
             raise ValueError(
-                f'value for type "{type}" and name "{name}" is already registered'
+                f'value for type "{type}" '
+                f'and name "{name}" is already registered'
             )
 
-        self.value_by_type_and_name[type][name] = value
+        self._value_by_type_and_name[type][name] = value
 
     def lookup_value(self, type: str, name: str) -> Any | None:
```
Irillit, line 217: Don't use `type` as a variable. It's a built-in, so, please, rename `type` to something else..

py/packages/genkit/src/genkit/core/registry.py
```diff
@@ -179,34 +196,38 @@ def register_value(self, type: str, name: str, value: Any):
...
 
         Returns:
-          The value or None if not found.
+            The value or None if not found.
         """
         if (
-            type in self.value_by_type_and_name
-            and name in self.value_by_type_and_name[type]
+            type in self._value_by_type_and_name
+            and name in self._value_by_type_and_name[type]
```
Irillit, line 229: What if you use `.get()`  dictionary method here? It might be shorter.

PR #2277, opened 2025-03-10
feat(py): return an empty response message when streaming completes

timere-nemo:
Description:

Updated generate stream to return a `GenerateResponse` when streaming completes, ensuring consistency with core system expectations.

Checklist (if applicable):

### DIFF
--- a/py/plugins/compat-oai/src/genkit/plugins/compat_oai/models/model.py
+++ b/py/plugins/compat-oai/src/genkit/plugins/compat_oai/models/model.py
@@ -92,7 +92,17 @@ class OpenAIModel:
 
     def generate_stream(
         self, request: GenerateRequest, callback: Callable
-    ) -> None:
+    ) -> GenerateResponse:
+        """
+        Generates a streaming response from the OpenAI client and processes it in chunks.
+
+        Args:
+            request (GenerateRequest): The request object containing generation parameters.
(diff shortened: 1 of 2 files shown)


py/plugins/compat-oai/src/genkit/plugins/compat_oai/models/model.py
```diff
@@ -110,3 +120,9 @@ def generate_stream(
             )
 
             callback(response_chunk)
+
+        # Return an empty response when streaming is complete
+        return GenerateResponse(
```
pavelgj, line 125: you can take a look what 3p openai plugin does:
https://github.com/TheFireCo/genkit-plugins/blob/d0b3886cd784582dee8fe55c80235cc2909db2eb/plugins/openai/src/gpt.ts#L628

I'm not sure what `finalChatCompletion` returns or whether the same is available for python.

timere-nemo: In Python, a stream's  returns nothing (`None`) when it is complete. However, .generate must return a `GenerateResponse` because the response wrapper expects an object that has a `.message` attribute: https://github.com/firebase/genkit/blob/main/py/packages/genkit/src/genkit/ai/model.py#L94

pavelgj (approved):

Irillit (approved):

PR #2279, opened 2025-03-10
chore(typing): cleanup more json schema and python model generation

pavelgj:
Checklist (if applicable):


ssbushi:
I know it is unnecessary, but just for the sake of maintaining parity, we should update the `js/ai` versions too, right?

ssbushi (approved):
tooling LGTM

pavelgj:
> I know it is unnecessary, but just for the sake of maintaining parity, we should update the `js/ai` versions too, right?

Good point. Done.

PR #2283, opened 2025-03-10
fix(go/ai): error if multiple tool requests share Ref

### LINKED ISSUE #2224: [Go] Multiple tool calls should be using the `ref` field to disambiguate
**Describe the bug**
`ToolRequest.Ref` is used but there is no erroring if there are multiple calls to the same tool but the ref is not set.

kwjw:
Resolves #2224.

If the same tool is called multiple times in the same request, then the tool requests must have distinct Ref values.

Checklist (if applicable):

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -532,10 +532,24 @@ func cloneMessage(m *Message) *Message {
 // either a new request to continue the conversation or nil if no tool requests
 // need handling.
 func handleToolRequests(ctx context.Context, r *registry.Registry, req *ModelRequest, resp *ModelResponse, cb ModelStreamCallback) (*ModelRequest, *Message, error) {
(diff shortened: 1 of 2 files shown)


apascal07 (changes requested):
This is one way to solve this but any model-generated refs aren't significant, we can just fill them in ourselves with an index or similar if they're not filled in by the model.

pavelgj:
> This is one way to solve this but any model-generated refs aren't significant, we can just fill them in ourselves with an index or similar if they're not filled in by the model.

Can use https://github.com/firebase/genkit/pull/2250 for reference

MichaelDoyle:
Hey @kwjw it's been over a year since we've seen any movement on this. Let us know if you want to pick it back up.

PR #2302, opened 2025-03-12
feat(js/ai/generate): retain all raw data from LLM

### LINKED ISSUE #2303: [JS] Current GenerateResponse implementation doesn't retain all data
My current project requires access to the citation data returned from Vertex AI / Gemini. Unfortunately, this data is not exposed to the flows that the project uses.

It would be nice if all data from an LLM were retained and accessible to callers. I like that Genkit has streamlined LLM responses to a useful subset, but sometimes we need a little more info. Plus, as the LLMs themselves and their services evolve, it might be that the `GenerateReponse` class doesn't stay in-sync with the LLM service.

I traced through the calls in my project in a debugger. I found that the `generate()` function, called (implicitly) by `ExecutePrompt`, only returns a `GenerateResponse` object. Looking at the resulting object, I saw that the citation metadata that was present in the `response` argument to the `GenerateResponse` constructor was not present in the resulting object.

telpirion:
Fixes #2303 

This PR adds a field to the `GenerateResponse` class to retain all information returned by the LLM. The current implementation of `GenerateResponse` selectively retains specific fields from the raw response returned from the LLM.

Case in point: the [`GenerateContentCandidate`](https://cloud.google.com/vertex-ai/generative-ai/docs/reference/nodejs/latest/vertexai/generatecontentcandidate) type in Vertex AI returns a list of citation metadata. My project requires access to this data. Rather than add a new field just for citation metadata, this change is more future proof and flexible.

Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate/response.ts
+++ b/js/ai/src/generate/response.ts
@@ -46,6 +46,8 @@ export class GenerateResponse<O = unknown> implements ModelResponseData {
   custom: unknown;
   /** The request that generated this response. */
   request?: GenerateRequest;
+  /** The raw response from the LLM. */
+  rawResponse?: unknown;
   /** The parser for output parsing of this response. */
   parser?: MessageParser<O>;
 
@@ -64,6 +66,7 @@ export class GenerateResponse<O = unknown> implements ModelResponseData {
         parser: options?.parser,
       });
     }
+    this.rawResponse = response;
     this.finishReason =
       response.finishReason || response.candidates?.[0]?.finishReason!;
     this.finishMessage =


MichaelDoyle:
We typically put the raw data into `custom`.

Citation specific: https://github.com/firebase/genkit/blob/main/js/plugins/vertexai/src/gemini.ts#L665-L667
Entire raw Gemini response: https://github.com/firebase/genkit/blob/main/js/plugins/vertexai/src/gemini.ts#L1019

Can you have a look and see if that works for you?

MichaelDoyle:
FWIW, I see `custom` is in the process of being deprecated in favor of `raw`. Maybe we just didn't get that change all the way over the line. cc @pavelgj.

telpirion:
Where is `this.raw` set? It doesn't look like it's set in the constructor.

telpirion:
Additional: does `custom` always have the complete raw response from the LLM? For example, candidates other than the very first?

Additional question:  from looking in the source code`fromGeminiPart()` doesn't seem to be used anywhere, just exposed. Is that intended?

MichaelDoyle:
Raw is a "get" method, essentially it's an alias to "custom". It's always the full response.

telpirion:
I think that we're still going to run into issues, but perhaps `custom` is enough. We shall see.

Issue #2308, opened 2025-03-12
[Go] Allow users to provide custom configuration to go-genai client


hugoaguirre:
Allow users to provide custom HTTP configuration and Google Credentials to go-genai client
https://github.com/googleapis/go-genai/blob/main/client.go#L70C3-L72C4

Currently, `API_KEY`, `ProjectID` and `Location` are available to be used but HTTP config and Google Credentials are not.

shubham-nair:
Hi, I had raised a PR for this issue, it's PR #324. First open source contribution.

nair-shubh:
Hello, I have been working on a fix for this issue. I believe the change is needed in the underlying go-genai library, so I have opened a pull request there: https://github.com/googleapis/go-genai/pull/324 . Could you please advise if this is the correct approach, or if I should re-submit this PR to the firebase/genkit repository? Thank you!

hugoaguirre:
Hello @iks1, @shubham-nair. Thanks for your interest in contributing to this project!

Unfortunately this issue was created when we were using custom types to interact with the `go-genai` SDK. Now that we are directly using `go-genai` types (#2958 ), this issue is no longer applicable. 

However, @shubham-nair your contribution to go-genai is still valuable!

I'll try to tag issues as `good-first-issue` for future contributions!

nair-shubh:
@hugoaguirre Thank you. I will be looking forward to contributing more!

PR #2312, opened 2025-03-12
feat(go/plugins)!: add unified google plugin (GoogleAI + VertexAI)

hugoaguirre:
Unify VertexAI and GoogleAI plugins into a single Google plugin

Checklist (if applicable):


go/plugins/google/google_test.go
```diff
@@ -317,6 +506,8 @@ func TestCacheHelper(t *testing.T) {
 }
 
 func TestHeader(t *testing.T) {
+	// TODO: (haguirre) re-enable this test when issue #2308 is solved
```
apascal07, line 509: Seems like it's already there? I don't quite understand why this test is skipped.

https://github.com/googleapis/go-genai/blob/main/client.go#L70C3-L72C4

hugoaguirre: Ah yes, the reason I created #2308 was because initially, the struct Config would only be accepting the `apiKey` for Gemini or `projectID` and `location` for Vertex.

Now, since we are going to split configurations, it might be a good idea to just expand it to accept custom http settings.

hugoaguirre: Closing conversation with the following agreement: 
> For now, let's remove this test since there's no actual requirement to provide support for a custom http server.

hugoaguirre:
Moved to https://github.com/firebase/genkit/pull/2450

PR #2359, opened 2025-03-17
fix(js/core): add listValues to reflection api

MichaelDoyle:
Provide defaultModel and other defined values to the Dev UI

#### Description

We need access to the default model for the dev-ui. I figured we ought to just expose everything under "values" in the registry. I ended up converting the registry value for the defaultModel into the actual action, since otherwise it has some funky zod schema stuff in it (from the original ModelArgument type).

Alternatively, we could just add something either (a) in the registry or (b) listActions where we add some metadata to the model action that is currently registered as default.

"
Checklist (if applicable):
- [X] Tested (manually)
- [N/A] Docs updated (updated docs or a docs bug required)

### DIFF
--- a/js/core/src/reflection.ts
+++ b/js/core/src/reflection.ts
@@ -131,25 +131,9 @@ export class ReflectionServer {
         const actions = await this.registry.listActions();
         const convertedActions = {};
         Object.keys(actions).forEach((key) => {
-          const action = actions[key].__action;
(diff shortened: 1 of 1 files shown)


MichaelDoyle:
We decided on a different approach, since "values" generally cannot be considered serializable and are used for internal implementation details.

PR #2377, opened 2025-03-19
feat(py): Gemini API tools

Irillit:
- Works only in case VertexAI is set

Checklist (if applicable):

### DIFF
--- a/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
+++ b/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
@@ -3,10 +3,12 @@
 
 from enum import StrEnum
 from functools import cached_property
+from typing import Any
 
+import google.genai.types as genai_types
 from google import genai
 
-from genkit.core.action import ActionRunContext
+from genkit.core.action import ActionKind, ActionRunContext
 from genkit.core.typing import (
     CustomPart,
     GenerateRequest,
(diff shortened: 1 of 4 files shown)


py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
```diff
@@ -36,7 +40,6 @@
         tools=True,
         tool_choice=True,
         system_role=True,
-        constrained='no-tools',
```
pavelgj, line 39: please keep this

py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
```diff
@@ -265,8 +380,23 @@ async def generate(
         request_cfg = (
             self._genkit_to_googleai_cfg(request.config)
             if request.config
-            else None
```
kamilkorski, line 268: Do we need an empty dict? annotation says config is optional with default None

kamilkorski, line 268: According to annotations config is optional. I think we should pass None if we dont set it

Irillit: I need it for my section, but I can re-create it, if it is better left None here.

Irillit: Ok, I changed it back.

Issue #2054, opened 2025-02-21
[Go] Add evaluator primitive support


apascal07:
**Describe the bug**
Framework support: https://github.com/firebase/genkit/blob/main/js/ai/src/evaluator.ts

PR #2388, opened 2025-03-19
feat(go): add batch evaluator

ssbushi:
Fixes: #2054 

Checklist (if applicable):

### DIFF
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -284,8 +284,23 @@ func LookupEmbedder(g *Genkit, provider, name string) ai.Embedder {
 	return ai.LookupEmbedder(g.reg, provider, name)
 }
 
+// DefineEvaluator registers the given evaluator function as an action, and
+// returns a [Evaluator] that runs it. This method process the input dataset
+// one-by-one.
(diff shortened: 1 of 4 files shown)


ssbushi:
@apascal07 I'm unable to understand if the test failures is related to the current PR, can you PTAL? Thanks!

go/genkit/genkit.go
```diff
@@ -292,6 +295,18 @@ func DefineEvaluator(g *Genkit, provider, name string, options *ai.EvaluatorOpti
 	return evaluator
 }
 
+// DefineBatchEvaluator registers the given evaluator function as an action, and
+// returns a [Evaluator] that runs it. This method provide the full
+// [EvaluatorRequest] to the callback function, giving more flexibilty to the
+// user for processing the data, such as batching or parallelization.
+func DefineBatchEvaluator(g *Genkit, provider, name string, options *ai.EvaluatorOptions, eval func(context.Context, *ai.EvaluatorRequest) (*ai.EvaluatorResponse, error)) ai.Evaluator {
+	evaluator, err := ai.DefineBatchEvaluator(g.reg, provider, name, options, eval)
+	if err != nil {
```
apascal07, line 304: I just noticed this, we should be returning the error here and in `DefineEvaluator`. We don't need to keep the interface the same as other primitives if it's possible to error during definition.

apascal07 (approved):
LGTM other than error return change.

PR #2399, opened 2025-03-20
fix(go): Added simulated system prompt support

sahdev77:
Addition of simulateSystemPrompt middleware to support the system prompt coming in request for the model which are not supporting the system role.
Resolving : #1927 

Checklist (if applicable):


go/ai/generate.go
```diff
@@ -99,7 +99,7 @@ func DefineModel(r *registry.Registry, provider, name string, info *ModelInfo, f
 		metadata["label"] = info.Label
 	}
 
-	fn = core.ChainMiddleware(ValidateSupport(name, info))(fn)
+	fn = core.ChainMiddleware(simulateSystemPrompt(info, nil), ValidateSupport(name, info))(fn)
```
apascal07, line 102: It probably makes sense to not export these since we don't want users to accidentally call them twice. Can you adjust `ValidateSupport` to be unexported as well?

go/ai/model_middleware.go
```diff
@@ -10,6 +10,48 @@ import (
...
+			copy(modifiedMessages, input.Messages)
+			for i, message := range input.Messages {
+				if message.Role == "system" {
+					systemPrompt := message.Content
+					userMessage := &Message{
+						Role:    "user",
+						Content: append([]*Part{NewTextPart(preface)}, systemPrompt...),
+					}
+					modelMessage := &Message{
+						Role:    "model",
```
apascal07, line 42: Can use `NewModelTextMessage` here instead.

apascal07 (approved):
Two minor changes but overall, LGTM!

go/ai/generate.go
sahdev77: Done

PR #2402, opened 2025-03-20
feat(go): Added media download middleware (enabled for Google GenAI)

sahdev77:
Addition of downloadRequestMedia middleware to download the media  as per the url in request for the model which  supporting the media.
Resolving : https://github.com/firebase/genkit/issues/1928

Checklist (if applicable):


go/ai/generate.go
```diff
@@ -99,7 +99,7 @@ func DefineModel(r *registry.Registry, provider, name string, info *ModelInfo, f
 		metadata["label"] = info.Label
 	}
 
-	fn = core.ChainMiddleware(ValidateSupport(name, info))(fn)
+	fn = core.ChainMiddleware(ValidateSupport(name, info), DownloadRequestMedia(nil))(fn)
```
apascal07: We could support this at the framework level but it's possible that a user does not want to download and inline media for a particular model or use case.

Let's add this directly to the Google AI and Vertex AI plugins like in JS: https://github.com/firebase/genkit/blob/b0151b256a0c18e2685b8ddad8f9104872425632/js/plugins/googleai/src/gemini.ts#L762

Go (intentionally) does not support action-level middleware like JS does so we'll want to do the same `core.ChainMiddleware` around the actual generation function implemented there.

apascal07 (commented):
Implementation and tests look good but could you please write or modify a sample in `go/samples` with a prompt that does inject a media URL?

go/ai/generate.go
sahdev77: Got it or can we provide  config to set list of host and absolute urls which can be auto applied in the filtering criteria to exclude from download request.

apascal07: I would say this middleware is more niche and the user should have full control over it. For example, their prompts may just be working with URLs but do not care at all about their contents. We want to provide the functionality but it's not as broad as some of the others to be included at the framework level.

sahdev77: The middleware is now importable and is being removed to be generic applicable . Also added in the gemini go with the same configuration that we have in type script.

apascal07 (approved):

Issue #2240, opened 2025-03-04
[Go] Add dotprompt `DefinePartial` and `DefineHelper`


apascal07:
**Describe the bug**
https://github.com/firebase/genkit/blob/c450b638f522f5bd831d1146538fab9df2fd8d04/js/ai/src/prompt.ts#L717

PR #2413, opened 2025-03-21
feat(go/genkit): Added loading of dotprompt files on `Genkit` init

rutujaD1999:
 - https://github.com/firebase/genkit/issues/2316

Checklist (if applicable):

Functionalities added:
- Added a new field called dotprompt in the registry structure
- Added loadprompt functionality to load prompts by default if there exists any during genkit init()
- Some fields are added by creating a structure and then sent to DefinePrompt function because the expected type in the With functions are different than that returned by Dotprompt. Directly assigning these fields due to conversion issues one type to another.
- Definepartials function is not complete yet and will be updated as part of issue #2240 
- The promptOptions fr definePrompt function call has been added similar to structure returned here : https://github.com/firebase/genkit/blob/main/js/ai/src/prompt.ts#L770


go/ai/prompt.go
```diff
@@ -426,3 +430,149 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+		log.Printf("Failed to parse prompt: %v", err)
+		return
+	}
+	metadata, err := registry.Dotprompt.RenderMetadata(string(source), &parsedPrompt.PromptMetadata)
+	if err != nil {
+		log.Printf("Failed to render metadata: %v", err)
+		return
+	}
+
+	commonOpts := commonOptions{
```
apascal07, line 512: This is perfectly fine how you are doing it here but I think it would be best to use the option interface like a user, so instead of creating the options struct manually, create a list of options that you append to like:

```go
commonOpts := []CommonOption{WithModelName(metadata.Model)}

if toolChoice, ok := metadata.Raw["toolChoice"].(ToolChoice); ok {
	commonOpts = append(commonOpts, WithToolChoice(toolChoice))
}

...
```

This way if we ever add any processing in those option functions, it will take affect here too. The only one that it won't work for is the default input/schema and it's ok to just have the option like you have it there.

go/ai/prompt.go
```diff
@@ -426,3 +430,149 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
 }
+
+// LoadPromptFolder loads prompts and partials from the input directory for the given namespace
+func LoadPromptFolder(registry *registry.Registry, dir string, ns string) {
+	if dir == "" {
+		dir = "./prompts"
+	}
+	promptsPath, err := filepath.Abs(dir)
+	if err != nil {
+		log.Printf("Failed to resolve directory: %v", err)
```
apascal07, line 441: Let's return an error if the prompt dir specified doesn't exist then only log errors if the prompts can't be parsed. The directory might be tough to specify correctly so you'd want to know if it's wrong but then you might have some broken .prompt files in there that you don't care about so you don't want to necessarily error on init for those.

go/ai/prompt.go
```diff
@@ -426,3 +430,149 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+	for _, entry := range entries {
+		fileName := entry.Name()
+		if entry.IsDir() {
+			LoadPromptFolderRecursively(registry, dir, ns, filepath.Join(subDir, fileName))
+		} else if strings.HasSuffix(fileName, ".prompt") {
+			if strings.HasPrefix(fileName, "_") {
+				partialName := strings.TrimSuffix(fileName[1:], ".prompt")
+				source, err := os.ReadFile(filepath.Join(parentPath, fileName))
+				if err != nil {
+					log.Printf("Failed to read partial file: %v", err)
```
apascal07, line 469: Please use `slog.Error()` here and elsewhere when logging errors.

go/ai/prompt.go
```diff
@@ -426,3 +430,149 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+					continue
+				}
+				definePartial(registry, partialName, string(source))
+				log.Printf("Registered Dotprompt partial \"%s\" from \"%s\"", partialName, filepath.Join(parentPath, fileName))
```
apascal07, line 473:
```suggestion
				slog.Debug("Registered Dotprompt partial \"%s\" from \"%s\"", partialName, filepath.Join(parentPath, fileName))
```

apascal07 (changes requested):
Please add a sample to `go/samples` that has an actual .prompt file in a prompts directory.

go/ai/prompt.go
```diff
@@ -426,3 +430,149 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+	promptOpt := promptOptions{
+		commonOptions: commonOpts,
+		DefaultInput:  metadata.Input.Default,
+	}
+
+	if inputSchema, ok := metadata.Input.Schema.(*jsonschema.Schema); ok {
+		promptOpt.InputSchema = inputSchema
+	}
+
+	outputSchemaMap := map[string]any{}
```
apascal07, line 542: Just noticed, this isn't quite right when combined with `WithOutputType`. `WithOutputType` will reflect the type into a JSON schema. You will want to create the options struct manually here as well like with input.

go/ai/prompt.go
```diff
@@ -426,3 +430,149 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+			log.Printf("Error while unmarshaling output schema: %v", err)
+			return
+		}
+	}
+
+	toolRefs := make([]ToolRef, len(metadata.Tools))
+	for i, tool := range metadata.Tools {
+		toolRefs[i] = ToolName(tool) // Assuming ToolRef has a Name field
+	}
+	DefinePrompt(registry, registryDefinitionKey(name, variant, ns), &promptOpt, WithMetadata(metadata.Metadata), WithDescription(metadata.Description), WithOutputType(outputSchemaMap), WithConfig(metadata.Config), WithTools(toolRefs...))
```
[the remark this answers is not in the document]
apascal07: `DefinePrompt` returns an err, let's return it out of this function as well.

go/genkit/genkit.go
apascal07, line 1: Please add a comment on `Genkit.PromptDir` here explaining that prompts will be automatically loaded from this directory on Genkit initialization. Invalid prompt files will log errors whereas valid prompt files that result in invalid prompt definitions will result in errors.

Maybe move the comment to above the field instead of next to it since it'll be hard to read otherwise.

go/genkit/genkit.go
```diff
@@ -107,6 +107,8 @@ func Init(ctx context.Context, opts ...genkitOption) (*Genkit, error) {
 		}
 	}
 
+	ai.LoadPromptFolder(r, params.PromptDir, "")
```
apascal07, line 110: Let's put this before the reflection server starts up.

go/ai/prompt.go
apascal07: Actually, strike that.

Since it's not just input, let's forego all `WithX` and create a `promptOptions` struct directly and set all options that way. I think it'll be worse having a mix long term since we'll still have to make sure that these exception fields behave the same.

go/ai/prompt.go
apascal07, line 560: Missing output format here I think.

go/ai/prompt.go
```diff
@@ -426,3 +430,149 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+
+}
+
+// registryDefinitionKey generates a unique key for the prompt in the registry.
+func registryDefinitionKey(name string, variant string, ns string) string {
```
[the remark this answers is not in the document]
rutujaD1999: Replaced it with namespace

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+		dir = "./prompts"
+		useDefaultDir = true
+	}
+	promptsPath, err := filepath.Abs(dir)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to resolve directory: %v", err)
+		if !useDefaultDir {
+			return errors.New(errMsg)
+		}
+		slog.Error(errMsg)
```
apascal07, line 447:
```suggestion
		if !useDefaultDir {
			return fmt.Errorf("failed to resolve prompt directory %q: %w, dir, err)
		}
		slog.Debug("default prompt directory not found, skipping loading .prompt files", "dir", dir)
```

We probably don't want to be spamming error logs when the user doesn't even want to use this feature. The differentiation between default prompt dir and a manually set one is good.

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+		return nil
+	}
+	if _, err := os.Stat(promptsPath); os.IsNotExist(err) {
+		errMsg := fmt.Sprintf("Prompts directory does not exist: %s", promptsPath)
+		if !useDefaultDir {
+			return errors.New(errMsg)
+		}
+		slog.Error(errMsg)
```
apascal07, line 455:
```suggestion
		if !useDefaultDir {
			return fmt.Errorf("failed to resolve prompt directory %q: %w, dir, err)
		}
		slog.Debug("Default prompt directory not found, skipping loading .prompt files", "dir", dir)
```

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+	return LoadPromptFolderRecursively(r, promptsPath, namespace, "")
+}
+
+// LoadPromptFolderRecursively recursively loads prompts and partials from the directory.
+func LoadPromptFolderRecursively(r *registry.Registry, dir string, namespace string, subDir string) error {
+	parentPath := filepath.Join(dir, subDir)
+	entries, err := os.ReadDir(parentPath)
+	if err != nil {
+		slog.Error("Failed to read directory", "error", err.Error())
+		return nil
```
apascal07, line 467:
```suggestion
		return fmt.Errorf("failed to read prompt directory structure: %w, err)
```

This is unlikely to happen so it's ok for it to error out. It's probably some permissions issue or something more obscure.

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+	}
+
+	for _, entry := range entries {
+		fileName := entry.Name()
+		if entry.IsDir() {
+			err := LoadPromptFolderRecursively(r, dir, namespace, filepath.Join(subDir, fileName))
+			if err != nil {
+				return err
+			}
```
apascal07, line 476:
```suggestion
			if err := LoadPromptFolderRecursively(r, dir, namespace, filepath.Join(subDir, fileName)); err != nil {
				return err
			}
```

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+		} else if strings.HasSuffix(fileName, ".prompt") {
+			if strings.HasPrefix(fileName, "_") {
+				partialName := strings.TrimSuffix(fileName[1:], ".prompt")
+				source, err := os.ReadFile(filepath.Join(parentPath, fileName))
+				if err != nil {
+					slog.Error("Failed to read partial file", "error", err.Error())
+					continue
+				}
+				definePartial(r, partialName, string(source))
+				slog.Debug("Registered Dotprompt partial \"%s\" from \"%s\"", partialName, filepath.Join(parentPath, fileName))
```
apascal07, line 486:
```suggestion
				slog.Debug("Registered Dotprompt partial %q from %q", partialName, filepath.Join(parentPath, fileName))
```

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+	return nil
+}
+
+// definePartial registers a partial template in the registry.
+func definePartial(r *registry.Registry, name string, source string) {
+	// TODO: Add this functionality
+}
+
+// loadPrompt loads a single prompt into the registry.
+func loadPrompt(r *registry.Registry, path string, filename string, prefix string, namespace string) error {
```
apascal07, line 504:
```suggestion
func LoadPrompt(r *registry.Registry, path, filename, prefix, namespace string) (*Prompt, error) {
```

You can omit the type if consecutive args have the same one. Also we should export this and expose it as a function under `Genkit` (like `genkit.DefinePrompt`) with this signature:

```go
func LoadPrompt(g *genkit.Genkit, filepath string) (*Prompt, error)
```

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+	name := strings.TrimSuffix(filename, ".prompt")
+	var variant string
+	if strings.Contains(name, ".") {
+		parts := strings.SplitN(name, ".", 2)
+		name = parts[0]
+		variant = parts[1]
```
apascal07, line 510:
```suggestion
	name, variant, _ := strings.Cut(name, ".")
```

I think this will do the same, please verify.

go/ai/prompt.go
```diff
@@ -426,3 +430,174 @@ func renderDotprompt(templateText string, variables map[string]any, defaultInput
...
+	}
+
+	source, err := os.ReadFile(filepath.Join(path, prefix, filename))
+	if err != nil {
+		slog.Error("Failed to read prompt file", "error", err.Error())
```
apascal07, line 515: To this error and the ones below, can you add the prompt's relative filepath including filename too so the user know which file failed?

go/samples/prompts-dir/main.go
```diff
@@ -0,0 +1,64 @@
...
+	// Initialize the Google AI plugin. When you pass nil for the
+	// Config parameter, the Google AI plugin will get the API key from the
+	// GOOGLE_GENAI_API_KEY environment variable, which is the recommended
+	// practice.
+	if err := googleai.Init(ctx, g, nil); err != nil {
+		log.Fatal(err)
+	}
+
+	// Define a simple flow that prompts an LLM to generate menu suggestions.
+	genkit.DefineFlow(g, "menuSuggestionFlow", func(ctx context.Context, input string) (string, error) {
```
apascal07, line 36: Can you adjust this flow to lookup the prompt we loaded in automatically and use it? That is the user journey we're targeting here so it'd be good to have that covered fully in a sample.

apascal07 (changes requested):

go/samples/prompts-dir/main.go
rutujaD1999: Removed the LoadPromptModel function as the model is available in the prompt itself. The modelname was being overwritten in the Execute() due to which it was not being detected. Fixed this issue.

PR #2417, opened 2025-03-21
feat(go/genkit): Added support for formatters and custom instructions

dysrama:
Checklist (if applicable):


go/ai/option.go
```diff
@@ -430,6 +446,20 @@ func WithOutputFormat(format OutputFormat) OutputOption {
 	return &outputOptions{OutputFormat: format}
 }
 
+// WithOutputInstructions sets whether output instructions should be added.
+func WithOutputInstructions(addInstructions bool) OutputOption {
```
apascal07, line 450: This is supposed to take in a string, which is the user's desired instructions to use in lieu of the default instructions. We store it as a pointer to a string. If it's nil, that means the user wants default instructions. If it's an empty string (not nil), the user does not want instructions to be added. If it's not an empty string, we use that as the instructions.

go/ai/option.go
```diff
@@ -430,6 +446,20 @@ func WithOutputFormat(format OutputFormat) OutputOption {
...
+	var instructions string
+	if addInstructions {
+		instructions = ""
+	}
+	return &outputOptions{OutputInstructions: &instructions}
+}
+
+// WithOutputNativeConstrained sets whether to use model native constrained output generation. Defaults to true.
```
apascal07, line 458:
```suggestion
// WithOutputNativeConstrained sets whether to use model native constrained output generation. If this is set to true and the model does not support native constrained generation, it will be simulated instead. Defaults to true.
```

go/ai/format_json.go
```diff
@@ -0,0 +1,101 @@
...
+	if schema != nil {
+		jsonBytes, err := json.Marshal(schema)
+		if err != nil {
+			instructions = fmt.Sprintf("Error marshalling schema to JSON: %v", err)
+		} else {
+			escapedJSON := strconv.Quote(string(jsonBytes))
+			instructions = fmt.Sprintf("Output should be in JSON format and conform to the following schema:\n\n```%s```", escapedJSON)
+		}
+	}
+	constrained := true
```
apascal07, line 58: Inline this?

go/ai/format_jsonl.go
```diff
@@ -0,0 +1,155 @@
...
+	if schema != nil && isJSONL(schema) {
+		jsonBytes, err := json.Marshal(schema["items"])
+		if err != nil {
+			instructions = fmt.Sprintf("Error marshalling schema to JSONL: %v", err)
+		} else {
+			escapedJSON := strconv.Quote(string(jsonBytes))
+			instructions = fmt.Sprintf("Output should be JSONL format, a sequence of JSON objects (one per line) separated by a newline '\\n' character. Each line should be a JSON object conforming to the following schema:\n\n```%s```", escapedJSON)
+		}
+	} else {
+		instructions = "Error, schema not valid JSONL"
```
apascal07, line 59: Like the other one, we should probably panic instead.

go/ai/option.go
```diff
@@ -430,6 +446,20 @@ func WithOutputFormat(format OutputFormat) OutputOption {
...
+// Nil/not set: default instructions will be added.
+// Empty string: no instructions will be added.
+// Non-empty string: will be used as instructions.
+func WithOutputInstructions(instructions *string) OutputOption {
+	return &outputOptions{OutputInstructions: instructions}
+}
+
+// WithOutputNativeConstrained sets whether to use model native constrained output generation.
+// If this is set to true and the model does not support native constrained generation, it will be simulated instead. Defaults to true.
+func WithOutputNativeConstrained(constrained bool) OutputOption {
```
apascal07, line 459: I wonder if this should be something like `WithUseSimulatedConstrained()` with no arg. Since native is the default and we still fallback to simulated if the model chosen doesn't support native, there's no much point to passing in a bool.

PR #2423, opened 2025-03-21
refactor(go): Refactored plugin interface and added `WithPlugins` to `Genkit`.

apascal07:
Checklist (if applicable):

### DIFF
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -24,8 +24,6 @@ import (
 
 // Plugin is a common interface for plugins.
 type Plugin interface {
-	// Name returns the name of the plugin.
-	Name() string
 	// Init initializes the plugin.
 	Init(ctx context.Context, g *Genkit) error
 }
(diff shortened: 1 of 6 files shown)


xavidop (approved):

go/genkit/genkit.go
```diff
@@ -35,65 +35,100 @@ import (
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 )
 
+// Plugin is a common interface for plugins.
+type Plugin interface {
```
pavelgj, line 39: wondering if this is better as an abstract class.We plan to add `ResolveAction(ctx context.Context, g *Genkit, n: string)` method, which should be optional.

pavelgj (approved):

go/genkit/genkit.go
apascal07: As discussed, dynamic resolution will be implemented as a separate `DynamicPlugin` interface that we will try to cast to during generation.

PR #2461, opened 2025-03-25
chore(js/plugins/googleai): expose gemini-2.5-pro-exp-03-25 (`gemini25ProExp0325`)

pavelgj:
```ts
import { gemini25ProExp0325, googleAI } from '@genkit-ai/googleai';

const ai = genkit({ 
  plugins: [ googleAI() ],
  model: gemini25ProExp0325,
});

const { text } = await await ai.generate('hi');
```

Checklist (if applicable):

### DIFF
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -250,6 +250,23 @@ export const gemini20ProExp0205 = modelRef({
   configSchema: GeminiConfigSchema,
 });
 
+export const gemini25ProExp0325 = modelRef({
+  name: 'googleai/gemini-2.5-pro-exp-03-25',
+  info: {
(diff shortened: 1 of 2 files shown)


apascal07 (approved):

GitTom:
The way this PR handles config.versions, to me, it seems inconsistent with the dotprompt documentation and with the way the OpenAI plugin does it.

I don't know which is right, but this has already tripped me up and I think there is going to a fair bit of confusion on this issue.

pavelgj:
> The way this PR handles config.versions, to me, it seems inconsistent with the dotprompt documentation and with the way the OpenAI plugin does it.
> 
> I don't know which is right, but this has already tripped me up and I think there is going to a fair bit of confusion on this issue.

There are multiple ways of accessing different versions of the model... we are working on a bit of a redesign for it. Stay tuned.

PR #2481, opened 2025-03-27
feat(go/plugins): Add model `stage` support

hugoaguirre:
Models now expose their current release state (deprecated, unstable, legacy, featured, etc) 

Checklist (if applicable):

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log/slog"
 	"strconv"
 	"strings"
 
@@ -172,18 +171,6 @@ func GenerateWithRequest(ctx context.Context, r *registry.Registry, opts *Genera
 	model, err := LookupModelByName(r, opts.Model)
 	if err != nil {
 		return nil, err
-	}
-
-	metadata, err := modelMetadata(r, opts.Model)
-	if err != nil {
-		return nil, err
-	}
-
-	switch metadata.Stage {
-	case ModelStageDeprecated:
-		slog.Warn(fmt.Sprintf("model: %s is deprecated and may be removed in a future release", opts.Model))
-	case ModelStageUnstable:
-		slog.Warn(fmt.Sprintf("model: %s is unstable and functionality might be compromised", opts.Model))
 	}
 
 	toolDefMap := make(map[string]*ToolDefinition)
@@ -392,7 +379,6 @@ func modelMetadata(r *registry.Registry, name string) (*ModelInfo, error) {
 	if name == "" {
 		return nil, errors.New("ai.modelMetadata: model name is empty")
 	}
-
 	provider, name, found := strings.Cut(name, "/")
 	if !found {
 		name = provider
@@ -408,7 +394,6 @@ func modelMetadata(r *registry.Registry, name string) (*ModelInfo, error) {
 	}
 
 	metadata := action.Desc().Metadata
-
 	m, ok := metadata["model"].(map[string]any)
 	if !ok {
 		return nil, fmt.Errorf("ai.modelMetadata: model metadata not definedd")
@@ -436,6 +421,7 @@ func modelMetadata(r *registry.Registry, name string) (*ModelInfo, error) {
 	if ms, ok := m["stage"].(ModelStage); ok {
 		info.Stage = ms
 	}
+
 	return info, nil
 }
 
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


go/ai/generate.go
```diff
@@ -172,6 +174,18 @@ func GenerateWithRequest(ctx context.Context, r *registry.Registry, opts *Genera
...
 	}
 
+	metadata, err := modelMetadata(r, opts.Model)
+	if err != nil {
+		return nil, err
+	}
+
+	switch metadata.Stage {
+	case ModelStageDeprecated:
+		slog.Warn(fmt.Sprintf("model: %s is deprecated and may be removed in a future release", opts.Model))
```
hugoaguirre, line 184: In JS the warning flag is always set to `true`, that's why we are always logging here.

QQ: Is this the right way to log these types of warnings?

go/plugins/googlegenai/models.go
```diff
@@ -58,6 +58,7 @@ var (
 				"gemini-1.5-flash-002",
 			},
 			Supports: &gemini.Multimodal,
+			Stage:    ai.ModelStageStable,
```
hugoaguirre, line 61: A more accurate way to determine the current stage of a model would be to set the stage value for every version available in the model.
https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions#stable-version

go/ai/generate.go
```diff
@@ -172,6 +174,18 @@ func GenerateWithRequest(ctx context.Context, r *registry.Registry, opts *Genera
 		return nil, err
 	}
 
+	metadata, err := modelMetadata(r, opts.Model)
```
apascal07, line 177: Any reason why this can't be in `validateSupport` in `model_middleware.go`?

hugoaguirre: Hmm no particular reason, I was kind of biased by the constrained generation check we do at this place and wanted to make use of it as well. But definitely makes more sense to place it in the middleware 🤔

go/ai/generate.go
```diff
@@ -373,6 +374,57 @@ func (m *modelActionDef) Generate(ctx context.Context, req *ModelRequest, cb Mod
 	return (*ModelAction)(m).Run(ctx, req, cb)
 }
 
+// modelMetadata retrieves the model's metadata
+func modelMetadata(r *registry.Registry, name string) (*ModelInfo, error) {
```
apascal07, line 378: Revert?

go/plugins/googlegenai/models.go
```diff
@@ -75,40 +77,47 @@ var (
...
 			Label:    "Gemini 2.0 Pro Exp 02-05",
 			Versions: []string{},
 			Supports: &gemini.Multimodal,
+			Stage:    ai.ModelStageStable,
 		},
 		gemini20FlashThinkingExp0121: {
 			Label:    "Gemini 2.0 Flash Thinking Exp 01-21",
 			Versions: []string{},
 			Supports: &gemini.Multimodal,
+			Stage:    ai.ModelStageStable,
```
apascal07, line 114: Anything marked with "exp" or "preview" is experimental so should be `ai.ModelStageUnstable`.

go/ai/model_middleware.go
```diff
@@ -125,6 +127,15 @@ func validateSupport(model string, info *ModelInfo) ModelMiddleware {
...
 			}
 
+			if len(info.Stage) > 0 {
+				switch info.Stage {
+				case ModelStageDeprecated:
+					logger.FromContext(ctx).Warn("model is deprecated and may be removed in a future release", "model", model)
+				case ModelStageUnstable:
+					logger.FromContext(ctx).Warn("model is unstable and functionality might be compromised", "model", model)
+				}
+			}
```
apascal07, line 137:
```suggestion
				switch info.Stage {
				case ModelStageDeprecated:
					logger.FromContext(ctx).Warn("model is deprecated and may be removed in a future release", "model", model)
				case ModelStageUnstable:
					logger.FromContext(ctx).Warn("model is unstable and functionality might be compromised", "model", model)
				default:
				}
```

go/ai/model_middleware.go
```diff
@@ -125,6 +127,15 @@ func validateSupport(model string, info *ModelInfo) ModelMiddleware {
 				}
 			}
 
+			if len(info.Stage) > 0 {
```
apascal07, line 130: This is a weird check for a string.

go/ai/model_middleware.go
```diff
@@ -125,6 +127,15 @@ func validateSupport(model string, info *ModelInfo) ModelMiddleware {
...
+				switch info.Stage {
+				case ModelStageDeprecated:
+					logger.FromContext(ctx).Warn("model is deprecated and may be removed in a future release", "model", model)
+				case ModelStageUnstable:
+					logger.FromContext(ctx).Warn("model is unstable and functionality might be compromised", "model", model)
```
apascal07, line 135:
```suggestion
					logger.FromContext(ctx).Info("model is experimental or unstable", "model", model)
```

apascal07 (approved):

PR #2487, opened 2025-03-27
feat(go): Added `ModelArg` interface and `ModelRef`

huangjeff5:


go/ai/generate.go
```diff
@@ -39,6 +39,20 @@ type (
...
+	// ModelArg is the interface for model arguments.
+	ModelArg interface {
+		Name() string
+		Resolve(context.Context, *registry.Registry) (Model, error)
+	}
+
+	// ModelRef is a struct to hold model name and configuration.
+	ModelRef struct {
+		name   string
+		config any
```
apascal07, line 55:
```suggestion
		Config any
```

go/plugins/googlegenai/gemini.go
```diff
@@ -69,6 +72,265 @@ type EmbedOptions struct {
...
+
+// extractConfigFromInput converts any supported config type to GoogleAIConfig
+func extractConfigFromInput(input *ai.ModelRequest) (GeminiConfig, error) {
+	var result GeminiConfig
+	switch config := input.Config.(type) {
+	case GeminiConfig:
+		return config, nil
+	case *GeminiConfig:
+		return *config, nil
+	case ai.GenerationCommonConfig:
```
apascal07, line 277: Instead of duplicating this logic, can we either have a helper function or just detect that it is/isn't a pointer and take the pointer of it? Doesn't seem necessary to be this many lines.

go/plugins/googlegenai/gemini.go
```diff
@@ -69,6 +72,265 @@ type EmbedOptions struct {
...
+			}
+			if config.TopP != 0 {
+				v := float32(config.TopP)
+				result.TopP = &v
+			}
+			result.Version = config.Version
+		}
+		return result, nil
+	case map[string]any:
+		// Todo: this will silently fail if extra parameters are passed, may want to expose errors
```
apascal07, line 320: Can you log warnings for now?

go/plugins/googlegenai/gemini.go
```diff
@@ -69,6 +72,265 @@ type EmbedOptions struct {
...
+		if err := mapToStruct(config, &result); err == nil {
+			return result, nil
+		} else {
+			return result, err
```
apascal07, line 324: Let's return nil here and elsewhere anytime there's an error. Even though it's nil currently, it's not easy to tell at a glance.

go/plugins/googlegenai/gemini.go
```diff
@@ -69,6 +73,152 @@ type EmbedOptions struct {
...
+		return config, nil
+	case ai.GenerationCommonConfig:
+		return convertFromCommonConfig(config), nil
+	case *ai.GenerationCommonConfig:
+		if config == nil {
+			return &result, nil
+		}
+		return convertFromCommonConfig(*config), nil
+	case map[string]any:
+		log.Printf("Warning: Using map[string]any for config may silently ignore unknown parameters")
```
apascal07, line 196: I was imagining we would check for unknown fields here and warn only then but that might be complex. Let's add the TODO back in and get rid of this -- there's nothing wrong with using a map *if* there are no unknown fields.

go/plugins/googlegenai/gemini.go
```diff
@@ -69,6 +73,152 @@ type EmbedOptions struct {
...
+
+// GeminiConfig mirrors GenerateContentConfig without direct genai dependency
+type GeminiConfig struct {
+	ai.GenerationCommonConfig
+
+	// Safety settings
+	SafetySettings []*SafetySetting `json:"safetySettings,omitempty"`
+
+	// Model version
+	Version string `json:"version,omitempty"`
```
apascal07, line 177: This is in `GenerationCommonConfig` no?

go/plugins/googlegenai/gemini.go
```diff
@@ -69,6 +73,152 @@ type EmbedOptions struct {
...
+		return &config, nil
+	case *GeminiConfig:
+		return config, nil
+	case ai.GenerationCommonConfig:
+		return convertFromCommonConfig(config), nil
+	case *ai.GenerationCommonConfig:
+		if config == nil {
+			return &result, nil
+		}
+		return convertFromCommonConfig(*config), nil
```
apascal07, line 194: I don't think we need this anymore since we embedded it in `GeminiConfig`.

apascal07: Did you decide not to do this?

apascal07: Let's fix this in a follow up PR because we need to unblock Hugo's PR.

apascal07 (approved):

go/plugins/googlegenai/gemini.go
huangjeff5: I thought I addressed this here: https://github.com/firebase/genkit/pull/2487/commits/8437142b6b39792b027e75f615cacbd370cd1182

PR #2524, opened 2025-03-28
feat(py): add system instructions support for openai-compat plugin

timere-nemo:
Add system instructions support: https://github.com/firebase/genkit/issues/2500

Checklist (if applicable):

### DIFF
--- a/py/plugins/compat-oai/src/genkit/plugins/compat_oai/models/model.py
+++ b/py/plugins/compat-oai/src/genkit/plugins/compat_oai/models/model.py
@@ -46,6 +46,8 @@ class OpenAIModel:
     Handles OpenAI API interactions for the Genkit plugin.
     """
 
+    _role_map = {Role.SYSTEM: 'developer', Role.MODEL: 'assistant'}
+
     def __init__(self, model: str, client: OpenAI, registry: GenkitRegistry):
         """
         Initializes the OpenAIModel instance with the specified model and OpenAI client parameters.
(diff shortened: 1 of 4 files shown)


py/plugins/compat-oai/src/genkit/plugins/compat_oai/models/model.py
```diff
@@ -46,6 +46,8 @@ class OpenAIModel:
     Handles OpenAI API interactions for the Genkit plugin.
     """
 
+    _role_map = {Role.SYSTEM: 'developer', Role.MODEL: 'assistant'}
```
pavelgj, line 49: openai has `system` role. why are you using `developer`?

timere-nemo: I was using this documentation: https://platform.openai.com/docs/guides/prompt-engineering#messages-and-roles

Role description:
developer: Instructions to the model that are prioritized ahead of user messages, following [chain of command](https://cdn.openai.com/spec/model-spec-2024-05-08.html#follow-the-chain-of-command). Previously called the system prompt.

pavelgj: IIUC you should still use `system` -- https://community.openai.com/t/how-is-developer-message-better-than-system-prompt/1062784

timere-nemo: @pavelgj As I understand, I can use both, and they handle the role themselves. But to keep our code cleaner, I will remove the `developer` role in the next PR. Is that okay?

Issue #2365, opened 2025-03-17
[Go] Integrate `google/dotprompt` into Genkit


apascal07:
**Describe the bug**
Genkit should use `google/dotprompt` package.

PR #2526, opened 2025-03-28
feat(go/genkit): Use `dotprompt` package for rendering prompt templates

rutujaD1999:
 - #2365 
 - Add dotprompt functions for rendering prompts
 - Modify metadata to pass template data as well 
 - Add input from ActionCtx as input for rendering the templates
 - Removed predefined helpers from prompt.go since they are already declared in the dotprompt function : https://github.com/google/dotprompt/blob/main/go/dotprompt/helper.go#L26

Checklist (if applicable):

### DIFF
--- a/go/ai/prompt.go
+++ b/go/ai/prompt.go
@@ -26,11 +26,11 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/aymerick/raymond"
 	"github.com/firebase/genkit/go/core"
 	"github.com/firebase/genkit/go/core/logger"
 	"github.com/firebase/genkit/go/internal/atype"
 	"github.com/firebase/genkit/go/internal/registry"
+	"github.com/google/dotprompt/go/dotprompt"
 	"github.com/invopop/jsonschema"
 )
 
@@ -260,15 +260,15 @@ func (p *Prompt) buildRequest(ctx context.Context, input any) (*GenerateActionOp
 	}
 
 	messages := []*Message{}
-	messages, err = renderSystemPrompt(ctx, p.promptOptions, messages, m, input)
+	messages, err = renderSystemPrompt(ctx, p.promptOptions, messages, m, input, p.registry.Dotprompt)
 	if err != nil {
 		return nil, err
 	}
-	messages, err = renderMessages(ctx, p.promptOptions, messages, m, input)
+	messages, err = renderMessages(ctx, p.promptOptions, messages, m, input, p.registry.Dotprompt)
 	if err != nil {
 		return nil, err
 	}
-	messages, err = renderUserPrompt(ctx, p.promptOptions, messages, m, input)
+	messages, err = renderUserPrompt(ctx, p.promptOptions, messages, m, input, p.registry.Dotprompt)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


go/ai/prompt.go
```diff
@@ -353,10 +353,14 @@ func renderMessages(ctx context.Context, opts promptOptions, messages []*Message
...
 		for _, part := range msg.Content {
 			if part.IsText() {
-				rendered, err := renderDotprompt(part.Text, input, opts.DefaultInput)
+				parts, err := renderPrompt(ctx, opts, part.Text, input, dp)
 				if err != nil {
 					return nil, err
 				}
+				var rendered string
+				if len(parts) != 0 {
+					rendered = parts[0].Text
```
apascal07, line 362: I'm not quite following this logic, we are iterating over messages and over parts but then we're setting a fixed 0 index on the parts. Is this right?

rutujaD1999: Here we are traversing through every msg and every part for that msg. For each part the renderPrompt function is called which gives []*Part as an output. Since we are already going through every part for every message it can be assumed that the text given as input to renderPrompt function will give output []*Part with only one element. That is why 0 index is used.

go/go.mod
```diff
@@ -13,8 +13,8 @@ require (
 	firebase.google.com/go/v4 v4.14.1
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.46.0
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace v1.22.0
-	github.com/aymerick/raymond v2.0.2+incompatible
-	github.com/google/dotprompt/go v0.0.0-20250320235217-796c6442a3c1
+	github.com/blues/jsonata-go v1.5.4
```
apascal07, line 16: Do we use jsonata directly?

go/ai/prompt.go
```diff
@@ -496,9 +491,24 @@ func loadPromptDir(r *registry.Registry, dir string, namespace string) error {
...
-func definePartial(r *registry.Registry, name string, source string) {
-	// TODO: Add this functionality
+// DefinePartial adds the partial to the list of partials to the dotprompt instance
+func DefinePartial(r *registry.Registry, name string, source string) {
+	if r.Dotprompt == nil {
+		r.Dotprompt = dotprompt.NewDotprompt(&dotprompt.DotpromptOptions{
+			Partials: map[string]string{},
+		})
+	}
+	r.Dotprompt.Partials[name] = source
```
apascal07, line 501: Throw an error if the key already exists. Same as below.

go/ai/prompt.go
```diff
@@ -496,9 +491,24 @@ func loadPromptDir(r *registry.Registry, dir string, namespace string) error {
 	return nil
 }
 
-// definePartial registers a partial template in the registry.
-func definePartial(r *registry.Registry, name string, source string) {
-	// TODO: Add this functionality
+// DefinePartial adds the partial to the list of partials to the dotprompt instance
+func DefinePartial(r *registry.Registry, name string, source string) {
+	if r.Dotprompt == nil {
```
apascal07, line 496: I would say this logic should live in the registry and this function should call `registry.RegisterPartial` and `registry.RegisterHelper`.

go/ai/prompt_test.go
```diff
@@ -1017,3 +1017,32 @@ func TestLoadPromptFolder_DirectoryNotFound(t *testing.T) {
 		t.Fatalf("Prompt should not have been registered for a non-existent directory")
 	}
 }
+
+// TestDefinePartialAndHelperJourney demonstrates a complete user journey for defining
+// and using both partials and helpers.
+func TestDefinePartialAndHelperJourney(t *testing.T) {
```
apascal07, line 1023:
```suggestion
func TestDefinePartialAndHelper(t *testing.T) {
```

go/genkit/genkit.go
```diff
@@ -596,3 +596,31 @@ func LoadPrompt(g *Genkit, path string, namespace string) (*ai.Prompt, error) {
 func RegisterSpanProcessor(g *Genkit, sp sdktrace.SpanProcessor) {
 	g.reg.RegisterSpanProcessor(sp)
 }
+
+// DefinePartial wraps DefinePartial to register a partial template with the given name and source.
+// Partials can be referenced in templates with the syntax {{>partialName}}.
+func DefinePartial(g *Genkit, name string, source string) {
+	if g == nil || g.reg == nil {
+		return
+	}
```
apascal07, line 605:
```suggestion
```

go/genkit/genkit.go
```diff
@@ -596,3 +596,31 @@ func LoadPrompt(g *Genkit, path string, namespace string) (*ai.Prompt, error) {
...
+//		return strings.ToUpper(s)
+//	})
+//
+// In a template, you would use it as:
+//
+//	{{uppercase "hello"}} => "HELLO"
+func DefineHelper(g *Genkit, name string, fn any) {
+	if g == nil || g.reg == nil {
+		return
+	}
```
apascal07, line 624:
```suggestion
```

apascal07 (approved):

PR #2572, opened 2025-03-31
feat(js/plugins/firebase): Data Connect tools (beta), serverApp context.

mbleigh:
- Adds `@genkit-ai/firebase/beta/data-connect` entrypoint to use Firebase Data Connect connector operations as tools in Genkit.
- Adds `serverAppConfig` option to `firebaseContext()` to allow automatic population of a FirebaseServerApp into context.
- Refactors existing entrypoints to use `kebab-case` consistently (keeping old entrypoint for backwards compatibility).


js/plugins/firebase/src/beta/data-connect.ts
```diff
@@ -0,0 +1,203 @@
...
+
+            const { data } = await executeMutation(
+              mutationRef(dc, tool.name, input)
+            );
+            logger.debug(
+              `[dataConnectTools] ${tool.name}(${JSON.stringify(input)}) -> ${JSON.stringify(data)}`
+            );
+            return data;
+          } catch (e) {
+            logger.info('[dataConnectTools] error on tool call:', e);
```
pavelgj, line 192:
```suggestion
            logger.error('[dataConnectTools] error on tool call:', e);
```

js/plugins/firebase/src/context.ts
```diff
@@ -181,6 +208,25 @@ export function firebaseContext<I = any>(
       instanceIdToken = request.headers['firebase-instance-id-token'];
     }
     const context: FirebaseContext = {};
+
+    if (typeof policy === 'object' && policy.serverAppConfig) {
+      // we dynamically import here to keep `firebase` an optional peer dep
+      console.log(
```
pavelgj, line 214: remove?

pavelgj (approved):

js/plugins/firebase/src/beta/data-connect.ts
mbleigh: Explicitly leaving this as an `info` because of the configurable behavior. FWIW I think we might want to adopt this `onError` for tools generally.

js/plugins/firebase/src/context.ts
mbleigh: Done.

PR #2600, opened 2025-04-02
fix(js/plugins/express): Check request.body is defined before reading data from it

ifielker:
If the plugin user misconfigured things, there was a possibility that request.body was undefined.

Checklist (if applicable):

### DIFF
--- a/js/plugins/express/src/index.ts
+++ b/js/plugins/express/src/index.ts
@@ -57,8 +57,8 @@ export function expressHandler<
     if (!request.body) {
       const errMsg =
         `Error: request.body is undefined. ` +
-        `Are you missing 'content-type: application/json' in your headers? ` +
-        `Or did you forget to use 'express.json()'? `;
+        `Possible reasons: missing 'content-type: application/json' in request ` +
+        `headers or misconfigured JSON middleware ('app.use(express.json()')? `;
       logger.error(errMsg);
       response
         .status(400)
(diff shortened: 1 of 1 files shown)


js/plugins/express/src/index.ts
```diff
@@ -54,6 +54,19 @@ export function expressHandler<
     response: express.Response
   ): Promise<void> => {
     const { stream } = request.query;
+    if (!request.body) {
+      const errMsg =
+        `Error: request.body is undefined. ` +
+        `Are you missing 'content-type: application/json' in your headers? ` +
+        `Or did you forget to use 'express.json()'? `;
```
pavelgj, line 61:
```suggestion
        `Error: request.body is undefined. ` +
        `Possible reasons: missing 'content-type: application/json' in request headers or ` +
        `misconfigured JSON middleware ('app.use(express.json()')? `;
```

js/plugins/express/src/index.ts
```diff
@@ -54,6 +54,19 @@ export function expressHandler<
...
+      logger.error(errMsg);
+      response
+        .status(400)
+        .json({ message: errMsg, status: 'BAD REQUEST' })
```
pavelgj, line 65: is this status correct? `INVALID_ARGUMENT`?

ifielker: done

PR #2615, opened 2025-04-03
fix(py/genkit): use the function docstring as the description unless you want to override; saves duplication

yesudeep:
Python already provides a way to provide function descriptions
as docstrings. Let's use those when available to avoid duplication
for action descriptions. The `description` kwarg can be used to
override.

CHANGELOG:
- [ ] Update docs to reflect this.
- [ ] Update decorators and tests.


pavelgj (approved):

pavelgj:
Nice!

PR #2623, opened 2025-04-03
feat(bin/killports): add a convenience script to kill processes bound to a range of ports

yesudeep:
Useful for debugging 'ports in use' errors.

CHANGELOG:
- [ ] add `bin/killports` to kill any processes listening on a range of
  ports.
- [ ] Documentation and example usage is in the script.


zarinn3pal (approved):
LGTM

PR #2643, opened 2025-04-04
feat(go): use `GenkitError` instead of regular `error`

sahdev77:
Addition of new Genkit error in go:
 - Handling Error using Genkit Error.
 - Removal of HTTPError from misc.
Resolving : [2476](https://github.com/firebase/genkit/issues/2476) 
Checklist (if applicable):


apascal07 (commented):
Can you modify a sample to return errors in various places and verify in the Dev UI that the errors look correct and have a useful stack? Conversely, the stack should not appear when calling the prod server, only the reflection server.

go/core/error.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// ToHTTPError returns a JSON-serializable representation for callable responses.
+func (e *GenkitError) ToHTTPError() HTTPError {
+	msg := e.Message
+	return HTTPError{
+		Details: e.Details,
+		Status:  e.Status,
+		Message: msg,
+	}
```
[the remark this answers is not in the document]
sahdev77: As per our discussion i had deleted it as it is not in use

go/core/error.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// getErrorStack extracts stack trace from an error object.
+// This captures the stack trace of the current goroutine when called.
+func getErrorStack(err error) string {
+	if err == nil {
+		return ""
+	}
+	return string(debug.Stack())
```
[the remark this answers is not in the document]
sahdev77: Fixed by creating a NewGenkitError and capture the errStack at inside it.

go/core/status_types.go
```diff
@@ -0,0 +1,153 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package status defines canonical status codes, names, and related types
+// inspired by gRPC status codes.
+package core
```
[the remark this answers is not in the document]
sahdev77: so are you suggesting to create a error package all together ?
As of now it is in sync with the structure that we had in Js and py but go behaves differently so we can think of keeping it as a separate package itself(genkitErr)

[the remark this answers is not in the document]
apascal07: JS and Python allow importing things piecemeal instead of the entire package like in Go so it's more important for things to be encapsulated well.

go/core/error.go
```diff
@@ -0,0 +1,138 @@
...
+	return ge
+}
+
+// Error implements the standard error interface.
+func (e *GenkitError) Error() string {
+	if e == nil {
+		return "<nil GenkitError>"
+	}
+	return e.Message
+}
```
apascal07, line 94:
```suggestion
func (e GenkitError) Error() string {
	return e.Message
}
```

go/core/error.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+// getErrorStack extracts stack trace from an error object.
+// This captures the stack trace of the current goroutine when called.
+func getErrorStack(err error) string {
+	if err == nil {
+		return ""
+	}
+	return string(debug.Stack())
```
apascal07, line 137: This doesn't do what the comment says. There's not really a point to passing in err here, and this whole function can be removed and `debug.Stack()` called directly in its place.

go/core/error.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+// ToReflectionError gets the JSON representation for reflection API Error responses.
+func ToReflectionError(err error) ReflectionError {
+	if ge, ok := err.(*GenkitError); ok {
+		return ge.ToReflectionError()
+	}
+
+	stack := getErrorStack(err)
```
apascal07, line 118: This is still creating an error stack where the `error` it converted to a `ReflectionError`, not where the error is created which is what we want the stack to point to.

go/core/error.go
sahdev77: done

go/core/error.go
sahdev77: done

go/core/error.go
sahdev77: This holds tru in the case where we have any other error other than genkit error. 
Particularly in those cases it become tricky to capture the debug stack for the source of error.

apascal07 (approved):

PR #2645, opened 2025-04-04
fix(py): use the runtime specified by nox with pytest

yesudeep:
CHANGELOG:
- [ ] Use the runtime specified by nox with pytest in `noxfile.py`


zarinn3pal (approved):
LGTM

PR #2659, opened 2025-04-05
chore[js]: move `@types` packages to `devDependencies` without modifying packages version

samuel871211:

Checklist (if applicable):


mbleigh:
Why do this? Most guidance I've seen is to put this in deps, not devDeps. I
think there's too much risk for no enumerated benefit

samuel871211:
@mbleigh 
Thanks for your comment

Based on [npmjs official docs](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file)

```
"devDependencies": Packages that are only needed for local development and testing.
```

`@types/xxx` packages are only needed for local development and testing, so I think these packages should be moved to devDependencies

PR #2695, opened 2025-04-07
fix(py/genkit): use ai.run instead of ai.run_async

yesudeep:
Public APIs are forever. This is the best time to fix this.

Since we decided that Genkit classes should be bifurcated into sync
and async variants, removing this wrapper `ai.run_async`. 
The equivalent functionality is in `ai.run`. The method on the sync
instance should be called the same.


zarinn3pal (approved):
LGTM

PR #2707, opened 2025-04-07
feat(py): context caching sample

kirgrim:


py/samples/context-caching/src/context_caching.py
```diff
@@ -16,15 +16,84 @@
 
 """A stub for the sample to come."""
 
+import requests
+import structlog
+from pydantic import BaseModel, Field
 
-def main() -> None:
+from genkit.ai import Genkit
+from genkit.core.typing import GenerationCommonConfig, Message, Role, TextPart
```
pavelgj, line 24:
```suggestion
from genkit.types import GenerationCommonConfig, Message, Role, TextPart
```

py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
```diff
@@ -528,6 +532,38 @@ def _call_tool(self, call: genai_types.FunctionCall) -> genai_types.Content:
...
+        validate_context_cache_request(request=request, model_name=model_name)
+
+        ttl = cache_config.get('ttl_seconds', DEFAULT_TTL)
+        cache_key = generate_cache_key(request=request)
+
+        iterator_config = genai_types.ListCachedContentsConfig()
+        cache = None
+        pages = await self._client.aio.caches.list(config=iterator_config)
+
+        async for item in pages:
```
pavelgj, line 547: is there a way to look up the cache by name? I don't remember exactly why in JS we also list... would be nice if we didn't have to do it.

pavelgj (approved):

py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
kirgrim: Yeah, so the problem is that we cannot force to set particular `name` for the context caching instance. Instead, I am setting `display_name` and then have to iterate over cache to find matching `display_name` and this way to get the actual name.

In JS its similar, perhaps once we did this operation in our runtime we can do some caching but no native option afaik

pavelgj (approved):

py/samples/context-caching/pyproject.toml
pavelgj, line 1: please rename this sample `google-genai-context-caching`

pavelgj, line 1: could you please also rename the sample folder -- `py/samples/google-genai-context-caching`

pavelgj (approved):

PR #2735, opened 2025-04-08
chore: update model config parameters

hugomurillomtz:
Here's the link to the issue: https://github.com/firebase/genkit/issues/2333. This updates the following plugins: `Ollama`, `vertexai/gemini` and `vertexai/mistral`.

I checked that I am getting the new values on the dev UI by logging out the configuration:

**Ollama:**
image

**vertexai/gemini:**
image

**vertexai/mistral:**
image

Checklist (if applicable):

### DIFF
--- a/js/plugins/ollama/src/index.ts
+++ b/js/plugins/ollama/src/index.ts
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 
-import { Genkit, ToolRequest, ToolRequestPart, ToolResponse } from 'genkit';
+import { Genkit, ToolRequest, ToolRequestPart, ToolResponse, z } from 'genkit';
 import { logger } from 'genkit/logging';
 import {
   GenerateRequest,
   GenerateResponseData,
+  GenerationCommonConfigDescriptions,
   GenerationCommonConfigSchema,
   MessageData,
   ToolDefinition,
@@ -59,6 +60,38 @@ export function ollama(params: OllamaPluginParams): GenkitPlugin {
   });
 }
 
+/**
+ * Please refer to: https://github.com/ollama/ollama/blob/main/docs/modelfile.md
+ * for further information.
+ */
+export const OllamaConfigSchema = GenerationCommonConfigSchema.extend({
+  temperature: z
+    .number()
+    .min(0.0)
+    .max(1.0)
+    .describe(
+      GenerationCommonConfigDescriptions.temperature +
+        ' The default value is 0.8.'
+    )
+    .optional(),
+  topK: z
+    .number()
+    .min(1)
+    .max(40)
+    .describe(
+      GenerationCommonConfigDescriptions.topK + ' The default value is 40.'
+    )
+    .optional(),
+  topP: z
+    .number()
+    .min(0)
+    .max(1.0)
+    .describe(
+      GenerationCommonConfigDescriptions.topP + ' The default value is 0.9.'
+    )
+    .optional(),
+});
+
 function ollamaModel(
   ai: Genkit,
   model: ModelDefinition,
@@ -69,7 +102,7 @@ function ollamaModel(
     {
       name: `ollama/${model.name}`,
       label: `Ollama - ${model.name}`,
-      configSchema: GenerationCommonConfigSchema,
+      configSchema: OllamaConfigSchema,
       supports: {
         multiturn: !model.type || model.type === 'chat',
         systemRole: true,
--- a/js/plugins/vertexai/src/gemini.ts
+++ b/js/plugins/vertexai/src/gemini.ts
@@ -42,6 +42,7 @@ import {
 import {
   CandidateData,
   GenerateRequest,
+  GenerationCommonConfigDescriptions,
   GenerationCommonConfigSchema,
   MediaPart,
   MessageData,
@@ -60,6 +61,7 @@ import {
 } from 'genkit/model/middleware';
 import { runInNewSpan } from 'genkit/tracing';
 import { GoogleAuth } from 'google-auth-library';
+
 import { PluginOptions } from './common/types.js';
 import { handleCacheIfNeeded } from './context-caching/index.js';
 import { extractCacheConfig } from './context-caching/utils.js';
@@ -108,8 +110,40 @@ const GoogleSearchRetrievalSchema = z.object({
 
 /**
  * Zod schema of Gemini model options.
+ * Please refer to: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/medlm, for further information.
  */
 export const GeminiConfigSchema = GenerationCommonConfigSchema.extend({
+  maxOutputTokens: z
+    .number()
+    .min(1)
+    .max(8192)
+    .describe(GenerationCommonConfigDescriptions.maxOutputTokens)
+    .optional(),
+  temperature: z
+    .number()
+    .min(0.0)
+    .max(1.0)
+    .describe(
+      GenerationCommonConfigDescriptions.temperature +
+        ' The default value is 0.2.'
+    )
+    .optional(),
+  topK: z
+    .number()
+    .min(1)
+    .max(40)
+    .describe(
+      GenerationCommonConfigDescriptions.topK + ' The default value is 40.'
+    )
+    .optional(),
+  topP: z
+    .number()
+    .min(0)
+    .max(1.0)
+    .describe(
+      GenerationCommonConfigDescriptions.topP + ' The default value is 0.8.'
+    )
+    .optional(),
   location: z
     .string()
     .describe('Google Cloud region e.g. us-central1.')
--- a/js/plugins/vertexai/src/modelgarden/mistral.ts
+++ b/js/plugins/vertexai/src/modelgarden/mistral.ts
@@ -42,7 +42,11 @@ import {
   ToolRequestPart,
   z,
 } from 'genkit';
-import { ModelAction, modelRef } from 'genkit/model';
+import {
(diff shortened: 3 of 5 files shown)


MichaelDoyle:
Now that we're re-using the descriptions in many places, it probably makes sense to extract these into constants. Maybe here: https://github.com/firebase/genkit/blob/198759e74f02b17dbbacf265fb0d714ccc05c6df/js/ai/src/model.ts ?

js/plugins/ollama/src/index.ts
```diff
@@ -59,6 +59,40 @@ export function ollama(params: OllamaPluginParams): GenkitPlugin {
...
+  topP: z
+    .number()
+    .min(0)
+    .max(1.0)
+    .describe(
+      'Decides how many possible words to consider. A higher value means ' +
+        'that the model looks at more possible words, even the less likely ' +
+        'ones, which makes the generated text more diverse.'
+    )
+    .default(0.9),
```
MichaelDoyle, line 93:
```suggestion
    .max(1.0)
    .default(0.9)
    .describe(
      'Decides how many possible words to consider. A higher value means ' +
        'that the model looks at more possible words, even the less likely ' +
        'ones, which makes the generated text more diverse.'
    ),
```

nit: consistency in the ordering

hugomurillomtz:
> Now that we're re-using the descriptions in many places, it probably makes sense to extract these into constants. Maybe here: https://github.com/firebase/genkit/blob/198759e74f02b17dbbacf265fb0d714ccc05c6df/js/ai/src/model.ts ?

I tried doing that, but I'm getting this error:
image

I take it, I have to specify the file extension. So, knowing that, and following what I found on the project, I put `js` at the end, like so:
image

However, I'm still getting the error :/. Not sure what is it that I'm doing wrong.

js/plugins/vertexai/src/modelgarden/mistral.ts
```diff
@@ -51,12 +51,27 @@ export const MistralConfigSchema = GenerationCommonConfigSchema.extend({
...
-  stopSequences: z.array(z.string()).optional(),
+  temperature: z
+    .number()
+    .describe(
+      'What sampling temperature to use, we recommend between 0.0 and 0.7. ' +
+        'Higher values like 0.7 will make the output more random, ' +
+        'while lower values like 0.2 will make it more focused and deterministic. ' +
+        'We generally recommend altering this or top_p but not both. ' +
+        'The default value varies depending on the model you are targeting. ' +
+        'Call the /models endpoint to retrieve the appropriate value.'
```
shrutip90, line 62: Remove this line, since this is not an genkit endpoint.

js/plugins/vertexai/src/modelgarden/mistral.ts
```diff
@@ -51,12 +51,27 @@ export const MistralConfigSchema = GenerationCommonConfigSchema.extend({
...
-  maxOutputTokens: z.number().optional(),
-  temperature: z.number().optional(),
-  //   TODO: is this supported?
-  //   topK: z.number().optional(),
-  topP: z.number().optional(),
-  stopSequences: z.array(z.string()).optional(),
+  temperature: z
+    .number()
+    .describe(
+      'What sampling temperature to use, we recommend between 0.0 and 0.7. ' +
```
shrutip90, line 57: "we recommend" -> "recommended value for Mistral models is"

js/plugins/ollama/src/index.ts
```diff
@@ -59,6 +60,32 @@ export function ollama(params: OllamaPluginParams): GenkitPlugin {
...
+/**
+ * Please refer to: https://github.com/ollama/ollama/blob/main/docs/modelfile.md
+ * for further information.
+ */
+export const OllamaConfigSchema = GenerationCommonConfigSchema.extend({
+  temperature: z
+    .number()
+    .min(0.0)
+    .max(1.0)
+    .default(0.8)
```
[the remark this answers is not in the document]
hugomurillomtz: Yes! I see the value on the dev UI:
image

js/plugins/vertexai/src/gemini.ts
```diff
@@ -108,8 +109,39 @@ const GoogleSearchRetrievalSchema = z.object({
...
+    .optional(),
+  topP: z
+    .number()
+    .min(0)
+    .max(1.0)
+    .default(0.8)
+    .describe(ModelConfigurationParameters.topP)
+    .optional(),
+  /**
+   * GCP region (e.g. us-central1)
```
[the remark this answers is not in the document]
hugomurillomtz: The description from `topP`?

js/plugins/ollama/src/index.ts
```diff
@@ -59,6 +60,32 @@ export function ollama(params: OllamaPluginParams): GenkitPlugin {
...
+    .min(1)
+    .max(40)
+    .default(40)
+    .describe(ModelConfigurationParameters.topK),
+  topP: z
+    .number()
+    .min(0)
+    .max(1.0)
+    .default(0.9)
+    .describe(ModelConfigurationParameters.topP),
```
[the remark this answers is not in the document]
MichaelDoyle: IIRC, when you set a default, it becomes "optional" in the json schema. Kind of a funny thing with zod.

js/plugins/vertexai/src/modelgarden/mistral.ts
hugomurillomtz: Done! Let me know what you think.

js/plugins/ollama/src/index.ts
```diff
@@ -59,6 +60,32 @@ export function ollama(params: OllamaPluginParams): GenkitPlugin {
...
+    .max(1.0)
+    .default(0.8)
+    .describe(ModelConfigurationParameters.temperature)
+    .optional(),
+  topK: z
+    .number()
+    .min(1)
+    .max(40)
+    .default(40)
+    .describe(ModelConfigurationParameters.topK),
```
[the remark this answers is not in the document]
MichaelDoyle: As I'm thinking about it, this might be a good time to pause and reflect what we want the behavior to be here.

Although this will set the value in the UI (which is nice) *an* issue we might run into with setting the default here, is that when the model provider changes their default(s), we'll still hold on to the old value(s). There should probably be a way in the dev-ui to not specify optional values, and fall back to the model default.

js/plugins/vertexai/src/modelgarden/mistral.ts
MichaelDoyle: Why aren't we using the general purpose descriptions here? :)

MichaelDoyle:
This is a good change to add model specific validations. 

I think potentially we need to hold off on setting `defaults` because this could lead to some unexpected results. Whenever we have a `default` set, the developer will not be able to set it as undefined and "fall through" to the model behavior (we'll *always* set the value to whatever we provide in default).

js/plugins/vertexai/src/modelgarden/mistral.ts
hugomurillomtz: The description from this model family in particular is slightly different from the general purpose descriptions. For example, it specifies that a value between 0.0 and 0.7 is recommended, whilst other model families like vertexai/gemini recommend something between 0.0 and 1.0.

**vertex/ai** (https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/medlm):
image

**mistral** (https://docs.mistral.ai/api/#tag/chat/operation/chat_completion_v1_chat_completions_post):
image

So, I thought it might be good to add this new one and override it. However, if you don't deem this description includes major changes, let me know so I can put the general purpose ones.

js/plugins/ollama/src/index.ts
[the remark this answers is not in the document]
shrutip90: thanks!

js/plugins/ollama/src/index.ts
```diff
@@ -59,6 +60,31 @@ export function ollama(params: OllamaPluginParams): GenkitPlugin {
...
+/**
+ * Please refer to: https://github.com/ollama/ollama/blob/main/docs/modelfile.md
+ * for further information.
+ */
+export const OllamaConfigSchema = GenerationCommonConfigSchema.extend({
+  temperature: z
+    .number()
+    .min(0.0)
+    .max(1.0)
+    .default(0.8)
```
shrutip90, line 72: As discussed on the group chat, can you please remove the default value setting here and add a line in the description instead? So, you can use "GenerationCommonConfigDescriptions.temperature + 'The default value is 0.8." in line 73.

Please do a similar change for the other defaults in this PR.

hugomurillomtz: Done!

js/plugins/ollama/src/index.ts
```diff
@@ -59,6 +60,35 @@ export function ollama(params: OllamaPluginParams): GenkitPlugin {
...
+ */
+export const OllamaConfigSchema = GenerationCommonConfigSchema.extend({
+  temperature: z
+    .number()
+    .min(0.0)
+    .max(1.0)
+    .describe(
+      GenerationCommonConfigDescriptions.temperature +
+        ' The default value is 0.8.'
+    ),
```
shrutip90, line 75: Need to add back optional now? Also applies to other parameters below.

shrutip90 (approved):

js/plugins/vertexai/src/modelgarden/mistral.ts
hugomurillomtz: Done!

image

PR #2739, opened 2025-04-09
bug(go): Refactoring Firestore retriever plugin for usability improvements

sahdev77:
Refactoring of Firbase retriever plugin.
Resolving : https://github.com/firebase/genkit/issues/2381

Checklist (if applicable):

### DIFF
--- a/go/plugins/firebase/firebase.go
+++ b/go/plugins/firebase/firebase.go
@@ -31,29 +31,29 @@ import (
 const provider = "firebase"
 
 var appState struct {
-	mu      sync.Mutex      // Mutex to control access.
-	app     *firebasev4.App // Holds the Firebase app instance
-	initted bool            // Whether the plugin has been initialized.
+	app *firebasev4.App // Holds the Firebase app instance
 }
 
 // FireStore passes configuration options to the plugin.
-type FireStore struct {
+type Firebase struct {
 	App           *firebasev4.App
 	RetrieverOpts RetrieverOptions
 	retriever     ai.Retriever
+	mu            sync.Mutex // Mutex to control access.
+	initted       bool       // Whether the plugin has been initialized.
 }
 
 // Name returns the name of the plugin.
-func (f *FireStore) Name() string {
+func (f *Firebase) Name() string {
 	return provider
 }
 
 // Init initializes the Firebase plugin..
-func (f *FireStore) Init(ctx context.Context, g *genkit.Genkit) error {
-	appState.mu.Lock()
-	defer appState.mu.Unlock()
+func (f *Firebase) Init(ctx context.Context, g *genkit.Genkit) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
 
-	if appState.initted {
+	if f.initted {
 		log.Println("firebase.Init: plugin already initialized, returning without reinitializing")
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


go/plugins/firebase/firebase.go
```diff
@@ -22,87 +22,83 @@ import (
...
 
-var state struct {
-	mu         sync.Mutex     // Ensures thread-safe access to state
-	initted    bool           // Tracks if the plugin has been initialized
-	app        *firebase.App  // Holds the Firebase app instance
-	retrievers []ai.Retriever // Holds the list of initialized retrievers
+// The provider used in the registry.
+const provider = "firebase"
+
+var appState struct {
```
apascal07, line 33: Please follow the pattern in the other plugins, where state is part of the plugin struct. There should be no global state at package-level.

go/plugins/firebase/firebase.go
```diff
@@ -22,87 +22,83 @@ import (
...
-	state.mu.Lock()
-	defer state.mu.Unlock()
+// Init initializes the Firebase plugin..
+func (f *FireStore) Init(ctx context.Context, g *genkit.Genkit) error {
+	appState.mu.Lock()
+	defer appState.mu.Unlock()
 
-	if state.initted {
+	if appState.initted {
 		log.Println("firebase.Init: plugin already initialized, returning without reinitializing")
```
apascal07, line 57: It should error if it's already been initted, like other plugins.

go/plugins/firebase/firebase.go
```diff
@@ -22,87 +22,70 @@ import (
...
+	mu            sync.Mutex // Mutex to control access.
+	initted       bool       // Whether the plugin has been initialized.
 }
 
-// FirebasePluginConfig is the configuration for the Firebase plugin.
-type FirebasePluginConfig struct {
-	App        *firebase.App      // Pre-initialized Firebase app
-	Retrievers []RetrieverOptions // Array of retriever options
+// Name returns the name of the plugin.
+func (f *FireStore) Name() string {
```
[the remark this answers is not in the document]
sahdev77: Resolved

go/plugins/firebase/firebase.go
sahdev77: Resolved

go/samples/firebase-retrievers/main.go
```diff
@@ -43,37 +41,50 @@ func main() {
...
 	if err != nil {
-		log.Fatalf("Error creating Firestore client: %v", err)
+		log.Fatal(err)
 	}
-	defer firestoreClient.Close()
 
-	// Firebase app configuration and initialization
-	firebaseApp, err := firebasev4.NewApp(ctx, nil)
+	// Mock embedder
+	embedder := &MockEmbedder{}
```
apascal07, line 50: Why not a real embedder?

go/plugins/firebase/firebase.go
apascal07: I think the reason for this is because Go generally doesn't like global state and things are managed directly right in the main code path. I think we should probably do the same. We should expect that a user will create their own `App` and pass it to the plugin.

go/plugins/firebase/retriever.go
```diff
@@ -29,12 +29,9 @@ const (
 	Vector64 VectorType = iota
 )
 
-const provider = "firebase"
-
 type RetrieverOptions struct {
```
apascal07, line 32: Can we add comments to all of these fields? This is a high touch point.

go/samples/firebase-retrievers/main.go
```diff
@@ -43,37 +41,50 @@ func main() {
...
+	}
+
+	f := &firebase.Firebase{
+		App:           firebaseApp,
+		RetrieverOpts: retrieverOptions,
 	}
 
 	// Initialize Firebase plugin
-	if err := firebase.Init(ctx, g, firebaseConfig); err != nil {
+	if err := f.Init(ctx, g); err != nil {
```
apascal07, line 84: The intended way for plugins to be used is to be passed to `genkit.Init()` in `WithPlugins()`. We should design around that use case.

sahdev77:
Closing PR as we had complete change in the plugin implementation.

Issue #2419, opened 2025-03-21
[Go] Allow accessing `ActionContext` in dotprompt templates


apascal07:
**Describe the bug**
`ActionContext` should be passed into the dotprompt template and top level keys should be accessible via `@keyName`.

PR #2746, opened 2025-04-10
feat(go/genkit): Add action context example

rutujaD1999:
ISSUES:
[ ] [#2419 ]

CHANGELOG:
[ ] Modified the sample to include input data from action context instead of WithInput()

Checklist (if applicable):


apascal07 (commented):
Context should be accessible with `@name` not `name`, can you update to test that? In fact, I would expect this to have failed to work.

rutujaD1999:
Context and input are being treated the same way here : https://github.com/google/dotprompt/blob/main/go/dotprompt/dotprompt.go#L219

I have figured out how to fix this and will be submitting a PR on the dotprompt repo for the same

PR #2748, opened 2025-04-10
feat: support deleting evalRuns

ssbushi:
API changes for https://github.com/firebase/genkit/issues/2305

Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/server/router.ts
+++ b/genkit-tools/common/src/server/router.ts
@@ -191,6 +191,15 @@ export const TOOLS_SERVER_ROUTER = (manager: RuntimeManager) =>
         return evalRun;
       }),
 
+    /** Deletes a single eval run by ID */
+    deleteEvalRun: loggedProcedure
+      .input(apis.DeleteEvalRunRequestSchema)
+      .mutation(async ({ input }) => {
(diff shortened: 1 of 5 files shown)


genkit-tools/common/src/server/router.ts
```diff
@@ -191,6 +191,15 @@ export const TOOLS_SERVER_ROUTER = (manager: RuntimeManager) =>
         return evalRun;
       }),
 
+    /** Deletes a single eval run by ID */
+    deleteEvalRun: loggedProcedure
+      .input(apis.DeleteEvalRunRequestSchema)
+      .mutation(async ({ input }) => {
+        const parts = input.name.split('/');
+        const evalRunId = parts[1];
```
shrutip90, line 199: Let's also verify parts[0] and throw incorrect format errors?

Please also add a test for this.

shrutip90 (approved):

genkit-tools/common/src/server/router.ts
ssbushi: I don't think this is super crucial. This is not user facing directly so the benefit of having extra validations is very little. I'll leave this as it is to match other routes here.

PR #2763, opened 2025-04-14
feat: Add experimental filter in tooling listTraces API

shrutip90:
Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/manager/manager.ts
+++ b/genkit-tools/common/src/manager/manager.ts
@@ -268,7 +268,7 @@ export class RuntimeManager {
   async listTraces(
     input: apis.ListTracesRequest
   ): Promise<apis.ListTracesResponse> {
-    const { limit, continuationToken } = input;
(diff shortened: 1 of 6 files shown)


genkit-tools/common/src/manager/manager.ts
```diff
@@ -279,6 +279,9 @@ export class RuntimeManager {
       }
       query += `continuationToken=${continuationToken}`;
     }
+    if (filter) {
+      query += `filter=${filter}`;
```
pavelgj, line 283: It should be like this, no?

### 
```suggestion
      query += `filter=${encodeURI(JSON.stringify(filter))}`;
```

pavelgj (approved):

genkit-tools/common/src/manager/manager.ts
shrutip90: Thanks! Done.

PR #2767, opened 2025-04-15
fix(cli): Set type in trace index as the action subtype

shrutip90:
Checklist (if applicable):

### DIFF
--- a/genkit-tools/telemetry-server/src/localFileTraceStore.ts
+++ b/genkit-tools/telemetry-server/src/localFileTraceStore.ts
@@ -255,7 +255,8 @@ export class Index {
     const indexData = {
       id: traceData.traceId,
     } as Record<string, string | number>;
-    indexData['type'] = `${rootSpan?.attributes?.['genkit:type'] ?? 'UNKNOWN'}`;
+    indexData['type'] =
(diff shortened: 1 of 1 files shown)


genkit-tools/telemetry-server/src/localFileTraceStore.ts
```diff
@@ -255,7 +255,8 @@ export class Index {
     const indexData = {
       id: traceData.traceId,
     } as Record<string, string | number>;
-    indexData['type'] = `${rootSpan?.attributes?.['genkit:type'] ?? 'UNKNOWN'}`;
+    indexData['type'] =
+      `${rootSpan?.attributes?.['genkit:metadata:subtype'] ?? 'UNKNOWN'}`;
```
pavelgj, line 259: Maybe something like this?
```suggestion
      `${rootSpan?.attributes?.['genkit:metadata:subtype'] || rootSpan?.attributes?.['genkit:type'] || 'UNKNOWN'}`;
```

shrutip90: Thanks, done.

ssbushi (approved):

PR #2768, opened 2025-04-15
feat(cli/evals): Generate metricSummaries

ssbushi:
Tooling part of https://github.com/firebase/genkit/issues/1642

Checklist (if applicable):


genkit-tools/common/src/eval/parser.ts
```diff
@@ -78,3 +79,111 @@ export function extractMetricsMetadata(evaluatorActions: Action[]) {
...
+            undefinedCount,
+            distribution,
+            statusDistribution,
+          };
+        }
+      }
+    }
+    return {
+      evaluator,
+      errorCount,
```
shrutip90, line 182: Also add a testCaseCount?

genkit-tools/common/tests/eval/parser_test.ts
```diff
@@ -113,4 +120,553 @@ describe('parser', () => {
...
+        expect(result).toEqual({
+          evaluator: '/evaluator/genkit/context_relevancy',
+          errorCount: 0,
+          undefinedCount: 0,
+          statusDistribution: { undefined: 3 },
+          average: 22.0 / 3,
+        });
+      });
+
+      it('distribution for simple boolean scores', () => {
```
shrutip90, line 164: distribution -> score distribution.

Same for other testcases below.

pavelgj:
nit: PR & commit messags title should be `feat(cli)` or `feat(cli/evals)`

genkit-tools/common/src/eval/parser.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import * as _ from 'lodash';
```
pavelgj, line 17: Let's avoid adding lodash. Its use it increasingly discouraged.
Many of the things can be done with vanilla JS.

ssbushi:
> nit: PR & commit messags title should be feat(cli) or feat(cli/evals)

This is not specific to the CLI, but okay.

genkit-tools/common/tests/eval/parser_test.ts
```diff
@@ -113,4 +120,553 @@ describe('parser', () => {
...
+        const booleanScores = reMapScores(simpleEvalOutput, (response, i) => ({
+          testCaseId: response.testCaseId,
+          evaluation: {
+            score: i % 2 === 0,
```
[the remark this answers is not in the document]
shrutip90: Thank you, this is better, but I would prefer having a different helper like I suggested below: "Maybe instead of reMapScores you can have a helper that takes an array of Score objects and generates Record<string, EvalResponse> with testcaseIds. That might be concise enough and easier to read."

Please check if we can do this.

shrutip90 (approved):

PR #2781, opened 2025-04-17
feat(evals): Support metrics on datasetMetadata

ssbushi:
API fix for https://github.com/firebase/genkit/issues/2306

Checklist (if applicable):
- [x] Tested (unit tested)

### DIFF
--- a/genkit-tools/common/src/eval/localFileDatasetStore.ts
+++ b/genkit-tools/common/src/eval/localFileDatasetStore.ts
@@ -90,6 +90,7 @@ export class LocalFileDatasetStore implements DatasetStore {
       datasetId: id,
       schema,
       targetAction,
+      metrics: req.metrics,
       size: dataset.length,
       version: 1,
       datasetType: req.datasetType,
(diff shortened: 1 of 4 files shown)


genkit-tools/common/src/eval/localFileDatasetStore.ts
```diff
@@ -90,6 +90,7 @@ export class LocalFileDatasetStore implements DatasetStore {
       datasetId: id,
       schema,
       targetAction,
+      metrics: req.metrics,
```
shrutip90, line 93: nit: Would be nice to indicate that these are names of metrics evaluators that will be used for evaluation. For anything named metrics, i usually expect to see numbers in it. Maybe thats just me haha. But how about metricsForEval or metricNames?

shrutip90 (approved):

genkit-tools/common/src/types/apis.ts
```diff
@@ -140,6 +141,8 @@ export const UpdateDatasetRequestSchema = z.object({
   datasetId: z.string(),
   data: InferenceDatasetSchema.optional(),
   schema: DatasetSchemaSchema.optional(),
+  // Set to undefined if no changes in `metrics`.
```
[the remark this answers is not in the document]
ssbushi: I've specified this for `metrics` because it is an array. For other fields their truthiness is easy to figure out, undefined or truthy, whereas for arrays, it is easy to confuse between `[]`, `undefined` and truthy, setting `metrics` to `[]` will erase existing metrics on the dataset, not a no-op.

genkit-tools/common/src/eval/localFileDatasetStore.ts
ssbushi: Good point. Made it `metricRefs`

PR #2807, opened 2025-04-23
feat(js): Dynamic action resolver support

ifielker:
Checklist (if applicable):

### DIFF
--- a/js/core/src/registry.ts
+++ b/js/core/src/registry.ts
@@ -61,7 +61,9 @@ function parsePluginName(registryKey: string) {
 // Parses e.g. /model/googleai/gemini-2.0-flash into parts
 function parseRegistryKey(
   registryKey: string
-): { actionType: ActionType; pluginName: string; actionName: string } | undefined {
+):
+  | { actionType: ActionType; pluginName: string; actionName: string }
+  | undefined {
   const tokens = registryKey.split('/');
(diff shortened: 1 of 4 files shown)


js/core/src/registry.ts
```diff
@@ -58,6 +58,21 @@ function parsePluginName(registryKey: string) {
...
+// Parses e.g. /model/googleai/gemini-2.0-flash into parts
+function parseRegistryKey(
+  registryKey: string
+): { pluginName: string; action: ActionType; target: string } | undefined {
+  const tokens = registryKey.split('/');
+  if (tokens.length >= 4) {
+    return {
+      action: tokens[1] as ActionType,
+      pluginName: tokens[2],
+      target: tokens[3],
```
pavelgj, line 70: let's call this actionName... we literally never used the term "target" anywhere.

ifielker: I was getting confused by ActionName and ActionType. I thought others might have the same problem, so I tried for clearer naming. I was understandably thinking "ActionName" was the name of the action, so "embedder" or "model", but that's actually the ActionType. Target makes it clear that it's the target of the action. So actionType is the string representing the name or type of the action, and target is the string representing the thing the action acts on.

pavelgj (approved):

PR #2817, opened 2025-04-24
fix(go/genkit): remove colons from runtime filename

hugoaguirre:
Having colons in the runtime causes Genkit to fail during startup in non-POSIX systems.

Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/manager/manager.ts
+++ b/genkit-tools/common/src/manager/manager.ts
@@ -547,12 +547,19 @@ export class RuntimeManager {
  * Checks if the runtime file is valid.
  */
 function isValidRuntimeInfo(data: any): data is RuntimeInfo {
+  let timestamp = '';
+  // runtime filename might come with underscores due OS filename restrictions
+  // revert the underscores so the timestamp gets parsed correctly
+  if (typeof data.timestamp === 'string') {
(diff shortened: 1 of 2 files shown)


genkit-tools/common/src/manager/manager.ts
```diff
@@ -547,12 +547,19 @@ export class RuntimeManager {
  * Checks if the runtime file is valid.
  */
 function isValidRuntimeInfo(data: any): data is RuntimeInfo {
+  let timestamp = '';
+  // runtime filename might come with underscores due OS filename restrictions
+  // revert the underscores so the timestamp gets parsed correctly
+  if (typeof data.timestamp === 'string') {
+    timestamp = data.timestamp.replaceAll('_', ':');
```
apascal07, line 554: I'm confused, why does the timestamp inside the file need to be modified like this? Isn't it just the filename?

hugoaguirre: So, the runtime info is obtained right from the filename and the timestamp is part of it. When a new runtime file is detected, it gets validated this way.

Keeping just the go changes will cause an error when validating the date format

apascal07 (approved):

PR #2839, opened 2025-04-29
feat(js/plugins/{googleai,vertexai}): implemented dynamic model listing for googleai, and new model lookup API for googleai & vertexai

pavelgj:
model listing for vertexai is coming next (more work than expected).

new model lookup API:

```ts
import { googleAI } from '@genkit-ai/googleai';

await ai.generate({
  prompt: '...',
  model: googleAI.model('gemini-2.0-flash-001')
})
// can set config right on the model ref
googleAI.model('gemini-2.0-flash-001', { temperature: 0.7 })
```

similarly for vertexAI

```ts
import { vertexAI } from '@genkit-ai/vertexai';

await ai.generate({
  prompt: '...',
  model: vertexAI.model('gemini-2.0-flash-001')
})
// can set config right on the model ref
vertexAI.model('gemini-2.0-flash-001', { temperature: 0.7 })
```

similarly for embedders:

```ts
googleAI.embedder('text-embedder-004')
vertexAI.embedder('text-embedder-004')
```

models are dynamically listed in the dev ui, pulled from the plugin's `listActions`
image

Checklist (if applicable):


js/plugins/googleai/src/index.ts
```diff
@@ -161,48 +176,159 @@ async function resolver(
...
-      name: modelRef.name,
-      apiKey: options?.apiKey,
-      baseUrl: options?.baseUrl,
-      info: {
-        ...modelRef.info,
-        label: `Google AI - ${actionName}`,
-      },
-      debugTraces: options?.experimental_debugTraces,
-    });
+  defineGoogleAIEmbedder(ai, `googleai/${actionName}`, {
```
ifielker, line 211: This if actionName is not already known, this defines an embedder with text only capability. I thought you mentioned that you wanted the generic model/embedder to "allow everything"? Shouldn't you modify the generic embedder definition as well?

js/plugins/vertexai/src/index.ts
```diff
@@ -157,28 +173,35 @@ async function resolveModel(
...
-      vertexClientFactory,
-      options: {
-        projectId,
-        location,
-      },
-      debugTraces: options?.experimental_debugTraces,
-    });
-  }
-  // TODO: Support other models
+  const modelRef = gemini(actionName);
```
ifielker, line 176: Why did you remove the check for gemini in the name? What if it's Imagen or veo or something else?

js/plugins/vertexai/src/index.ts
```diff
@@ -157,28 +173,35 @@ async function resolveModel(
...
+}
+
+async function resolveEmbedder(
+  ai: Genkit,
+  actionName: string,
+  options?: PluginOptions
+) {
+  const { projectId, location, authClient } = await getDerivedParams(options);
+
+  defineVertexAIEmbedder(ai, actionName, authClient, { projectId, location });
```
ifielker, line 198: Is the default embedder still text only? Should it be multimodal for future embedders so people can use all the capabilities of the embedders that are available?

js/testapps/flow-simple-ai/src/index.ts
```diff
@@ -139,16 +139,36 @@ export const drawPictureFlow = ai.defineFlow(
...
+export const streamFlowGemini = ai.defineFlow(
+  {
+    name: 'streamFlowGemini',
+    inputSchema: z.string(),
+    outputSchema: z.string(),
+    streamSchema: z.string(),
+  },
+  async (prompt, { sendChunk }) => {
+    const { response, stream } = ai.generateStream({
+      model: googleAI.model('gemini-2.0-flash-001', { temperature: 0.77 }),
```
ifielker, line 171: Why not make this input part of the inputSchema? you could take a string model and a config { temperature } etc. and set the inputSchema defaults to be 'gemini-2.0-flash-001' etc. and then this is more flexible for testing and you don't have to type it in every time you refresh the page.

js/testapps/flow-simple-ai/src/index.ts
```diff
@@ -868,3 +888,18 @@ ai.defineFlow('geminiEnum', async (thing) => {
 
   return output;
 });
+
+ai.defineFlow('embedders-tester', async () => {
+  console.log(
+    await ai.embed({
+      content: 'hello world',
+      embedder: googleAI.embedder('text-embedding-004'),
```
ifielker, line 896: You could also take the content string and embedder name e.g. (vertexai/text-embedding-004) as input here, set these values as defaults, and make this more flexible while still not having to type anything every time you run it.

js/plugins/vertexai/src/index.ts
pavelgj: in the list actions we check `supportedGenerationMethods` for `generateContent`, so all models that support that work just fine. Gemini API has a number of non-gemini models, like `gemma` and `learnlm` which work just fine with this implementation.
If someone tries to use imagen or veo they will get an error, and that's fine.

js/plugins/vertexai/src/index.ts
pavelgj: Good point. Made generic embedder metadata multimodal.

js/plugins/googleai/src/index.ts
pavelgj: Done. made generic embedder multimodal.

js/testapps/flow-simple-ai/src/index.ts
pavelgj: because I'm testing `googleAI.model` function here

js/testapps/flow-simple-ai/src/index.ts
pavelgj: same here, I'm testing `googleAI.embedder` function, not specific embedders.

ifielker: Sure, but this way you are only testing one possible input to the googleAI.embedder function. the other way, you can test various inputs. e.g. does it handle "googleai/text-embedding-004" or just "text-embedding-004"? What happens if you specify "vertexai/text-embedding-004"? What kind of errors do you get if you try with "text-embedding-banana"? Does it fail where you expect it to fail? What about special characters? Any other weird input you can think of to somehow screw it up?

js/testapps/flow-simple-ai/src/index.ts
ifielker: Yes, and you should be able to test more than one input to it...

js/testapps/flow-simple-ai/src/index.ts
ifielker: Same here. The ability to test more inputs is good.

js/testapps/flow-simple-ai/src/index.ts
pavelgj: Many flows in this testapp are using this pattern and I never use any of them because I find them hard to use mainly because I need to spend time parsing the input schema and how those fields wire into logic of the flow to figure out what the flow is capable of and then find the right inputs. Sure it will take maybe extra 10-20 seconds, but I don't like that. It's a personal preference, I prefer simple (non-parameterized) tests.

js/testapps/flow-simple-ai/src/index.ts
pavelgj: Same as above. This is not an e2e test. This is closer to sample code that we can refer user to showing them how to use these apis. Should be easy to read.

ifielker (approved):

PR #2843, opened 2025-04-30
fix(js/prompts): fixed .prompt input handling on prompt render

pavelgj:
wrapExecutablePromptPromise was  incorrectly handling input for render.

Checklist (if applicable):


js/genkit/src/genkit.ts
```diff
@@ -263,11 +263,10 @@ export class Genkit implements HasRegistry {
     }) as ExecutablePrompt<z.infer<I>, O, CustomOptions>;
 
     executablePrompt.render = async (
```
pavelgj, line 265: this wrapped the wrong (old) interface.

mbleigh (approved):

Issue #2788, opened 2025-04-19
Thinking budget


rouapps:
I am trying to upgrade from Gemini-2.0-Flash to 2.5 flash but would like to disable thinking, however thinking budget doesn't seem to be available yet on Genkit. I have tried from the Google AI documentation but it doesn't seem to work on Genkit:
config: {
      thinkingConfig: {
        thinkingBudget: 0,
      },
    }

Thanks!

alexklimin89:
facing the same issue right now

Rytis-J:
Any solution to this?

rvh70:
Thanks for that link to the 1.8.0 release! Still getting the hang of this package (and Genkit in general)

dosstx:
Trying to understand. SO, how to enable thinking budget for JS version of Genkit? Docs have no info.

Rytis-J:
Hey, this worked for me

```
    const ai = new GoogleGenAI({ apiKey: "YOUR_KEY" });


    const response = await ai.models.generateContent({
      model,
      contents,
      config: {
        ...
        thinkingConfig: { thinkingBudget: 8000 }
      }
    });
```

hilariie:
Thinking budget support landed shortly after this was filed. Config passthrough shipped in [#2848](https://github.com/genkit-ai/genkit/pull/2848) (released in 1.8.0), so the exact snippet from the original post, config: { thinkingConfig: { thinkingBudget: 0 } }, works to disable thinking on the 2.5 models. The current @genkit-ai/google-genai plugin also types thinkingConfig.thinkingBudget directly (0 disables thinking), and Go got the same in [#2846](https://github.com/genkit-ai/genkit/pull/2846). Closing as fixed.

PR #2851, opened 2025-05-01
feat(js): thinking mode support

EduardWayland:
## Description
Thinking mode is becoming a standard feature in modern AI models, including the latest Gemini 2.5.
This PR introduces the new capability in the default `ModelInfoSchema` named `thinkingMode` and support for configuring thinking behavior in the `googleai` plugin with the goal of aligning with [Gemini's official specs](https://ai.google.dev/api/generate-content#ThinkingConfig).

### Example
Here some examples of the usage of the new configuration fully retro-compatible:
**Through `.withConfig`:**
```typescript
// Genkit initialization...
const prompt = ai.prompt("test", {});
const chat = session.chat(prompt, {
    model: gemini25FlashPreview0417.withConfig({
        thinkingConfig: {
            includeThoughts: false,
            thinkingBudget: 0,
        }
    }),
    config: {
        temperature: 0.1,
    }
});
```
**Through basic configuration:**
```typescript
const prompt = ai.prompt("survey-generation", {});
const chat = session.chat(prompt, {
    model: gemini25FlashPreview0417,
    config: {
        temperature: 0.1,
        thinkingConfig: {
            includeThoughts: true,
            thinkingBudget: 1000,
        }
    }
});
```

Note for reviewers:
- It would be cool to have the UI behave similar to [Google AI Studio dashboard](https://aistudio.google.com/) where in the `Run setup` the thinking mode shows up parameters conditionally.

Checklist (if applicable):

This feature addresses issue #2788 and aligns with the aligns with the broader trend of making AI reasoning more explicit and tunable.


js/plugins/googleai/src/gemini.ts
```diff
@@ -952,11 +976,15 @@ export function defineGoogleAIModel({
...
         responseMimeType: jsonMode ? 'application/json' : undefined,
       };
+
+      // HACK: cast to any since those options aren't officially supported in the old SDK yet
       if (requestConfig.responseModalities) {
-        // HACK: cast to any since this isn't officially supported in the old SDK yet
         (generationConfig as any).responseModalities =
           requestConfig.responseModalities;
       }
+      if (model.info?.supports?.thinkingMode && requestConfig.thinkingConfig) {
```
pavelgj, line 985: I don't think this supports check is necessary. API will ignore the fields if the model doesn't support it. And if it doesn't it better to let the user know (with and error) that the option they are setting is not supported.

```suggestion
      if (requestConfig.thinkingConfig) {
```

js/plugins/vertexai/src/gemini.ts
```diff
@@ -496,6 +519,7 @@ export const gemini25FlashPreview0417 = modelRef({
       tools: true,
       toolChoice: true,
       systemRole: true,
+      thinkingMode: true,
```
pavelgj, line 522: could you please also add it for GENERIC_GEMINI_MODEL on line 539

js/plugins/vertexai/src/gemini.ts
```diff
@@ -1117,6 +1141,11 @@ export function defineGeminiModel({
         });
       }
 
+      if (modelInfo?.supports?.thinkingMode && request.config?.thinkingConfig) {
```
pavelgj, line 1144: same here.

```suggestion
      if (request.config?.thinkingConfig) {
```

alexklimin89:
I always had a type error with thinkingConfig but got it to work with @ts-ignore comment

genkit-tools/common/src/types/model.ts
```diff
@@ -120,6 +120,8 @@ export const ModelInfoSchema = z.object({
       constrained: z.enum(['none', 'all', 'no-tools']).optional(),
       /** Model supports controlling tool choice, e.g. forced tool calling. */
       toolChoice: z.boolean().optional(),
+      /** Model can use an internal "thinking process" during response generation. */
+      thinkingMode: z.boolean().optional(),
```
pavelgj, line 124: Let's exclude this from this PR. When we add first class support for thinking mode we'll do. I'm not sure if a simple boolean is going to sufficient. For example, see `constrained` option above, which needs to support other values.

```suggestion
```

genkit-tools/genkit-schema.json
```diff
@@ -898,6 +898,9 @@
             },
             "toolChoice": {
               "type": "boolean"
+            },
+            "thinkingMode": {
+              "type": "boolean"
```
pavelgj, line 903:
```suggestion
```

js/plugins/googleai/src/gemini.ts
```diff
@@ -292,6 +315,7 @@ export const gemini25FlashPreview0417 = modelRef({
       tools: true,
       toolChoice: true,
       systemRole: true,
+      thinkingMode: true,
```
pavelgj, line 318:
```suggestion
```

js/plugins/vertexai/src/gemini.ts
```diff
@@ -513,6 +536,7 @@ export const gemini25FlashPreview0417 = modelRef({
       tools: true,
       toolChoice: true,
       systemRole: true,
+      thinkingMode: true,
```
pavelgj, line 539:
```suggestion
```

js/plugins/vertexai/src/gemini.ts
```diff
@@ -564,6 +588,7 @@ export const GENERIC_GEMINI_MODEL = modelRef({
...
```
pavelgj, line 591:
```suggestion
```

pavelgj (approved):

PR #2854, opened 2025-05-01
feat(js/ai): added dynamic tools that can be dynamically defined at runtime

pavelgj:
Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -334,6 +334,8 @@ export async function generate<
   };
   const resolvedFormat = await resolveFormat(registry, resolvedOptions.output);
 
+  registry = maybeRegisterDynamicTools(registry, resolvedOptions);
+
   const params = await toGenerateActionOptions(registry, resolvedOptions);
 
   const tools = await toolsToActionRefs(registry, resolvedOptions.tools);
@@ -360,6 +362,29 @@ export async function generate<
       });
     }
   );
+}
+
+function maybeRegisterDynamicTools<
+  O extends z.ZodTypeAny = z.ZodTypeAny,
+  CustomOptions extends z.ZodTypeAny = typeof GenerationCommonConfigSchema,
+>(registry: Registry, options: GenerateOptions<O, CustomOptions>): Registry {
+  let hasDynamicTools = false;
+  options?.tools?.forEach((t) => {
+    if (
+      (t as Action).__action &&
+      (t as Action).__action.metadata?.type === 'tool' &&
+      (t as Action).__action.metadata?.dynamicTool
+    ) {
+      if (!hasDynamicTools) {
+        hasDynamicTools = true;
+        // Create a temporary registry with dynamic tools for the duration of this
+        // generate request.
+        registry = Registry.withParent(registry);
+      }
+      registry.registerAction('tool', t as Action);
+    }
+  });
+  return registry;
 }
 
 export async function toGenerateActionOptions<
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -373,7 +373,7 @@ function maybeRegisterDynamicTools<
     if (
       (t as Action).__action &&
       (t as Action).__action.metadata?.type === 'tool' &&
-      (t as Action).__action.metadata?.dynamicTool
+      (t as Action).__action.metadata?.dynamic
     ) {
       if (!hasDynamicTools) {
         hasDynamicTools = true;
--- a/js/ai/src/tool.ts
+++ b/js/ai/src/tool.ts
@@ -405,19 +405,18 @@ export function dynamicTool<I extends z.ZodTypeAny, O extends z.ZodTypeAny>(
     {
       ...config,
       actionType: 'tool',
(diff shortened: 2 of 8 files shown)


js/ai/src/generate.ts
```diff
@@ -362,6 +364,29 @@ export async function generate<
...
+function maybeRegisterDynamicTools<
+  O extends z.ZodTypeAny = z.ZodTypeAny,
+  CustomOptions extends z.ZodTypeAny = typeof GenerationCommonConfigSchema,
+>(registry: Registry, options: GenerateOptions<O, CustomOptions>): Registry {
+  let hasDynamicTools = false;
+  options?.tools?.forEach((t) => {
+    if (
+      (t as Action).__action &&
+      (t as Action).__action.metadata?.type === 'tool' &&
+      (t as Action).__action.metadata?.dynamicTool
```
mbleigh, line 376: Maybe just

```suggestion
      (t as Action).__action.metadata?.dynamic
```

js/ai/src/generate.ts
```diff
@@ -362,6 +364,29 @@ export async function generate<
...
+    ) {
+      if (!hasDynamicTools) {
+        hasDynamicTools = true;
+        // Create a temporary registry with dynamic tools for the duration of this
+        // generate request.
+        registry = Registry.withParent(registry);
```
mbleigh, line 382: For my edification...why does it need to be registered at all? And how does the child registry not just disappear since it's not being returned or anything? Will this leak memory or be otherwise weird over time?

js/ai/src/tool.ts
```diff
@@ -381,3 +390,36 @@ function interruptTool(registry: Registry) {
     throw new ToolInterruptError(metadata);
   };
 }
+
+/**
+ * Defines a dynamic tool. Dynamic tools are just like regular tools but will not be registered in the
+ * Genkit registry and can be defined dynamically at runtime.
+ */
+export function dynamicTool<I extends z.ZodTypeAny, O extends z.ZodTypeAny>(
```
mbleigh, line 398: If the call signature doesn't include the interrupt version, will it work with a non-beta registry? If so, could we maybe allow it to work with stable and use assertUnstable only if it's an interrupt?

js/core/src/registry.ts
```diff
@@ -115,21 +115,30 @@ export class Registry {
...
-      }
-      return toJsonSchema(resolvedSchema);
-    },
-  });
-
-  constructor(public parent?: Registry) {}
+  readonly asyncStore: AsyncStore;
+  readonly dotprompt: Dotprompt;
+
+  constructor(public parent?: Registry) {
```
mbleigh, line 121: Please declare variable separately (I want to turn on "erasable syntax only" sometime)

js/ai/src/generate.ts

js/ai/src/generate.ts
pavelgj: this child registry instance is returned (line 387) and then used (see line 337) for the rest of the generate call.
Once this generate call is done, the child registry is garbage collected, it should not be referenced by anything anymore.

js/ai/src/tool.ts
pavelgj: yes, beta is only asserted if interrupt is involved. It will work with `fn` in stable.

js/core/src/registry.ts
pavelgj: Done.

js/ai/src/generate.ts
mbleigh: Ah, I see. Why do the tools need to be registered at all (just curious)? Is it just to hook into the existing lifecycle stuff?

js/genkit/src/tool.ts
```diff
@@ -16,6 +16,7 @@
 
 export {
   asTool,
+  dynamicTool,
```
mbleigh, line 19: This should probably be exported from `"genkit"`, no? vs. a sub-export

mbleigh (approved):

js/ai/src/generate.ts
pavelgj: well... they don't "need" to be registered, but I think it's cleaner this way. Specifically, one of my goals is to keep all of the "generate" implementation cleanly encapsulated inside generate action behind the GenerateActionOptions interface: https://github.com/firebase/genkit/blob/0e8fd8cb78f50debffa907ab6858163147819cb9/js/ai/src/model.ts#L715
(generate is already standardized across JS, Go and Python and Dev UI is migrated to use it)

Since generate action is an action it takes serializable input -- tools names (and we could maybe allow passing in tool definitions)... which might work for functionless (interrupting) dynamic tools, but we can't pass in tool implementations. The way it works is that it needs to resolve the implementation somewhere... it already does it via the registry. And we have child/ephemeral registries...

js/genkit/src/tool.ts
pavelgj: done.

pavelgj: I mean, exported from `genkit` as well

PR #2884, opened 2025-05-07
feat(js/plugins/mcp): Refactor to a client manager structure, with dynamic tool calling.

ssbushi:
- Modified the current McpClient class to a Client manager class
- GenkitMcpClient is now a dedicated single client-server connection.
- Tools are now fetched dynamically to support adding/removing/disabling clients in the manager.


js/plugins/mcp/README.md
```diff
@@ -13,53 +15,94 @@ To get started, you'll need Genkit and the MCP plugin:
...
+const mcpManager = createMcpClientManager({
+  name: 'myMcpClients', // A name for the manager plugin itself
+  mcpClients: {
+    // Each key (e.g., 'fs', 'git') becomes a namespace for the server's tools.
+    fs: { // Configuration for the first MCP server
+      serverProcess: {
+        command: 'npx',
+        args: ['-y', '@modelcontextprotocol/server-everything', ...ALLOWED_DIRS],
+      },
+    },
```
mbleigh, line 39: It's much more important to me that we match the evolving pseudo-standard of MCP server config than that we stay consistent with the existing plugin.

This should be configured as I had in `mcp-redux`:

```ts
createMcpManager({
  name: 'mcp',
  version: '0.1.0',
  mcpServers: {
    fs: {command: 'npx', args: [...]. env: {...}},
    remote: {url: "https://..."}
  }
});

// let's do this for a single client
createMcpClient({
  name: 'firebase',
  version: '0.1.0',
  command: 'firebase',
  args: [...]
});
```

See e.g. [cursor docs](https://docs.cursor.com/context/model-context-protocol#configuring-mcp-servers).

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,94 @@ To get started, you'll need Genkit and the MCP plugin:
 npm i genkit genkitx-mcp
 ```
 
-## MCP Client
+## MCP Client Manager
 
-To create an MCP client, you call the `mcpClient` function to generate a Genkit plugin for an MCP server. For example, to use MCP's example [filesystem server](https://github.com/modelcontextprotocol/servers/tree/main/src/filesystem):
+To connect to one or more MCP servers, you use the `createMcpClientManager` function. This function returns a Genkit plugin that manages connections to the configured MCP servers.
```
mbleigh, line 20: Naming wise let's just go with `createMcpClient` and `createMcpManager` which is the multi-client manager.

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,94 @@ To get started, you'll need Genkit and the MCP plugin:
...
 });
 
 const ai = genkit({
   plugins: [
-    filesystemClient /* ... other plugins such as model providers ...*/,
+    /* ... install plugins such as model providers ...*/
   ],
 });
+
+const mcpTools = await mcpManager.getAllTools(ai);
```
mbleigh, line 53: Let's go with `getActiveTools` since it shouldn't return tools from disabled/inactive servers.

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,94 @@ To get started, you'll need Genkit and the MCP plugin:
...
+});
 ```
 
+The `createMcpClientManager` function initializes a `GenkitMcpClientManager` instance, which handles the lifecycle and communication with the defined MCP servers.
+
+### `createMcpClientManager()` Options
+
+-   **`name`**: (required, string) A name for the client manager plugin itself.
+-   **`version`**: (optional, string) The version of the client manager plugin. Defaults to "1.0.0".
+-   **`mcpClients`**: (required, object) An object where each key is a client-side name (namespace) for an MCP server, and the value is the configuration for that server.
```
mbleigh, line 69: This should be `mcpServers` (you're specifying the list of servers to connect to, also that's consistent with how this is configured in other tools).

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,94 @@ To get started, you'll need Genkit and the MCP plugin:
...
+            -   **`args`**: (optional, string[]) Array of string arguments to pass to the command.
+            -   **`env`**: (optional, Record<string, string>) Key-value object of environment variables.
+        -   **`serverUrl`**: (string) The URL of a remote server to connect to using the SSE MCP transport.
+        -   **`serverWebsocketUrl`**: (string) The URL of a remote server to connect to using the WebSocket MCP transport.
+        -   **`transport`**: An existing MCP transport object for connecting to the server.
+
 Most MCP servers are built to run as spawned processes on the same machine using the `stdio` transport. When you supply the `serverProcess` option, you are specifying the command, arguments, and environment variables for spawning the server as a subprocess.
 
-### mcpClient() Options
+### Legacy `mcpClient()`
```
mbleigh, line 85: Let's mark `mcpClient` as deprecated but keep the old API, and have `createMcpClient` as not deprecated with the new API suggested above.

js/plugins/mcp/src/client/client.ts
```diff
@@ -0,0 +1,291 @@
...
+ * when dealing with multiple MCP server connections.
+ */
+export class GenkitMcpClient {
+  private name: string;
+  private version: string;
+  private serverConfig: McpServerConfig;
+  private rawToolResponses?: boolean;
+  private disabled: boolean;
+
+  private _server?: McpServerRef;
```
mbleigh, line 104: Let's make this public so that there's an escape hatch for manually doing things with the underlying MCP SDK.

js/plugins/mcp/package.json
mbleigh, line 1: Thinking about it, let's move the MCP SDK to be a `peerDependency` instead of a normal dep.

js/plugins/mcp/src/client/client.ts
```diff
@@ -0,0 +1,291 @@
...
+    await this.ready();
+    let tools: ToolAction[] = [];
+
+    if (this._server) {
+      const capabilities = await this._server.client.getServerCapabilities();
+      if (capabilities?.tools)
+        tools.push(
+          ...(await fetchDynamicTools(ai, this._server.client, {
+            rawToolResponses: this.rawToolResponses,
+            serverName: this.name + '-mcp-server',
```
mbleigh, line 272: I don't think we need to append `-mcp-server` here do we?

js/plugins/mcp/src/client/index.ts
mbleigh, line 1: Let's call this `manager.ts`

js/plugins/mcp/src/client/index.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  /**
+   * Connects to a single MCP server defined by the provided configuration.
+   * If a server with the same name already exists, it will be disconnected first.
+   * Stores the client and transport references internally. Handles connection errors
+   * by marking the server as disabled.
+   * @param serverName The name to assign to this server connection.
+   * @param config The configuration object for the server.
+   */
+  async connectClient(serverName: string, config: McpClientOptions) {
```
mbleigh, line 84: From a nomenclature perspective I think we should consider ourselves the client and the thing the dev is trying to interact with the server.

So you're connecting a server, not a client -- the client is an implementation detail.

But in terms of names I'd just keep it simple with `connect(serverName: string, config: {...})` and do the same throughout.

mbleigh (commented):
One more high-level piece of food for thought: in the client manager context, I think we probably want to be very robust to misconfiguration errors. It's possible people will use this for dynamic, user-configured tools and when that happens we should recover gracefully as best we can.

So maybe in the client manager there should be an `{error?: {message, status, details}}` and if there's an error trying to connect to the server it is put in a disabled state, an error-level log is emitted, and the error is also saved with the server.

mbleigh:
Oh also I think we should put this work into a launch branch - I created `launch/mcp-redux` and made #2896 to squash my own work and create a chain of commits. Can you rebase onto the `launch/mcp-redux` as your merge target?

js/plugins/mcp/README.md
ssbushi: Thanks for the reference/examples. Done.

js/plugins/mcp/src/client/client.ts
ssbushi: I don't know what I was thinking. Removed.

js/plugins/mcp/src/client/index.ts
ssbushi: Thanks, I was a bit confused myself with where I was landing with "client" vs "server" terminology. Updated.

ssbushi:
> One more high-level piece of food for thought: in the client manager context, I think we probably want to be very robust to misconfiguration errors. It's possible people will use this for dynamic, user-configured tools and when that happens we should recover gracefully as best we can.
> 
> So maybe in the client manager there should be an {error?: {message, status, details}} and if there's an error trying to connect to the server it is put in a disabled state, an error-level log is emitted, and the error is also saved with the server.

I'll work on this in a new commit, updating the PR to reflect the changes so far.

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,99 @@ To get started, you'll need Genkit and the MCP plugin:
...
+```
+This function takes similar options to a single server configuration within `createMcpManager` (e.g., `name`, `version`, `serverProcess`, `rawToolResponses`) and directly returns a Genkit plugin for that single client. It is recommended to use `createMcpManager` for new projects.
 
 ### Using MCP Actions
 
-The Genkit MCP client automatically discovers available tools and prompts and registers them with Genkit making them available anywhere other tools and prompts can be used. To access resources, special `list_resources` and `read_resource` tools are registered that will access resources for the server.
+The Genkit MCP client manager, through its `getActiveTools()` method, discovers available tools from each connected and enabled server. These tools can then be provided to Genkit models for use in generation requests. Resources and MCP Tools are both wrapped as GenkitTools. 
 
-All MCP actions are namespaced under the name you supply, so a client called `filesystem` will register tools such as `filesystem/read_file`.
+<!-- TODO: Note about MCP prompts. -->
```
ssbushi, line 106: @mbleigh @pavelgj 

How should we approach prompts with the newer dynamic API? Tools and resources can be wrapped as dynamic tools, but prompts should be treated differently, I suppose....

We don't have dynamic prompt support today and I'm wondering how to expose them in this new iteration of the plugin.

Should we also provide the user to install the MCP client as a plugin like we did for `mcpClient({..})` and provide prompts there?

mbleigh: Let's think short-term / long-term.

Short-term, maybe we should essentially follow the API of `ExecutablePrompt` but not actually register them? So basically:

```ts
const mcpPrompt = mcpManager.prompt('some-prompt');
await mcpPrompt(input, options);
await mcpPrompt.stream(input, options);
await mcpPrompt.render(input, options);
```

and under the hood that's just rendering the prompt to `MessageData[]` and then either executing it, streaming it, or returning the rendered options?

Long-term I think we should have an MCP manager able to be used as a plugin with full dynamic registry integration.

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,99 @@ To get started, you'll need Genkit and the MCP plugin:
...
+const mcpTools = await mcpManager.getActiveTools(genkit);
+
+// Provide MCP tools to the model of your choice.
+const response = await ai.generate({
+  model: gemini15Flash,
+  prompt: 'What are the last 5 commits in the repo `/home/yourusername/Desktop/test-repo/?`',
+  tools: mcpTools,
+});
+
+console.log(response.text());
```
pavelgj, line 62:
```suggestion
console.log(response.text);
```

js/plugins/mcp/package.json
```diff
@@ -39,7 +39,8 @@
   "license": "Apache-2.0",
   "peerDependencies": {
     "@genkit-ai/core": "workspace:^",
```
pavelgj, line 41:
```suggestion
```

js/plugins/mcp/README.md
ssbushi: I like the idea of bespoke prompt references. One thing I'd suggest for the short term is to use a serverName qualifer to narrow it down:

```
const mcpPrompt = mcpManager.prompt({serverName: 'my-mcp-server', prompt: 'some-prompt'});
```

We can add support for dynamic look-ups as we go forward and handle server resolution when there is not conflicts.

> Long-term I think we should have an MCP manager able to be used as a plugin with full dynamic registry integration.

Agreed. We need to discuss more on how to support dynamic registrations via plugins.

js/plugins/mcp/src/client/client.ts
```diff
@@ -0,0 +1,295 @@
...
+      this.disabled = false;
+    }
+  }
+
+  /**
+   * Closes and then restarts the transport connection for the specified server.
+   * Useful for attempting to recover from connection issues without full
+   * reconfiguration.
+   */
+  async reconnect() {
```
ssbushi, line 252: Is this the `reconnect` behavior we want:

- `reconnect` does not help establish a connection after calling `disconnect` since the server config is deleted on disconnection. 
- This method documents that it disconnects and reconnects, for recovering connections (and behaves accordingly). 
- If the user wants to "reconnect" after a `disconnect`, they have to call the `connect` method with the serverConfig.

js/plugins/mcp/README.md
mbleigh: Maybe `mcpManager.prompt('server-name', 'some-prompt')`?

js/plugins/mcp/src/client/client.ts
ssbushi: Done. (Renamed to `restart`)

js/core/package.json
```diff
@@ -41,7 +41,7 @@
     "express": "^4.21.0",
     "get-port": "^5.1.0",
     "json-schema": "^0.4.0",
-    "zod": "^3.23.8",
+    "zod": "^3.22.4",
```
pavelgj, line 44: Up is fine, down may not be...

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,150 @@ To get started, you'll need Genkit and the MCP plugin:
...
   ],
 });
+
+
+// Retrieve tools from all active MCP servers
+const mcpTools = await mcpManager.getActiveTools(genkit);
+
+// Provide MCP tools to the model of your choice.
+const response = await ai.generate({
+  model: gemini15Flash,
```
pavelgj, line 57: Please use the new syntax. And at least 2.0

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,150 @@ To get started, you'll need Genkit and the MCP plugin:
...
+    // Individual clients are not Genkit plugins themselves.
+  ],
+});
+
+async function runSingleClient() {
+  // Retrieve tools from this specific client
+  const fsTools = await myFsClient.getActiveTools(genkit);
+
+  const response = await genkit.generate({
+    model: gemini15Flash, // Replace with your model
```
pavelgj, line 115: Ditto

js/plugins/mcp/src/client/manager.ts
```diff
@@ -0,0 +1,345 @@
...
+      client.disable();
+    }
+  }
+
+  /**
+   * Re-enables a previously disabled server connection. Attempts to reconnect
+   * using the stored transport. Does nothing if the server is not disabled.
+   * @param serverName The name of the server to re-enable.
+   */
+  async reenable(serverName: string) {
```
mbleigh, line 166: Let's just do `enable` as suggested above (I know this was my name originally)

js/plugins/mcp/src/client/manager.ts
```diff
@@ -0,0 +1,345 @@
...
+   * const mcpManager = createMcpManager({ ... });
+   * // In your Genkit configuration:
+   * // const allMcpTools = await mcpManager.getActiveTools(ai);
+   * // Then, these tools can be used or registered with Genkit.
+   * ```
+   *
+   * @param ai The Genkit instance, used by individual clients to define dynamic tools.
+   * @returns A Promise that resolves to an array of `ToolAction` from all active MCP clients.
+   */
+  async getActiveTools(ai: Genkit): Promise<ToolAction[]> {
```
mbleigh, line 269: Add the `{resourceTools?: boolean}` here as well.

js/plugins/mcp/src/client/manager.ts
```diff
@@ -0,0 +1,345 @@
...
+  private setError(
+    serverName: string,
+    error: {
+      message: string;
+      detail?: any;
+    }
+  ) {
+    this._clientStates[serverName] = { error };
+    logger.error(
+      `An error has occured while managing your MCP client '${serverName}'. The client may be disabled to avoid further issues. Please resolve the issue and reenable the client '${serverName}' to continue using its resources.`
```
mbleigh, line 335: Probably a `warn` level log here.

js/plugins/mcp/src/util/prompts.ts
```diff
@@ -0,0 +1,242 @@
...
+      status: 'UNIMPLEMENTED',
+      message: `[MCP] prompt.asTool not supported with MCP`,
+    });
+  };
+
+  return callPrompt as ExecutablePrompt<z.infer<I>, O, CustomOptions>;
+}
+
+/**
+ * Lookup all tools available in the server and register each as a Genkit tool.
```
mbleigh, line 184:
```suggestion
 * Lookup all prompts available in the server and register each as a Genkit prompt.
```

js/plugins/mcp/src/util/resources.ts
```diff
@@ -0,0 +1,160 @@
...
+
+import type {
+  Resource,
+  ResourceTemplate,
+} from '@modelcontextprotocol/sdk/types.js';
+import { Genkit, z } from 'genkit';
+
+function listResourcesTool(
+  ai: Genkit,
+  client: any, // Use 'any' or let TS infer; removing specific type import
```
mbleigh, line 25: Why any? If it's a circular dependency issue, you can do `import type` without triggering circ dependencies.

js/plugins/mcp/src/util/resources.ts
mbleigh, line 1: I think resources need to be handled a bit differently in the Manager. I don't want there to be a tool per client to list and read resources - there should be a single tool at the manager level (perhaps with optional list of server names to include) for listing / reading across all connected servers.

js/plugins/mcp/package.json
mbleigh, line 1: Let's go ahead and change the package name to `@genkit-ai/mcp` so we don't forget to do it later. Should be safe since we're merging to a launch branch.

mbleigh (commented):
This is mostly looking really good, thanks for all the work! I think probably the most important thing after this will be adding a good suite of tests to give us more confidence that it's robust and ready for an upgrade.

js/plugins/mcp/src/client/manager.ts
pavelgj: Done.

js/plugins/mcp/src/client/manager.ts
pavelgj: Done.

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,150 @@ To get started, you'll need Genkit and the MCP plugin:
...
-### Using MCP Actions
+```ts
+import { mcpClient } from 'genkitx-mcp/legacy'; // Note the import path
+
+const legacyFilesystemClient = mcpClient({
+  name: 'filesystemLegacy',
+  serverProcess: { /* ... */ },
+});
+```
+This function takes similar options to a single server configuration within `createMcpManager` (e.g., `name`, `version`, `serverProcess`, `rawToolResponses`) and directly returns a Genkit plugin for that single client. It is recommended to use `createMcpManager` for new projects.
```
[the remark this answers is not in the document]
pavelgj: Removed.

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,150 @@ To get started, you'll need Genkit and the MCP plugin:
...
+    -   One of the following server connection configurations:
+        -   Parameters for launching a local server process using the stdio MCP transport.
+            -   **`command`**: (required, string) Shell command path for launching the MCP server (e.g., `npx`, `python`).
+            -   **`args`**: (optional, string[]) Array of string arguments to pass to the command.
+            -   **`env`**: (optional, Record<string, string>) Key-value object of environment variables.
+        -   **`url`**: (string) The URL of a remote server to connect to using the SSE MCP transport.
+        -   **`serverWebsocketUrl`**: (string) The URL of a remote server to connect to using the WebSocket MCP transport.
+        -   **`transport`**: An existing MCP transport object for connecting to the server.
+
 Most MCP servers are built to run as spawned processes on the same machine using the `stdio` transport. When you supply the `serverProcess` option, you are specifying the command, arguments, and environment variables for spawning the server as a subprocess.
```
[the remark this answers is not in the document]
pavelgj: removed

js/plugins/mcp/src/client/client.ts
```diff
@@ -0,0 +1,314 @@
...
+    return new Promise<void>((resolve, reject) => {
+      this._readyListeners.push({ resolve, reject });
+    });
+  }
+
+  /**
+   * Connects to a single MCP server defined by the provided configuration.
+   * If a server with the same name already exists, it will be disconnected first.
+   * Stores the client and transport references internally. Handles connection errors
+   * by marking the server as disabled.
```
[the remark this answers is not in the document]
pavelgj: Made connect on the client private/internal.

js/plugins/mcp/src/client/client.ts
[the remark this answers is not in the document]
pavelgj: Done.

js/plugins/mcp/README.md
```diff
@@ -13,53 +15,150 @@ To get started, you'll need Genkit and the MCP plugin:
...
-  serverProcess: {
-    command: 'npx',
-    args: ['-y', '@modelcontextprotocol/server-everything', ...ALLOWED_DIRS],
+const mcpManager = createMcpManager({
+  name: 'myMcpClients', // A name for the manager plugin itself
+  mcpServers: {
+    // Each key (e.g., 'fs', 'git') becomes a namespace for the server's tools.
+    fs: {
+        command: 'npx',
+        args: ['-y', '@modelcontextprotocol/server-everything', ...ALLOWED_DIRS],
```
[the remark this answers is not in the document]
pavelgj: changed to @modelcontextprotocol/server-filesystem

js/plugins/mcp/README.md
pavelgj: Updated prompts.

js/plugins/mcp/src/client/client.ts
```diff
@@ -0,0 +1,314 @@
...
+      await this.disconnect();
+      await this.connect(this.serverConfig);
+    }
+  }
+
+  /**
+   * Fetches all tools available through this client, if the server
+   * configuration is not disabled.
+   */
+  async getActiveTools(ai: Genkit): Promise<ToolAction[]> {
```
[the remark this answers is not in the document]
pavelgj: Done.

js/plugins/mcp/src/client/manager.ts
pavelgj: Done.

js/plugins/mcp/src/util/resources.ts
pavelgj: Done.

js/plugins/mcp/src/client/client.ts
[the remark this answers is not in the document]
pavelgj: Ack.

js/plugins/mcp/src/index.ts
```diff
@@ -77,6 +115,6 @@ export function createMcpServer(
 /**
  * @deprecated use `createMcpServer` instead.
  */
-export function mcpServer(ai: Genkit, options: McpServerOptions) {
+export function mcpServer(ai: Genkit, options: McpClientOptions) {
```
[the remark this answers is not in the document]
pavelgj: removed it altogether, don't need to deal with legacy stuff since we're renaming the package

PR #2887, opened 2025-05-08
chore: typechecking errors 2

hendrixmar:
Fixing type hint errors of mypy


py/packages/genkit/src/genkit/core/typing.py
```diff
@@ -671,7 +671,7 @@ class Resume(BaseModel):
     metadata: dict[str, Any] | None = None
 
 
-class Part(RootModel[TextPart | MediaPart | ToolRequestPart | ToolResponsePart | DataPart | CustomPart]):
+class Part(RootModel[tuple[TextPart | MediaPart | ToolRequestPart | ToolResponsePart | DataPart | CustomPart]]):
```
AbeJLazaro, line 674: why is tuple type added?

py/packages/genkit/src/genkit/web/typing.py
```diff
@@ -29,12 +29,12 @@
 from __future__ import annotations
 
 from collections.abc import Awaitable, Callable
-from typing import Union
+from typing import Type, Union
```
AbeJLazaro, line 32: Why is Type imported? where is it used?

PR #2895, opened 2025-05-09
Add Roots support to the MCP client and server

gspencergoog:
## Description

This adds roots support to the MCP server. It supports setting client roots when creating the client, but not modifying the roots once the client plugin is created.  That will have to come later. I did add support for the server to request the roots from the client (because some do that even if the roots don't change).

## Tests
 - I also added some basic configuration unit tests for the client and server (they don't actually test that much, just the setup).


gspencergoog:
@mbleigh  Here is the roots support PR I said I'd send your way.  Let me know if it's what you were expecting.

I've tested the client against a server that supports roots, and it seems to work as expected.  It should probably have some integration tests too, so that the client (and server) would be tested against a fake server (and client).

mbleigh:
@gspencergoog I know it's been forever on this -- we're in the middle of revamping the MCP plugin with a pretty big rewrite and this got caught in the middle. We're in the home stretch and I'm making sure roots support is added before we finish, so hopefully the long wait is over soon.

gspencergoog:
Ahh, OK.  Shall I just close this then?

mbleigh:
I'll close it out - #3074 adds roots to the new implementation 😄  but thank you!

PR #2918, opened 2025-05-14
feat(cli): made telemetry index persistent and self-maintaining

pavelgj:
Checklist (if applicable):


genkit-tools/telemetry-server/src/localFileTraceStore.ts
```diff
@@ -240,11 +240,41 @@ export class Index {
...
+      fs.readFileSync(this.metadataFileName(), { encoding: 'utf8' })
+    );
+  }
+
+  private newIndexFileName() {
+    return `idx_${(Date.now() + '').padStart(17, '0')}.json`;
+  }
+
+  listIndexFiles() {
+    return fs.readdirSync(this.indexRoot);
```
shrutip90, line 277: metadata file will also be in indexRoot dir? So this is not strictly listing index files?

genkit-tools/telemetry-server/src/localFileTraceStore.ts
```diff
@@ -47,32 +48,35 @@ export class LocalFileTraceStore implements TraceStore {
...
   async init() {
-    if (!this.index) {
-      return;
+    const metadata = this.index.getMetadata();
+    // if the metadata file doesn't exist or it was for the older version or if
+    // there are too many index files we recreate the index.
+    if (
+      !metadata ||
+      metadata.version !== currentVersion ||
+      this.index.listIndexFiles().length > MAX_INDEX_FILES
```
shrutip90, line 72: Why not delete the oldest files in this case?

genkit-tools/telemetry-server/src/localFileTraceStore.ts
```diff
@@ -240,11 +240,41 @@ export class Index {
...
+
+  private newIndexFileName() {
+    return `idx_${(Date.now() + '').padStart(17, '0')}.json`;
+  }
+
+  listIndexFiles() {
+    return fs.readdirSync(this.indexRoot);
   }
 
   add(traceData: TraceData) {
```
shrutip90, line 280: Can we please index on status too?

genkit-tools/telemetry-server/src/localFileTraceStore.ts
pavelgj: Good catch! I had filter, but must have inadvertently removed it during "cleanup". Added back.

genkit-tools/telemetry-server/src/localFileTraceStore.ts
pavelgj: Done.

genkit-tools/telemetry-server/src/localFileTraceStore.ts
pavelgj: Because an index file may contain 1 trace or 1000+ traces. For example what might happen is that the latest 9 files have 1 trace each, but the 10th has 100. If we just delete the oldest one, the index will only have 9 traces, but missing the rest. There are weirder corner cases when there are concurrent telemetry server instances running at the same time...

shrutip90 (approved):

PR #2926, opened 2025-05-15
feat(py): Adding vertexai plugin dynamic model registry

AbeJLazaro:
I added dynamic model registry for vertex ai

- Simple comparison to determine if it is a image model or text model
- Rename some methods that are common among GoogleAI and VertexAI plugin. 
- Adding api key
- Adding doc strings
- Adding resolve_action for VertexAI Plugin

Checklist (if applicable):
- [X] All samples successfully tested (only google-genai-image fails but same from main)


py/plugins/google-genai/src/genkit/plugins/google_genai/google.py
```diff
@@ -21,6 +21,7 @@
 from google.genai.client import DebugConfig
 from google.genai.types import HttpOptions, HttpOptionsDict
 
+import genkit.plugins.google_genai.constants as const
```
yesudeep, line 24: It seems weird to call this module `constants`. Should we call it `config` instead?

yesudeep (approved):

py/plugins/google-genai/src/genkit/plugins/google_genai/google.py
AbeJLazaro: I was thinking of it, and I will keep the name of constants, maybe later I will move the constants of the plugins to this file and keep it. I think it is better for maintainability.

PR #2931, opened 2025-05-16
feat(go/plugins/googlegenai): Add dynamic model registration

hugoaguirre:
Checklist (if applicable):


go/genkit/reflection.go
```diff
@@ -256,6 +256,61 @@ func wrapReflectionHandler(h func(w http.ResponseWriter, r *http.Request) error)
...
+	err := json.Unmarshal(input, &inputMap)
+	if err != nil {
+		return nil
+	}
+
+	var modelName string
+	var found bool
+	modelName, ok := inputMap["model"].(string)
+	if !ok {
+		// reflection API tests do not set "model" attribute but it is provided in the key
```
apascal07: I don't quite understand the situation here. Is this referring to the end to end tests?

go/plugins/googlegenai/googlegenai.go
```diff
@@ -310,3 +313,129 @@ func GoogleAIEmbedder(g *genkit.Genkit, name string) ai.Embedder {
 func VertexAIEmbedder(g *genkit.Genkit, name string) ai.Embedder {
 	return genkit.LookupEmbedder(g, vertexAIProvider, name)
 }
+
+func (ga *GoogleAI) ListActions() []core.ActionDesc {
```
apascal07: Let's change this to accept `context.Context` in the `DynamicPlugin` interface and pipe it through.

go/genkit/reflection.go
hugoaguirre: Yes, specifically talking about the test in the [reflection_api.yaml](https://github.com/firebase/genkit/blob/main/tests/specs/reflection_api.yaml#L27) file

go/genkit/reflection.go
```diff
@@ -357,9 +416,9 @@ func handleNotify(g *Genkit) func(w http.ResponseWriter, r *http.Request) error
 
 // handleListActions lists all the registered actions.
 // The list is sorted by action name and contains unique action names.
-func handleListActions(g *Genkit) func(w http.ResponseWriter, r *http.Request) error {
+func handleListActions(ctx context.Context, g *Genkit) func(w http.ResponseWriter, r *http.Request) error {
 	return func(w http.ResponseWriter, r *http.Request) error {
-		ads := listResolvableActions(g)
+		ads := listResolvableActions(ctx, g)
```
apascal07: Revert the `ctx` argument throughout there, should use `r.Context()` instead.

go/genkit/reflection.go
```diff
@@ -302,6 +357,10 @@ func handleRunAction(g *Genkit) func(w http.ResponseWriter, r *http.Request) err
 			json.Unmarshal(body.Context, &contextMap)
 		}
 
+		if err = resolveModel(ctx, g, body.Input); err != nil {
```
apascal07: Shouldn't this be in `runAction` after it fails to resolve the action and right before it returns an error saying it doesn't exist?

go/genkit/reflection.go
```diff
@@ -256,6 +255,85 @@ func wrapReflectionHandler(h func(w http.ResponseWriter, r *http.Request) error)
...
+			dp, ok := plugin.(DynamicPlugin)
+			if !ok {
+				continue
+			}
+			if dp.Name() != provider {
+				continue
+			}
+			if a := g.reg.LookupAction(fmt.Sprintf("/%s/%s/%s", atype, provider, name)); a != nil {
+				return g.reg.LookupAction(key).(core.Action), nil
+			}
```
apascal07:
```suggestion
			if a := g.reg.LookupAction(fmt.Sprintf("/%s/%s/%s", atype, provider, name)); a != nil {
				return a.(core.Action), nil
			}
```

Is this what you meant? Doesn't make sense as is.

hugoaguirre: You are right.

apascal07 (approved):

PR #2949, opened 2025-05-21
feat(py): Add list models for dev UI Gemini

AbeJLazaro:
Add list models for Dev UI

- Add method in Plugin interface
- Add override of list_models method for GoogleAI and VertexAI plugin
- Add fix typo in resolve_action implementations.
- Add patter of resolve_action for list_model functions

Checklist (if applicable):

### DIFF
--- a/py/packages/genkit/src/genkit/ai/_plugin.py
+++ b/py/packages/genkit/src/genkit/ai/_plugin.py
@@ -75,11 +75,11 @@ class Plugin(abc.ABC):
         """
         pass
 
-    def list_models(
+    def list_actions(
         self,
         kind: ActionKind,
     ) -> list[str]:
-        """Generate a list of available models.
+        """Generate a list of available actions or models.
 
         Args:
             kind: Supported kind
--- a/py/packages/genkit/src/genkit/ai/_plugin.py
+++ b/py/packages/genkit/src/genkit/ai/_plugin.py
@@ -74,3 +74,17 @@ class Plugin(abc.ABC):
             None, initialization is done by side-effect on the registry.
         """
         pass
+
+    def list_models(
+        self,
+        kind: ActionKind,
+    ) -> list[str]:
+        """Generate a list of available models.
+
+        Args:
+            kind: Supported kind
+
+        Returns:
+            List of model names.
+        """
+        return []
--- a/py/packages/genkit/src/genkit/ai/_plugin.py
+++ b/py/packages/genkit/src/genkit/ai/_plugin.py
@@ -75,16 +75,14 @@ class Plugin(abc.ABC):
         """
         pass
 
-    def list_actions(
-        self,
-        kind: ActionKind,
-    ) -> list[str]:
+    def list_actions(self) -> list[dict[str, str]]:
         """Generate a list of available actions or models.
 
-        Args:
-            kind: Supported kind
-
         Returns:
-            List of model names.
+            list of actions dicts with the following shape:
+            {
+                'name': str,
+                'kind': ActionKind,
+            }
         """
         return []
--- a/py/packages/genkit/src/genkit/core/registry.py
+++ b/py/packages/genkit/src/genkit/core/registry.py
@@ -233,7 +233,7 @@ class Registry:
         actions: dict[str, Action] | None = None,
         allowed_kinds: set[ActionKind] | None = None,
     ) -> dict[str, Action] | None:
-        """List more models into a dictionary.
+        """Add actions or models.
 
         Args:
             actions: dictionary of serializable actions.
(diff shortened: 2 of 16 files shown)


yesudeep (approved):
Interesting. model resolution is also dynamic now?

AbeJLazaro:
> Interesting. model resolution is also dynamic now?

Yes! It is for the UI only. But it just shows the names, it does not register the action for all those new models. When you choose one that has not its action registered, the resolve_action method is executed.

py/packages/genkit/src/genkit/ai/_plugin.py
```diff
@@ -74,3 +74,17 @@ def initialize(self, ai: GenkitRegistry) -> None:
             None, initialization is done by side-effect on the registry.
         """
         pass
+
+    def list_models(
+        self,
+        kind: ActionKind,
```
pavelgj, line 80: this is probably not necessary... we usually just ask for everything.

py/packages/genkit/src/genkit/core/registry.py
```diff
@@ -97,6 +98,21 @@ def register_action_resolver(self, plugin_name: str, resolver: ActionResolver):
                 raise ValueError(f'Plugin {plugin_name} already registered')
             self._action_resolvers[plugin_name] = resolver
 
+    def register_list_models_resolver(self, plugin_name: str, resolver: Callable) -> None:
```
pavelgj, line 101:
```suggestion
    def register_list_action_resolver(self, plugin_name: str, resolver: Callable) -> None:
```

py/packages/genkit/src/genkit/core/registry.py
```diff
@@ -212,6 +228,38 @@ def list_serializable_actions(self, allowed_kinds: set[ActionKind] | None = None
                         }
             return actions
 
+    def list_models(
```
pavelgj, line 231: list_actions

py/packages/genkit/src/genkit/ai/_plugin.py
```diff
@@ -74,3 +74,17 @@ def initialize(self, ai: GenkitRegistry) -> None:
             None, initialization is done by side-effect on the registry.
         """
         pass
+
+    def list_models(
```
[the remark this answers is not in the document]
pavelgj: it's important that it's `actions` not `models` -- this can return embedders, tools, etc.

py/packages/genkit/src/genkit/ai/_plugin.py
AbeJLazaro: I added it just to avoid hard implementation in other plugins.

pavelgj: what do you mean? I just meant that this method is only used by the dev ui and dev ui always pull all actions.

AbeJLazaro: aaaahhhhhh, okokok, I got it. But it requires it to generate the key of the action. All the actions retrived from the endpoint has a suffix model/ or embedder/, that's why I placed the parameter.

AbeJLazaro: More because the list of actions is also used in other views like embedders. Other approch could be to filter the registry, but I wouldn't like to set that kind of logic in the registry.

py/plugins/google-genai/src/genkit/plugins/google_genai/google.py
```diff
@@ -206,6 +206,33 @@ def _resolve_embedder(self, ai: GenkitRegistry, name: str) -> None:
...
+            {
+                'name': str,
+                'kind': ActionKind,
+            }
+        """
+        actions_list = list()
+        for m in self._client.models.list():
+            name = m.name.replace('models/', '')
+            if 'generateContent' in m.supported_actions:
+                actions_list.append({
```
pavelgj, line 223: we talked about `modelActionMetadata` and `embedderActionMetadata` helpers:
https://github.com/firebase/genkit/blob/ae1549c627c02dfb4ab9a7d011b1f3fcaeb76b90/js/plugins/googleai/src/index.ts#L235
this metadata must contain input and output schema, as well as all the necessary model/embedder info.

py/plugins/google-genai/src/genkit/plugins/google_genai/google.py
```diff
@@ -204,8 +212,43 @@ def _resolve_embedder(self, ai: GenkitRegistry, name: str) -> None:
         ai.define_embedder(
             name=googleai_name(_clean_name),
             fn=embedder.generate,
+            metadata=default_embedder_info(_clean_name),
+            config_schema=EmbedContentConfig,
         )
 
+    def list_actions(self) -> list[dict[str, str]]:
```
pavelgj, line 219: Could you please cache this. This method is quite slow and there's not need to re-fetch actions every time. Dev UI is restarted often, so it's OK making devs restart the dev ui to get new models.

PR #2950, opened 2025-05-21
Chore: Remove old docs content

chrisraygill:
The PR does the following:

* Changes references to the old Firebase docs to point toward genkit.dev instead.
* Deletes the Go docs directory
* Deletes all docs content from the docs directory - but keeps the assets used in the root README and adds a docs/README file that explains the move to https://github.com/genkit-ai/docsite.
* Update the repo README to include Python support and more closely reflect our latest messaging.


apascal07 (approved):
LGTM for Go.

pavelgj (approved):

PR #2956, opened 2025-05-22
feat(js/genkit): Package Genkit CLI as platform-specific binaries

chrisraygill:
Includes changes to relevant scripts in `genkit-tools` and the project root to package the Genkit CLI as platform-specific binaries alongside the creation of the NPM packages.


chrisraygill:
Cleaned up PR here: https://github.com/firebase/genkit/pull/2957

PR #2979, opened 2025-05-27
fix(cli): fixed telemetry server saving spans that get filtered

pavelgj:
Checklist (if applicable):

### DIFF
--- a/genkit-tools/telemetry-server/src/localFileTraceStore.ts
+++ b/genkit-tools/telemetry-server/src/localFileTraceStore.ts
@@ -115,18 +115,31 @@ export class LocalFileTraceStore implements TraceStore {
 
   async save(id: string, rawTrace: TraceData): Promise<void> {
     let trace = this.filter(rawTrace);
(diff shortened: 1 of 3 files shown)


genkit-tools/telemetry-server/src/localFileTraceStore.ts
```diff
@@ -115,18 +115,31 @@ export class LocalFileTraceStore implements TraceStore {
...
     await mutex.waitForUnlock();
     const release = await mutex.acquire();
     try {
-      const existing = await this.load(id);
+      const existing = (await this.load(id)) || trace;
       if (existing) {
         Object.keys(trace.spans).forEach(
           (spanId) => (existing.spans[spanId] = trace.spans[spanId])
         );
+        // If we it's one of those wierd roots (internal span that we filter) we try to fix
```
pavelgj, line 129:
```suggestion
        // If it's one of those weird roots (internal span that we filter) we try to fix
```

apascal07 (approved):

PR #3015, opened 2025-06-02
fix(js/plugins/vertexai): Handle Gemini API thought signatures

pavelgj:
followup for https://github.com/firebase/genkit/pull/3014

Checklist (if applicable):

### DIFF
--- a/js/plugins/vertexai/src/gemini.ts
+++ b/js/plugins/vertexai/src/gemini.ts
@@ -736,7 +736,7 @@ function toGeminiPart(part: Part): GeminiPart {
   if (part.media) return toGeminiFileDataPart(part);
   if (part.toolRequest) return toGeminiToolRequestPart(part);
   if (part.toolResponse) return toGeminiToolResponsePart(part);
-  if (typeof part.reasoning === 'string') return toThought(part);
+  if (typeof part.reasoning === 'string') return toGeminiThought(part);
 
   throw new Error('unsupported type');
 }
@@ -761,7 +761,7 @@ function fromGeminiInlineDataPart(part: GeminiPart): MediaPart {
   };
 }
 
-function toThought(part: Part) {
+function toGeminiThought(part: Part) {
   const outPart: any = { thought: true };
   if (part.metadata?.thoughtSignature)
     outPart.thoughtSignature = part.metadata.thoughtSignature;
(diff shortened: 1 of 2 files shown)


ifielker (approved):

js/plugins/vertexai/src/gemini.ts
```diff
@@ -732,17 +732,13 @@ function fromGeminiFinishReason(
...
-  } else if (part.toolResponse) {
-    return toGeminiToolResponsePart(part);
-  } else {
-    throw new Error('unsupported type');
-  }
+  if (part.text) return { text: part.text };
+  if (part.media) return toGeminiFileDataPart(part);
+  if (part.toolRequest) return toGeminiToolRequestPart(part);
+  if (part.toolResponse) return toGeminiToolResponsePart(part);
+  if (typeof part.reasoning === 'string') return toThought(part);
```
ifielker: Shouldn't it follow the naming conventions? i.e. 'toGeminiThoughtPart'?

js/plugins/vertexai/src/gemini.ts
```diff
@@ -858,6 +858,18 @@ export function fromGeminiCandidate(
...
 }
+
+function fromThought(part: {
+  thought: boolean;
+  text?: string;
+  thoughtSignature?: string;
+}): Part {
+  return {
+    reasoning: part.text || '',
+    metadata: { thoughtSignature: (part as any).thoughtSignature },
```
ifielker, line 869: Why do you need the `(part as any)` in here? part has a type that has a thoughtSignature...

js/plugins/vertexai/src/gemini.ts
```diff
@@ -732,17 +732,13 @@ function fromGeminiFinishReason(
...
-  } else {
-    throw new Error('unsupported type');
-  }
+  if (part.text) return { text: part.text };
+  if (part.media) return toGeminiFileDataPart(part);
+  if (part.toolRequest) return toGeminiToolRequestPart(part);
+  if (part.toolResponse) return toGeminiToolResponsePart(part);
+  if (typeof part.reasoning === 'string') return toThought(part);
+
+  throw new Error('unsupported type');
```
mbleigh: I'd recommend improving this to `Unsupported Gemini part type: ${JSON.stringify(part)}` to make debugging easier.

js/plugins/vertexai/src/gemini.ts
pavelgj: good point, renamed

js/plugins/vertexai/src/gemini.ts
pavelgj: Done.

js/plugins/vertexai/src/gemini.ts
pavelgj: old vertex sdk doesn't have that field in their types... but still has the field in the underlying object.

PR #3021, opened 2025-06-03
fix: log error with duplicate action key

### LINKED ISSUE #26: Duplicate names in `defineFlow` overrides the original definition rather than failing.

gioboa:
This commit change the behavior with actions with the same name.
Based on the issue description I changed warnings into errors.

Close #26 

Checklist (if applicable):

### DIFF
--- a/js/core/src/registry.ts
+++ b/js/core/src/registry.ts
@@ -291,9 +291,8 @@ export class Registry {
     const key = `/${type}/${action.__action.name}`;
     logger.debug(`registering ${key}`);
     if (this.actionsById.hasOwnProperty(key)) {
-      // TODO: Make this an error!
-      logger.warn(
-        `WARNING: ${key} already has an entry in the registry. Overwriting.`
+      logger.error(
+        `ERROR: ${key} already has an entry in the registry. Overwriting.`
       );
     }
     this.actionsById[key] = action;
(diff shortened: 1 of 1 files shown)


MichaelDoyle:
Hey @gioboa - just checking - it looks like this PR only changes the "log level" from warning to error, and does not actually raise any error. Based on the PR title, though, I think you intend to throw an error?

gioboa:
> Hey @gioboa - just checking - it looks like this PR only changes the "log level" from warning to error, and does not actually raise any error. Based on the PR title, though, I think you intend to throw an error?

Yep, you are right. I fixed the code to get the desired behavior.

gioboa:
Thanks for your review @MichaelDoyle 
I'm looking forward to getting your feedback about the change.

gioboa:
@pavelgj Can you review this simple change please?

js/core/src/registry.ts
```diff
@@ -291,9 +291,8 @@ export class Registry {
     const key = `/${type}/${action.__action.name}`;
     logger.debug(`registering ${key}`);
     if (this.actionsById.hasOwnProperty(key)) {
-      // TODO: Make this an error!
-      logger.warn(
-        `WARNING: ${key} already has an entry in the registry. Overwriting.`
+      logger.error(
+        `Failed to register ${key} action because it's already in the registry.`
```
pavelgj: this message is not accurate -- `logger.error` does not throw an error, it just logs at "error" log level, so it would not fail to register a duplicate key. 

I'm not sure if we can actually make this throw an error at this point... I know for sure there would need to be a number of tests that will need fixing.

s/logger.warn/logger.error/ is a good idea and a low risk change.

gioboa: I see, I changed the message to `ERROR`.

pavelgj (approved):

Issue #2438, opened 2025-03-23
[Py] Implement resolve_action for all plugins.


yesudeep:
```
FAILED plugins/compat-oai/tests/test_plugin.py::test_openai_plugin_initialize - TypeError: Can't instantiate abstract class OpenAI without an implementation for abstract method 'resolve_action'
```

PR #3022, opened 2025-06-03
feat(py/genkit): added the resolve method dummy for vectorstore plugin.

hendrixmar:
**Description**:

This PR introduces a dummy implementation of the resolve method in the vectorstore plugin. The process has been added to satisfy interface requirements for the base plugin base class, which will enable the abstract method for resolve_method in the future.

### DIFF
--- a/py/plugins/dev-local-vectorstore/src/genkit/plugins/dev_local_vectorstore/plugin_api.py
+++ b/py/plugins/dev-local-vectorstore/src/genkit/plugins/dev_local_vectorstore/plugin_api.py
@@ -67,9 +67,18 @@ class DevLocalVectorStore(Plugin):
         kind: ActionKind,
         name: str,
     ) -> None:
-        ...
 
-    def _configure_dev_local_retriever(self, ai: GenkitRegistry) -> Action:
+        if kind is ActionKind.RETRIEVER:
+            self._configure_dev_local_retriever(ai, name)
+            return None
+
+        if kind is ActionKind.INDEXER:
+            self._configure_dev_local_indexer(ai, name)
+
+
+        return None
+
+    def _configure_dev_local_retriever(self, ai: GenkitRegistry, name: str | None = None) -> Action:
         """Registers Local Vector Store retriever for provided parameters.
 
         Args:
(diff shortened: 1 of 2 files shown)


py/plugins/dev-local-vectorstore/src/genkit/plugins/dev_local_vectorstore/plugin_api.py
```diff
@@ -60,6 +61,14 @@ def initialize(self, ai: GenkitRegistry) -> None:
         self._configure_dev_local_retriever(ai=ai)
         self._configure_dev_local_indexer(ai=ai)
 
+    def resolve_action(  # noqa: B027
```
AbeJLazaro: is it required the ` # noqa: B027`?

hendrixmar: Is just for the mypy ignore. I saw it in the plugin base class.

AbeJLazaro: okokok, I think it is not required here then, you can delete it then.

py/plugins/dev-local-vectorstore/src/genkit/plugins/dev_local_vectorstore/plugin_api.py
```diff
@@ -60,6 +61,14 @@ def initialize(self, ai: GenkitRegistry) -> None:
         self._configure_dev_local_retriever(ai=ai)
         self._configure_dev_local_indexer(ai=ai)
 
+    def resolve_action(
```
pavelgj, line 64: I'm a bit torn on this... my original thinking was that these plugins methods are optional to implement. Can you explain again why this is needed?

AbeJLazaro: The goal is to make `resolve_action` an abstract method in the `Plugin` parent class. This would force all plugins to implement it.

However, we can't do that immediately because not all existing plugins currently have this method. If we mark it abstract now, those plugins would break. This is precisely what issue #2438 addresses.

To move forward, I've asked Hendrik to implement resolve_action in this and VertexAIVectorSearch as a dummy method in this specific plugin.

yesudeep:
@hendrixmar is this PR still relevant?

hendrixmar:
Not really

yesudeep:
@hendrixmar  please close it if it is obsolete

PR #3026, opened 2025-06-04
feat(js): added dynamicTool factory function that does not depend on genkit instance

pavelgj:
```ts
import { dynamicTool } from 'genkit';

const dynamicGablorkenTool = dynamicTool(
  {
    name: 'dynamicGablorkenTool',
    inputSchema: z.object({
      value: z.number(),
    }),
    description: 'can be used to calculate gablorken value',
  },
  async (input) => {
    return input.value * 3 - 4;
  }
);

const { text } = ai.generateStream({
  model: googleAI.model('gemini-2.0-flash'),
  tools: [dynamicGablorkenTool],
  prompt: `what is a gablorken of ${input}`,
});
```

also added global (Genkit instance level) context:

```ts
const ai = genkit({
  context: { something: 'extra' }
});
```

Checklist (if applicable):


js/ai/src/generate.ts
```diff
@@ -271,8 +276,6 @@ async function toolsToActionRefs(
     } else if (typeof (t as ExecutablePrompt).asTool === 'function') {
       const promptToolAction = await (t as ExecutablePrompt).asTool();
       tools.push(`/prompt/${promptToolAction.__action.name}`);
-    } else if (t.name) {
```
ifielker, line 274: Can you explain why you removed this? I'm not sure I understand... is it to avoid registering dynamic tools in the non-ephemeral registry?

js/ai/src/generate.ts
```diff
@@ -375,6 +378,9 @@ function maybeRegisterDynamicTools<
       (t as Action).__action.metadata?.type === 'tool' &&
```
ifielker: Might this if statement be better as a general type guard? Like:
function isDynamicToolAction(t: unknown): t is DynamicToolAction {
  if (!t || typeof t !== 'object' || ...

possibly stacked with a isToolAction typeguard to simplify it:
if (!isToolAction(t)) { return false; }
return !!(t.__action.metadata?.dynamic)

js/ai/src/tool.ts
```diff
@@ -90,7 +83,37 @@ export type ToolAction<
...
+> = Action<I, O, z.ZodTypeAny, ToolRunOptions> &
+  Resumable<I, O> & {
+    __action: {
+      metadata: {
+        type: 'tool';
+      };
+    };
+  };
+
+/**
```
ifielker, line 105: The comments for these two types should not be the same. You should explain the difference between them and when each is used.

js/ai/src/tool.ts
```diff
@@ -174,10 +202,11 @@ export async function resolveTools<
         return asTool(registry, ref as Action);
       } else if (typeof (ref as ExecutablePrompt).asTool === 'function') {
         return await (ref as ExecutablePrompt).asTool();
-      } else if (ref.name) {
+      } else if ((ref as ToolDefinition).name) {
```
ifielker, line 207: This function would be nicer with type guards instead of all the casting

js/core/tests/action_test.ts
```diff
@@ -118,6 +118,48 @@ describe('action', () => {
     assert.deepStrictEqual(chunks, [1, 2, 3]);
   });
 
+  it('run the action with context plus registry global context', async () => {
```
ifielker: s/run/runs "it runs ..."

js/genkit/tests/flow_test.ts
```diff
@@ -78,4 +80,15 @@ describe('flow', () => {
     // a "streaming" flow can be invoked in non-streaming mode.
     assert.strictEqual(await streamingBananaFlow('banana2'), 'banana2');
   });
+
+  it('pass thought the context', async () => {
```
ifielker: s/pass/passes "it passes...
s/thought/through "...through the context"

js/genkit/tests/generate_test.ts
```diff
@@ -399,13 +406,59 @@ describe('generate', () => {
       );
     });
 
+    it('call the tool with context', async () => {
+      ai.defineTool(
+        { name: 'testTool', description: 'description' },
+        async (_, { context }) => JSON.stringify(context)
+      );
+
+      // first response be tools call, the subsequent just text response from agent b.
```
ifielker: The comment needs more English sentence structure. :)

js/ai/src/generate.ts
pavelgj: it is obsolete -- we used to allow passing `ToolDefinition` here as well, but it was removed and this resolver was never updated.

js/ai/src/generate.ts
pavelgj: good idea. done.

js/ai/src/tool.ts
pavelgj: Done.

PR #3030, opened 2025-06-04
fix: filter promptTemplate root spans from traceStore

MichaelDoyle:


MichaelDoyle:
This also filtered from inside traces.

PR #3049, opened 2025-06-10
fix(py): Support more complex schemas gemini

AbeJLazaro:
I was doing a recommender system by myself, and I found that gemini do not support this kind of output_schemas:
```
class Product(BaseModel):
    product_name: str = Field(description='name of the product')
    product_description: str = Field(description='description of the product')

class StoreSchema(BaseModel):
    store_name: str = Field(description='name of the store')
    products: list[Product] = Field(description='list of products in the store (5 to 10)')
```
So I decided to generate the bugfix

Checklist (if applicable):

### DIFF
--- a/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
+++ b/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
@@ -544,17 +544,23 @@ class GeminiModel:
         if defs is None:
             defs = input_schema.get('$defs') if '$defs' in input_schema else {}
 
+        if '$ref' in input_schema:
+            ref_path = input_schema['$ref']
+            ref_tokens = ref_path.split('/')
+            ref_name = ref_tokens[-1]
+
+            if ref_name not in defs:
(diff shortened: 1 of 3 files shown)


py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
```diff
@@ -538,12 +538,30 @@ def _convert_schema_property(
...
             Schema or None
         """
-        if input_schema is None or 'type' not in input_schema:
+        if input_schema is None:
             return None
 
         if defs is None:
             defs = input_schema.get('$defs') if '$defs' in input_schema else {}
 
+        if '$ref' in input_schema:
```
pavelgj, line 547: could you please add a unittest for this fix: https://github.com/firebase/genkit/blob/927dc6aa1f70ccefa54225663f20fbe353cf0a45/py/plugins/google-genai/test/models/test_googlegenai_gemini.py#L193

AbeJLazaro: Done!

PR #3050, opened 2025-06-10
fix: Update metric attribute length truncation to match Trace attribute max length.

bryanatkinson:
Attribute truncation should be consistent between metric data written and trace data written so that data can be correlated between aggregate metric data and specific traces.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/metrics.ts
+++ b/js/plugins/google-cloud/src/metrics.ts
@@ -25,7 +25,6 @@ import type { PathMetadata } from 'genkit/tracing';
 
 export const METER_NAME = 'genkit';
 export const METRIC_NAME_PREFIX = 'genkit';
-const METRIC_DIMENSION_MAX_CHARS = 32;
 
 export function internalMetricNamespaceWrap(...namespaces) {
   return [METRIC_NAME_PREFIX, ...namespaces].join('/');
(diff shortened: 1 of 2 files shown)


js/plugins/google-cloud/src/metrics.ts
```diff
@@ -103,15 +103,15 @@ export class MetricHistogram extends Metric<Histogram> {
...
+ * data can be properly correlated. This will truncate all dimensions for
+ * metrics such that they will be consistent with data included in traces.
  */
 function truncateDimensions(opts?: any) {
   if (opts) {
     Object.keys(opts).forEach((k) => {
-      // We don't want to truncate paths. They are known to be long but with
-      // relatively low cardinality, and are useful for downstream monitoring.
-      if (!k.startsWith('path') && typeof opts[k] == 'string') {
+      if (typeof opts[k] == 'string') {
```
schnecle: I think this should be an ===

schnecle (commented):
Do we need to do anything to the trace attributes as well?

bryanatkinson:
> Do we need to do anything to the trace attributes as well?

They are already being truncated at 256 by default

js/plugins/google-cloud/src/metrics.ts
bryanatkinson: Updated

schnecle (approved):

PR #3055, opened 2025-06-10
feat(py/genkit): added the resolve_method for openai compatible plugin

hendrixmar:
Description:

This PR introduces the concrete implementation of the resolve_method for the OpenAI compat plugin.

### DIFF
--- a/py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
+++ b/py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
@@ -16,7 +16,7 @@
 
 
 """OpenAI OpenAI API Compatible Plugin for Genkit."""
-from typing import Any
+from typing import Any, Callable
 
 from openai import Client, OpenAI as OpenAIClient
 
@@ -24,14 +24,25 @@ from genkit.ai._plugin import Plugin
 from genkit.ai._registry import GenkitRegistry
 from genkit.core.action.types import ActionKind
 from genkit.plugins.compat_oai.models import (
+    SUPPORTED_OPENAI_COMPAT_MODELS,
     SUPPORTED_OPENAI_MODELS,
     OpenAIModel,
     OpenAIModelHandler,
 )
+from genkit.plugins.compat_oai.models.model_info import get_default_openai_model_info
 from genkit.plugins.compat_oai.typing import OpenAIConfig
-from genkit.plugins.ollama.models import ModelDefinition
 
-OPENAI_PLUGIN_NAME = 'openai'
+
+def open_ai_name(name: str) -> str:
+    """Create a OpenAi-Compat action name.
+
+    Args:
+        name: Base name for the action.
+
+    Returns:
+        The fully qualified OpenAi-Compat action name.
+    """
+    return f'openai/{name}'
 
 def default_openai_metadata(name: str) -> dict[str, Any]:
     return {
@@ -83,6 +94,31 @@ class OpenAI(Plugin):
                 },
             )
 
+    def get_model_info(self, name: str) -> dict[str, str] | None:
+        """Retrieves metadata and supported features for the specified model.
+
+        This method looks up the model's information from a predefined list
+        of supported OpenAI-compatible models or provides default information.
+
+        Returns:
+            A dictionary containing the model's 'name' and 'supports' features,
+            or None if no information can be found (though typically, a default
+            is provided). The 'supports' key contains a dictionary representing
+            the model's capabilities (e.g., tools, streaming).
+        """
+
+        if model_supported := SUPPORTED_OPENAI_MODELS.get(name):
+            return {
+                'label': model_supported.label,
+                'supports': model_supported.supports.model_dump(exclude_none=True),
+            }
+
+        model_info = SUPPORTED_OPENAI_COMPAT_MODELS.get(name, get_default_openai_model_info(self))
+        return {
+            'label': model_info.label,
+            'supports': model_info.supports.model_dump(exclude_none=True),
+        }
+
     def resolve_action(  # noqa: B027
         self,
         ai: GenkitRegistry,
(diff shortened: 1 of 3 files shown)


py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -68,6 +72,50 @@ def initialize(self, ai: GenkitRegistry) -> None:
                     },
                 },
             )
+    def resolve_action(  # noqa: B027
```
AbeJLazaro, line 122: Let's add a blank line before any new method.

py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -68,6 +72,50 @@ def initialize(self, ai: GenkitRegistry) -> None:
...
+        self,
+        ai: GenkitRegistry,
+        kind: ActionKind,
+        name: str,
+    ) -> None:
+
+        if kind is not ActionKind.MODEL:
+            return None
+
+        if name not in SUPPORTED_OPENAI_MODELS:
```
AbeJLazaro: This line is not supposed to be here. The objective of resolve_action method is that the user could register new models that are not hard coded in the supported models.

py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -69,6 +73,80 @@ def initialize(self, ai: GenkitRegistry) -> None:
...
+        """
+
+        handler = OpenAIModelHandler(OpenAIModel(name, self._openai_client, ai)).generate
+        ai.define_model(
+            name=f'openai/{name}',
+            fn=handler,
+            config_schema=OpenAIConfig,
+            metadata={
+                'model': {
+                    'label': f"OpenAI - {name}",
```
AbeJLazaro: let's define the default supports for this metadata field.

py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -69,6 +73,80 @@ def initialize(self, ai: GenkitRegistry) -> None:
...
+                },
+            },
+        )
+
+    def resolve_action(  # noqa: B027
```
AbeJLazaro, line 122: The code is duplicated.

py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -69,6 +73,80 @@ def initialize(self, ai: GenkitRegistry) -> None:
                 },
             )
 
+    def resolve_action(  # noqa: B027
```
AbeJLazaro, line 122: let's also remove the # noqa: B027 comment

AbeJLazaro:
Let's check if it is required to keep using the handler. I think the model generate method is enough, and the handler is only a wrapper to determine which models are valid based on the supported models variable. We are going to deprecate those validations, so maybe the handler is not required.

AbeJLazaro (approved):

py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -69,6 +73,45 @@ def initialize(self, ai: GenkitRegistry) -> None:
...
+
+        handler = OpenAIModelHandler(OpenAIModel(name, self._openai_client, ai)).generate
+        ai.define_model(
+            name=f'openai/{name}',
+            fn=handler,
+            config_schema=OpenAIConfig,
+            metadata={
+                'model': {
+                    'label': f"OpenAI - {name}",
+                    'supports': {'multiturn': True}
```
pavelgj: you probably want to add other `supports` fields as well.. like system prompt?

AbeJLazaro: we saw that the initializate method only add the multiturn field in the metadata field. Should we change both methods to add all the supports object in the metadata? 

image

pavelgj: yes. A shared constant would be great.

py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -16,16 +16,30 @@
...
+from genkit.core.action.types import ActionKind
 from genkit.plugins.compat_oai.models import (
     SUPPORTED_OPENAI_MODELS,
+    OpenAIModel,
     OpenAIModelHandler,
 )
 from genkit.plugins.compat_oai.typing import OpenAIConfig
+from genkit.plugins.ollama.models import ModelDefinition
+
+OPENAI_PLUGIN_NAME = 'openai'
```
AbeJLazaro: what is the purpose of this constant?

py/plugins/compat-oai/src/genkit/plugins/compat_oai/openai_plugin.py
```diff
@@ -69,6 +83,40 @@ def initialize(self, ai: GenkitRegistry) -> None:
...
+            ai: The Genkit AI registry instance.
+            name: The name of the model to be registered.
+        """
+
+        handler = OpenAIModelHandler(OpenAIModel(name, self._openai_client, ai)).generate
+        ai.define_model(
+            name=f'openai/{name}',
+            fn=handler,
+            config_schema=OpenAIConfig,
+            metadata=default_openai_metadata(name)
```
AbeJLazaro: Can we validate that the model name is not in the supported models? like for model garden here: https://github.com/firebase/genkit/blob/main/py/plugins/vertex-ai/src/genkit/plugins/vertex_ai/model_garden/model_garden.py#L91

PR #3061, opened 2025-06-11
refactor(js): moved ALS code into separate node specific package

pavelgj:
Checklist (if applicable):


js/core/src/registry.ts
```diff
@@ -145,8 +145,17 @@ export class Registry {
     }
   }
 
+  /**
+   * Creates a new child registry overlaid onto this registry instance.
+   */
+  child(): Registry {
+    throw new Error('Method not implemented.');
```
ifielker, line 152: Shouldn't we implement the new way before deprecating the old one? (Or at least a the same time). It can be frustrating to have something be deprecated and the replacement unimplemented.

js/core/tests/registry_test.ts
```diff
@@ -264,7 +265,7 @@ describe('registry class', () => {
     });
 
     it('returns all registered actions, including parent', async () => {
-      const child = Registry.withParent(registry);
+      const child = registry.child();
```
ifielker, line 268: If child is unimplemented, how does this test work?

ifielker (approved):

MichaelDoyle:
@pavelgj you planning to pick this PR back up? or close it out, and revisit?

PR #3070, opened 2025-06-13
refactor(js): introduced background actions and reimplemented veo using them

pavelgj:
Checklist (if applicable):


js/ai/src/check-operation.ts
```diff
@@ -14,42 +14,27 @@
  * limitations under the License.
  */
 
-import { GenkitError } from '@genkit-ai/core';
+import { GenkitError, Operation } from '@genkit-ai/core';
 import { Registry } from '@genkit-ai/core/registry';
-import { GenerateRequest, ModelAction, ModelOperation } from './model';
 
-export async function checkOperation(
+export async function checkOperation<T>(
```
mbleigh:
```suggestion
export async function checkOperation<T = unknown>(
```

Always good to have generics have a fallback imo.

js/core/src/background-action.ts
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Background operation.
+ */
+export interface Operation<O = any> {
+  action?: string;
+  id: string;
+  done?: boolean;
+  output?: O;
+  metadata?: Record<string, any>;
+}
```
mbleigh: What happens if the operation errors? I think we can choose between:

1. The `check` operation throws.
2. We add an `error?: {message: string, [key: string]: unknown}`.

I think (2) is probably the way to go -- which I think should be a small change to the existing code and just some additional code on the Veo integration side.

mbleigh (commented):
How are we going to tackle background actions in Dev UI?

js/core/src/background-action.ts
pavelgj: I was kinda thinking #1, but agree #2 is better. Done.

pavelgj:
> How are we going to tackle background actions in Dev UI?

It's a new model type (`background-model`)... so will have to add the handling. I'm thinking custom UI, with polling.

PR #3084, opened 2025-06-16
feat(js): added id and metadata to executable prompts

pavelgj:
Checklist (if applicable):

### DIFF
--- a/js/ai/src/prompt.ts
+++ b/js/ai/src/prompt.ts
@@ -147,11 +147,8 @@ export interface ExecutablePrompt<
   O extends z.ZodTypeAny = z.ZodTypeAny,
   CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
 > {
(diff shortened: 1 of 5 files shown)


js/ai/src/prompt.ts
```diff
@@ -147,6 +147,12 @@ export interface ExecutablePrompt<
   O extends z.ZodTypeAny = z.ZodTypeAny,
   CustomOptions extends z.ZodTypeAny = z.ZodTypeAny,
 > {
+  /** Prompt id. */
+  id: string;
```
mbleigh: Should this be `name` (and I know it can't be `name` because conflicts with function prototype, but should this be maybe `__prompt: {name: string, metadata?: Record<string, any>`?

pavelgj: updated to ref

mbleigh (approved):

PR #3086, opened 2025-06-16
feat(compat_oai): Refactor to match proposal.

ssbushi:

Checklist (if applicable):


js/plugins/compat-oai/src/audio.ts
```diff
@@ -0,0 +1,217 @@
...
+  TranscriptionCreateParams,
+} from 'openai/resources/audio/index.mjs';
+
+export const RESPONSE_FORMAT_MEDIA_TYPES = {
+  mp3: 'audio/mpeg',
+  opus: 'audio/opus',
+  aac: 'audio/aac',
+  flac: 'audio/flac',
+  wav: 'audio/wav',
+  pcm: 'audio/L16',
```
ssbushi, line 41: Should this be `audio/pcm`?

pavelgj: I believe L16 and PCM are technically different... although might be the same thing in practice. I'm not sure.

js/plugins/compat-oai/src/openai/index.ts
```diff
@@ -0,0 +1,169 @@
...
+  o1Mini,
+  o1Preview,
+  o3,
+  o3Mini,
+  o4Mini,
+} from './gpt.js';
+import { SUPPORTED_TTS_MODELS, gpt4oMiniTts, tts1, tts1Hd } from './tts.js';
+import { SUPPORTED_STT_MODELS, gpt4oTranscribe, whisper1 } from './whisper.js';
+export {
+  dallE3,
```
pavelgj: I don't think we should export these constants anymore.
Models should be defined only via:
 - `'openai/gpt4'` (as string)
 - `openAI.model('gpt4')`

pavelgj: I saw you have a TODO for `openAI.model`. That's fine. Let's do string refs for now.

js/plugins/compat-oai/src/openai/index.ts
```diff
@@ -0,0 +1,169 @@
...
+ *    ... // other plugins
+ *  ]
+ * });
+ * ```
+ */
+export const openAI = (options?: OpenAIPluginOptions) =>
+  openAICompatible({
+    name: 'openai',
+    ...options,
+    initializer: async (ai, client) => {
```
pavelgj, line 67: how about dynamic resolution? still TODO?

pavelgj:
PR title: `feat(js/plugins/compat_oai): ...`  please.

js/plugins/compat-oai/src/openai/index.ts
ssbushi: Yes, I'll do that along with the other TODOs.

js/plugins/compat-oai/src/audio.ts
ssbushi: Gemini says, this is preferred for Open AI, I will leave it as it is for now.

pavelgj (approved):

PR #3090, opened 2025-06-16
feat(js): added abort signal to actions/flows

### LINKED ISSUE #3023: [JS] Support for AbortSignal
Currently there doesn't seem to be an idiomatic way to pass down a cancellation token / abort signal to an action (e g. flow or stream call), e.g. if a user presses a Stop button. 

Ideally we'd be able to pass down an AbortSignal as a separate option (outside any sort of input schema) to an action.

Using global variables and hooking into specific plugins may be possible, but quite hacky :-/

pavelgj:
Fixes #3023

```ts
const testFlow = defineFlow(
  registry,
  {
    name: 'testFlow',
  },
  async (input, { abortSignal }) => {
    if (abortSignal.aborted) return; // ABORT
    abortSignal.onabort = () => {
      // ABORT: abortSignal.reason
    };
  }
);
const ctrl = new AbortController();
const responsePromise = testFlow('foo', {
  abortSignal: ctrl.signal,
});
ctrl.abort('reason');
```

Checklist (if applicable):


js/ai/src/model.ts
```diff
@@ -464,7 +481,19 @@ export function defineModel<
   options: DefineModelOptions<CustomOptionsSchema>,
   runner: (
     request: GenerateRequest<CustomOptionsSchema>,
-    streamingCallback?: StreamingCallback<GenerateResponseChunkData>
+    streamingCallback: StreamingCallback<GenerateResponseChunkData> | undefined
```
ifielker: This is a breaking change, and fairly certain it's user facing. Everyone who uses defineModel (e.g. plugins etc.) will need to update all their call sites. This is for non v2 models and streamingCallback is going from optional to required, even if the value is undefined. So before runner: { request } is ok, and now it needs to be runner: { request, streamingCallback: undefined }.

js/ai/src/model.ts
```diff
@@ -464,7 +481,19 @@ export function defineModel<
...
+): ModelAction<CustomOptionsSchema>;
+
+export function defineModel<
+  CustomOptionsSchema extends z.ZodTypeAny = z.ZodTypeAny,
+>(
+  registry: Registry,
+  options: any,
+  runner: (
+    request: GenerateRequest<CustomOptionsSchema>,
+    // GenerateRequest<CustomOptionsSchema> | DefineModelOptionsV2<CustomOptionsSchema>
```
ifielker: Not sure if you left this line by accident?

js/ai/src/model.ts
```diff
@@ -489,10 +518,16 @@ export function defineModel<
...
       use: middleware,
     },
-    (input) => {
+    (input, ctx) => {
       const startTimeMs = performance.now();
-
-      return runner(input, getStreamingCallback(registry)).then((response) => {
+      return runner(
+        input,
+        options.apiVersion === 'v2'
```
ifielker, line 523: This second parameter is a bit difficult to read. I think it might be better to have 
const streamingCallback = options.apiVersion === 'v2' ? ... 
and then 
return runner(input, streamingCallback).then(...
because then it's clear what that whole expression is about.

js/core/src/action.ts
```diff
@@ -368,12 +390,16 @@ export function detachedAction<
...
+                  streamingRequested:
+                    !!options?.onChunk &&
+                    options.onChunk !== sentinelNoopStreamingCallback,
                   sendChunk: options?.onChunk ?? sentinelNoopStreamingCallback,
                   trace: {
                     traceId,
                     spanId,
                   },
                   registry,
+                  abortSignal: options?.abortSignal ?? makeNoopAbortSignal(),
```
ifielker, line 402: why do we need a no-op abort signal? Why not just undefined?

js/core/src/action.ts
```diff
@@ -101,6 +118,11 @@ export interface ActionFnArg<S> {
     traceId: string;
     spanId: string;
   };
+
+  /**
+   * Abort signal for the action request.
+   */
+  abortSignal: AbortSignal;
```
ifielker, line 125: Why not potentially undefined if there's no abort signal?

js/core/src/flow.ts
```diff
@@ -55,26 +54,8 @@ export interface FlowConfig<
  * side-channel context data. The context itself is a function, a short-cut
  * for streaming callback.
  */
-export interface FlowSideChannel<S> {
+export interface FlowSideChannel<S> extends ActionFnArg<S> {
```
ifielker, line 57: Isn't this a breaking change? ActionFnArg has new required fields... so this does too?

js/genkit/src/genkit.ts
```diff
@@ -241,14 +242,40 @@ export class Genkit implements HasRegistry {
...
+
   /**
    * Defines a new model and adds it to the registry.
    */
   defineModel<CustomOptionsSchema extends z.ZodTypeAny = z.ZodTypeAny>(
     options: DefineModelOptions<CustomOptionsSchema>,
     runner: (
       request: GenerateRequest<CustomOptionsSchema>,
-      streamingCallback?: StreamingCallback<GenerateResponseChunkData>
+      streamingCallback:
```
ifielker: breaking change here too.

js/ai/src/model.ts
pavelgj: I actually don't need this, was wrestling with TS type inference with function overloading... 
JS is weird with function signatures... it treats them all as optional (`?`), so in this case it actually makes no difference (we have tons of tests for all the cases).

js/ai/src/model.ts
pavelgj: no, this is on purpose. With method overloading I couldn't get a more narrow type than `any`, so the comment just shows the actual union type.

js/core/src/action.ts
pavelgj: it makes it easier to not have to worry about whether these things are passed or not, we just pass in a noop abort signal if caller did not provide one. It should not impact the logic of the action.

js/core/src/action.ts
pavelgj: explained in the previous comment.

js/core/src/action.ts
pavelgj: we actually follow this pattern with streaming callbacks. It makes the code cleaner not having to do null checks everywhere.

js/core/src/flow.ts
pavelgj: we set the field, this interface is not meant for the user to implement. We guarantee that the field will always be implemented.

js/ai/src/model.ts
```diff
@@ -489,10 +518,15 @@ export function defineModel<
       },
       use: middleware,
     },
-    (input) => {
+    (input, ctx) => {
       const startTimeMs = performance.now();
-
-      return runner(input, getStreamingCallback(registry)).then((response) => {
+      const secondParam =
```
ifielker, line 522: What actually is the secondParam? I looks like it is context | callback | boolean | undefined... which seems really weird... am I missing something?

pavelgj: yes, second param is `ActionFnArg<GenerateResponseChunkData> | StreamingCallback<GenerateResponseChunkData> | undefined`. That is the overload. If you define model as:

```ts
defineModel({
  name: 'foo',
  apiVersion: 'v2'
}, async (input, ctx) => {...})
```

then `ctx` is `ActionFnArg<GenerateResponseChunkData>`. Otherwise:

```ts
defineModel({
  name: 'foo',
}, async (input, streamingCallback) => {...})
```

`streamingCallback` is `StreamingCallback<GenerateResponseChunkData> | undefined`.

The overload signatures above will ensure that these are the only options.

js/ai/src/model.ts
```diff
@@ -466,6 +483,18 @@ export function defineModel<
...
+): ModelAction<CustomOptionsSchema>;
+
+export function defineModel<
+  CustomOptionsSchema extends z.ZodTypeAny = z.ZodTypeAny,
+>(
+  registry: Registry,
+  options: any,
+  runner: (
+    request: GenerateRequest<CustomOptionsSchema>,
+    // GenerateRequest<CustomOptionsSchema> | DefineModelOptionsV2<CustomOptionsSchema>
```
pavelgj:
```suggestion
```

zerolethanh:
@pavelgj If I send a abortSignal from client (like web or app) as describe here (https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal), this abort function might working?

PR #3091, opened 2025-06-16
fix(js/plugins/googleai): list veo models in actions list

pavelgj:
Checklist (if applicable):

### DIFF
--- a/js/ai/src/model.ts
+++ b/js/ai/src/model.ts
@@ -622,16 +622,18 @@ export function modelActionMetadata({
   name,
   info,
   configSchema,
+  background,
 }: {
   name: string;
   info?: ModelInfo;
   configSchema?: z.ZodTypeAny;
+  background?: boolean;
 }): ActionMetadata {
   return {
-    actionType: 'model',
+    actionType: background ? 'background-model' : 'model',
     name: name,
     inputJsonSchema: toJsonSchema({ schema: GenerateRequestSchema }),
-    outputJsonSchema: toJsonSchema({ schema: GenerateResponseSchema }),
+    outputJsonSchema: toJsonSchema({ schema: OperationSchema }),
     metadata: {
       model: {
         ...info,
--- a/js/plugins/googleai/src/index.ts
+++ b/js/plugins/googleai/src/index.ts
@@ -64,7 +64,12 @@ import {
   type KNOWN_IMAGEN_MODELS,
 } from './imagen.js';
 import { listModels } from './list-models.js';
-import { KNOWN_VEO_MODELS, VeoConfigSchema, defineVeoModel } from './veo.js';
+import {
+  GENERIC_VEO_INFO,
+  KNOWN_VEO_MODELS,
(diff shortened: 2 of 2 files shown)


js/ai/src/model.ts
```diff
@@ -622,16 +622,18 @@ export function modelActionMetadata({
...
   configSchema?: z.ZodTypeAny;
+  background?: boolean;
 }): ActionMetadata {
   return {
-    actionType: 'model',
+    actionType: background ? 'background-model' : 'model',
     name: name,
     inputJsonSchema: toJsonSchema({ schema: GenerateRequestSchema }),
-    outputJsonSchema: toJsonSchema({ schema: GenerateResponseSchema }),
+    outputJsonSchema: toJsonSchema({ schema: OperationSchema }),
```
shrutip90: I see GenerateResponseSchema now includes operation. Do we need this change?

js/plugins/googleai/src/index.ts
```diff
@@ -173,15 +178,17 @@ async function resolver(
...
-      break;
-    default:
-    // no-op
+  if (actionType === 'embedder') {
+    resolveEmbedder(ai, actionName, options);
+  } else if (actionName.startsWith('veo')) {
+    // we do it this way because the request may come in for
+    // action type 'model' and action name 'veo-...'. That case should
+    // be a noop. It's just the order or model lookup.
+    if (actionType === 'background-model') {
```
ifielker, line 187: It might be nicer to just have 3 switch cases 'model', 'background-model' and 'embedder' and each goes to its respective resolve\<X\>. Then in those resolveX functions you can look at the actionName and decide what you want to do with it if anything. (e.g. resolveModel does nothing with Veo, and resolveBackgroundModel does nothing without Veo ... for now... until there are more background models). This is more expandable in future without becoming a complicated page of nested conditions.

ifielker (approved):

js/ai/src/model.ts
pavelgj: This was an oops. Fixed -- it needed to be conditional.

js/plugins/googleai/src/index.ts
pavelgj: I don't think there's a way to avoid nested conditions. With 3 switch cases it would be: case statements with nested if conditions, which I'm not sure is better... because it was the source of the original issue -- had to handle veo models in two separate code paths.

```ts
switch (actionType) {
  case 'model':
    if (!actionName.startsWith('veo')) {
      resolveModel(ai, actionName, options);
    }
    break;
  case 'background-model':
    if (actionName.startsWith('veo')) {
      defineVeoModel(ai, actionName, options?.apiKey);
    }
    break;
  case ...
}
```

I'm going to leave as is. Not set in stone, can refactor later.

js/ai/src/model.ts
pavelgj: Background models returns an operation, always.

PR #3107, opened 2025-06-19
fix(js/plugins/google-cloud): Write input and output logs for generate step

bryanatkinson:
Previously the 'generate' helpers wrote their full input/output logs only when it is a root node (outside a flow). This change ensures that those logs always get written. They provide useful information in the Firebase console.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/telemetry/action.ts
+++ b/js/plugins/google-cloud/src/telemetry/action.ts
@@ -32,24 +32,23 @@ class ActionTelemetry implements Telemetry {
     logInputAndOutput: boolean,
     projectId?: string
   ): void {
+    if (!logInputAndOutput) {
+      return;
+    }
     const attributes = span.attributes;
-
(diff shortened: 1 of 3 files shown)


js/plugins/google-cloud/src/telemetry/action.ts
```diff
@@ -42,7 +42,10 @@ class ActionTelemetry implements Telemetry {
       featureName = actionName;
     }
 
-    if (subtype === 'tool' && logInputAndOutput) {
+    if (
+      (subtype === 'tool' || actionName === 'generate') &&
```
schnecle: Are there other times where logInputAndOutput is true on different action types?

schnecle (approved):

js/plugins/google-cloud/src/telemetry/action.ts
bryanatkinson: logInputAndOutput is a global on/off flag for writing logs. Updated to early-out this function if it's false.

PR #3108, opened 2025-06-19
feat(js/plugins/google-genai): Added embedders for GoogleAI and VertexAI

ifielker:
Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/googleai/embedder.ts
+++ b/js/plugins/google-genai/src/googleai/embedder.ts
@@ -1,0 +1,142 @@
+/**
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


js/plugins/google-genai/src/googleai/embedder.ts
```diff
@@ -0,0 +1,142 @@
...
+});
+
+export type GeminiEmbeddingConfig = z.infer<
+  typeof GoogleAIEmbeddingConfigSchema
+>;
+
+// TODO(ifielker): Update embedders to be current models.
+// (textEmbeddingGecko001 is outdated).
+
+const textEmbedding004 = embedderRef({
```
pavelgj, line 56: for future cleanup -- we probably don't need full reference objects, only ModelInfo

ifielker: Noted.

PR #3113, opened 2025-06-19
fix(js): process spans in isolation to improve accuracy of telemetry

andrewbrook:
Span processing in the google-cloud plugin is currently done in batches, which are chosen by the OTel libraries. We erroneously rely on each batch being a complete representation of a trace, and to contain only spans from a single trace. Since these assumptions are incorrect there are currently subtle bugs where span metrics and logs are being mis-attributed, or dropped. This change removes the need for the above assumptions, by processing each span entirely independently. 

We also remove some unused logs/metrics for simplicity, and improve the test coverage around "path" telemetry.

Checklist (if applicable):


js/core/src/tracing/types.ts
```diff
@@ -43,6 +43,8 @@ export const SpanMetadataSchema = z.object({
   isRoot: z.boolean().optional(),
   metadata: z.record(z.string(), z.string()).optional(),
   path: z.string().optional(),
+  // Indicates a "leaf" span that is the source of a failure.
+  isFailedSpan: z.boolean().optional(),
```
bryanatkinson: This is a little confusing because of the state attribute above. The comment helps, but I wonder if we named it something like `isFailureSourceSpan` or `isErrorOriginSpan`

bryanatkinson (approved):

js/core/src/tracing/types.ts
schnecle: I like `isFailureSource` to keep it shorter, but agree that it's a little confusing without the context that it is the source of the failure.

schnecle (approved):
This looks great - the simplification is so good! Thanks Andrew!

PR #3115, opened 2025-06-20
Update gemini models

Julioevm:
 Add the final gemini model, and update the latest preview ones. Also include Flash lite preview.
Checklist (if applicable):

I wonder if these need to be added somewhere else?


MichaelDoyle:
googleai plugin was deprecated in favor of google genai. these models are available (and old at this point)

PR #3116, opened 2025-06-20
fix(js/plugins/google-genai): Fix tests for node 23

ifielker:
Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/common/types.ts
+++ b/js/plugins/google-genai/src/common/types.ts
@@ -119,7 +119,7 @@ export enum DynamicRetrievalMode {
 /**
  * Schema for parameters passed to {@link FunctionDeclaration.parameters}.
  */
-export interface FunctionDeclarationSchema {
+export declare interface FunctionDeclarationSchema {
   /** The type of the parameter. */
   type: SchemaType;
   /** The format of the parameter. */
@@ -252,7 +252,7 @@ export declare interface PromptFeedback {
 /**
  * URI based data.
  */
-export interface FileData {
+export declare interface FileData {
   /** The IANA standard MIME type of the source data. */
   mimeType: string;
   /** URI of the file. */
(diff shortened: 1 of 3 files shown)


js/plugins/google-genai/src/common/types.ts
```diff
@@ -119,7 +119,7 @@ export enum DynamicRetrievalMode {
 /**
  * Schema for parameters passed to {@link FunctionDeclaration.parameters}.
  */
-export interface FunctionDeclarationSchema {
+export declare interface FunctionDeclarationSchema {
```
pavelgj: just noticed, Genkit naming conventions reserves `*Schema` suffix for Zod schema. Call this `FunctionDeclaration`.

pavelgj: you actually don't need `declare` if you're importing as `type` (which I see that you do as well)

pavelgj: oh, you don't... just update imports to `import { type FunctionDeclarationSchema } from '...'`

js/plugins/google-genai/src/common/types.ts
```diff
@@ -334,7 +334,7 @@ export interface Schema {
...
```
pavelgj, line 315: wait... you have FunctionDeclarationSchema twice

js/plugins/google-genai/src/common/types.ts
ifielker: Do actually need it, (discussed offline).

js/plugins/google-genai/src/common/types.ts
ifielker: oops! fixed.

js/plugins/google-genai/src/common/types.ts
ifielker: Discussed offline. JSON schema is a reasonable exception to the rule.

pavelgj (approved):

PR #3117, opened 2025-06-20
feat(js/plugins/compat-oai): Support dynamic model resolution

ssbushi:
Checklist (if applicable):

### DIFF
--- a/js/plugins/compat-oai/src/openai/index.ts
+++ b/js/plugins/compat-oai/src/openai/index.ts
@@ -16,6 +16,20 @@
  */
 
 import {
+  ActionMetadata,
+  embedderActionMetadata,
+  embedderRef,
+  EmbedderReference,
+  Genkit,
+  modelActionMetadata,
+  modelRef,
+  ModelReference,
+  z,
+} from 'genkit';
+import { GenkitPlugin } from 'genkit/plugin';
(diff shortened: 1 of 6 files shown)


js/plugins/compat-oai/src/openai/index.ts
```diff
@@ -101,6 +205,59 @@ export const openAI = (options?: OpenAIPluginOptions) =>
...
     },
+    resolver,
+    listActions,
+  });
+}
+
+export type OpenAIPlugin = {
+  (params?: OpenAIPluginOptions): GenkitPlugin;
+  model(name: string, config?: any): ModelReference<z.ZodTypeAny>;
+  embedder(name: string, config?: any): EmbedderReference<z.ZodTypeAny>;
```
ssbushi, line 216: Should this params be more tightly typed, like in googleai?

pavelgj: probably a good idea because otherwise code completion for configs on various model types won't work well... feel free to do it as a follow up.

pavelgj (approved):

js/plugins/compat-oai/src/openai/index.ts
ssbushi: Ack.

PR #3124, opened 2025-06-23
fix(js/plugins/google-cloud): Add input and output logs for prompt rendering.

bryanatkinson:
Logs for the inputs to and output from prompt rendering now written.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
+++ b/js/plugins/google-cloud/src/gcpOpenTelemetry.ts
@@ -351,27 +351,15 @@ class AdjustingTraceExporter implements SpanExporter {
       featuresTelemetry.tick(span, this.logInputAndOutput, this.projectId);
       // Set root status explicitly
       span.attributes['genkit:rootState'] = span.attributes['genkit:state'];
-    } else {
-      if (type === 'action' && subtype === 'model') {
-        // Report generate metrics () for all model actions
-        generateTelemetry.tick(span, this.logInputAndOutput, this.projectId);
-      }
-      if (type === 'action' && subtype === 'tool') {
-        // TODO: Report input and output for tool actions
-      }
-      if (
-        type === 'action' ||
-        type === 'flow' ||
-        type == 'flowStep' ||
-        type == 'util'
-      ) {
-        // Report request and latency metrics for all actions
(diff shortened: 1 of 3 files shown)


js/plugins/google-cloud/src/gcpOpenTelemetry.ts
```diff
@@ -351,27 +351,15 @@ class AdjustingTraceExporter implements SpanExporter {
...
-    if (type === 'userEngagement') {
+    } else if (type === 'action' && subtype === 'model') {
+      // Report generate metrics () for all model actions
+      generateTelemetry.tick(span, this.logInputAndOutput, this.projectId);
+    } else if (type === 'userEngagement') {
       // Report user acceptance and feedback metrics
       engagementTelemetry.tick(span, this.logInputAndOutput, this.projectId);
+    } else {
+      // Report request and latency metrics for all actions
+      actionTelemetry.tick(span, this.logInputAndOutput, this.projectId);
```
andrewbrook, line 362: This block is no longer hit if [type=action, subtype=model]. Is that intended?

andrewbrook (approved):

js/plugins/google-cloud/src/gcpOpenTelemetry.ts
bryanatkinson: Yep! The action telemetry now only writes logs, and only for specific action types (prompt, generate action wrapper, and tools), and so although this was previously called for the model actions it was a no-op. We write logs for those actions above with the 'generateTelemetry'.

schnecle (approved):
I love this. Are there any other action types that we should include. For instance in beta the send action in chat should likely also fall into this category, and it's action type is "helper" I believe?

bryanatkinson:
> I love this. Are there any other action types that we should include. For instance in beta the send action in chat should likely also fall into this category, and it's action type is "helper" I believe?

I could either add 'helper' to the check, or remove the check altogether so that any action type with input or output would write the data. Do you think there are any action types we explicitly wouldn't want to write input and output for?

MichaelDoyle:
@bryanatkinson @schnecle @andrewbrook worth merging? or close

MichaelDoyle:
closed per feedback

PR #3131, opened 2025-06-25
feat(js/plugins/compat-oai): Add `raw` field in responses, support config.tools

ssbushi:
Also manually tested web search built in tool.

Checklist (if applicable):

### DIFF
--- a/js/testapps/compat-oai/src/index.ts
+++ b/js/testapps/compat-oai/src/index.ts
@@ -41,6 +41,23 @@ export const jokeFlow = ai.defineFlow(
   }
 );
 
+export const toolFlow = ai.defineFlow(
+  {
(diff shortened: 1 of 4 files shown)


js/testapps/compat-oai/src/index.ts
```diff
@@ -41,6 +41,23 @@ export const jokeFlow = ai.defineFlow(
   }
 );
 
+export const toolFlow = ai.defineFlow(
+  {
+    name: 'toolFlow',
+    outputSchema: z.string(),
```
pavelgj:
```suggestion
export const webSearchFlow = ai.defineFlow(
  {
    name: 'webSearchFlow',
    outputSchema: z.string(),
```

PR #3133, opened 2025-06-26
feat(js/ai): added support for dynamically resolvable resources

pavelgj:
```ts
ai.defineResource({
  uri: 'my://resource/{param}',
  description: 'provides my resource',
}, async ({param}) => {
  return [{ text: `resource ${param}` }]
});

await ai.generate({
  prompt: [{ resource: 'my://resource/value' }]
})
```

Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate/action.ts
+++ b/js/ai/src/generate/action.ts
@@ -42,6 +42,7 @@ import {
   GenerateActionOptionsSchema,
   GenerateResponseChunkSchema,
   GenerateResponseSchema,
+  MessageData,
   resolveModel,
   type GenerateActionOptions,
   type GenerateActionOutputConfig,
@@ -56,6 +57,7 @@ import {
   type Part,
   type Role,
 } from '../model.js';
+import { findMatchingResource } from '../resource.js';
 import { resolveTools, toToolDefinition, type ToolAction } from '../tool.js';
 import {
   assertValidToolNames,
@@ -245,6 +247,7 @@ async function generate(
     rawRequest
   );
   rawRequest = applyFormat(rawRequest, format);
+  rawRequest = await applyResources(registry, rawRequest);
 
   // check to make sure we don't have overlapping tool names *before* generation
   await assertValidToolNames(tools);
@@ -470,4 +473,51 @@ function getRoleFromPart(part: Part): Role {
   if (part.media !== undefined) return 'user';
   if (part.data !== undefined) return 'user';
   throw new Error('No recognized fields in content');
+}
+
+async function applyResources(
+  registry: Registry,
+  rawRequest: GenerateActionOptions
+): Promise<GenerateActionOptions> {
+  // quick check, if no resources bail.
+  if (!rawRequest.messages.find((m) => !!m.content.find((c) => c.resource))) {
+    return rawRequest;
+  }
+
+  const updatedMessages = [] as MessageData[];
+  for (const m of rawRequest.messages) {
+    if (!m.content.find((c) => c.resource)) {
+      updatedMessages.push(m);
+      continue;
+    }
+    const updatedContent = [] as Part[];
+    for (const p of m.content) {
+      if (!p.resource) {
+        updatedContent.push(p);
+        continue;
+      }
+      const resource = await findMatchingResource(registry, p.resource);
+      if (!resource) {
+        throw new GenkitError({
+          status: 'NOT_FOUND',
+          message: `failed to find matching resource for ${p.resource}`,
+        });
+      }
+      const resourceParts = await resource(p.resource);
+      resourceParts.forEach((r) => {
+        r.metadata = { resource: resource.__action.name, uri: p.resource };
+      });
+      updatedContent.push(...resourceParts);
+    }
+
+    updatedMessages.push({
+      ...m,
+      content: updatedContent,
+    });
+  }
+
+  return {
+    ...rawRequest,
+    messages: updatedMessages,
+  };
 }
--- a/js/ai/src/resource.ts
+++ b/js/ai/src/resource.ts
@@ -1,0 +1,126 @@
+/**
+ * Copyright 2025 Google LLC
+ *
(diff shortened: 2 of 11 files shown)


js/ai/src/generate/action.ts
```diff
@@ -471,3 +474,50 @@ function getRoleFromPart(part: Part): Role {
...
+      const resource = await findMatchingResource(registry, p.resource);
+      if (!resource) {
+        throw new GenkitError({
+          status: 'NOT_FOUND',
+          message: `failed to find matching resource for ${p.resource}`,
+        });
+      }
+      const resourceParts = await resource(p.resource);
+      resourceParts.forEach((r) => {
+        r.metadata = { resource: resource.__action.name, uri: p.resource };
```
mbleigh:
```suggestion
        r.metadata = { resource: {name: resource.__action.name, uri: p.resource }};
```

Let's bundle all the resource-related metadata together.

js/ai/src/resource.ts
```diff
@@ -0,0 +1,126 @@
...
+import { Part, PartSchema } from './model.js';
+
+/**
+ * Options for defining a resource.
+ */
+export interface ResourceOptions {
+  /**
+   * The URI of the resource. Can contain template variables.
+   */
+  uri: string;
```
mbleigh: I wanted to differentiate between template and non-template resources because MCP does. I want this to map super-easily to MCP and unless we have a way to detect whether a URI does or doesn't have template params I'd rather have `uri` option for one and `template` option for the other -- it can resolve to the same kind of action etc. I just want to be able to trivially provide to MCP.

js/ai/src/resource.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * A description of the resource.
+   */
+  description?: string;
+}
```
mbleigh, line 56: Missing `metadata` here (important because I need to be able to annotate MCP metadata with `metadata: {mcp: {_meta: {...}}}`)

mbleigh (commented):
Looking good, few comments and higher-level questions:

1. Will resource resolution show up in traces?
2. Should we put this in beta or go straight to stable?
3. At the moment the action name is the URI but the metadata gets annotated with both the name and the URI - should we either allow an (optional?) name for a resource template or have metadata just be `{resource: {uri: "..."}}`
4. Should we consider the part type being `{resource: {uri: "..."}}` instead of `{resource:  string}`? I'm torn, because I like the simplicity a lot but worry if we might add something later...maybe better to keep it simple.

js/ai/src/resource.ts
pavelgj: Done.

js/ai/src/resource.ts
pavelgj: Added.

pavelgj:
> Looking good, few comments and higher-level questions:
> 
> 1. Will resource resolution show up in traces?

Yes:
image

> 2. Should we put this in beta or go straight to stable?

I'm fine with stable if you're happy with the API.

> 3. At the moment the action name is the URI but the metadata gets annotated with both the name and the URI - should we either allow an (optional?) name for a resource template or have metadata just be `{resource: {uri: "..."}}`

you're talking about generate message metadata (on the part)?
right now it's:

```ts
        metadata: {
          resource: {
            name: 'test://resource/{param}',
            uri: 'test://resource/value',
          },
        },
```

`name` referring the action name. Action name is derived from either `uri` or `template` (whichever one is provided). 
The `uri` in metadata might be misleading... I'm thinking `value` instead.

```ts
        metadata: {
          resource: {
            name: 'test://resource/{param}',
            value: 'test://resource/value',
          },
        },
```


> 4. Should we consider the part type being `{resource: {uri: "..."}}` instead of `{resource:  string}`? I'm torn, because I like the simplicity a lot but worry if we might add something later...maybe better to keep it simple.

yeah... judgement call. It's conceivable that some other options might get added....

js/ai/src/resource.ts
```diff
@@ -0,0 +1,184 @@
...
+   */
+  name?: string;
+
+  /**
+   * The URI of the resource. Can contain template variables.
+   */
+  uri?: string;
+
+  /**
+   * The URI tempalate (ex. `my://resource/{id}`). See RFC6570 for specification.
```
mbleigh:
```suggestion
   * The URI template (ex. `my://resource/{id}`). See RFC6570 for specification.
```

js/ai/src/resource.ts
```diff
@@ -0,0 +1,184 @@
...
+  const uri = opts.uri ?? opts.template;
+  if (!uri) {
+    throw new GenkitError({
+      status: 'INVALID_ARGUMENT',
+      message: `must specify either url or template options`,
+    });
+  }
+  const template = opts.template ? uriTemplate(opts.template) : undefined;
+  const matcher = opts.uri
+    ? (input: string) => (input === opts.uri ? {} : undefined)
```
mbleigh: Annoying pedantic nit (maybe just worth a `TODO` or a low-pri issue): should we be normalizing the URI before checking for equality?

`foo://bar?baz=1&qux=2` and `foo://bar?qux=2&baz=1` are equivalent URIs but would not match.

js/ai/src/resource.ts
```diff
@@ -0,0 +1,184 @@
...
+      };
+
+  const action = defineAction(
+    registry,
+    {
+      actionType: 'resource',
+      name: opts.name ?? uri,
+      description: opts.description,
+      inputSchema: ResourceInputSchema,
+      outputSchema: z.array(PartSchema),
```
mbleigh: Should we do `{content: Part[]}` here just to (again) leave ourselves wiggle room for future things? You can call this in either direction and I'm good.

pavelgj: Yes, was thinking the same thing. Done.

js/ai/src/resource.ts
pavelgj: good point. Added a todo.

PR #3136, opened 2025-06-26
patch(#3133): added resource name, handle parent resources

pavelgj:
Checklist (if applicable):

### DIFF
--- a/js/ai/src/resource.ts
+++ b/js/ai/src/resource.ts
@@ -142,9 +142,9 @@ export function defineResource(
             (p.metadata as any).resource.parent = {
               uri: input.uri,
             };
-          if (opts.template) {
-            (p.metadata as any).resource.parent.template = opts.template;
-          }
+            if (opts.template) {
+              (p.metadata as any).resource.parent.template = opts.template;
+            }
           }
         } else {
           (p.metadata as any).resource = {
--- a/js/ai/src/resource.ts
+++ b/js/ai/src/resource.ts
@@ -30,6 +30,11 @@ import { Part, PartSchema } from './model.js';
  */
 export interface ResourceOptions {
   /**
+   * Resource name.
+   */
+  name: string;
+
+  /**
    * The URI of the resource. Can contain template variables.
    */
   uri?: string;
@@ -54,7 +59,7 @@ export interface ResourceOptions {
  * A function that returns parts for a given resource.
  */
 export type ResourceFn = (
-  vars: Record<string, string>,
+  input: string,
   ctx: ActionContext
 ) => Part[] | Promise<Part[]>;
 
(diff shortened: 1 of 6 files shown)


js/ai/src/resource.ts
```diff
@@ -121,7 +126,25 @@ export function defineResource(
...
+        }
+        if (p.metadata?.resource) {
+          (p.metadata as any).resource.parent = {
+            uri: input,
+            ...(p.metadata as any).resource.parent,
+          };
+        } else {
+          (p.metadata as any).resource = {
+            name: opts.name,
+            uri: input,
```
mbleigh: I think the important metadata is `uri` and `template` (I'm ambivalent about including name, but do think template should be included if it exists).

js/ai/src/resource.ts
```diff
@@ -54,7 +59,7 @@ export interface ResourceOptions {
  * A function that returns parts for a given resource.
  */
 export type ResourceFn = (
-  vars: Record<string, string>,
+  input: string,
```
mbleigh: let's make input `{uri: string}` instead of just `string` so we leave ourselves options to enhance in the future

js/ai/src/resource.ts
```diff
@@ -29,6 +29,11 @@ import { Part, PartSchema } from './model.js';
  * Options for defining a resource.
  */
 export interface ResourceOptions {
+  /**
+   * Resource name.
+   */
+  name: string;
```
mbleigh: let's make this optional with the name defaulting to the specified uri or template

js/ai/src/resource.ts
```diff
@@ -121,7 +126,25 @@ export function defineResource(
...
+      parts.map((p) => {
+        if (!p.metadata) {
+          p.metadata = {};
+        }
+        if (p.metadata?.resource) {
+          (p.metadata as any).resource.parent = {
+            uri: input,
+            ...(p.metadata as any).resource.parent,
+          };
+        } else {
```
mbleigh: How about:

1. if metadata.resource.parent already exists, leave it alone entirely
2. if metadata.resource already exists, populate metadata.resource.parent with {uri,template}
3. if metadata.resource doesn't exist, populate metadata.resource with {uri,template}

js/ai/src/resource.ts
pavelgj: Done.

js/ai/src/resource.ts
pavelgj: Done.

js/ai/src/resource.ts
pavelgj: Done.

mbleigh (approved):

PR #3165, opened 2025-07-07
fix(go): exclude reasoning part when return value from Text() function

jcooky:
I’ve noticed that the `Text()` function is meant to operate on `TextPart` or `DataPart` only. If a developer wants to retrieve the reasoning part as text, they should use `Reasoning()` instead.

However, `Text()` currently returns a merged value that includes the reasoning part, which does not match the intended behavior.

Checklist (if applicable):

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -768,13 +768,16 @@ func (c *ModelResponseChunk) Text() string {
 	}
 	var sb strings.Builder
 	for _, p := range c.Content {
-		sb.WriteString(p.Text)
+		if p.IsText() || p.IsData() {
+			sb.WriteString(p.Text)
+		}
 	}
 	return sb.String()
 }
 
 // Text returns the contents of a [Message] as a string. It
 // returns an empty string if the message has no content.
+// If you want to get reasoning from the message, use Reasoning() instead.
 func (m *Message) Text() string {
 	if m == nil {
 		return ""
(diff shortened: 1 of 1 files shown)


huangjeff5:
Good catch! But should we be including Data() parts in Text()?

jcooky:
@huangjeff5 The `Output()` function currently calls `Text()` for JSON parsing, but I believed the intended design is to use the Data part for that.

For backward compatibility, `Text()` still includes the Data part. In both the OpenAI and Gemini codebases, JSON output is written into the Text part.

You can find the relevant code here:
- https://github.com/firebase/genkit/blob/310f45807bf84c71b2c443d562b3a82247e60350/go/plugins/compat_oai/generate.go#L338-L385
- https://github.com/firebase/genkit/blob/310f45807bf84c71b2c443d562b3a82247e60350/go/plugins/googlegenai/gemini.go#L682-L702

These implementations should use the Data part instead of the Text part when the output mode is JSON. If we remove the Data part from `Text()`, we need to update them to support the OutputMode and write output to the Data part instead of Text.

apascal07:
We call `Text()` for JSON parsing because JSON is usually (always?) returned as markdown which is text not valid JSON data, which we then strip the markdown delimiters to turn it into "data".

jcooky:
@apascal07 Thank you for your insight. I've added a comment about you mentioned.

Can you review it?

apascal07 (approved):

PR #3171, opened 2025-07-08
feat(js/ai): added a way to define resources dynamically

pavelgj:
Similar to `dynamicTool`

```ts
const myResource = dynamicResource(
  {
    uri: 'foo://foo',
    description: 'description',
  },
  async () => ({ content: [...] })
);

  prompt: [
    { text: 'analyze this: ' },
    { resource: { uri: 'foo://foo' } },
  ],
  resources: [myResource],
});

```

Checklist (if applicable):

### DIFF
--- a/js/ai/src/resource.ts
+++ b/js/ai/src/resource.ts
@@ -141,11 +141,13 @@ export async function findMatchingResource(
 }
 
 /** Checks whether provided object is a dynamic resource. */
-export function isDynamicResource(t: unknown): t is DynamicResourceAction {
+export function isDynamicResourceAction(
+  t: unknown
+): t is DynamicResourceAction {
   return (
     (isDetachedAction(t) || isAction(t)) &&
     t.__action.metadata?.type === 'resource' &&
-    t.__action.metadata?.dynamic
+    !!t.__action.metadata?.dynamic
   );
 }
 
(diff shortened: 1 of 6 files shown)


js/ai/src/resource.ts
```diff
@@ -96,24 +99,74 @@ export function defineResource(
...
+      if (resource.matches(input)) {
+        return resource;
+      }
+    }
+  }
+  return undefined;
+}
+
+/** Checks whether provided object is a dynamic resource. */
+export function isDynamicResource(t: unknown): t is DynamicResourceAction {
```
ifielker: The function name and the type name should match in a type guard, so either `function isDynamicResourceAction ...` or rename the type to DynamicResource ?

js/ai/tests/resource/resource_test.ts
ifielker, line 1: Shouldn't there be new tests?

js/ai/src/resource.ts
pavelgj: renamed the function

js/ai/tests/resource/resource_test.ts
pavelgj: the tests are in `js/genkit/tests/generate_test.ts` -- it's testing via the actual user-facing API.

PR #3172, opened 2025-07-08
feat(js/plugins/google-genai): Added GoogleAI index 

ifielker:
also fixed embedder and clients

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/googleai/index.ts
+++ b/js/plugins/google-genai/src/googleai/index.ts
@@ -1,0 +1,232 @@
+/**
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -0,0 +1,232 @@
...
+      if (listActionsCache) return listActionsCache;
+      listActionsCache = await listActions(options);
+      return listActionsCache;
+    }
+  );
+}
+
+export type GoogleAIPlugin = {
+  (params?: PluginOptions): GenkitPlugin;
+  model(
```
pavelgj, line 191: you are probably aware, but still calling it out that these will need overloads for gemma, imagen, veo models.

pavelgj (approved):

js/plugins/google-genai/src/googleai/index.ts
ifielker: Yes, the next PR completely refactors how that works because it's currently broken (not just in the new plugin). But I found a good solution that has both correct config typing and also shows all the known model names in autocomplete. It took like 6 hours to figure out, but it's beautiful now.. or rather will be in the next PR :)

PR #3186, opened 2025-07-09
fix(js): Add OperationData type in tools common for use in dev UI

shrutip90:
Checklist (if applicable):

### DIFF
--- a/js/core/src/background-action.ts
+++ b/js/core/src/background-action.ts
@@ -35,11 +35,6 @@ export const OperationSchema = z.object({
 });
 
 /**
- * Operation data.
- */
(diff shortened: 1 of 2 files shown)


js/core/src/background-action.ts
```diff
@@ -34,11 +34,6 @@ export const OperationSchema = z.object({
   metadata: z.record(z.string(), z.any()).optional(),
 });
 
-/**
- * Operation data.
```
pavelgj, line 38: why remove this?

shrutip90: i had added it yesterday for the same purpose but in the wrong place :/

pavelgj (approved):

PR #3193, opened 2025-07-10
feat(js/plugins/google-genai): Added Veo models

ifielker:
Checklist (if applicable):


js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -54,9 +57,13 @@ async function resolver(
 ) {
   switch (actionType) {
     case 'model':
-      if (imagen.isImagenModelName(actionName)) {
+    case 'background-model':
```
ifielker: I know we sometimes lookup `model/veo-`, but that should be fine since veo will define itself as a background-model and then the next lookup for 'background-model/veo' will succeed without resolution.

pavelgj: there's one problem with this. Imagine this order of model lookup from the registry:

 - lookup `model/veo-foo`, the model does not exist, resolver is called and we create `background-model/veo-foo`
 - next time, lookup `model/veo-foo`, the model does not exist, resolver is called and we create `background-model/veo-foo` (again)
 - next time, lookup `model/veo-foo`, the model does not exist, resolver is called and we create `background-model/veo-foo` (again)
 - etc....

that's why I had to do this very specific matching:
https://github.com/firebase/genkit/blob/9a5ca3a4c93cf1a20ff3057adbda4849c7e54ec7/js/plugins/googleai/src/index.ts#L182

ifielker: Good point. Thanks.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -298,14 +298,22 @@ const GENERIC_GEMMA_MODEL = commonRef(
   GemmaConfigSchema
 );
 
-const KNOWN_MODELS = {
+const KNOWN_GEMINI_MODELS = {
   'gemini-2.0-flash': commonRef('gemini-2.0-flash'),
```
pavelgj, line 302:
```suggestion
  'gemini-2.0-flash': commonRef('gemini-2.0-flash'),
  'gemini-2.0-pro': commonRef('gemini-2.0-pro'),
```

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -298,14 +298,22 @@ const GENERIC_GEMMA_MODEL = commonRef(
...
   'gemini-2.0-flash-lite': commonRef('gemini-2.0-flash-lite'),
   'gemini-2.0-pro-exp-02-05': commonRef('gemini-2.0-pro-exp-02-05'),
   'gemini-2.0-flash-exp': commonRef('gemini-2.0-flash-exp'),
   'gemini-2.5-pro-exp-03-25': commonRef('gemini-2.5-pro-exp-03-25'),
   'gemini-2.5-pro-preview-03-25': commonRef('gemini-2.5-pro-preview-03-25'),
   'gemini-2.5-flash-preview-04-17': commonRef('gemini-2.5-flash-preview-04-17'),
```
pavelgj:
```suggestion
  'gemini-2.5-flash-preview-04-17': commonRef('gemini-2.5-flash-preview-04-17'),
  'gemini-2.5-flash': commonRef('gemini-2.5-flash'),
  'gemini-2.5-pro': commonRef('gemini-2.5-pro'),
```

pavelgj: could you please update these lists to match https://ai.google.dev/gemini-api/docs/models#model-variations
all the old models can be summarily removed

ifielker: done.

PR #3211, opened 2025-07-11
fix(js/plugins/compat-oai): Bug bash fixes

ssbushi:
Fix warning: `WARNING: /model/openai/o1 already has an entry in the registry. Overwriting.`
Fix: Filter old models

Checklist (if applicable):
- [x] Tested (manually)


js/plugins/compat-oai/src/openai/gpt.ts
```diff
@@ -15,286 +15,143 @@
...
-      multiturn: true,
-      tools: true,
-      media: false,
-      systemRole: true,
-      output: ['text'],
-    },
-  },
-  configSchema: ChatCompletionCommonConfigSchema,
+const gpt4oMini = commonRef('openai/gpt-4o-mini', {
+  versions: ['gpt-4o-mini', 'gpt-4o-mini-2024-07-18'],
```
pavelgj: I think we're trying to make version obsolete. Please define these as separate models

js/plugins/compat-oai/src/openai/gpt.ts
```diff
@@ -15,286 +15,143 @@
...
-      multiturn: true,
-      tools: true,
-      media: true,
-      systemRole: true,
-      output: ['text', 'json'],
-    },
-  },
-  configSchema: ChatCompletionCommonConfigSchema,
+const gpt4Turbo = commonRef('openai/gpt-4-turbo', {
+  versions: [
```
pavelgj: Same here.... And below

js/plugins/compat-oai/src/openai/gpt.ts
```diff
@@ -90,68 +88,95 @@ const o4Mini = commonRef('openai/o4-mini', {
...
+  'openai/gpt-3.5-turbo-0125',
+  GPT_35_MODEL_INFO
+);
+const gpt35Turbo1106 = commonRef(
+  'openai/gpt-3.5-turbo-1106',
+  GPT_35_MODEL_INFO
+);
+
+const ALL_GPT_MODELS = [
+  gpt45,
```
pavelgj, line 153: nit: you can inline all these

ssbushi: it looks very messy if inline. I'd rather leave the definitions alone and keep this array neat.

js/plugins/compat-oai/README.md
```diff
@@ -142,7 +142,6 @@ import { ModelInfo } from 'genkit/model';
 import openAI from '@genkit-ai/compat-oai';
 
 const modelInfo: ModelInfo = {
```
pavelgj, line 144: this is optional, right? if not specified a generic (everything supported) model info would be used, right?

ssbushi: it is optional but I don't set a default explicitly in this plugin. I thought the framework had a default by itself.... I might be wrong.

ssbushi: I'll address this in a follow-up.

PR #3223, opened 2025-07-14
Fix JSON Schema parsing when object has no properties

gspencergoog:
## Description

This fixes an issue in parsing JSON schemas where `googleai` was forcing the schema object to have `properties`. It's legal for a JSON schema to not provide any properties (e.g. `{"type": "object"}` is a legal schema), so it should be robust to that input.

## Tests
 - Added a test for this case.


gspencergoog:
Ahh! Looks like someone else beat me to it.

PR #3239, opened 2025-07-17
fix(js,py): Export ResourcePartSchema through genkit-tools

shrutip90:
Checklist (if applicable):
- [x] Tested (manually)


genkit-tools/genkit-schema.json
pavelgj, line 1: need to update python as well

shrutip90: Done

PR #3250, opened 2025-07-19
refactor(js): introduced @genkit-ai/shared package that will contain shared type, plugin and action framework

pavelgj:
@genkit-ai/shared  is going to be runtime agnostic. It may largely replace @genkit-ai/core in the future except @genkit-ai/core may remain the node specific isntrumentation.

Checklist (if applicable):


MichaelDoyle:
@pavelgj you still want this? or close/revisit in another pass?

PR #3262, opened 2025-07-21
feat(go/genkit): added background action and model support

sahdev77:
 - Addition of Background model support in go.
 - Addition of veo model to validate the background model
 - Addition of sample for veo.

Checklist (if applicable):

### DIFF
--- a/py/plugins/vertex-ai/tests/model_garden/test_model_garden.py
+++ b/py/plugins/vertex-ai/tests/model_garden/test_model_garden.py
@@ -65,6 +65,7 @@ def model_garden_instance(client):
                     'constrained': None,
                     'content_type': None,
                     'context': None,
+                    'long_running': None,
                     'multiturn': True,
                     'media': True,
                     'tools': True,


go/ai/generate.go
```diff
@@ -1053,3 +1067,155 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+		},
+	)
+
+	return bgAction
+}
+
+// GenerateOperation generates a model response as a long-running operation based on the provided options.
+// It returns an error if the model does not support long-running operations.
+//
+// This is a beta feature and requires the model to explicitly support long-running operations.
```
apascal07: We should return an error if the model does not support LRO.

go/ai/generate.go
```diff
@@ -1053,3 +1067,155 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+
+	var modelName string
+	if genOpts.Model != nil {
+		modelName = genOpts.Model.Name()
+	} else {
+		modelName = genOpts.ModelName
+	}
+
+	provider, name, _ := strings.Cut(modelName, "/")
+	backAction := LookupBackgroundModel(r, provider, name)
```
apascal07:
```suggestion
	bgAction := LookupBackgroundModel(r, provider, name)
```

go/ai/generate.go
```diff
@@ -1053,3 +1067,155 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+		genOpts.Config = modelRef.Config()
+	}
+
+	startOps, err := backAction.Start(ctx, &ModelRequest{Messages: messages, Config: genOpts.Config})
+	if err != nil {
+		return nil, err
+	}
+
+	currentOp := startOps
+	for {
```
apascal07: We don't want to wait for the operation to be done here, we want to return it right away and have the caller poll on it, otherwise there's no point to having LRO.

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/firebase/genkit/go/internal/registry"
+	"github.com/invopop/jsonschema"
+)
+
+// Operation represents a background task operation
+type Operation struct {
```
apascal07: This should take a type parameter that's used for the type of `Output`.

go/core/action.go
```diff
@@ -66,6 +66,9 @@ const (
 	ActionTypeTool             ActionType = "tool"
 	ActionTypeUtil             ActionType = "util"
 	ActionTypeCustom           ActionType = "custom"
+	ActionTypeBackgroundModel  ActionType = "background-model"
+	ActionTypeBackgroundCheck  ActionType = "check-operation"
+	ActionTypeBackgroundCancel ActionType = "cancel-operation"
```
apascal07:
```suggestion
	ActionTypeCheckOperation  ActionType = "check-operation"
	ActionTypeCancelOperation ActionType = "cancel-operation"
```

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+	Output   any            `json:"output,omitempty"`   // Result when done
+	Error    string         `json:"error,omitempty"`    // Error if failed
+	Metadata map[string]any `json:"metadata,omitempty"` // Additional info
+}
+
+// BackgroundModelOptions holds configuration for defining a background model
+type BackgroundModelOptions struct {
+	Versions       []string           // Known version names for this model
+	Supports       any                // Capabilities this model supports
+	ConfigSchema   *jsonschema.Schema // Custom options schema for this model
```
apascal07:
```suggestion
	ConfigSchema   map[string]any // Custom options schema for this model
```

We should use `map[string]any` instead of the jsonschema struct, I'm changing it elsewhere to match.

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+	Label          string             // Descriptive name for this model
+	SupportsCancel bool               // Whether the model supports cancellation
+}
+
+// BackgroundStartFunc starts a background operation
+type BackgroundStartFunc[In, Out any] func(ctx context.Context, input In) (*Operation, error)
+
+// BackgroundCheckFunc checks the status of a background operation
+type BackgroundCheckFunc[Out any] func(ctx context.Context, operation *Operation) (*Operation, error)
```
apascal07:
```suggestion
type CheckOperationFunc[Out any] func(ctx context.Context, operation *Operation) (*Operation, error)
```

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+
+// BackgroundCancelFunc cancels a background operation
+type BackgroundCancelFunc[Out any] func(ctx context.Context, operation *Operation) (*Operation, error)
```
apascal07:
```suggestion
type CancelOperationFunc[Out any] func(ctx context.Context, operation *Operation) (*Operation, error)
```

go/ai/generate.go
```diff
@@ -1053,3 +1067,155 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+		}
+
+		messages = append(messages, NewUserTextMessage(prompt))
+	}
+
+	if modelRef, ok := genOpts.Model.(ModelRef); ok && genOpts.Config == nil {
+		genOpts.Config = modelRef.Config()
+	}
+
+	startOps, err := backAction.Start(ctx, &ModelRequest{Messages: messages, Config: genOpts.Config})
```
apascal07:
```suggestion
	op, err := backAction.Start(ctx, &ModelRequest{Messages: messages, Config: genOpts.Config})
```

go/ai/generate.go
sahdev77: Done

go/core/action.go
sahdev77: Done

go/core/background_action.go
sahdev77: Done

go/core/background_action.go
sahdev77: Done

go/core/background_action.go
sahdev77: Done

go/ai/generate.go
sahdev77: Done

go/ai/generate.go
sahdev77: This is fixed and also the sample is fixed accordingly

go/core/background_action.go
sahdev77: Can you provide more detail on it and i haven't seen this in ts implementation .

go/ai/generate.go
sahdev77: add a function to check support for LRO as LookupBackgroundModel contain complete BackgroundAction.

go/ai/generate.go
```diff
@@ -1058,3 +1072,177 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+func GenerateOperation(ctx context.Context, r *registry.Registry, opts ...GenerateOption) (*core.Operation, error) {
+	genOpts := &generateOptions{}
+	for _, opt := range opts {
+		if err := opt.applyGenerate(genOpts); err != nil {
+			return nil, core.NewError(core.INVALID_ARGUMENT, "ai.Generate: error applying options: %v", err)
+		}
+	}
+
+	if !SupportsLongRunning(r, genOpts.ModelName) {
+		return nil, core.NewError(core.UNIMPLEMENTED, "model %q does not support long-running operations", genOpts.ModelName)
```
apascal07:
```suggestion
		return nil, core.NewError(core.INVALID_ARGUMENT, "model %q does not support long-running operations", genOpts.ModelName)
```

UNIMPLEMENTED means something else.

go/core/background_action.go
apascal07: Right now you have type parameters in `CheckOperationFunc` and `CancelOperationFunc` but they're not used. It should be:

```go
type Operation[Out any] struct {
    ...
    Output Out
    ...
}
```

go/core/background_action.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/firebase/genkit/go/internal/registry"
+)
+
+// Operation represents a background task operation
+type Operation struct {
+	Action   string         `json:"action,omitempty"`   // The action that created this operation
+	ID       string         `json:"id"`                 // Unique identifier for tracking
+	Done     bool           `json:"done,omitempty"`     // Whether the operation is complete
+	Output   any            `json:"output,omitempty"`   // Result when done
+	Error    string         `json:"error,omitempty"`    // Error if failed
```
apascal07:
```suggestion
	Error    error         `json:"error,omitempty"`    // Error if failed
```

go/core/background_action.go
apascal07: Actually I see that we have `Operation` in `genkit-tools` schema, can we just generate the schema from there and put it in `gen.go`? Then you can keep it as `any` since we likely need it to be wire compatible.

go/core/background_action.go
apascal07: Looks unchanged?

go/core/background_action.go
apascal07: Same here, did you push your changes?

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+	Metadata map[string]any `json:"metadata,omitempty"` // Additional info
+}
+
+// BackgroundModelOptions holds configuration for defining a background model
+type BackgroundModelOptions struct {
+	Versions       []string           // Known version names for this model
+	Supports       any                // Capabilities this model supports
```
apascal07:
```suggestion
	Supports       ModelSupports                // Capabilities this model supports
```

Why not this?

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+type BackgroundAction[In, Out any] interface {
+	Start(ctx context.Context, input In) (*Operation, error)
+	Check(ctx context.Context, operation *Operation) (*Operation, error)
+	Cancel(ctx context.Context, operation *Operation) (*Operation, error)
+	SupportsCancel() bool
+	Name() string
+}
+
+// backgroundActionImpl implements BackgroundAction
+type backgroundActionImpl[In, Out any] struct {
```
apascal07:
```suggestion
type backgroundActionDef[In, Out any] struct {
```

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+type BackgroundModelOptions struct {
+	Versions       []string           // Known version names for this model
+	Supports       any                // Capabilities this model supports
+	ConfigSchema   *jsonschema.Schema // Custom options schema for this model
+	Label          string             // Descriptive name for this model
+	SupportsCancel bool               // Whether the model supports cancellation
+}
+
+// BackgroundStartFunc starts a background operation
+type BackgroundStartFunc[In, Out any] func(ctx context.Context, input In) (*Operation, error)
```
apascal07:
```suggestion
type BackgroundStartFunc[In, Out any] = func(ctx context.Context, input In) (*Operation, error)
```

Here and on the other ones.

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+type BackgroundCheckFunc[Out any] func(ctx context.Context, operation *Operation) (*Operation, error)
+
+// BackgroundCancelFunc cancels a background operation
+type BackgroundCancelFunc[Out any] func(ctx context.Context, operation *Operation) (*Operation, error)
+
+// BackgroundAction interface represents a background action
+type BackgroundAction[In, Out any] interface {
+	Start(ctx context.Context, input In) (*Operation, error)
+	Check(ctx context.Context, operation *Operation) (*Operation, error)
+	Cancel(ctx context.Context, operation *Operation) (*Operation, error)
```
apascal07:
```suggestion
	Check(ctx context.Context, op *Operation) (*Operation, error)
	Cancel(ctx context.Context, op *Operation) (*Operation, error)
```

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+	Name() string
+}
+
+// backgroundActionImpl implements BackgroundAction
+type backgroundActionImpl[In, Out any] struct {
+	startAction    *ActionDef[In, *Operation, struct{}]
+	checkAction    *ActionDef[*Operation, *Operation, struct{}]
+	cancelAction   *ActionDef[*Operation, *Operation, struct{}]
+	name           string
+	supportsCancel bool
```
apascal07: Do we even need this? Can't we just check for the presence of `cancelAction` like we're already doing anyway?

go/core/background_action.go
sahdev77: Done

go/core/background_action.go
sahdev77: Done for all 3

go/core/background_action.go
sahdev77: Done

go/core/background_action.go
sahdev77: As it is resulting in cyclic import with ai package otherwise we need to declare it locally in file itself which is not correct so keep any.
Please do suggest what we can do in this.

go/core/background_action.go
sahdev77: done

go/ai/generate.go
sahdev77: done

go/core/background_action.go
sahdev77: Are you talking about removal of supportsCancel ?

go/core/background_action.go
sahdev77: chanegs done

go/core/background_action.go
apascal07: `BackgroundModelOptions` should be in `ai`. Only `BackgroundAction` and related is "core".

go/core/background_action.go
apascal07: Yes, just the field, since the boolean can be derived from the presence of `cancelAction`. Unless I'm missing something?

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+
+// DefineBackgroundAction creates and registers a background action with three component actions
+func DefineBackgroundAction[In, Out any](
+	r *registry.Registry,
+	provider, name string,
+	config BackgroundModelOptions,
+	metadata map[string]any,
+	start BackgroundStartFunc[In, Out],
+	check BackgroundCheckFunc[Out], // Function to check operation status
+	cancel BackgroundCancelFunc[Out], // Optional function to cancel operation
```
apascal07: Move all of these to `BackgroundModelOptions`. The arg list should be `provider`, `name`, `opts`.

go/core/background_action.go
```diff
@@ -0,0 +1,188 @@
...
+) BackgroundAction[In, Out] {
+	// Create the start action - initiates the background operation
```
apascal07: Panic with a descriptive message if start or check actions are not defined.

go/core/background_action.go
apascal07: I see the issue, you're mixing background action and background model. Background model is just one type of background action (and belongs in `ai`). Background action is a core framework concept and belongs in `core`. We should clean up the types/naming so that is clear. There shouldn't be any circular dependencies. `DefineBackgroundModel` should be in `ai` and have a `BackgroundModelOptions` arg.

go/ai/generate.go
```diff
@@ -92,6 +93,12 @@ type (
 		toolResponse *Part
 		interrupt    *Part
 	}
+
+	StartModelFunc  = core.StartOperationFunc[*ModelRequest, *ModelResponse] // Function to start a model background operation
+	CheckModelFunc  = core.CheckOperationFunc[*ModelResponse]                // Function to check model operation status
+	CancelModelFunc = core.CancelOperationFunc[*ModelResponse]               // Function to cancel model operation
+
+	BackgroundAction = core.BackgroundAction[*ModelRequest, *ModelResponse] // Background action for model operations
```
apascal07:
```suggestion
	BackgroundModel = core.BackgroundAction[*ModelRequest, *ModelResponse] // Background action for model operations
```

go/ai/gen.go
```diff
@@ -397,3 +404,24 @@ type TraceMetadata struct {
...
+type StartOperationFunc[In, Out any] = func(ctx context.Context, input In) (*core.Operation[Out], error)
+
+// CheckOperationFunc checks the status of a background operation
+type CheckOperationFunc[Out any] = func(ctx context.Context, operation *core.Operation[Out]) (*core.Operation[Out], error)
+
+// CancelOperationFunc cancels a background operation
+type CancelOperationFunc[Out any] = func(ctx context.Context, operation *core.Operation[Out]) (*core.Operation[Out], error)
+
+// BackgroundModelOptions holds configuration for defining a background model
+type BackgroundModelOptions[In, Out any] struct {
```
apascal07: This actually doesn't need to be parameterized, it's always going to be `ModelRequest` for `In` and `ModelResponse` for `Out`.

Also, please move this to `generate.go`, this file is generated from the schema.

go/ai/generate.go
```diff
@@ -92,6 +92,12 @@ type (
 		toolResponse *Part
 		interrupt    *Part
 	}
+
+	//StartModelFunc  = core.StartOperationFunc[*ModelRequest, *ModelResponse] // Function to start a model background operation
+	//CheckModelFunc  = core.CheckOperationFunc[*ModelResponse]                // Function to check model operation status
+	//CancelModelFunc = core.CancelOperationFunc[*ModelResponse]               // Function to cancel model operation
+
```
apascal07:
```suggestion
```

go/ai/generate.go
```diff
@@ -1058,3 +1071,146 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+	metadata := map[string]any{
+		"model": map[string]any{
+			"label":    label,
+			"versions": opts.Versions,
+			"supports": opts.Supports,
+		},
+	}
+
+	if opts.ConfigSchema != nil {
+		metadata["model"].(map[string]any)["customOptions"] = opts.ConfigSchema
```
apascal07: This can panic, do a check on whether it cast correctly:

`if meta, ok := metadata["model"].(map[string]any); ok {`

go/ai/generate.go
```diff
@@ -1058,3 +1071,146 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+		}
+
+		messages = append(messages, NewUserTextMessage(prompt))
+	}
+
+	if modelRef, ok := genOpts.Model.(ModelRef); ok && genOpts.Config == nil {
+		genOpts.Config = modelRef.Config()
+	}
+
+	op, err := bgAction.Start(ctx, &ModelRequest{Messages: messages, Config: genOpts.Config})
```
apascal07: This is incomplete. We need to be calling `Generate()` from here with the full option list that we were given so that we can share the business logic in there. We need to regen `gen.go` that has the `Operation` field added to it (https://github.com/firebase/genkit/blob/96591062e16daef1b7fdb93a23eba486c0c26b3f/genkit-tools/genkit-schema.json#L1122) and then we will be able to take it out of the `ModelResponse` from `Generate()` to return here as a typed `Operation[ModelResponse]`.

go/ai/gen.go
sahdev77: Done

go/ai/generate.go
sahdev77: changed similar to DefineModel

go/ai/gen.go
```diff
@@ -123,6 +123,7 @@ const (
 type GenerateActionOptions struct {
 	Config             any                         `json:"config,omitempty"`
 	Docs               []*Document                 `json:"docs,omitempty"`
+	LongRunning        bool                        `json:"longRunning,omitempty"`
```
apascal07: We can't add this here, it's generated from JS and this interface needs to remain clean.

sahdev77: ok so in order to identify the longrunning operation in GenerateWithRequest should is check with support long running model function

apascal07: Yes I think that would be the simplest way to do it for now.

apascal07: Alternatively, just look up model, and if it's nil, try to look up background model (or vice versa). That's how JS does it.

sahdev77: That might be also fragile i had checked the code with SupportsLongRunning I think we need to think through to have a proper identifier for the same.

go/ai/generate.go
sahdev77: Code changes done please review

go/ai/generate.go
```diff
@@ -106,6 +106,29 @@ type ModelOptions struct {
...
+type BackgroundModelOptions struct {
+	Versions     []string       `json:"versions,omitempty"`
+	Supports     *ModelSupports `json:"supports,omitempty"`
+	ConfigSchema map[string]any `json:"configSchema,omitempty"`
+	Metadata     map[string]any `json:"metadata,omitempty"`
+	Label        string         `json:"label,omitempty"`
+	Start        StartOperationFunc[*ModelRequest, *ModelResponse]
+	Check        CheckOperationFunc[*ModelResponse]
+	Cancel       CancelOperationFunc[*ModelResponse]
+}
```
apascal07:
```suggestion
type BackgroundModelOptions struct {
	ModelOptions
	Start        StartOperationFunc[*ModelRequest, *ModelResponse]
	Check        CheckOperationFunc[*ModelResponse]
	Cancel       CancelOperationFunc[*ModelResponse]
}
```

go/core/background_action.go
```diff
@@ -0,0 +1,160 @@
...
+	return b.cancelAction.Run(ctx, op, nil)
+}
+
+// Name returns the action name
+func (b *backgroundActionDef[In, Out]) Name() string {
+	return b.name
+}
+
+// DefineBackgroundAction creates and registers a background action with three component actions
+func DefineBackgroundAction[In, Out any](
```
apascal07, line 88: We also need `NewBackgroundAction` which calls `NewAction` instead of `DefineAction` and in `generate.go` we need `NewBackgroundModel`. Check the other primitives for reference.

go/core/background_action.go
```diff
@@ -0,0 +1,160 @@
...
+	Metadata map[string]any `json:"metadata,omitempty"` // Additional info
+}
+
+// BackgroundAction interface represents a background action
+type BackgroundAction[In, Out any] interface {
+	Start(ctx context.Context, input In) (*Operation[Out], error)
+	Check(ctx context.Context, op *Operation[Out]) (*Operation[Out], error)
+	Cancel(ctx context.Context, op *Operation[Out]) (*Operation[Out], error)
+	Name() string
+}
```
apascal07: I don't think we need this interface. We should export `backgroundActionDef` and return it from the `DefineBackgroundAction` and `NewBackgroundAction` functions.

go/ai/generate.go
```diff
@@ -106,6 +106,29 @@ type ModelOptions struct {
 	Versions []string `json:"versions,omitempty"`
 }
 
+type BackgroundModel = core.BackgroundAction[*ModelRequest, *ModelResponse] // Background action for model operations
```
apascal07:
```suggestion
type backgroundModel struct {
    core.BackgroundActionDef[*ModelRequest, *ModelResponse]
}
```

Based on the other comment below.

Then we'll want to have an interface called `BackgroundModel` that has `Start()`, `Check()`, `Cancel()`, and `Name()`. See the model types/interfaces in the main branch.

go/ai/generate.go
```diff
@@ -111,6 +112,29 @@ type ModelOptions struct {
 	Versions []string `json:"versions,omitempty"`
 }
 
+// BackgroundModel is a background action for model operations.
+type BackgroundModel struct {
```
apascal07:
```suggestion
type backgroundModel struct {
```

Then we should define an interface called `BackgroundModel` that has the functions specific to this, like `Model`, `Embedder`, etc and return that from the definition function.

go/ai/generate.go
```diff
@@ -1257,3 +1342,152 @@ func executeResourcePart(ctx context.Context, r api.Registry, resourceURI string
...
+		"label":    label,
+		"versions": opts.Versions,
+		"supports": opts.Supports,
+	}
+	if opts.ConfigSchema != nil {
+		mergedMetadata["customOptions"] = opts.ConfigSchema
+		if modelMeta, ok := mergedMetadata["model"].(map[string]any); ok {
+			modelMeta["customOptions"] = opts.ConfigSchema
+		}
+	}
```
apascal07: This should all be done in `NewBackgroundModel` and this function should just register the action. Check `NewModel` and `DefineModel`.

go/ai/generate.go
```diff
@@ -215,10 +250,12 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
 	}
 
 	m := LookupModel(r, opts.Model)
+
 	if m == nil {
 		return nil, core.NewError(core.NOT_FOUND, "ai.GenerateWithRequest: model %q not found", opts.Model)
 	}
 
+	isLongRunning := SupportsLongRunning(r, opts.Model)
```
apascal07: Do we need this function? Can't we just try to look up the model and then try background model? It's a long and brittle function and we are able to get that information in other ways.

go/ai/generate.go
sahdev77: This is done and i had separated the code for background model in background_model.go to simplify the things

go/ai/generate.go
sahdev77: This is done as per your comments

go/ai/generate.go
sahdev77: Although i had removed the SupportsLongRunning but checking for each model whether it is background model or not now the question arises why this was coming in normal model as if i touch that peice to have veo not to be the part of normal model list then so many code piece need  to be touched upon for that.
Need your opinion on that.

go/ai/generate.go
apascal07: Wait, why is this redefined again separately from `background_model.go`? How does even this build?

sahdev77: I found the issue it arises because of some merge conflict while merging with main the original code was in commit (dbaeaa4852) but now fixed. Test are still failing because of genkit-schema.json which you can have a look i am not sure how to fix it.

apascal07:
There is a test failing due to generated code not matching, can you take a look?

apascal07:
@sahdev77 We're still quite a bit ways off where we want to be in this PR so I'm going to take over the last set of changes.

sahdev77:
Ok sure

sahdev77:
@apascal07 Few code validation added and also few minor issues found during the sample app those are fixed also added new video models veo3 and veo3 fast.
The only issue left is that build issue of schema please have a look. From long operation and veo perspective all is fine now.

PR #3264, opened 2025-07-22
feat(go): update telemetry for compatibility with Firebase AIM

huangjeff5:
Metrics
unnamed

(Will add thought metrics as a fast follow; they were added recently and don't want to keep exploding this PR)

### Basic success flow

Screenshot 2025-08-19 at 10 40 28 AM

Screenshot 2025-08-19 at 10 40 18 AM

### Error flow
Screenshot 2025-08-19 at 10 46 23 AM

Screenshot 2025-08-19 at 10 45 52 AM

### Substep flow

Screenshot 2025-08-19 at 10 39 37 AM

Screenshot 2025-08-19 at 10 39 33 AM

### Tool flow

Screenshot 2025-08-19 at 10 41 49 AM

Screenshot 2025-08-19 at 10 41 44 AM

(The Tool View looks slightly different because our tool loop inside the Generate() call is implemented slightly differently, that requires some more refactoring if we want the views to look exactly the same)


go/core/tracing/tracing.go
```diff
@@ -80,55 +116,147 @@ const (
...
-		Name:   name,
-		Input:  input,
-		IsRoot: isRoot,
+		Name:     metadata.Name,
+		Input:    input,
+		IsRoot:   metadata.IsRoot,
+		Type:     metadata.Type,
+		Subtype:  metadata.Subtype,
+		Metadata: metadata.Metadata,
+		mu:       sync.Mutex{},
```
jeffdh5, line 164: double check if lock is needed

go/plugins/googlecloud/utils.go
```diff
@@ -0,0 +1,271 @@
...
+					return first[1:end]
+				}
+			}
+		}
+	}
+
+	return "<unknown>"
+}
+
+// Functional option helpers for common configurations
```
jeffdh5, line 164: Remove htese helpers

go/ai/generate.go
```diff
@@ -1053,3 +1060,208 @@ func handleResumeOption(ctx context.Context, r *registry.Registry, genOpts *Gene
...
+}
+
+// calculateOutputCharacters counts the total characters in the output response.
+func calculateOutputCharacters(resp *ModelResponse) int {
+	if resp == nil || resp.Message == nil {
+		return 0
+	}
+
+	totalChars := 0
+	for _, part := range resp.Message.Content {
```
jeffdh5, line 1156: factor in all parts?

go/plugins/googlecloud/slog_handler.go
```diff
@@ -20,62 +20,225 @@ package googlecloud
...
 func (h *handler) recordToEntry(ctx context.Context, r slog.Record) logging.Entry {
-	return logging.Entry{
+	span := trace.SpanFromContext(ctx)
+
+	// Extract message and data from slog record
+	message := r.Message
+	var metadata map[string]interface{}
+
+	// Process record attributes to separate message from metadata
+	r.Attrs(func(a slog.Attr) bool {
```
huangjeff5, line 195: see if we can clean this part up

go/samples/telemetry-test/kitchen_sink/kitchen_sink.go
```diff
@@ -0,0 +1,165 @@
...
+			} else {
+				results[topic] = resp.Text()
+			}
+		}
+
+		return results, nil
+	})
+
+	// Start the server
+	fmt.Println("Kitchen Sink Telemetry Demo")
```
huangjeff5, line 138: simplify

go/internal/doc-snippets/telemetryplugin/telemetryplugin.go
```diff
@@ -50,40 +47,71 @@ type Config struct {
 
 // [END config]
 
-func Init(cfg Config) error {
-	ctx := context.Background()
-	g, err := genkit.Init(ctx)
-	if err != nil {
-		return err
+// [START enablecustom]
```
huangjeff5: unclear on exactly how START / END works and how it impacts docs

go/internal/doc-snippets/telemetryplugin/exporters.go
```diff
@@ -26,62 +26,131 @@ import (
 	"go.opentelemetry.io/otel/sdk/trace"
```
huangjeff5: not exaclty sure how these doc snippets work; need to make sure they display well

go/core/tracing/tracing.go
```diff
@@ -80,55 +206,176 @@ const (
...
+	// Metadata are genkit-specific metadata that get "genkit:metadata:" prefix
+	// (e.g., "subtype": "tool" becomes "genkit:metadata:subtype": "tool")
+	Metadata map[string]string
+}
+
+// RunInNewSpan runs f on input in a new span with the provided metadata.
+// The metadata contains all span configuration including name, type, labels, etc.
 func RunInNewSpan[I, O any](
 	ctx context.Context,
 	tstate *State,
```
apascal07: Do we still need to take this in? Right now what's checked in `main` is just a global getter for the tracer provider, so it could just be called within this function, right?

go/core/flow.go
```diff
@@ -40,8 +40,11 @@ type StreamingFlowValue[Out, Stream any] struct {
 // flowContextKey is a context key that indicates whether the current context is a flow context.
 var flowContextKey = base.NewContextKey[*flowContext]()
 
-// flowContext is a context that contains flow-specific information.
-type flowContext struct{}
+// flowContext is a context that contains the tracing state for a flow.
+type flowContext struct {
+	tracingState *tracing.State
```
apascal07: Do we need to store this? Isn't tracing state global anyway?

go/samples/telemetry-test/kitchen_sink/kitchen_sink.go
```diff
@@ -0,0 +1,165 @@
...
+func main() {
+	ctx := context.Background()
+
+	// Initialize Firebase telemetry
+	firebase.EnableFirebaseTelemetry(&firebase.FirebaseTelemetryOptions{
+		ForceDevExport: true, // Force telemetry export in development
+	})
+
+	// Initialize Genkit with plugins
+	g, err := genkit.Init(ctx, genkit.WithPlugins(
```
apascal07: Latest Genkit doesn't return error from here anymore, go.sum is probably outdated?

go/core/tracing/tracing.go
```diff
@@ -91,35 +201,103 @@ const (
...
+	Metadata map[string]string
+}
+
+// RunInNewSpan runs f on input in a new span with the provided metadata.
+// The metadata contains all span configuration including name, type, labels, etc.
 func RunInNewSpan[I, O any](
 	ctx context.Context,
-	name, spanType string,
-	isRoot bool,
+	state *State,
```
apascal07: Isn't this nil most places? Can we just always operate on the global `otel.GetTracerProvider()`?

go/plugins/firebase/telemetry.go
```diff
@@ -0,0 +1,117 @@
...
+//	g, err := genkit.Init(ctx, genkit.WithPlugins(&googlegenai.GoogleAI{}))
+//
+//	// With configuration (all modules still enabled)
+//	firebase.EnableFirebaseTelemetry(&firebase.FirebaseTelemetryOptions{
+//		ProjectID:            "my-firebase-project",
+//		ForceExport:          true,
+//		ExportInputAndOutput: true,
+//	})
+//	g, err := genkit.Init(ctx, genkit.WithPlugins(&googlegenai.GoogleAI{}))
+func EnableFirebaseTelemetry(options ...*FirebaseTelemetryOptions) {
```
apascal07: This is a weird signature if we're passing in an options struct, no? Usually it's reserved for options like we have in `option.go`. I think it's more confusing to have this and allow passing in multiple structs than to just require passing in nil for default.

go/genkit/reflection.go
```diff
@@ -436,20 +436,26 @@ func runAction(ctx context.Context, g *Genkit, key string, input json.RawMessage
...
+	if telemetryLabels != nil {
+		err := json.Unmarshal(telemetryLabels, &telemetryAttributes)
+		if err != nil {
+			return nil, core.NewError(core.INTERNAL, "Error unmarshalling telemetryLabels: %v", err)
 		}
+	}
+
+	var traceID string
+	spanMetadata := &tracing.SpanMetadata{
+		Name:            "dev-run-action-wrapper",
```
apascal07: Do we still want this? I thought we removed this in JS (but not 100% certain).

go/genkit/reflection.go
```diff
@@ -436,20 +436,26 @@ func runAction(ctx context.Context, g *Genkit, key string, input json.RawMessage
...
+		Type:            "action", // Wrapper action gets type "action"
```
apascal07: Nit: can we avoid comments like this and below? I don't think they add much. Outside of function/type/field definition, comments should focus on the why not the what.

go/plugins/firebase/telemetry.go
jeffdh5: Hmm.. This was the only way I could think of to allow EnableFirebaseTelemetry() OR EnableFirebaseTelemetry(options), which I thought was better than EnableFirebaseTelemetry(nil), but I'll defer to you.

How come you're concerned about passing in multiple structs? I could see maybe adding in logic to return an error in that case.

apascal07: It kinda feels like a weird pattern given that you can pass an argument that is either ignored (bad) or you have to panic at runtime (also bad given that this could easily be a compile time error).

huangjeff5: OK, I modified it to accept EnableFirebaseTelemetry(nil)

go/core/tracing/tracing.go
```diff
@@ -31,6 +36,99 @@ import (
...
+			break
+		}
+	}
+
+	return strings.Join(cleanLines, "\n")
+}
+
+// State holds OpenTelemetry values for creating traces.
+type State struct {
+	tracer trace.Tracer // cached tracer for creating spans
```
huangjeff5: check if we need this still

go/plugins/googlecloud/googlecloud.go
```diff
@@ -18,121 +18,547 @@
...
+//	googlecloud.EnableGoogleCloudTelemetry()
+//	g, err := genkit.Init(ctx, genkit.WithPlugins(&googlegenai.GoogleAI{}))
+//
+//	// With configuration
+//	googlecloud.EnableGoogleCloudTelemetry(&googlecloud.GoogleCloudTelemetryOptions{
+//		ProjectID:      "my-project",
+//		ForceDevExport: true,
+//	})
+//	g, err := genkit.Init(ctx, genkit.WithPlugins(&googlegenai.GoogleAI{}))
+func EnableGoogleCloudTelemetry(options ...*GoogleCloudTelemetryOptions) {
```
huangjeff5: convert to *GoogleCloudTelemetryOptions

go/plugins/googlecloud/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// NewMetricHistogram creates a new histogram metric with the given name and options
+func NewMetricHistogram(name string, opts MetricHistogramOptions) *MetricHistogram {
+	meter := otel.Meter("genkit")
+	histogram, err := meter.Float64Histogram(name,
+		metric.WithDescription(opts.Description),
+		metric.WithUnit(opts.Unit))
+	if err != nil {
+		panic(fmt.Sprintf("failed to create histogram %s: %v", name, err))
```
huangjeff5, line 78: is panic the right thing to do?

go/core/tracing/tracing.go
huangjeff5: can deprecate

go/plugins/googlecloud/metrics.go
apascal07: Seems fine yeah.

go/core/error.go
```diff
@@ -112,6 +113,11 @@ func ToReflectionError(err error) ReflectionError {
 		return ge.ToReflectionError()
 	}
 
+	var ge *GenkitError
```
apascal07, line 119: What is the difference here from above? Do we need both?

huangjeff5: We do, because error can be either a markedError or GenkitError type. (markedError wraps GenkitError and is used to indicate whether the error has already been attributed to a span).

Casting markedError directly fails because it is indeed a different type. errors.As() unwraps markedError and funds the GenkitError underneath.

PR #3269, opened 2025-07-23
feat: add instance name to GenkitOptions

MichaelDoyle:
Will be displayed in developer tooling

Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/manager/manager.ts
+++ b/genkit-tools/common/src/manager/manager.ts
@@ -444,6 +444,9 @@ export class RuntimeManager {
       );
 
       if (isValidRuntimeInfo(runtimeInfo)) {
+        if (!runtimeInfo.name) {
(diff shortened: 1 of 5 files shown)


genkit-tools/common/src/manager/manager.ts
```diff
@@ -569,6 +569,7 @@ function isValidRuntimeInfo(data: any): data is RuntimeInfo {
     typeof data === 'object' &&
     typeof data.id === 'string' &&
     typeof data.pid === 'number' &&
+    typeof data.name === 'string' &&
```
pavelgj: let's make sure it's optional, for backwards compatibility with older runtimes... and fall back to id it not set.

MichaelDoyle: done

pavelgj (approved):

PR #3279, opened 2025-07-23
fix(js/plugins/google-genai): Fixed imagen, veo, and streaming thoughts

ifielker:
Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/common/utils.ts
+++ b/js/plugins/google-genai/src/common/utils.ts
@@ -14,7 +14,13 @@
  * limitations under the License.
  */
 
-import { GenkitError, JSONSchema } from 'genkit';
+import {
+  EmbedderReference,
+  GenkitError,
+  JSONSchema,
+  ModelReference,
+  z,
+} from 'genkit';
 import { GenerateRequest } from 'genkit/model';
 import { ImagenInstance } from './types';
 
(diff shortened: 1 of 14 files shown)


js/plugins/google-genai/src/googleai/imagen.ts
```diff
@@ -89,6 +89,7 @@ function commonRef(
 ): ModelReference<ConfigSchemaType> {
   return modelRef({
     name: `googleai/${name}`,
+    version: name,
```
pavelgj: This should not be necessary. It should work without the version. Version is always optional, and in fact we're working towards making it obsolete. If set, use it as model name, if not it should use whatever is in the ref name.

pavelgj (changes requested):

js/plugins/google-genai/src/googleai/imagen.ts
ifielker: refactored

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -40,6 +46,19 @@ export function extractErrMsg(e: unknown): string {
   return errorMessage;
 }
 
+/**
+ * Gets the un-prefixed model name from a modelReference
+ */
+export function extractVersion(
+  model: ModelReference<z.ZodTypeAny> | EmbedderReference<z.ZodTypeAny>
+): string {
+  return model.name.startsWith('googleai/')
```
pavelgj: You can make this more generic: if there's a slash, return the part after the slash.

pavelgj: The logic should be:
if version is set, return the version
Else extract the version from name

ifielker: done

ifielker: done

pavelgj (approved):

PR #3339, opened 2025-07-29
fix: add thinking tokens to usage stats for gemini models

### LINKED ISSUE #3335: [JS] Add thought (reasoning) token "usage" counts to telemetry
Currently the Gemini API returns thought token "usage" counts, however, we're not capturing it and/or displaying it in the tooling. We need to show it, because it effects the price of API calls to Gemini, even in cases where thought tokens are not returned in the response.

Screenshot below from a trace shows that `usage` does not contain throught tokens, however the Gemini response `custom.usageMetadata` does.

Image

MichaelDoyle:
Fixes #3335

- Adds to genai plugin
- Backports to gemini & vertex plugins.

### DIFF
--- a/js/plugins/google-genai/src/googleai/gemini.ts
+++ b/js/plugins/google-genai/src/googleai/gemini.ts
@@ -621,6 +621,7 @@ export function defineModel(
             ...getBasicUsageStats(request.messages, candidateData),
             inputTokens: response.usageMetadata?.promptTokenCount,
             outputTokens: response.usageMetadata?.candidatesTokenCount,
+            thoughtsTokens: response.usageMetadata?.thoughtsTokenCount,
             totalTokens: response.usageMetadata?.totalTokenCount,
             cachedContentTokens:
               response.usageMetadata?.cachedContentTokenCount,
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -36,6 +36,7 @@ import {
   type StartChatParams,
   type Tool,
   type ToolConfig,
+  type UsageMetadata,
 } from '@google/generative-ai';
 import {
   GENKIT_CLIENT_HEADER,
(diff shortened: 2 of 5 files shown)


js/plugins/googleai/src/gemini.ts
```diff
@@ -67,6 +68,10 @@ import { getApiKeyFromEnvVar } from './common';
 import { handleCacheIfNeeded } from './context-caching';
 import { extractCacheConfig } from './context-caching/utils';
 
+type ExtendedUsageMetadata = UsageMetadata & {
```
ifielker, line 71: Why does this need to be a new type? What happens when we have another field, is it then ExtendedExtendedUsageMetadata? :)

js/plugins/vertexai/src/gemini.ts
```diff
@@ -63,11 +64,14 @@ import {
 } from 'genkit/model/middleware';
 import { runInNewSpan } from 'genkit/tracing';
 import { GoogleAuth } from 'google-auth-library';
-
 import type { PluginOptions } from './common/types.js';
 import { handleCacheIfNeeded } from './context-caching/index.js';
 import { extractCacheConfig } from './context-caching/utils.js';
 
+type ExtendedUsageMetadata = UsageMetadata & {
```
ifielker, line 71: Same here

ifielker (changes requested):

ifielker (approved):

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -621,6 +621,7 @@ export function defineModel(
             ...getBasicUsageStats(request.messages, candidateData),
             inputTokens: response.usageMetadata?.promptTokenCount,
             outputTokens: response.usageMetadata?.candidatesTokenCount,
+            reasoningTokens: response.usageMetadata?.thoughtsTokenCount,
```
ifielker: This needs to match GenerationUsageSchema
so it needs to be  thoughtsTokens: z.number().optional(),

ifielker (changes requested):

js/plugins/google-genai/src/googleai/gemini.ts
MichaelDoyle: Done -- thanks!

schnecle (approved):

PR #3342, opened 2025-07-29
feat(go): Add support for genkit resources

huangjeff5:


go/genkit/resource.go
```diff
@@ -0,0 +1,331 @@
...
+		action:  action,
+		matcher: d.matcher,
+	}
+
+	// Register as a resource
+	resourceKey := fmt.Sprintf("resource/%s", d.name)
+	r.RegisterValue(resourceKey, resourceAct)
+
+	// Return cleanup function
+	return func() {
```
huangjeff5: need to discard attached handlers

go/genkit/resource.go
```diff
@@ -0,0 +1,326 @@
...
+//	}, func(ctx context.Context, input core.ResourceInput) (ResourceOutput, error) {
+//	  content, err := os.ReadFile("/docs/handbook.pdf")
+//	  if err != nil {
+//	    return ResourceOutput{}, err
+//	  }
+//	  return ResourceOutput{
+//	    Content: []*ai.Part{ai.NewTextPart(string(content))},
+//	  }, nil
+//	})
+func DefineResource(g *Genkit, opts ResourceOptions, fn ResourceFunc) error {
```
huangjeff5: May need to update interface to accept resourceName as separate param

go/genkit/resource.go
```diff
@@ -0,0 +1,326 @@
...
+	metadata["description"] = opts.Description
+	metadata["resource"] = map[string]any{
+		"uri":      opts.URI,
+		"template": opts.Template,
+	}
+
+	// Define the action
+	action := core.DefineAction(
+		g.reg,
+		"", // no provider for resources
```
huangjeff5: double check this is correct

go/genkit/resource.go
```diff
@@ -0,0 +1,326 @@
...
+//
+//	// Use in generation:
+//	ai.Generate(ctx, g,
+//	  ai.WithPrompt([]*ai.Part{
+//	    ai.NewTextPart("Analyze this user:"),
+//	    ai.NewResourcePart("user://profile/123"),
+//	  }),
+//	  ai.WithResources([]core.DetachedResourceAction{dynamicRes}),
+//	)
+func DynamicResource(opts ResourceOptions, fn ResourceFunc) (core.DetachedResourceAction, error) {
```
huangjeff5: do dynamic resources have the same name as regular resource?

huangjeff5: what would a genkit resource name look like? (does it have a parent)

go/genkit/resource.go
```diff
@@ -0,0 +1,326 @@
...
+type ResourceOptions struct {
+	Name        string         // Required: unique name for the resource
+	URI         string         // Static URI (mutually exclusive with Template)
+	Template    string         // URI template (mutually exclusive with URI)
+	Description string         // Optional description
+	Metadata    map[string]any // Optional metadata
+}
+
+// resourceAction implements a resource as a core.Action.
+type resourceAction struct {
```
huangjeff5: move this and detachedResourceAction to some internal library maybe?

go/ai/document.go
```diff
@@ -209,12 +227,18 @@ func (p *Part) MarshalJSON() ([]byte, error) {
 	}
 }
 
+type resourcePart struct {
```
apascal07: Shouldn't this come from the generated types in `gen.go`?

go/genkit/resource.go
apascal07: Yes, let's do this please, based on our discussion and the changes I made to the other `Define*` methods. Also no error return, panic on any errors.

go/genkit/resource.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/firebase/genkit/go/internal/registry"
+)
+
+// ResourceOutput with concrete ai.Part type for this package
+type ResourceOutput struct {
+	Content []*ai.Part `json:"content"` // The content parts returned by the resource
+}
+
+// ResourceFunc is a function that loads content for a resource.
+type ResourceFunc func(context.Context, core.ResourceInput) (ResourceOutput, error)
```
apascal07:
```suggestion
type ResourceFunc = func(context.Context, core.ResourceInput) (ResourceOutput, error)
```

Aliases preferred if not adding methods on it.

go/genkit/resource.go
apascal07, line 1: Curious, any reason why you broke out some stuff into the `genkit` package and other in `core`? I would expect all of it to be in `core` since `genkit` is largely an abstraction to hide direct access to the registry.

go/genkit/resource.go
apascal07: That's fine, this is consistent with flows, tools, and others. This arg will go away with https://github.com/firebase/genkit/pull/3403

go/genkit/resource.go
apascal07: Dynamic resources should be like dynamic tools (and in the near future, all primitives which will be detached from a registry by default and only registered "intentionally" via `Define*` rather than `New*`).

apascal07:
```suggestion
func NewResource(opts ResourceOptions, fn ResourceFunc) core.DetachedResourceAction {
```

This function name is more idiomatic to Go. `Define*` and `New*` shouldn't return an error, make it panic on invalid configs.

go/genkit/resource.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// ExtractVariables extracts variables from a URI using this resource's template.
+func (r *resourceAction) ExtractVariables(uri string) (map[string]string, error) {
+	return r.matcher.ExtractVariables(uri)
+}
+
+// detachedResourceAction represents a resource action that isn't registered in any registry.
+// It can be temporarily attached during generation and implements core.DetachedResourceAction.
+type detachedResourceAction struct {
```
apascal07: Have you taken a look at how tools are implemented? We have both registered tools and "detached/dynamic" tools and they reused the same types. Tools can be created as standalone and later registered if needed.

This is the pattern we'll be moving to in general per discussions with Pavel.

jeffdh5: Fixed

go/genkit/resource.go
jeffdh5: Ah I understand better now. It wasn't super clear to me how everything was split but that makes sense. Fixed

go/ai/resource.go
```diff
@@ -0,0 +1,240 @@
...
+	metadata, wrappedFn, matcher := implementResource(name, opts, fn)
+	resourceAction := core.DefineAction(r, "", name, core.ActionTypeResource, metadata, wrappedFn)
+
+	res := &resourceImpl{
+		Action:  resourceAction,
+		matcher: matcher,
+	}
+
+	// Register as a resource value for URI lookup
+	r.RegisterValue(fmt.Sprintf("resource/%s", name), res)
```
huangjeff5, line 84: Use RegisterAction, not RegisterValue directly

go/genkit/resource.go
huangjeff5: Cool, done

go/ai/document.go
huangjeff5: Fixed

go/genkit/resource.go
huangjeff5: fixed

go/genkit/resource.go
huangjeff5: fixed

go/genkit/resource.go
huangjeff5: Fixed

go/genkit/resource.go
huangjeff5: Fixed this, I wasn't fully clear on what fit inside core vs. ai vs. genkit but it all makes sense now.

go/ai/option.go
```diff
@@ -725,8 +725,9 @@ type generateOptions struct {
 	outputOptions
 	executionOptions
 	documentOptions
-	RespondParts []*Part // Tool responses to return from interrupted tool calls.
-	RestartParts []*Part // Tool requests to restart interrupted tools with.
+	RespondParts     []*Part    // Tool responses to return from interrupted tool calls.
+	RestartParts     []*Part    // Tool requests to restart interrupted tools with.
+	DynamicResources []Resource // Dynamic resources to be temporarily available during generation.
```
huangjeff5: rename to resources

huangjeff5: put in commongenoptions

go/ai/resource.go
```diff
@@ -0,0 +1,254 @@
...
+	// ExtractVariables extracts variables from a URI using this resource's template.
+	ExtractVariables(uri string) (map[string]string, error)
+	// Execute runs the resource with the given input.
+	Execute(ctx context.Context, input ResourceInput) (ResourceOutput, error)
+	// Register sets the tracing state on the action and registers it with the registry.
+	Register(r *registry.Registry)
+}
+
+// DefineResource creates a resource and registers it with the given Registry.
+func DefineResource(r *registry.Registry, name string, opts ResourceOptions, fn ResourceFunc) Resource {
```
apascal07:
```suggestion
func DefineResource(r *registry.Registry, name string, opts *ResourceOptions, fn ResourceFunc) Resource {
```

Just to be consistent with all other `Define*` functions.

go/ai/resource.go
```diff
@@ -0,0 +1,254 @@
...
+
+	// Add user metadata
+	if opts.Metadata != nil {
+		for k, v := range opts.Metadata {
+			metadata[k] = v
+		}
+	}
+
+	// Wrapped function - just call the user function directly
+	wrappedFn := func(ctx context.Context, input ResourceInput) (ResourceOutput, error) {
```
apascal07: Can we just use `fn` directly? Doesn't look like we augment it in any way.

go/ai/resource.go
```diff
@@ -0,0 +1,254 @@
...
+			"template": opts.Template,
+		},
+	}
+
+	// Add user metadata
+	if opts.Metadata != nil {
+		for k, v := range opts.Metadata {
+			metadata[k] = v
+		}
+	}
```
apascal07: I think you can use `maps.Copy` here.

go/ai/resource.go
```diff
@@ -0,0 +1,254 @@
...
+// ResourceOptions configures a resource definition.
+type ResourceOptions struct {
+	URI         string         // Static URI (mutually exclusive with Template)
+	Template    string         // URI template (mutually exclusive with URI)
+	Description string         // Optional description
+	Metadata    map[string]any // Optional metadata
+}
+
+// ResourceFunc is a function that loads content for a resource.
+type ResourceFunc = func(context.Context, ResourceInput) (ResourceOutput, error)
```
apascal07:
```suggestion
type ResourceFunc = func(context.Context, *ResourceInput) (*ResourceOutput, error)
```

We generally do pointers, so to be consistent.

go/genkit/resource.go
```diff
@@ -0,0 +1,107 @@
...
+//	  content, err := os.ReadFile("/docs/handbook.pdf")
+//	  if err != nil {
+//	    return ai.ResourceOutput{}, err
+//	  }
+//	  return ai.ResourceOutput{
+//	    Content: []*ai.Part{ai.NewTextPart(string(content))},
+//	  }, nil
+//	})
+func DefineResource(g *Genkit, resourceName string, opts ai.ResourceOptions, fn ai.ResourceFunc) ai.Resource {
+	// Delegate to ai implementation
```
apascal07:
```suggestion
```

go/genkit/resource.go
```diff
@@ -0,0 +1,107 @@
...
+	return ai.DefineResource(g.reg, resourceName, opts, fn)
+}
+
+// FindMatchingResource finds a resource that matches the given URI.
+func FindMatchingResource(g *Genkit, uri string) (ai.Resource, ai.ResourceInput, error) {
+	// Delegate to ai implementation
```
apascal07:
```suggestion
```

go/genkit/resource.go
```diff
@@ -0,0 +1,107 @@
...
+	return ai.FindMatchingResource(g.reg, uri)
+}
+
+// ExecuteResource is a helper to execute an ai.Resource.
+func ExecuteResource(ctx context.Context, resource ai.Resource, input ai.ResourceInput) (ai.ResourceOutput, error) {
+	// Delegate to ai implementation
```
apascal07:
```suggestion
```

go/genkit/resource.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// FindMatchingResource finds a resource that matches the given URI.
+func FindMatchingResource(g *Genkit, uri string) (ai.Resource, ai.ResourceInput, error) {
+	// Delegate to ai implementation
+	return ai.FindMatchingResource(g.reg, uri)
+}
+
+// ExecuteResource is a helper to execute an ai.Resource.
+func ExecuteResource(ctx context.Context, resource ai.Resource, input ai.ResourceInput) (ai.ResourceOutput, error) {
```
apascal07: Is there a reason to have this function vs just calling `resource.Execute()`?

go/genkit/resource.go
apascal07, line 1: There's nothing wrong with having a separate file for these but for now we have everything in `genkit.go` so let's keep it there for now until/if we break out all the primitives, just so it's consistent to find things.

go/genkit/resource.go
```diff
@@ -0,0 +1,107 @@
...
+//	}, func(ctx context.Context, input ai.ResourceInput) (ai.ResourceOutput, error) {
+//	  content, err := os.ReadFile("/docs/handbook.pdf")
+//	  if err != nil {
+//	    return ai.ResourceOutput{}, err
+//	  }
+//	  return ai.ResourceOutput{
+//	    Content: []*ai.Part{ai.NewTextPart(string(content))},
+//	  }, nil
+//	})
+func DefineResource(g *Genkit, resourceName string, opts ai.ResourceOptions, fn ai.ResourceFunc) ai.Resource {
```
apascal07:
```suggestion
func DefineResource(g *Genkit, name string, opts *ai.ResourceOptions, fn ai.ResourceFunc) ai.Resource {
```

go/genkit/resource.go
```diff
@@ -0,0 +1,107 @@
...
+//
+//	// Use in generation:
+//	ai.Generate(ctx, g,
+//	  ai.WithPrompt([]*ai.Part{
+//	    ai.NewTextPart("Analyze this user:"),
+//	    ai.NewResourcePart("user://profile/123"),
+//	  }),
+//	  ai.WithResources([]ai.Resource{dynamicRes}),
+//	)
+func NewResource(resourceName string, opts ai.ResourceOptions, fn ai.ResourceFunc) ai.Resource {
```
apascal07:
```suggestion
func NewResource(resourceName string, opts *ai.ResourceOptions, fn ai.ResourceFunc) ai.Resource {
```

go/genkit/resource.go
```diff
@@ -0,0 +1,107 @@
...
+//	  return ai.ResourceOutput{Content: []*ai.Part{ai.NewTextPart(userData)}}, nil
+//	})
+//
+//	// Use in generation:
+//	ai.Generate(ctx, g,
+//	  ai.WithPrompt([]*ai.Part{
+//	    ai.NewTextPart("Analyze this user:"),
+//	    ai.NewResourcePart("user://profile/123"),
+//	  }),
+//	  ai.WithResources([]ai.Resource{dynamicRes}),
```
apascal07:
```suggestion
//	  ai.WithResources(dynamicRes),
```

This should still work and is preferred for variadic args.

go/ai/resource.go
huangjeff5: Good catch, think this was some leftover code while refactoring

go/genkit/resource.go
huangjeff5: No strong preference, I initially just thought it might be a nice utility to have but probably can be removed

go/genkit/resource.go
huangjeff5: Sounds good

go/genkit/genkit.go
```diff
@@ -974,3 +974,77 @@ func DefineFormat(g *Genkit, name string, formatter ai.Formatter) {
...
+//
+//	// Use in generation:
+//	ai.Generate(ctx, g,
+//	  ai.WithPrompt([]*ai.Part{
+//	    ai.NewTextPart("Analyze this user:"),
+//	    ai.NewResourcePart("user://profile/123"),
+//	  }),
+//	  ai.WithResources(dynamicRes),
+//	)
+func NewResource(resourceName string, opts *ai.ResourceOptions, fn ai.ResourceFunc) ai.Resource {
```
apascal07:
```suggestion
func NewResource(name string, opts *ai.ResourceOptions, fn ai.ResourceFunc) ai.Resource {
```

apascal07 (approved):

PR #3355, opened 2025-07-31
fix(go): Fixed type assertions on LookupPlugin

gregfurman:
Description:
 - https://github.com/firebase/genkit/issues/3354
 - Performs a safe type-switch on the returned value of a registry's LookupPlugin. If a plugin does not exist (or is the incorrect type), this will no longer panic.

Checklist (if applicable):
- [x] Tested (manually) 


apascal07:
Thank you, Greg!

PR #3364, opened 2025-08-01
feat(js/plugins/google-genai): Added Veo and Lyria

ifielker:
(Plus some refactoring)

Checklist (if applicable):

### DIFF
--- a/js/ai/src/formats/types.ts
+++ b/js/ai/src/formats/types.ts
@@ -15,7 +15,7 @@
  */
 
 import type { JSONSchema } from '@genkit-ai/core';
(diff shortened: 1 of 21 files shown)


js/ai/src/formats/types.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import type { JSONSchema } from '@genkit-ai/core';
-import type { GenerateResponseChunk } from '../generate.js';
+import type { GenerateResponseChunk } from '../generate/chunk.js';
```
ifielker, line 18: It was complaining about re-exports so I got it from the original definition instead of the re-export

pavelgj (approved):

PR #3367, opened 2025-08-03
fix(compat-oai): dynamically add file extension for transcription workflow

nsrCodes:
- Tries to fix the first part of https://github.com/firebase/genkit/issues/3366
 - Required for whipser API else causes errors ([Ref](https://community.openai.com/t/whisper-api-cannot-read-files-correctly/93420/20))


MichaelDoyle:
@ssbushi can you take a peek at this?

ssbushi (approved):

PR #3368, opened 2025-08-03
chore: return `raw` in response generated by `generate`

nsrCodes:
- to make it possible to extract segments and words from verbose responses in transcription flows as describe in https://github.com/firebase/genkit/issues/3366

### DIFF
--- a/js/ai/src/generate/response.ts
+++ b/js/ai/src/generate/response.ts
@@ -210,6 +210,7 @@ export class GenerateResponse<O = unknown> implements ModelResponseData {
       custom: (this.custom as { toJSON?: () => any }).toJSON?.() || this.custom,
       request: this.request,
       operation: this.operation,
+      raw: this.raw,
     };
     if (!out.finishMessage) delete out.finishMessage;
     if (!out.request) delete out.request;


MichaelDoyle:
Hey @nsrCodes - typically "custom" should contain the raw response. It's possible the oai plugin needs to be adjusted to set custom as the raw response here.

I think we do want to move to using raw for this purpose (its clearer that way), but we are in a bit of a between state right now.

js/ai/src/generate/response.ts
```diff
@@ -210,6 +210,7 @@ export class GenerateResponse<O = unknown> implements ModelResponseData {
       custom: (this.custom as { toJSON?: () => any }).toJSON?.() || this.custom,
       request: this.request,
       operation: this.operation,
+      raw: this.raw,
```
pavelgj, line 213: I'm not sure what's the best way to do this... the `custom` field already contains the same data as `raw`... duplicating that data is not ideal.

nsrCodes: @pavelgj I had to end up doing this because there was no way of getting segment/word level transcriptions. 

these weren't present inside `custom` and so had to rely on pulling out `raw` like this to get the `{segments, words, text}` from oai whisper api

MichaelDoyle:
closing - as custom should contain the raw results, and if not - should be fixed there.

PR #3377, opened 2025-08-05
fix(js): `run()` calls the function with given input

### LINKED ISSUE #3376: [JS] `run()` doesn't pass the input to the passed function
**Describe the bug**

`run()` doesn't pass the input to the passed function.

This is caused by checking `arguments.length` in the `run()` implementation. That condition is always falsy.

https://github.com/firebase/genkit/blob/f18da1fd1adeb2c94764873f8e689508479b2c86/js/core/src/flow.ts#L200-L204

**To Reproduce**

```js
    it('should run the action', async () => {
      const act = async () => 'bar';

      const result = await run('action', act, registry);

      assert.equal(result, 'bar'); // => OK
    });

    it('should run the action with input', async () => {
      const act = async (input: string) => `${input} bar`;

      const result = await run('action', 'foo', act, registry);

      assert.equal(result, 'foo bar'); // => ERROR `undefined bar`.
    });
```

**Expected behavior**

Tests the above should pass.

**Screenshots**

 - OS: [e.g. Linux, MacOS]

** Node version
 -  run `node --version` at paste here

**Additional context**

lacolaco:
Fix #3376 

- Add tests for `run` function call.
- Avoid to use `arguments.length`

Checklist (if applicable):

### DIFF
--- a/js/core/src/flow.ts
+++ b/js/core/src/flow.ts
@@ -156,10 +156,12 @@ export function run<T>(
   let func;
   let input;
   let registry: Registry | undefined;
+  let hasInput = false;
   if (typeof funcOrInput === 'function') {
     func = funcOrInput;
   } else {
     input = funcOrInput;
+    hasInput = true;
   }
   if (typeof fnOrRegistry === 'function') {
     func = fnOrRegistry;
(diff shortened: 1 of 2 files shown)


pavelgj (approved):

pavelgj:
Thanks for fix!

PR #3379, opened 2025-08-05
refactor(js/plugins/google-genai): streaming aggregation logic...

ifielker:
between googleAI and vertexAI will now match each other (googleAI style) instead of previous implementations.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/common/utils.ts
+++ b/js/plugins/google-genai/src/common/utils.ts
@@ -310,7 +310,10 @@ function getResponseStream(
             let err = e;
             err.stack = e.stack;
             if (err.name === 'AbortError') {
-              err = new Error('Request aborted when reading from the stream');
+              err = new GenkitError({
+                status: 'ABORTED',
+                message: 'Request aborted when reading from the stream',
+              });
             } else {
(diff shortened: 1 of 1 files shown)


js/plugins/google-genai/src/common/utils.ts
```diff
@@ -232,3 +239,231 @@ export function cleanSchema(schema: JSONSchema): JSONSchema {
...
+              currentText = currentText.substring(match[0].length);
+              match = currentText.match(responseLineRE);
+            }
+            return pump();
+          })
+          .catch((e: Error) => {
+            let err = e;
+            err.stack = e.stack;
+            if (err.name === 'AbortError') {
+              err = new Error('Request aborted when reading from the stream');
```
pavelgj: we probably need to throw `GenkitError` instead of `Error`:

```ts
new GenkitError({
  status: 'ABORTED',
  message: 'Request aborted when reading from the stream',
})
```

pavelgj (approved):

js/plugins/google-genai/src/common/utils.ts
ifielker: done

PR #3393, opened 2025-08-09
fix: instantiate OpenAI client in all relevant async functions

### LINKED ISSUE #3387: [JS] [compat-oai] Error deploying using firebase cloud functions
Error when deploying cloud functions event when the firebase functions:secrets:set OPENAI_API_KEY is set

```
i  functions: Loading and analyzing source code for codebase default to determine what to deploy
Serving at port 8568

OpenAIError: The OPENAI_API_KEY environment variable is missing or empty; either provide it, or instantiate the OpenAI client with an apiKey option, like new OpenAI({ apiKey: 'My API Key' }).
```

mambisi:
Fixes early open ai library constructor call. When deploying firebase cloud functions the environmental variables are not initiated till functions is deployed therefore process.env.OPENAI_API_KEY is not available even if user has set the secretes.

Fixes #3387 

Replaces #3392 failed checks

### DIFF
--- a/js/plugins/compat-oai/src/index.ts
+++ b/js/plugins/compat-oai/src/index.ts
@@ -109,23 +109,25 @@ export interface PluginOptions extends Partial<ClientOptions> {
  * ```
  */
 export const openAICompatible = (options: PluginOptions) => {
-  const client = new OpenAI(options);
   let listActionsCache;
   return genkitPlugin(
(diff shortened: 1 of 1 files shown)


MichaelDoyle:
Excellent, thanks! Looping in @ssbushi  who is more familiar with this code.

PR #3413, opened 2025-08-13
fix(install_cli):  `sudo: command not found` on certain machines

cabljac:
Firebase Studio integration was reportedly getting the following:

Screenshot 2025-08-13 at 11 00 11

This PR:

- euid/sudo aware installs: if not root and install dir isn’t writable, use sudo when available; otherwise fall back to ~/.local/bin. Uninstall path uses the same logic.
- Correct install detection: separated desired target from existing binary; only treat as installed if an executable exists, preventing false “corrupted/installed” when GENKIT_BINARY is set.
- Repro + validation: added .github/workflows/repro-install-cli.yml to reproduce the sudo-less failure and confirm the fixed script installs locally; verified locally with `act`.

### DIFF
--- a/.github/workflows/repro-install-cli.yml
+++ b/.github/workflows/repro-install-cli.yml
@@ -1,0 +1,32 @@
+name: Repro install_cli without sudo
+
+on:
+  workflow_dispatch: {}
+
+jobs:
+  repro:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:22.04
+    steps:
+      - name: Print environment info
+        run: |
+          id
+          uname -a
+          echo "PATH=$PATH"
+      - name: Install curl
+        run: |
+          apt-get update
+          DEBIAN_FRONTEND=noninteractive apt-get install -y curl ca-certificates
+      - name: Reproduce failure (no sudo, non-root)
+        run: |
+          useradd -m runner
+          su - runner -c 'set -x; curl -sL cli.genkit.dev | bash' || true
+      - name: Workaround (user-local install)
+        run: |
+          su - runner -c 'set -x; mkdir -p ~/.local/bin && curl -sL cli.genkit.dev | GENKIT_BINARY=$HOME/.local/bin/genkit bash && ~/.local/bin/genkit --version'
+      - name: Show where genkit installed
+        run: |
+          su - runner -c 'command -v genkit || true; ls -l ~/.local/bin/genkit || true'
+
+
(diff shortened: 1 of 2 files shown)


CorieW (approved):

pavelgj:
@cabljac were you able to repro this in Firebase Studio? You can just create a workspace in https://idx.google.com

IDX is nix. I think in their setup the problem might be that you don't get root access at all (can't sudo nor su). `/usr/local/bin` also doesn't exist and you can't create it.

I wonder if we need a "local=true" install option that writes the binary to `$HOME/.local/bin` instead of `/usr/local/bin`?

.github/workflows/repro-install-cli.yml
```diff
@@ -0,0 +1,35 @@
+name: Repro install_cli without sudo
+
+on:
+  workflow_dispatch: {}
+
+jobs:
+  repro:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:22.04
```
pavelgj: firebase studio (idx) is nix, without root access. I don't think this is a representative repro.

cabljac: ah interesting, will update to address this

bin/install_cli
```diff
@@ -16,6 +16,12 @@
 #
 :   curl -sL cli.genkit.dev | bash
 #
+# For local execution with command-line options:
```
pavelgj: why do this? we don't allow running the script this way

bin/install_cli
```diff
@@ -74,20 +87,48 @@
...
 
+: -----------------------------------------
+:  Local install switch
+: -----------------------------------------
+
+# To force a user-local install into $HOME/.local/bin (useful on systems without sudo),
+# use the local option:
+#
+# Curl execution:
+: curl -sL cli.genkit.dev | GENKIT_INSTALL=local bash
```
pavelgj: why no make it consistent with other options? Ex: `curl -sL cli.genkit.dev | local=true bash` or something like that?

bin/install_cli
cabljac: this is fair, my thoughts were for manual testing purposes, but on reflection it's just more maintenance overhead

bin/install_cli
cabljac: ah here i was concerned with local being reserved, but we're doing `#!/usr/bin/env bash` so we should be fine, I was overthinking this.

cabljac: also `local` is a keyword in bash, so was thinking it could get confusing.

pavelgj: Let's stick to one of these options (at least in docs). I'd prefer `local=true`

Issue #3378, opened 2025-08-05
[Dev UI] Span name for `ai.generate()` step


lacolaco:
## Overview

In building a multi-step flow, each `ai.generate()` step cannot be identified at the trace view because they all are labeled with `generate`.

<img width="2610" height="1720" alt="Image" src="https://github.com/user-attachments/assets/4010ee2f-1ede-4176-924b-a86c66a12677" />

I'd like to distinguish them with explicit naming, as well as `ai.run()`.

### User goal(s)

- A developer can set a name to `ai.generate()` step.

### Acceptance Criteria
- 1
- 2
- 3

### Designs

- First draft: `ai.generate({ prompt: '...', stepName: 'foobar' })`

marcelfolaron:
@pavelgj I think this can be closed now.

PR #3416, opened 2025-08-13
Named generates

marcelfolaron:
Implements #3378 custom span names for generate steps. New optional parameter called `stepName` can be set to define the step name displayed in traces.

Checklist (if applicable):

### DIFF
--- a/js/ai/src/model-types.ts
+++ b/js/ai/src/model-types.ts
@@ -406,5 +406,7 @@ export const GenerateActionOptionsSchema = z.object({
   returnToolRequests: z.boolean().optional(),
   /** Maximum number of tool call iterations that can be performed in a single generate call (default 5). */
   maxTurns: z.number().optional(),
+  /** Custom step name for this generate call to display in trace views. Defaults to "generate". */
+  stepName: z.string().optional(),
 });
 export type GenerateActionOptions = z.infer<typeof GenerateActionOptionsSchema>;
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -170,6 +170,8 @@ export interface GenerateOptions<
   context?: ActionContext;
   /** Abort signal for the generate request. */
   abortSignal?: AbortSignal;
+  /** Custom step name for this generate call to display in trace views. Defaults to "generate". */
+  stepName?: string;
   /**
    * Additional metadata describing the GenerateOptions, used by tooling. If
    * this is an instance of a rendered dotprompt, will contain any prompt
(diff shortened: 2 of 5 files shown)


marcelfolaron:
Missed to commit my change to rename the parameter `name` in the testapp to `stepName`. Should build now

js/ai/src/model-types.ts
```diff
@@ -406,5 +406,7 @@ export const GenerateActionOptionsSchema = z.object({
   returnToolRequests: z.boolean().optional(),
   /** Maximum number of tool call iterations that can be performed in a single generate call (default 5). */
   maxTurns: z.number().optional(),
+  /** Custom step name for this generate call to display in trace views. Defaults to "generate". */
+  stepName: z.string().optional(),
```
pavelgj, line 410: These types need to be kept in sync with https://github.com/firebase/genkit/blob/main/genkit-tools/common/src/types/model.ts#L379

marcelfolaron: Thanks for calling that out. I  went and added stepName to the model.

marcelfolaron:
Huh, anything I need to do here re python checks? Seems the failed check is unrelated? 
@pavelgj

pavelgj:
> Huh, anything I need to do here re python checks? Seems the failed check is unrelated? @pavelgj

The data model in common is shared across runtimes (js, go and python). There are code generators that handle this stuff.

Run

```
pnpm build
py/bin/generate_schema_typing
```

(the py script may require uv)

marcelfolaron:
> pnpm build
> py/bin/generate_schema_typing

I see, thank you. That worked and I committed the 2 additional file changes.

marcelfolaron:
It's failing with license checks. When I run it locally it fails as well and running `./bin/add_license` fixes it. However that adds a lot of license content to quite a few files which seems outside the scope of this PR.

pavelgj:
> It's failing with license checks. When I run it locally it fails as well and running `./bin/add_license` fixes it. However that adds a lot of license content to quite a few files which seems outside the scope of this PR.

must be upstream change in the license checker... can you try adding *.toml to ignore list in https://github.com/firebase/genkit/blob/9d5e4edb348817ae196936b424d70cbe72051de7/bin/check_license#L60

marcelfolaron:
> > It's failing with license checks. When I run it locally it fails as well and running `./bin/add_license` fixes it. However that adds a lot of license content to quite a few files which seems outside the scope of this PR.
> 
> must be upstream change in the license checker... can you try adding *.toml to ignore list in
> 
> https://github.com/firebase/genkit/blob/9d5e4edb348817ae196936b424d70cbe72051de7/bin/check_license#L60

That worked. Though I had to add toml and nix files
```
  -ignore '**/*.toml' \
  -ignore '**/*.nix' \
```

PR #3426, opened 2025-08-14
fix(js):Add precondition as option configuration 

mambisi:
Fixes firebase deployment error: When deploying firebase cloud functions fails `i  functions: Loading and analyzing source code for codebase default to determine what to deploy` due to checking the api keys in environmental variable before deployed.


ssbushi:
Thanks for the fix, I'm kinda torn on this. Do we really need this to be a separate argument to the plugin interface?

I think the API key checking can be handled in the initializer, resolver methods respectively. WDYT @pavelgj ?

ssbushi:
I see it. The apiKey needs to be resolved before the client object is constructed, so this has to happen before `initializer` and `resolver` are invoked. However, without the apiKey in `xAI` or `deepseek` plugins, they will not work correctly.

js/plugins/compat-oai/src/deepseek/index.ts
```diff
@@ -79,19 +79,21 @@ const listActions = async (client: OpenAI): Promise<ActionMetadata[]> => {
...
-        'Please pass in the API key or set the DEEPSEEK_API_KEY environment variable.',
-    });
-  }
+  
   return openAICompatible({
     name: 'deepseek',
     baseURL: 'https://api.deepseek.com',
-    apiKey,
     ...options,
+    precondition: async () => {
```
ssbushi, line 87: one way I see this working is if this is not just a precondition but a`resolveOptions` function. It should return the resolved options which can be used to create the OpenAI client.

mambisi: that should be good

MichaelDoyle:
thanks! i'm going to close this out, as a different solution was recommended. let us know if you want to pick it back up in the future.

PR #3427, opened 2025-08-15
feat(js): new plugin API that does not rely on the registry

pavelgj:
Checklist (if applicable):


js/genkit/src/genkit.ts
```diff
@@ -916,26 +919,71 @@ export class Genkit implements HasRegistry {
...
-        },
-        async listActions() {
-          if (loadedPlugin.listActions) {
-            return await loadedPlugin.listActions();
-          }
-          return [];
-        },
-      });
+      const isV2 =
+        typeof plugin === 'object' && typeof plugin.resolve === 'function';
```
ssbushi: Why not check the `version` field?

ssbushi (approved):

js/genkit/src/genkit.ts
pavelgj: Ah, missed it. Done.

PR #3431, opened 2025-08-15
fix(js/plugin): added init function to plugin v2

pavelgj:
Checklist (if applicable):

### DIFF
--- a/js/ai/src/embedder.ts
+++ b/js/ai/src/embedder.ts
@@ -59,7 +59,7 @@ const EmbedRequestSchema = z.object({
 });
 
 export interface EmbedRequest<O = any> {
-  input: DocumentData[];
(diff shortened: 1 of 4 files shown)


js/ai/src/embedder.ts
```diff
@@ -120,7 +120,13 @@ export function embedder<ConfigSchema extends z.ZodTypeAny = z.ZodTypeAny>(
   ) => Promise<EmbedResponse>
 ) {
   const embedder = action(embedderActionOptions(options), (i, opts) =>
-    runner(i, opts)
+    runner(
```
pavelgj, line 123: this is an unrelated fix for the embedder fn -- making sure we're passing in Document objects instead if raw DocumentData

PR #3432, opened 2025-08-15
refactor(js/plugin/compat-oai): migrated compat-oai plugin to v2 plugin api

pavelgj:
Checklist (if applicable):


js/plugins/compat-oai/src/deepseek/index.ts
```diff
@@ -57,6 +54,7 @@ const resolver = async (
   } else {
     logger.warn('Only model actions are supported by the DeepSeek plugin');
   }
+  return undefined;
```
ssbushi:
```suggestion
  } else {
    logger.warn('Only model actions are supported by the DeepSeek plugin');
    return undefined;
  }
```

ssbushi: Is `undefined` an acceptable return value (without causing logs and errors)?

pavelgj: yes, undefined is fine. basically means that the plugin could not resolve the action.

Issue #3160, opened 2025-07-03
[JS] Explore and document workarounds or (better yet) fix Next.js build warnings


mbleigh:
**Describe the bug**
With a default Next.js app Genkit causes tons of warnings to be displayed at build time. While it does work, this is ugly and potentially worrisome for users. We should investigate and fix.

```
> next-genkit@0.1.0 build /Users/bleigh/sandbox/next-genkit
> next build

   ▲ Next.js 15.3.4

   Creating an optimized production build ...
 ⚠ Compiled with warnings in 2000ms

./node_modules/.pnpm/@opentelemetry+sdk-node@0.52.1_@opentelemetry+api@1.9.0/node_modules/@opentelemetry/sdk-node/build/src/TracerProviderWithEnvExporter.js
Module not found: Can't resolve '@opentelemetry/exporter-jaeger' in '/Users/bleigh/sandbox/next-genkit/node_modules/.pnpm/@opentelemetry+sdk-node@0.52.1_@opentelemetry+api@1.9.0/node_modules/@opentelemetry/sdk-node/build/src'

Import trace for requested module:
./node_modules/.pnpm/@opentelemetry+sdk-node@0.52.1_@opentelemetry+api@1.9.0/node_modules/@opentelemetry/sdk-node/build/src/TracerProviderWithEnvExporter.js
./node_modules/.pnpm/@opentelemetry+sdk-node@0.52.1_@opentelemetry+api@1.9.0/node_modules/@opentelemetry/sdk-node/build/src/sdk.js
./node_modules/.pnpm/@opentelemetry+sdk-node@0.52.1_@opentelemetry+api@1.9.0/node_modules/@opentelemetry/sdk-node/build/src/index.js
./node_modules/.pnpm/@genkit-ai+core@1.14.0/node_modules/@genkit-ai/core/lib/tracing.js
./node_modules/.pnpm/genkit@1.14.0/node_modules/genkit/lib/tracing.js
./node_modules/.pnpm/genkit@1.14.0/node_modules/genkit/lib/genkit.js
./node_modules/.pnpm/genkit@1.14.0/node_modules/genkit/lib/index.mjs
./src/app/api/route.ts

./node_modules/.pnpm/handlebars@4.7.8/node_modules/handlebars/lib/index.js
require.extensions is not supported by webpack. Use a loader instead.

Import trace for requested module:
./node_modules/.pnpm/handlebars@4.7.8/node_modules/handlebars/lib/index.js
[...]

PR #3459, opened 2025-08-20
fix(js) Fix next js warnings related to firebase

marcelfolaron:
The warnings in #3160 are primarily webpack related and can be mitigated using webpack magic comments

Checklist (if applicable):

### DIFF
--- a/js/core/src/tracing.ts
+++ b/js/core/src/tracing.ts
@@ -52,7 +52,10 @@ async function checkFirebaseMonitoringAutoInit() {
     process.env.ENABLE_FIREBASE_MONITORING === 'true'
   ) {
     try {
-      const firebaseModule = await require('@genkit-ai/firebase');
+      // Use webpack-friendly dynamic import for optional dependency
+      const firebaseModule = await import(
+        /* webpackIgnore: true */ '@genkit-ai/firebase'
+      );
       firebaseModule.enableFirebaseTelemetry();
     } catch (e) {
       logger.warn(


js/core/src/tracing.ts
```diff
@@ -52,7 +52,13 @@ async function checkFirebaseMonitoringAutoInit() {
     process.env.ENABLE_FIREBASE_MONITORING === 'true'
   ) {
     try {
-      const firebaseModule = await require('@genkit-ai/firebase');
+      // Use runtime dynamic import to bypass TypeScript module resolution
+      // @genkit-ai/firebase is declared as optionalDependencies for webpack resolution
+      const importModule = new Function(
+        'moduleName',
+        'return import(moduleName)'
```
pavelgj, line 57: does this need to be `import` or `require` would work?

marcelfolaron: `require` works as well and you don't need `await` with that.

schnecle (commented):
Nice! Can you describe how you tested this change?

marcelfolaron:
> Nice! Can you describe how you tested this change?

Mainly by running the use case @mbleigh used here: #3160
Anything in particular you'd want me to test?

marcelfolaron:
PS: Sorry apparently `require` cannot be used in this way and so we'd need to use import. (which I had before)

schnecle:
> > Nice! Can you describe how you tested this change?
> 
> Mainly by running the use case @mbleigh used here: #3160 Anything in particular you'd want me to test?

Can you make sure that you can initialize firebase telemetry with the environment variable:
`export ENABLE_FIREBASE_MONITORING=true`

And that if you have both a direct call to `enableFirebaseTelemetry()` that overrides configuration and ENABLE_FIREBASE_MONITORING=true, the overriding configuration wins?

Docs for reference: 
* [Enabling firebase telemetry with environment variables](https://genkit.dev/docs/observability/getting-started/#environment-based-configuration)
* [Overriding configuration to export from local](https://genkit.dev/docs/observability/advanced-configuration/#export-local-telemetry)

marcelfolaron:
> > > Nice! Can you describe how you tested this change?
> > 
> > 
> > Mainly by running the use case @mbleigh used here: #3160 Anything in particular you'd want me to test?
> 
> Can you make sure that you can initialize firebase telemetry with the environment variable: `export ENABLE_FIREBASE_MONITORING=true`
> 
> And that if you have both a direct call to `enableFirebaseTelemetry()` that overrides configuration and ENABLE_FIREBASE_MONITORING=true, the overriding configuration wins?
> 
> Docs for reference:
> 
> * [Enabling firebase telemetry with environment variables](https://genkit.dev/docs/observability/getting-started/#environment-based-configuration)
> * [Overriding configuration to export from local](https://genkit.dev/docs/observability/advanced-configuration/#export-local-telemetry)

Thanks for the pointer. I can confirm that the solution with import() appears to be working with `ENABLE_FIREBASE_MONITORING=true`  as well as `enableFirebaseTelemetry()` including the precedence test.

marcelfolaron:
@schnecle let me know if there is anything else I should test or do.

marcelfolaron:
Thank you for pointing it out!

PR #3462, opened 2025-08-20
breaking(go): refactored plugins, detached primitives from registry, refactored types

apascal07:
Summary of changes:

- Tracer provider is now lazily instantiated when `TracerProvider()` is called instead of setting the state on each action.
- Primitives have had underlying types changed to embed `ActionDef[In, Out, Stream]` which allows them to inherit all methods from the base `core.Action` interface. This was necessary to allow a common interface to be returned from plugins and passed around between `NewAction`, `DefineAction`, and `registry.RegisterAction`. This is a non-breaking change since all APIs operate on the interfaces only, not the underlying type.
- Primitives now all have `NewX` functions that create a detached primitive which can then be registered with `action.Register(registry)`.
- Plugins have been updated to the new plugin API which does not concern itself with the registry and only returns action instances.

Checklist (if applicable):


go/ai/embedder.go
```diff
@@ -82,13 +82,14 @@ type EmbedderOptions struct {
...
 // embedder is an action with functions specific to converting documents to multidimensional vectors such as Embed().
-type embedder core.ActionDef[*EmbedRequest, *EmbedResponse, struct{}]
+type embedder struct {
+	core.ActionDef[*EmbedRequest, *EmbedResponse, struct{}]
+}
 
-// DefineEmbedder registers the given embed function as an action, and returns an
-// [Embedder] that runs it.
-func DefineEmbedder(r *registry.Registry, name string, opts *EmbedderOptions, fn EmbedderFunc) Embedder {
+// NewEmbedder creates a new [Embedder] without registering it.
```
pavelgj:
```suggestion
// NewEmbedder creates a new [Embedder].
```

go/genkit/plugin.go
```diff
@@ -31,16 +31,14 @@ type Plugin interface {
...
 
 // DynamicPlugin is a [Plugin] that can dynamically resolve actions.
 type DynamicPlugin interface {
 	Plugin
 	// ListActions returns a list of action descriptors that the plugin is capable of resolving to [core.Action]s.
 	ListActions(ctx context.Context) []core.ActionDesc
-	// ResolveAction resolves an action type and name by defining and registering it in the registry.
-	ResolveAction(g *Genkit, atype core.ActionType, name string) error
+	// ResolveAction resolves an action type and name to a [core.Action].
+	ResolveAction(atype core.ActionType, name string) core.Action
```
pavelgj, line 43: don't need error?

apascal07: I don't think so? If we're sticking with panicking when defining/registering things, then this should either return a valid action, nil (if not able to be resolved by this plugin), or panic for all other cases. Even before this PR, this error was just logged, not returned in any way.

go/ai/tools.go
```diff
@@ -47,26 +47,25 @@ func (t ToolName) Name() string {
...
 
-// tool is the internal implementation of the Tool interface.
-// It holds the underlying core action and allows looking up tools
-// by name without knowing their specific input/output types.
+// tool is an action with functions specific to tools.
+// It embeds [core.Action] instead of [core.ActionDef] like other primitives
+// because the inputs/outputs can vary and the tool is only meant to be called
+// with JSON input anyway.
 type tool struct {
 	core.Action
```
huangjeff5, line 55: core.ActionDef[*ToolInput, *ToolOutput, struct{}]?

huangjeff5:
Resources + MCP changes look good to me.

huangjeff5 (approved):
Resources / MCP code changes LGTM

go/ai/tools.go
apascal07: `tool` is the exception since it does not have predefined inputs and outputs like the other `ai` primitives so we can't get hard types when reading back from the registry.

PR #3464, opened 2025-08-21
feat(cli): add update checks

CorieW:


genkit-tools/cli/src/utils/updates.ts
```diff
@@ -0,0 +1,246 @@
...
+    throw new GenkitToolsError(
+      `Failed to fetch npm versions: ${error?.message ?? error}`
+    );
+  }
+}
+
+/**
+ * Checks if update notifications are disabled via environment variable or user config.
+ */
+function areUpdateNotificationsDisabled(): boolean {
```
pavelgj: I know it's not grammatically correct, but let's stick with `is*`... 

```suggestion
function isUpdateNotificationsDisabled(): boolean {
```

genkit-tools/cli/src/utils/updates.ts
```diff
@@ -0,0 +1,246 @@
...
+
+  if (!platformData) {
+    logger.debug(`No update information for platform: ${machine}`);
+    return null;
+  }
+
+  const latestVersion = normalizeVersion(gcsLatestData.latestVersion);
+  const fileName = platformData.versionedUrl.split('/').pop() || '';
+  const downloadUrl = `${GCS_BUCKET_URL}/prod/${machine}/v${latestVersion}/${fileName}`;
+  const updateMessage = `${clc.dim('Run')} ${clc.bold(`curl -Lo ./genkit_bin ${downloadUrl}`)} ${clc.dim('to upgrade')}`;
```
pavelgj: This should be `curl -sL cli.genkit.dev | uninstall=true bash` and perhaps link to https://genkit.dev/docs/devtools/ for alternative options.

genkit-tools/cli/src/utils/updates.ts
```diff
@@ -0,0 +1,246 @@
...
+        `Invalid semver: current=${current}, latest=${latestVersion}`
+      );
+      return;
+    }
+
+    if (!semver.gt(latestVersion, current)) {
+      return;
+    }
+
+    logger.info(
```
pavelgj: lets make the message look very similar to firebase cli:
https://github.com/firebase/firebase-tools/blob/0f06446371dcd02174f3ea23d7715de3a2b30cdf/src/bin/cli.ts#L80

genkit-tools/cli/src/utils/updates.ts
```diff
@@ -0,0 +1,246 @@
...
+/**
+ * Gets the latest CLI version from npm registry for non-binary installations.
+ * @param ignoreRC - If true, ignore prerelease versions (default: true)
+ */
+export async function getLatestVersionFromNpm(
+  ignoreRC: boolean = true
+): Promise<string | null> {
+  try {
+    const response = await AXIOS_INSTANCE.get(
+      `https://registry.npmjs.org/${name}`
```
pavelgj: is this correct? `name` is not a defined var/const, is it?... so the name must resolve to function name...  so this must end up checking for `https://registry.npmjs.org/getLatestVersionFromNpm`?

pavelgj:
```suggestion
      `https://registry.npmjs.org/genkit-cli`
```

CorieW: It's the package name, generated in the version file, and imported [here](https://github.com/firebase/genkit/pull/3464/files#diff-bd9b2466bea273fbfd5ab71894a1a65e6141fc0381d32f62b99ae4e882ed4328R25)

CorieW: I'll rename this to `packageName` to clear things up

pavelgj: lol 🤦 ... yeah, a more specific variable name is a good idea.

pavelgj:
please sync to head on main, it should fix the go build error

PR #3478, opened 2025-08-25
docs: Update some docs and samples to use the new google-genai plugin

ifielker:

Checklist (if applicable):

### DIFF
--- a/samples/js-gemini/src/index.ts
+++ b/samples/js-gemini/src/index.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import googleAI from '@genkit-ai/googleai';
+import { googleAI } from '@genkit-ai/google-genai';
 import * as fs from 'fs';
 import { genkit, MediaPart, z } from 'genkit';
 import { Readable } from 'stream';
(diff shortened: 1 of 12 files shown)


samples/js-gemini/src/index.ts
```diff
@@ -358,10 +358,18 @@ ai.defineFlow('photo-move-veo', async (_, { sendChunk }) => {
   return operation;
 });
 
+function getApiKeyFromEnvVar(): string | undefined {
```
pavelgj, line 361: this is not necessary. GEMINI_API_KEY is the standard api key they are expected to use. GOOGLE_API_KEY and GOOGLE_GENAI_API_KEY are legacy api keys we still support in the plugins for backwards compatibility.

ifielker: It's tripped me up twice now, so I figured it might be causing problems for others as well. Calls to the models work fine. Everything works up until the download and then it doesn't because the download is looking for a different key. So maybe my setup is wrong, but also maybe others are using the same setup since it works except for the download...

PR #3487, opened 2025-08-28
refactor(plugins/vertexai): migrate to v2 API

CorieW:
Couple of issues still:
- Unsure of `ai.embed` alternative in `vector_search/indexers.ts`
- Unsure of `ai.embedMany` alternative in `vector_search/retrievers.ts`

Todo:
- [ ] Test

### DIFF
--- a/js/plugins/vertexai/src/embedder.ts
+++ b/js/plugins/vertexai/src/embedder.ts
@@ -14,15 +14,16 @@
  * limitations under the License.
  */
 
-import { z, type Document, type Genkit } from 'genkit';
+import { z, type Document } from 'genkit';
 import {
-  embedderRef,
+  embedderRef as createEmbedderRef,
   type EmbedderAction,
   type EmbedderReference,
 } from 'genkit/embedder';
 import type { GoogleAuth } from 'google-auth-library';
 import type { PluginOptions } from './common/types.js';
 import { predictModel, type PredictClient } from './predict.js';
+import { embedder } from 'genkit/plugin';
 
 export const TaskTypeSchema = z.enum([
   'RETRIEVAL_DOCUMENT',
@@ -61,7 +62,7 @@ function commonRef(
   name: string,
   input?: InputType[]
 ): EmbedderReference<typeof VertexEmbeddingConfigSchema> {
-  return embedderRef({
+  return createEmbedderRef({
     name: `vertexai/${name}`,
     configSchema: VertexEmbeddingConfigSchema,
     info: {
@@ -88,7 +89,7 @@ export const multimodalEmbedding001 = commonRef('multimodalembedding@001', [
   'image',
   'video',
 ]);
-export const geminiEmbedding001 = embedderRef({
+export const geminiEmbedding001 = createEmbedderRef({
   name: 'vertexai/gemini-embedding-001',
   configSchema: VertexEmbeddingConfigSchema,
   info: {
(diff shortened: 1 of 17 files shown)


js/plugins/vertexai/src/index.ts
```diff
@@ -268,23 +264,23 @@ async function listActions(options?: PluginOptions) {
...
   let listActionsCache;
-  return genkitPlugin(
-    'vertexai',
-    async (ai: Genkit) => await initializer(ai, options),
-    async (ai: Genkit, actionType: ActionType, actionName: string) =>
-      await resolver(ai, actionType, actionName, options),
-    async () => {
+  return genkitPluginV2({
+    name: 'vertexai',
+    init: async () => initializer(),
```
cabljac: this needs to accept options no?

cabljac: same as resolve?

js/plugins/vertexai/src/index.ts
```diff
@@ -150,50 +149,48 @@ async function initializer(ai: Genkit, options?: PluginOptions) {
...
   options?: PluginOptions
-) {
+): Promise<ResolvableAction | undefined> {
   const { projectId, location, vertexClientFactory, authClient } =
     await getDerivedParams(options);
 
   if (actionName.startsWith('imagen')) {
-    defineImagenModel(ai, actionName, authClient, { projectId, location });
+    defineImagenModel(actionName, authClient, { projectId, location });
     return;
```
cabljac: I think we need to return the model now?

js/plugins/vertexai/src/index.ts
```diff
@@ -111,24 +110,25 @@ export {
...
 };
 
-async function initializer(ai: Genkit, options?: PluginOptions) {
+async function initializer(options?: PluginOptions) {
   const { projectId, location, vertexClientFactory, authClient } =
     await getDerivedParams(options);
 
+  const actions: ResolvableAction[] = [];
+
   Object.keys(SUPPORTED_IMAGEN_MODELS).map((name) =>
```
cabljac: I think i'd prefer a `for ... of` loop here instead of a map, now that we're doing something with it.

js/plugins/vertexai/src/embedder.ts
```diff
@@ -298,18 +298,18 @@ export function defineVertexAIEmbedder(
...
+      name: embedderRef.name,
+      configSchema: embedderRef.configSchema,
+      info: embedderRef.info!,
     },
     async (input, options) => {
-      const predictClient = predictClientFactory(options);
+      const predictClient = predictClientFactory(embedderRef.config);
       const response = await predictClient(
-        input.map((doc: Document) => {
+        input.input.map((doc: Document) => {
```
cabljac: i think call it request (or something else) so we dont do `input.input`

I don't think we use options here? Could we just do 
```
async (request) => {
```

js/plugins/vertexai/src/embedder.ts
```diff
@@ -370,13 +370,13 @@ export function defineVertexAIEmbedder(
             // Text only embedder
             instance = {
               content: doc.text,
-              task_type: options?.taskType,
-              title: options?.title,
+              task_type: embedderRef.config?.taskType,
```
cabljac, line 376: I'm not sure about reading from the embedderRef here?

Before we used `options` - i'm thinking at request time we could have these options specified which would take precedence over the config of the ref maybe

js/plugins/vertexai/src/index.ts
CorieW: Ah yeah

js/plugins/vertexai/src/index.ts
CorieW: True, good spot

js/plugins/vertexai/src/embedder.ts
CorieW: Oh pretty sure I had, must have undone it somehow

PR #3488, opened 2025-08-28
fix(go/plugins/compat_oai): forward tool response to model

haibingtown:
Checklist:


apascal07:
Fixed!

PR #3490, opened 2025-08-28
feat(go/genkit/prompt): add support for code defined schemas

hugoaguirre:
Checklist (if applicable):

### DIFF
--- a/go/ai/option.go
+++ b/go/ai/option.go
@@ -877,7 +877,8 @@ type promptExecutionOptions struct {
 	commonGenOptions
 	executionOptions
 	documentOptions
-	Input any // Input fields for the prompt. If not nil this should be a struct that matches the prompt's input schema.
+	Input  any // Input fields for the prompt. If not nil this should be a struct that matches the prompt's input schema.
+	Output any // Output fields for the prompt. If not nil, this should be a struct that matches the prompt's desired output schema
 }
 
 // PromptExecuteOption is an option for executing a prompt. It applies only to [prompt.Execute].
@@ -913,4 +914,10 @@ func (o *promptExecutionOptions) applyPromptExecute(pgOpts *promptExecutionOptio
 // prompt's input schema and can either be a map[string]any or a struct of the same type.
 func WithInput(input any) PromptExecuteOption {
 	return &promptExecutionOptions{Input: input}
+}
+
+// WithOutput sets the output for the prompt desired output. Output must conform to the
+// prompt's output schema and should be a map[string]any or a struct of the same type.
+func WithOutput(output any) PromptExecuteOption {
+	return &promptExecutionOptions{Output: output}
 }
--- a/go/ai/prompt.go
+++ b/go/ai/prompt.go
@@ -120,6 +120,12 @@ func LookupPrompt(r *registry.Registry, name string) Prompt {
 	}
 }
 
+// DefineSchema defines a schema in the registry
+func DefineSchema(r *registry.Registry, name string, schema any) error {
+	// TODO: marshal the schema and store it into the registry
+	return nil
+}
+
 // Execute renders a prompt, does variable substitution and
 // passes the rendered template to the AI model specified by the prompt.
 func (p *prompt) Execute(ctx context.Context, opts ...PromptExecuteOption) (*ModelResponse, error) {
@@ -183,6 +189,8 @@ func (p *prompt) Render(ctx context.Context, input any) (*GenerateActionOptions,
 	if input == nil {
 		input = p.Desc().Metadata["prompt"].(map[string]any)["defaultInput"]
 	}
+
+	// TODO: should we consider output atp?
 
 	return p.Run(ctx, input, nil)
 }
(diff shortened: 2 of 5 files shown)


go/samples/prompts/main.go
```diff
@@ -328,6 +329,47 @@ func PromptWithMediaType(ctx context.Context, g *genkit.Genkit) {
...
+	Title       string       `json:"title" description:"Recipe name"`
+	Ingredients []Ingredient `json:"ingredients" description:"Recipe ingredients"`
+	Steps       []string     `json:"steps" description:"Recipe steps"`
+}
+
+func PromptWithSchema(ctx context.Context, g *genkit.Genkit) {
+	err := genkit.DefineSchema(g, "recipe", RecipeSchema{})
+	if err != nil {
+		log.Fatalf("error defining schema: %v", err)
+	}
```
apascal07:
```suggestion
	genkit.DefineSchema(g, "recipe", RecipeSchema{})
```

This should panic instead of returning an error.

go/samples/prompts/main.go
```diff
@@ -343,6 +344,46 @@ func PromptWithMediaType(ctx context.Context, g *genkit.Genkit) {
...
+		log.Fatal("empty prompt")
+	}
+
+	// prompt schemas can be referenced at any time
+	genkit.DefineSchema(g, "recipe", RecipeSchema{})
+
+	resp, err := prompt.Execute(ctx,
+		ai.WithModelName("googleai/gemini-2.5-pro"),
+		ai.WithInput(map[string]any{"food": "tacos", "ingredients": []string{"octopus", "shrimp"}}),
+		ai.WithOutput(RecipeSchema{}),
```
apascal07: The intention is to reference it by name, so this should be called something like `ai.WithOutputSchemaName("recipe")`.

go/ai/prompt.go
```diff
@@ -122,6 +122,18 @@ func LookupPrompt(r api.Registry, name string) Prompt {
 	}
 }
 
+// DefineSchema defines a schema in the registry
+// It panics if an error was encountered
+func DefineSchema(r api.Registry, name string, schema any) {
```
apascal07:
```suggestion
func DefineSchema(r api.Registry, name string, schema map[string]any) {
```

go/ai/prompt.go
```diff
@@ -122,6 +122,18 @@ func LookupPrompt(r api.Registry, name string) Prompt {
...
+	reflector := &jsonschema.Reflector{}
+	s := reflector.Reflect(schema)
+	b, err := json.Marshal(s)
+	if err != nil {
+		panic(fmt.Errorf("ai.DefineSchema: failed to marshal schema %q: %w", name, err))
+	}
```
apascal07: Shouldn't need to do this because it should already be a map.

go/samples/prompts/main.go
hugoaguirre: Added `WithOutputSchemaName` and kept the same core logic:
- Deferred schema resolution
- Output schema overrides

go/ai/option.go
```diff
@@ -913,3 +921,8 @@ func (o *promptExecutionOptions) applyPromptExecute(pgOpts *promptExecutionOptio
 func WithInput(input any) PromptExecuteOption {
 	return &promptExecutionOptions{Input: input}
 }
+
+// WithOutputSchemaName sets the schema name that will be used to render the prompt's desired output.
+func WithOutputSchemaName(schema string) PromptExecuteOption {
```
apascal07: This should be a `OutputOption` not a `PromptExecuteOption`. It should be alongside `OutputSchema`.

go/genkit/genkit.go
```diff
@@ -618,6 +618,11 @@ func LookupPrompt(g *Genkit, name string) ai.Prompt {
 	return ai.LookupPrompt(g.reg, name)
 }
 
+// DefineSchema defines a prompt schema programmatically
+func DefineSchema(g *Genkit, name string, schema map[string]any) {
```
apascal07, line 639: We probably also want `DefineSchemaWithType` that takes in a Go value and reflects it into a schema.

go/internal/registry/registry.go
```diff
@@ -46,10 +49,32 @@ func New() *Registry {
...
+		SchemaResolver: func(name string) (*jsonschema.Schema, error) {
+			s := r.LookupSchema(name)
+			if s == nil {
+				r.RegisterSchema(name, nil)
+				// Define a reference schema that Dotprompt can handle
+				// Use a custom URI scheme "genkit:" to denote registry references
+				return &jsonschema.Schema{
+					Ref: "genkit:" + name,
+				}, nil
+			}
```
apascal07, line 66: What does this do? When is `SchemaResolver` called? We should be erroring if we don't find the schema at execution.

hugoaguirre: This `SchemaResolver` is basically used when we are loading a prompt file, either with `LoadPrompt` or `LoadPromptDir`. This is in case there's a schema reference in the .prompt file that later on we will have to resolve (in execution time)

go/genkit/genkit.go
hugoaguirre: Adding it next

PR #3503, opened 2025-09-04
feat(py): Added the render_system_prompt function and the dotprompt i…

hendrixmar:
…nstance to registry

Checklist (if applicable):

### DIFF
--- a/py/packages/genkit/src/genkit/blocks/prompt.py
+++ b/py/packages/genkit/src/genkit/blocks/prompt.py
@@ -21,10 +21,13 @@ This module provides types and utilities for managing prompts and templates
 used with AI models in the Genkit framework. It enables consistent prompt
 generation and management across different parts of the application.
 """
-
+import asyncio
 from asyncio import Future
 from collections.abc import AsyncIterator
-from typing import Any
+from typing import Any, Awaitable, Protocol
+
+from dotpromptz.typing import PromptFunction
+from pydantic import BaseModel
 
 from genkit.aio import Channel
 from genkit.blocks.generate import (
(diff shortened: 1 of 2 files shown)


py/packages/genkit/src/genkit/blocks/prompt.py
```diff
@@ -408,3 +437,53 @@ def _normalize_prompt_arg(
...
+async def render_dotprompt_to_parts(
+    registry: Registry,
+    prompt_function: PromptFunction,
+    input_: Any,
+    options: PromptConfig,
+) -> list[Part]:
+    """Renders a prompt using dotprompt."""
+    rendered = await prompt_function(input_)
+
+    if 1 < len(rendered.messages):
```
pavelgj:
```suggestion
    if len(rendered.messages) > 1:
```

yesudeep:
@hendrixmar can we please address the python checks that are failing?

py/packages/genkit/src/genkit/blocks/prompt.py
```diff
@@ -408,3 +437,53 @@ def _normalize_prompt_arg(
         return prompt
     else:
         return [prompt]
+
+
+async def render_system_prompt(registry: Registry, input_: Any, options: PromptConfig) -> Message:
```
pavelgj: input type can be `str | Part | list[Part] | PartsResolver`, right?

py/packages/genkit/src/genkit/blocks/prompt.py
```diff
@@ -408,3 +437,53 @@ def _normalize_prompt_arg(
...
+        result = await options.system(input_, ActionRunContext._current_context())
+
+        return Message(
+            role=Role.SYSTEM,
+            content=_normalize_prompt_arg(result)
+        )
+
+    elif isinstance(options.system, str):
+
+        user_prompt = await registry.dotprompt.compile(input_)
```
pavelgj: It would be great to if we could cache the compiled dotprompt template. Otherwise we have to compile it on every invocation.

hendrixmar:
> @hendrixmar can we please address the python checks that are failing?

py/packages/genkit/src/genkit/blocks/prompt.py
hendrixmar: Yes, I reviewed that part from the JS implementation, and I was wondering if that object was already defined in the Python implementation. I searched using the same name and I haven't found anything close to that
image
image

py/packages/genkit/src/genkit/blocks/prompt.py
hendrixmar: In the JS implementation I see that it is **z.infer<I>** and in the Golang it  use **map[string]any** 
so not sure which one to use

py/packages/genkit/src/genkit/blocks/prompt.py
pavelgj, line 1: there are presubmit checks failing for this file

PR #3507, opened 2025-09-05
fix(ai): propagate abortSignal in recursive generateHelper calls

longfin:
When generateHelper recursively calls itself during tool iterations, the abortSignal was not being passed to the recursive call. This caused abort signals to be lost after the first tool iteration, preventing proper cancellation of long-running operations with multiple tool calls.

This fix ensures abortSignal is propagated through all recursive calls, maintaining cancellation capability throughout the entire generation lifecycle.

Fixes issue where task cancellation fails when maxTurns > 1

Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate/action.ts
+++ b/js/ai/src/generate/action.ts
@@ -408,6 +408,7 @@ async function generate(
     currentTurn: currentTurn + 1,
     messageIndex: messageIndex + 1,
     streamingCallback,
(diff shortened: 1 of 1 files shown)


longfin:
@pavelgj PTAL 🙏

pavelgj:
Good catch. Thanks!

PR #3512, opened 2025-09-06
chore(go): update samples and test cases

hugoaguirre:
Checklist (if applicable):


go/plugins/googlegenai/gemini.go
```diff
@@ -613,6 +613,11 @@ func translateCandidate(cand *genai.Candidate) *ai.ModelResponse {
 		m.FinishReason = ai.FinishReasonUnknown
 	}
 	msg := &ai.Message{}
+	if cand.Content == nil {
+		m.Message = msg
```
hugoaguirre: @apascal07 when testing MCP, there were some spurious responses that returned no errors but no candidates on them. When falling in this corner case, should we panic or silently return an empty response message?

apascal07: Which responses are you referring to? It's intermittent or it fails for certain requests consistently?

hugoaguirre: Yeah the responses from the plugin sometimes come with no candidates when running the `mcp-client` sample using the `client` example. I'd say 2 out of 5 times the issue is reproducible.

apascal07: Then create an error and return it saying no valid candidates were produced. That's pretty weird though that there is neither a response nor an error. Please sort this out with @huangjeff5.

huangjeff5: I believe the cause of the intermittent issue was this: https://github.com/firebase/genkit/commit/8094cbd6814d1fb1fddfb627b0b6c95f33840852

Can you give it another try?

hugoaguirre: Hi @huangjeff5. I encountered an issue at main in the mcp-client sample where the model ref was not getting resolved when calling generate(). 

The tools are always being listed now, but every time I try to run the generate call using the WithModelName() helper function, the model resolution does not happen successfully. 

I'll continue the investigation later today

go/plugins/googlegenai/gemini.go
```diff
@@ -339,11 +348,17 @@ func toGeminiRequest(input *ai.ModelRequest, cache *genai.CachedContent) (*genai
...
+	// candidate count might not be set to 1 and will keep its zero value if not set
+	// e.g. default value from reflection server is 0
+	if gcc.CandidateCount == 0 {
+		gcc.CandidateCount = 1
+	}
 
 	// Genkit primitive fields must be used instead of go-genai fields
 	// i.e.: system prompt, tools, cached content, response schema, etc
+	if gcc.CandidateCount != 1 {
+		return nil, fmt.Errorf("candidate count must be 1, got: %d", gcc.CandidateCount)
```
apascal07:
```suggestion
		return nil, errors.New("multiple candidates is not supported")
```

go/samples/gemini/main.go
```diff
@@ -0,0 +1,399 @@
...
+	ctx := context.Background()
+
+	// Initialize Genkit with the Google AI plugin. When you pass nil for the
+	// Config parameter, the Google AI plugin will get the API key from the
+	// GEMINI_API_KEY or GOOGLE_API_KEY environment variable, which is the recommended
+	// practice.
+	g := genkit.Init(ctx,
+		genkit.WithPlugins(
+			&googlegenai.GoogleAI{},
+			&googlegenai.VertexAI{},
```
apascal07: Separate this out so that it's easy to just run this without setting both credentials.

go/samples/gemini/main.go
```diff
@@ -0,0 +1,399 @@
...
+	})
+
+	// Define a simple flow that generates jokes about a given topic with a context of bananas
+	genkit.DefineFlow(g, "context", func(ctx context.Context, input string) (string, error) {
+		resp, err := genkit.Generate(ctx, g,
+			ai.WithModelName("googleai/gemini-2.0-flash"),
+			ai.WithConfig(&genai.GenerateContentConfig{
+				Temperature: genai.Ptr[float32](1.0),
+			}),
+			ai.WithPrompt(fmt.Sprintf(`Tell silly short jokes about %s`, input)),
```
apascal07: FYI, `WithPrompt` takes format args itself. Don't need `fmt.Sprintf`.

go/samples/gemini/main.go
apascal07, line 1: Make all flows streaming flows. We should do this by default, there's no reason not to support it in our simple examples.

apascal07:
Please rebase against `main` so that I test the recent telemetry changes.

hugoaguirre:
Not needed. Samples will have a different folder structure.

PR #3516, opened 2025-09-07
fix(js/generate): handle when finishReason is length

assafshemesh:
Currently when the model returns finishReason === 'length' the flow fails on schema validation (as output will probably be corrupted), hence hiding the real fail reason.
This fix will raise error properly on that case

Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate/response.ts
+++ b/js/ai/src/generate/response.ts
@@ -93,6 +93,13 @@ export class GenerateResponse<O = unknown> implements ModelResponseData {
       );
     }
 
+    if (this.finishReason === 'length') {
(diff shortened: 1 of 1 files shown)


assafshemesh:
@pavelgj can you please review this PR?

MichaelDoyle:
When does this happen? After max tokens is reached? If so, I don't think we can fail here - it could be expected behavior.

I assume plain text is fine. Are you using constrained output (e.g. json). Probably needs to be handled upstream where the parsing happens.

Let me know if you want to revisit. @assafshemesh

PR #3519, opened 2025-09-08
chore(cli/mcp): added analytics for internal mcp tools

pavelgj:
Checklist (if applicable):

### DIFF
--- a/genkit-tools/cli/src/mcp/docs.ts
+++ b/genkit-tools/cli/src/mcp/docs.ts
@@ -91,7 +91,7 @@ export async function defineDocsTool(server: McpServer) {
       },
     },
     async ({ language, files }) => {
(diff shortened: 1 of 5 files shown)


genkit-tools/cli/src/mcp/docs.ts
```diff
@@ -89,6 +91,8 @@ export async function defineDocsTool(server: McpServer) {
       },
     },
     async ({ language, files }) => {
+      await record(new McpRunToolEvent('lookup_genkit_docs'));
```
apascal07, line 94: Should these be constants somewhere central?

pavelgj: I considered it... but I decided no because we might want to maintain analytics continuity for some of these events even if we rename the actual tool name.

PR #3525, opened 2025-09-08
fix(js/google-genai): fixes imports missing .js extension

mbleigh:
As far as I can tell the `@genkit-ai/google-genai` plugin is currently not functional in Next.js apps because it has directory imports and doesn't have proper extensions on its imports.

This should fix that.

### DIFF
--- a/js/pnpm-lock.yaml
+++ b/js/pnpm-lock.yaml
@@ -181,7 +181,7 @@ importers:
     optionalDependencies:
       '@genkit-ai/firebase':
         specifier: ^1.16.1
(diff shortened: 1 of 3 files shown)


ifielker (approved):

js/plugins/google-genai/src/common/converters.ts
pavelgj, line 1: I could swear that I added it to our ESM test app (that runs on presubmit)...
could you please add it to https://github.com/firebase/Genkit/blob/main/js/testapps/esm/src/index.ts

pavelgj (approved):

PR #3532, opened 2025-09-09
fix(go): Fix API test

huangjeff5:


MichaelDoyle (approved):
@huangjeff5 is this still needed? if so, let's rebase and get it in.

PR #3542, opened 2025-09-10
refactor(plugins/google-genai): migrate to v2 API

CorieW:
Resolves https://github.com/firebase/genkit/issues/3452

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/googleai/embedder.ts
+++ b/js/plugins/google-genai/src/googleai/embedder.ts
@@ -23,7 +23,7 @@ import {
   z,
 } from 'genkit';
 import { embedderRef } from 'genkit/embedder';
-import { embedder } from 'genkit/plugin';
+import { embedder as pluginEmbedder } from 'genkit/plugin';
 import { embedContent } from './client.js';
 import {
   EmbedContentRequest,
@@ -90,7 +90,7 @@ const KNOWN_MODELS = {
 };
 export type KnownModels = keyof typeof KNOWN_MODELS; // For autocomplete
 
-export function createEmbedderRef(
+export function model(
   version: string,
   config: EmbeddingConfig = {}
 ): EmbedderReference<ConfigSchemaType> {
@@ -112,7 +112,7 @@ export function listActions(models: Model[]): ActionMetadata[] {
       // Filter out deprecated
       .filter((m) => !m.description || !m.description.includes('deprecated'))
       .map((m) => {
-        const ref = createEmbedderRef(m.name);
+        const ref = model(m.name);
         return embedderActionMetadata({
           name: ref.name,
           info: ref.info,
@@ -133,9 +133,9 @@ export function defineEmbedder(
   pluginOptions?: GoogleAIPluginOptions
 ): EmbedderAction {
   checkApiKey(pluginOptions?.apiKey);
-  const ref = createEmbedderRef(name);
+  const ref = model(name);
 
-  return embedder(
+  return pluginEmbedder(
     {
       name: ref.name,
       configSchema: ref.configSchema,
--- a/js/plugins/google-genai/src/googleai/gemini.ts
+++ b/js/plugins/google-genai/src/googleai/gemini.ts
@@ -26,7 +26,7 @@ import {
   modelRef,
 } from 'genkit/model';
 import { downloadRequestMedia } from 'genkit/model/middleware';
-import { model } from 'genkit/plugin';
+import { model as pluginModel } from 'genkit/plugin';
 import { runInNewSpan } from 'genkit/tracing';
 import {
   fromGeminiCandidate,
@@ -378,7 +378,7 @@ const KNOWN_MODELS = {
   ...KNOWN_GEMMA_MODELS,
 };
 
-export function createModelRef(
+export function model(
   version: string,
   config: GeminiConfig | GeminiTtsConfig | GemmaConfig = {}
 ): ModelReference<ConfigSchemaType> {
@@ -419,7 +419,7 @@ export function listActions(models: Model[]): ActionMetadata[] {
       // Filter out deprecated
       .filter((m) => !m.description || !m.description.includes('deprecated'))
       .map((m) => {
-        const ref = createModelRef(m.name);
+        const ref = model(m.name);
         return modelActionMetadata({
           name: ref.name,
           info: ref.info,
@@ -443,7 +443,7 @@ export function defineModel(
   pluginOptions?: GoogleAIPluginOptions
 ): ModelAction {
   checkApiKey(pluginOptions?.apiKey);
-  const ref = createModelRef(name);
+  const ref = model(name);
   const clientOptions: ClientOptions = {
     apiVersion: pluginOptions?.apiVersion,
     baseUrl: pluginOptions?.baseUrl,
@@ -476,7 +476,7 @@ export function defineModel(
     );
   }
 
-  return model(
+  return pluginModel(
     {
       name: ref.name,
       ...ref.info,
--- a/js/plugins/google-genai/src/googleai/imagen.ts
+++ b/js/plugins/google-genai/src/googleai/imagen.ts
@@ -29,7 +29,7 @@ import {
   type ModelInfo,
   type ModelReference,
 } from 'genkit/model';
-import { model } from 'genkit/plugin';
+import { model as pluginModel } from 'genkit/plugin';
 import { imagenPredict } from './client.js';
 import type {
   ClientOptions,
@@ -132,7 +132,7 @@ export function isImagenModelName(value?: string): value is ImagenModelName {
   return !!value?.startsWith('imagen-');
 }
 
-export function createModelRef(
+export function model(
   version: string,
   config: ImagenConfig = {}
 ): ModelReference<ConfigSchemaType> {
@@ -160,7 +160,7 @@ export function listActions(models: Model[]): ActionMetadata[] {
     )
     .filter((m) => !m.description || !m.description.includes('deprecated'))
     .map((m) => {
-      const ref = createModelRef(m.name);
+      const ref = model(m.name);
       return modelActionMetadata({
         name: ref.name,
         info: ref.info,
@@ -180,13 +180,13 @@ export function defineModel(
   pluginOptions?: GoogleAIPluginOptions
 ): ModelAction {
   checkApiKey(pluginOptions?.apiKey);
-  const ref = createModelRef(name);
+  const ref = model(name);
   const clientOptions: ClientOptions = {
     apiVersion: pluginOptions?.apiVersion,
     baseUrl: pluginOptions?.baseUrl,
   };
 
-  return model(
+  return pluginModel(
     {
       name: ref.name,
       ...ref.info,
--- a/js/plugins/google-genai/src/googleai/index.ts
+++ b/js/plugins/google-genai/src/googleai/index.ts
@@ -161,19 +161,19 @@ export const googleAI = googleAIPlugin as GoogleAIPlugin;
   config?: any
 ): ModelReference<z.ZodTypeAny> => {
   if (veo.isVeoModelName(name)) {
-    return veo.createModelRef(name, config);
+    return veo.model(name, config);
   }
   if (imagen.isImagenModelName(name)) {
-    return imagen.createModelRef(name, config);
+    return imagen.model(name, config);
   }
   // gemma, tts, gemini and unknown model families.
-  return gemini.createModelRef(name, config);
+  return gemini.model(name, config);
 };
 googleAI.embedder = (
   name: string,
   config?: embedder.EmbeddingConfig
 ): EmbedderReference<embedder.EmbeddingConfigSchemaType> => {
-  return embedder.createEmbedderRef(name, config);
+  return embedder.model(name, config);
 };
 
 export default googleAI;
--- a/js/plugins/google-genai/src/vertexai/embedder.ts
+++ b/js/plugins/google-genai/src/vertexai/embedder.ts
@@ -21,7 +21,7 @@ import {
   type EmbedderAction,
   type EmbedderReference,
 } from 'genkit/embedder';
-import { embedder } from 'genkit/plugin';
+import { embedder as pluginEmbedder } from 'genkit/plugin';
 import { embedContent } from './client.js';
 import {
   ClientOptions,
@@ -109,7 +109,7 @@ export const KNOWN_MODELS = {
   }),
 } as const;
 
-export function createEmbedderRef(
+export function model(
   version: string,
   config: EmbeddingConfig = {}
 ): EmbedderReference<ConfigSchemaType> {
@@ -160,9 +160,9 @@ export function defineEmbedder(
   clientOptions: ClientOptions,
   _?: VertexPluginOptions
 ): EmbedderAction<any> {
-  const ref = createEmbedderRef(name);
+  const ref = model(name);
 
-  return embedder(
+  return pluginEmbedder(
     {
       name: ref.name,
       configSchema: ref.configSchema,
--- a/js/plugins/google-genai/src/vertexai/index.ts
+++ b/js/plugins/google-genai/src/vertexai/index.ts
@@ -155,22 +155,22 @@ export const vertexAI = vertexAIPlugin as VertexAIPlugin;
   config?: any
 ): ModelReference<z.ZodTypeAny> => {
   if (imagen.isImagenModelName(name)) {
-    return imagen.createModelRef(name, config);
+    return imagen.model(name, config);
   }
   if (lyria.isLyriaModelName(name)) {
-    return lyria.createModelRef(name, config);
+    return lyria.model(name, config);
   }
   if (veo.isVeoModelName(name)) {
-    return veo.createModelRef(name, config);
+    return veo.model(name, config);
   }
   // gemini and unknown model families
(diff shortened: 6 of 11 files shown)


js/plugins/google-genai/src/googleai/imagen.ts
```diff
@@ -126,13 +126,13 @@ const KNOWN_MODELS = {
 } as const;
 export type KnownModels = keyof typeof KNOWN_MODELS; // For autocomplete
 
-// For conditional types in index.ts model()
+// For conditional types in index.ts model
```
cabljac:
```suggestion
// For conditional types in index.ts createModelRef
```

maybe i think

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
-  veo.defineKnownModels(ai, options);
-}
+/**
+ * Google Gemini Developer API plugin.
+ */
+export function googleAIPlugin(
+  options?: GoogleAIPluginOptions
+): GenkitPluginV2 {
+  // Cache for list function - per plugin instance
+  let listCache: any[] | null = null;
```
cabljac: Can we avoid introducing an explicit `any`

cabljac (changes requested):
generally lgtm - one comment about an `any` we've introduced, would rather not.

js/plugins/google-genai/src/googleai/imagen.ts
CorieW: Yeah, was wondering the same thing actually

js/plugins/google-genai/src/googleai/embedder.ts
```diff
@@ -90,7 +90,7 @@ const KNOWN_MODELS = {
 };
 export type KnownModels = keyof typeof KNOWN_MODELS; // For autocomplete
 
-export function model(
+export function createModelRef(
```
pavelgj: nit
```suggestion
export function createEmbedderRef(
```

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
+        ...embedder.defineKnownModels(options),
+      ];
+    },
+    resolve: async (actionType: ActionType, name: string) => {
+      // Extract the model name without the plugin prefix
+      const modelName = name.replace(/^googleai\//, '');
+
+      if (actionType === 'model') {
+        if (veo.isVeoModelName(modelName)) {
+          return veo.defineModel(modelName, options);
```
ifielker: No. This needs to be no-op like before.

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
+      return [
+        ...imagen.defineKnownModels(options),
+        ...gemini.defineKnownModels(options),
+        ...veo.defineKnownModels(options),
+        ...embedder.defineKnownModels(options),
+      ];
+    },
+    resolve: async (actionType: ActionType, name: string) => {
+      // Extract the model name without the plugin prefix
+      const modelName = name.replace(/^googleai\//, '');
```
ifielker: Why do we need this? It was fine before?

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
+
+      if (actionType === 'model') {
```
ifielker: Case is better, why switch to multiple if statements?

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
-async function listActions(
-  options?: GoogleAIPluginOptions
-): Promise<ActionMetadata[]> {
-  try {
-    const apiKey = calculateApiKey(options?.apiKey, undefined);
-    const allModels = await listModels(apiKey, {
-      baseUrl: options?.baseUrl,
-      apiVersion: options?.apiVersion,
-    });
+      const apiKey = options?.apiKey || getApiKeyFromEnvVar();
```
ifielker: No. Use the calculate like it was before. This will do the wrong thing if the options are set to False.

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
-      if (veo.isVeoModelName(actionName)) {
-        // no-op (not gemini)
-      } else if (imagen.isImagenModelName(actionName)) {
-        imagen.defineModel(ai, actionName, options);
-      } else {
-        // gemini, tts, gemma, unknown models
-        gemini.defineModel(ai, actionName, options);
+  return genkitPluginV2({
+    name: 'googleai',
+    init: async () => {
```
ifielker: This is really hard to read now. You have 3 fully defined functions in a single structure. Please leave it readable and have the functions separate from the GenkitPluginV2 like it was before.

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
+        return [];
+      }
+    },
+  });
 }
 
 export type GoogleAIPlugin = {
-  (pluginOptions?: GoogleAIPluginOptions): GenkitPlugin;
-  model(
+  (pluginOptions?: GoogleAIPluginOptions): GenkitPluginV2;
```
ifielker, line 126: Why are we renaming this? All the samples have things like ai.generate(..., googleai.model("gemini-2.5-flash"),... ) and now it will be ai.generate(..., googleai.createModelRef("gemini-2.5-flash"), ...) which just seems like an unnecessary change for a lot more typing...? Also it will be a breaking change... Is that intentional?

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,134 +14,124 @@
...
   ): ModelReference<imagen.ImagenConfigSchemaType>;
-  model(
+  createModelRef(
     name: veo.KnownModels | (veo.VeoModelName & {}),
     config?: veo.VeoConfig
   ): ModelReference<veo.VeoConfigSchemaType>;
-  model(name: string, config?: any): ModelReference<z.ZodTypeAny>;
+  createModelRef(name: string, config?: any): ModelReference<z.ZodTypeAny>;
 
   embedder(
```
ifielker, line 149: And now this is inconsistent. If you want to create a generative model then you do googleai.createModelRef(...) and if you want to create an embedding model then you do googleai.embedder(...).

CorieW:
@ifielker 🤦‍♂️ All good points here. I've definitely overcomplicated this when I was unfamiliar with migrating.

I've made the changes, so hopefully now it should be as you'd expect. I'll also double check my other migrations to make sure I haven't done something similar with those.

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -14,69 +14,69 @@
...
-import { listModels } from './client.js';
-import { GoogleAIPluginOptions } from './types.js';
-import { calculateApiKey } from './utils.js';
+  GenkitPluginV2,
+  ResolvableAction,
+  genkitPluginV2,
+} from 'genkit/plugin';
+import type { ActionType } from 'genkit/registry';
 
 // These are namespaced because they all intentionally have
```
ifielker, line 26: This comment only pertains to the 'import * as ...' imports. Please group them together as before (separate from the other imports)  so the comment still makes sense and we have a nice obvious place to add additional models in future.

js/plugins/google-genai/src/googleai/embedder.ts
```diff
@@ -90,7 +90,7 @@ const KNOWN_MODELS = {
 };
 export type KnownModels = keyof typeof KNOWN_MODELS; // For autocomplete
 
-export function model(
+export function createEmbedderRef(
```
ifielker: No need to rename this.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -32,6 +26,7 @@ import {
   modelRef,
 } from 'genkit/model';
 import { downloadRequestMedia } from 'genkit/model/middleware';
+import { model } from 'genkit/plugin';
```
ifielker: import { model as pluginModel }  ... and then you don't have to rename the original model function anywhere.

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -104,23 +104,25 @@ async function listActions(
...
-  return genkitPlugin(
-    'googleai',
-    async (ai: Genkit) => await initializer(ai, options),
-    async (ai: Genkit, actionType: ActionType, actionName: string) =>
-      await resolver(ai, actionType, actionName, options || {}),
-    async () => {
+  return genkitPluginV2({
+    name: 'googleai',
+    init: async () => await initializer(options),
+    resolve: async (actionType, actionName) =>
```
ifielker: Please put the types on the inputs as before (actionType: ActionType, actionName: string)

js/plugins/google-genai/src/vertexai/embedder.ts
```diff
@@ -107,8 +108,15 @@ export const KNOWN_MODELS = {
     supports: { input: ['text'] },
   }),
 } as const;
+export type KnownModels = keyof typeof KNOWN_MODELS;
+export type EmbedderModelName = `embedder=${string}`;
```
ifielker: 1) EmbedderModelName is wrong in multiple ways. a) Not all embedders start with 'embedder' and b) None of the model names contain an equal sign.
2) Why do we need this section? We didn't need it before?

js/plugins/google-genai/src/vertexai/embedder.ts
```diff
@@ -107,8 +108,15 @@ export const KNOWN_MODELS = {
...
+export function isEmbedderModelName(
+  value?: string
+): value is EmbedderModelName {
+  return !!value?.includes('embedding');
```
ifielker: And even if it was 'embedder=' for the embedder prefix... (it's not) this doesn't match that either...  Where is this coming from?

js/plugins/google-genai/src/vertexai/embedder.ts
```diff
@@ -146,35 +154,35 @@ export function model(
...
 
-  return ai.defineEmbedder(
+  return embedder(
     {
       name: ref.name,
       configSchema: ref.configSchema,
       info: ref.info!,
     },
-    async (input, options?: EmbeddingConfig) => {
+    async (request, options) => {
```
ifielker: What is the options parameter for? It looks like it's unused?

js/plugins/google-genai/src/vertexai/index.ts
```diff
@@ -20,17 +20,17 @@
  * @module /
  */
 
-import { EmbedderReference, Genkit, ModelReference, z } from 'genkit';
-import { GenkitPlugin, genkitPlugin } from 'genkit/plugin';
-import { ActionType } from 'genkit/registry';
-import { listModels } from './client.js';
+import { EmbedderReference, ModelReference, z } from 'genkit';
```
ifielker, line 23: There are 3 groups of imports:

1) things imported from non-local files (that don't end in .js) 
2) model imports: "import * as ... "
3) things imported from local files (that end in .js)

Please keep them organized the same way.

js/plugins/google-genai/src/vertexai/index.ts
```diff
@@ -41,56 +41,43 @@ export { type LyriaConfig } from './lyria.js';
...
+async function initializer(pluginOptions?: VertexPluginOptions) {
   const clientOptions = await getDerivedOptions(pluginOptions);
-  veo.defineKnownModels(ai, clientOptions, pluginOptions);
-  imagen.defineKnownModels(ai, clientOptions, pluginOptions);
-  lyria.defineKnownModels(ai, clientOptions, pluginOptions);
-  gemini.defineKnownModels(ai, clientOptions, pluginOptions);
-  embedder.defineKnownModels(ai, clientOptions, pluginOptions);
+
+  return [
+    ...imagen.defineKnownModels(clientOptions, pluginOptions),
```
ifielker, line 50: Why did you remove veo and lyria models? We need those.

js/plugins/google-genai/src/vertexai/index.ts
```diff
@@ -41,56 +41,43 @@ export { type LyriaConfig } from './lyria.js';
...
 async function resolver(
-  ai: Genkit,
   actionType: ActionType,
   actionName: string,
   pluginOptions?: VertexPluginOptions
 ) {
   const clientOptions = await getDerivedOptions(pluginOptions);
   switch (actionType) {
     case 'model':
-      if (lyria.isLyriaModelName(actionName)) {
```
ifielker, line 62: What is happening here? Why is all this removed?

js/plugins/google-genai/src/vertexai/index.ts
```diff
@@ -41,56 +41,43 @@ export { type LyriaConfig } from './lyria.js';
...
-      break;
+      return embedder.defineEmbedder(actionName, clientOptions, pluginOptions);
     default:
-    // no-op
+      return undefined;
   }
 }
 
 async function listActions(options?: VertexPluginOptions) {
+  const clientOptions = await getDerivedOptions(options);
```
ifielker: Why are you changing the logic? This was correct before, now it will incorrectly throw errors... put it back.

js/plugins/google-genai/src/vertexai/index.ts
```diff
@@ -41,56 +41,43 @@ export { type LyriaConfig } from './lyria.js';
...
   try {
-    const clientOptions = await getDerivedOptions(options);
     const models = await listModels(clientOptions);
     return [
       ...gemini.listActions(models),
       ...imagen.listActions(models),
-      ...lyria.listActions(models),
```
ifielker, line 92: Randomly removing things is not ok.

ifielker (changes requested):
I stopped my review partway through again. I'll pick up where I left off once you address these comments.

js/plugins/google-genai/src/googleai/index.ts
CorieW: ah, good spot

js/plugins/google-genai/src/vertexai/index.ts
CorieW: My mistake. Used a prior history of GoogleAI's index.ts when making the adjustments, then switched to VertexAI's index.ts without changing branch, so I've copied outdated code and adjusted it. I'll update the code.

js/plugins/google-genai/src/vertexai/index.ts
CorieW: Ah, I see. Which is your preference:
1. Keep them the same as [googleai index.ts](https://github.com/firebase/genkit/blob/efa743a4fd439a74344fe539444ed51f98ec596c/js/plugins/google-genai/src/googleai/index.ts#L27) and [vertexai index.ts](https://github.com/firebase/genkit/blob/efa743a4fd439a74344fe539444ed51f98ec596c/js/plugins/google-genai/src/vertexai/index.ts#L26).
2. Organise them as you've suggested.

js/plugins/google-genai/src/googleai/embedder.ts
CorieW: Just basing it on https://github.com/firebase/genkit/pull/3542#discussion_r2373308603. Also, it seems more obvious to me that it would follow the same pattern used everywhere else in the code. I am aware that you've suggested importing `model` as `pluginModel` to avoid these name changes, but doesn't the name `createModelRef`/`createEmbedderRef` provide a clearer description of what the function does?

js/plugins/google-genai/src/vertexai/embedder.ts
CorieW: Was meant to get rid of it earlier when getting rid of the other overcomplicated code, but missed. Good spot.

js/plugins/google-genai/src/vertexai/embedder.ts
CorieW: Was meant to get rid of it earlier when getting rid of the other overcomplicated code, but missed. Good spot.

js/plugins/google-genai/src/googleai/embedder.ts
cabljac: Hey @CorieW let's pair up on this and review today

js/plugins/google-genai/src/vertexai/index.ts
CorieW: I've updated the code and organised the imports as you've outlined.

js/plugins/google-genai/src/googleai/embedder.ts
cabljac: Hey @ifielker we're happy to revert to the original internal name `model` here if you prefer.

Originally I think the change to add the `Ref` suffix came from looking at the `compat-openai` plugin example (although that naming doesn't use `create` in it either actually, i think it uses e.g `openAIModelRef` etc for factories like this.)

cabljac: the public API has `model` return a reference so also it'd be good to have consistency and keep it as `model`

cabljac:
Some tests seem to need fixing, investigating this now.

CorieW:
Closing in favour of https://github.com/firebase/genkit/pull/3649

PR #3545, opened 2025-09-11
Fix(go) prompt.excute merge external incoming messages

### LINKED ISSUE #3544: [Go]prompt.excute did not handle the additional message
**Describe the bug**
When I use prompt.excute, if I append a message, it won't be rendered and sent to the large model when it is actually sent。

**Screenshots**
Image

eric642:
Fixes #3544

### DIFF
--- a/go/ai/prompt.go
+++ b/go/ai/prompt.go
@@ -176,7 +176,7 @@ func (p *prompt) Execute(ctx context.Context, opts ...PromptExecuteOption) (*Mod
 		if err != nil {
 			return nil, err
 		}
-		actionOpts.Messages = append(actionOpts.Messages, msg...)
+		actionOpts.Messages = msg
 	}
 
 	return GenerateWithRequest(ctx, p.registry, actionOpts, execOpts.Middleware, execOpts.Stream)


go/ai/prompt.go
```diff
@@ -136,10 +136,15 @@ func (p *prompt) Execute(ctx context.Context, opts ...PromptExecuteOption) (*Mod
 
 	p.MessagesFn = mergeMessagesFn(p.MessagesFn, execOpts.MessagesFn)
 
+	input, err := buildVariables(execOpts.Input)
```
apascal07, line 139: I don't quite see why this is necessary. Without this, we pipe through `execOpts.Input` to `buildRequest()` which does this as a first step.

go/ai/prompt.go
```diff
@@ -136,10 +136,15 @@ func (p *prompt) Execute(ctx context.Context, opts ...PromptExecuteOption) (*Mod
 
 	p.MessagesFn = mergeMessagesFn(p.MessagesFn, execOpts.MessagesFn)
```
apascal07, line 137: We want to delete this line since it'll double up the messages from the prompt when we render the messages at the bottom.

apascal07:
Fix here: https://github.com/firebase/genkit/pull/3565

PR #3547, opened 2025-09-11
feat(js/plugins/ollama): migrate ollama plugin to v2 plugin API

### LINKED ISSUE #3456: refactor: migrate ollama plugin to V2 Plugin API
This is a sub-issue to track migration of the ollama js plugin to V2.

HassanBahati:
Fixes https://github.com/firebase/genkit/issues/3456

[Plugin migration guide](https://docs.google.com/document/d/1s1YKwhqdrGN44IG7lSAwltxM3Bqh9p6jTcri8SNaBWk/edit?usp=sharing)

**Notable Changes**
- added interface `ResolveActionOptions` for cleaner destructuring of resolveActions args
- added a `OllamaEmbedderConfig` schema so that serverAddress comes through on request options.
- used OllamaConfigSchema in model<> for better type safety
- added several test suites for better coverage
- Updated and added integration tests with the core genkit framework, for streaming responses and tool calls.
- All existing plugin tests pass
 
**Testing**
  - [x] Checked that the models and embedders load with the same names/prefixes on the UI.
  - [x] Checked that at least one model works consistent with pre-migration on the UI.
image

  - [x] Using a model string (`'pluginName/modelName'`), checked using a flow works consistent with pre-migration.
image

  - [x] Tested that the trace on flows are the same.
image

  - [x] Tested error response from model.
image

  - [x] Tested `ollama().resolve(...)`.
image

- [x] Tested Embeddings functionality - Tested RAG flow with Pokemon data using `nomic-embed-text:latest` embedder   
                     - 768-dimensional embeddings generated correctly
                     - Embedder resolution works with `ollama/nomic-embed-text:latest` format
                     - RAG similarity search functions properly

image

 - [x] Tested direct plugin usage - Verified `ollama().model()` and `ollama().embedder()` work correctly
    - `ollama().model('phi3.5:latest')` resolves properly
    - `ollama().embedder('nomic-embed-text:latest')` resolves properly
    - Embedder registry shows: `/embedder/ollama/nomic-embed-text:latest`

### DIFF
--- a/js/plugins/ollama/package.json
+++ b/js/plugins/ollama/package.json
@@ -18,7 +18,7 @@
     "build:clean": "rimraf ./lib",
     "build": "npm-run-all build:clean check compile",
     "build:watch": "tsup-node --watch",
-    "test": "find tests -name 'model_test.ts' ! -name '*_live_test.ts' -exec node --import tsx --test {} +",
+    "test": "find tests -name '*_test.ts' ! -name '*_live_test.ts' -exec node --import tsx --test {} +",
     "test:live": "node --import tsx --test tests/*_test.ts"
   },
   "repository": {
--- a/js/plugins/ollama/package.json
+++ b/js/plugins/ollama/package.json
@@ -18,7 +18,7 @@
     "build:clean": "rimraf ./lib",
     "build": "npm-run-all build:clean check compile",
     "build:watch": "tsup-node --watch",
-    "test": "find tests -name '*_test.ts' ! -name '*_live_test.ts' -exec node --import tsx --test {} +",
+    "test": "find tests -name 'model_test.ts' ! -name '*_live_test.ts' -exec node --import tsx --test {} +",
     "test:live": "node --import tsx --test tests/*_test.ts"
   },
   "repository": {
--- a/js/plugins/ollama/src/embeddings.ts
+++ b/js/plugins/ollama/src/embeddings.ts
@@ -80,9 +80,6 @@ export function defineOllamaEmbedder({
       },
     },
     async (request, config) => {
-      console.log('request.options', request.options);
-      // request.options; might be the equivalent of config now
-
       const serverAddress =
         options.serverAddress || DEFAULT_OLLAMA_SERVER_ADDRESS;
 
--- a/js/plugins/ollama/src/embeddings.ts
+++ b/js/plugins/ollama/src/embeddings.ts
@@ -18,6 +18,7 @@ import { embedder } from 'genkit/plugin';
 import type { EmbedRequest, EmbedResponse } from 'ollama';
 import { DEFAULT_OLLAMA_SERVER_ADDRESS } from './constants.js';
 import type { DefineOllamaEmbeddingParams, RequestHeaders } from './types.js';
+import { OllamaEmbedderConfigSchema } from './types.js';
 
 export async function toOllamaEmbedRequest(
   modelName: string,
@@ -66,10 +67,13 @@ export function defineOllamaEmbedder({
   modelName,
   dimensions,
   options,
-}: DefineOllamaEmbeddingParams): EmbedderAction<any> {
+}: DefineOllamaEmbeddingParams): EmbedderAction<
+  typeof OllamaEmbedderConfigSchema
+> {
   return embedder(
     {
       name: `ollama/${name}`,
+      configSchema: OllamaEmbedderConfigSchema,
       info: {
         label: 'Ollama Embedding - ' + name,
         dimensions,
@@ -79,14 +83,15 @@ export function defineOllamaEmbedder({
         },
       },
     },
-    async (request, config) => {
+    async ({ input, options: requestOptions }, config) => {
       const serverAddress =
-        options.serverAddress || DEFAULT_OLLAMA_SERVER_ADDRESS;
-
+        requestOptions?.serverAddress ||
+        options.serverAddress ||
+        DEFAULT_OLLAMA_SERVER_ADDRESS;
       const { url, requestPayload, headers } = await toOllamaEmbedRequest(
         modelName,
         dimensions,
-        request.input,
+        input,
         serverAddress,
         options.requestHeaders
       );
--- a/js/plugins/ollama/src/embeddings.ts
+++ b/js/plugins/ollama/src/embeddings.ts
@@ -13,11 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { Document, EmbedderAction, Genkit } from 'genkit';
+import { embedder } from 'genkit/plugin';
+import type { Document, EmbedderAction } from 'genkit';
 import type { EmbedRequest, EmbedResponse } from 'ollama';
 import type { DefineOllamaEmbeddingParams, RequestHeaders } from './types.js';
 
-async function toOllamaEmbedRequest(
+export async function toOllamaEmbedRequest(
   modelName: string,
   dimensions: number,
   documents: Document[],
@@ -60,10 +61,9 @@ async function toOllamaEmbedRequest(
 }
 
 export function defineOllamaEmbedder(
-  ai: Genkit,
   { name, modelName, dimensions, options }: DefineOllamaEmbeddingParams
 ): EmbedderAction<any> {
-  return ai.defineEmbedder(
+  return embedder(
     {
       name: `ollama/${name}`,
       info: {
@@ -76,12 +76,12 @@ export function defineOllamaEmbedder(
       },
     },
     async (input, config) => {
-      const serverAddress = config?.serverAddress || options.serverAddress;
+      const serverAddress = options.serverAddress || 'http://localhost:11434';
 
       const { url, requestPayload, headers } = await toOllamaEmbedRequest(
         modelName,
         dimensions,
-        input,
+        input.input,
         serverAddress,
         options.requestHeaders
       );
--- a/js/plugins/ollama/src/embeddings.ts
+++ b/js/plugins/ollama/src/embeddings.ts
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { embedder } from 'genkit/plugin';
 import type { Document, EmbedderAction } from 'genkit';
+import { embedder } from 'genkit/plugin';
 import type { EmbedRequest, EmbedResponse } from 'ollama';
 import type { DefineOllamaEmbeddingParams, RequestHeaders } from './types.js';
 
@@ -60,9 +60,12 @@ export async function toOllamaEmbedRequest(
   };
 }
 
-export function defineOllamaEmbedder(
-  { name, modelName, dimensions, options }: DefineOllamaEmbeddingParams
-): EmbedderAction<any> {
+export function defineOllamaEmbedder({
+  name,
+  modelName,
+  dimensions,
+  options,
+}: DefineOllamaEmbeddingParams): EmbedderAction<any> {
   return embedder(
     {
       name: `ollama/${name}`,
--- a/js/plugins/ollama/src/embeddings.ts
+++ b/js/plugins/ollama/src/embeddings.ts
@@ -78,13 +78,13 @@ export function defineOllamaEmbedder({
         },
       },
     },
-    async (input, config) => {
+    async (request, config) => {
       const serverAddress = options.serverAddress || 'http://localhost:11434';
 
       const { url, requestPayload, headers } = await toOllamaEmbedRequest(
         modelName,
         dimensions,
-        input.input,
+        request.input,
         serverAddress,
         options.requestHeaders
       );
--- a/js/plugins/ollama/src/embeddings.ts
+++ b/js/plugins/ollama/src/embeddings.ts
@@ -16,6 +16,7 @@
 import type { Document, EmbedderAction } from 'genkit';
 import { embedder } from 'genkit/plugin';
 import type { EmbedRequest, EmbedResponse } from 'ollama';
+import { DEFAULT_OLLAMA_SERVER_ADDRESS } from './constants.js';
 import type { DefineOllamaEmbeddingParams, RequestHeaders } from './types.js';
 
 export async function toOllamaEmbedRequest(
@@ -79,7 +80,11 @@ export function defineOllamaEmbedder({
       },
     },
     async (request, config) => {
-      const serverAddress = options.serverAddress || 'http://localhost:11434';
+      console.log('request.options', request.options);
+      // request.options; might be the equivalent of config now
+
+      const serverAddress =
+        options.serverAddress || DEFAULT_OLLAMA_SERVER_ADDRESS;
 
       const { url, requestPayload, headers } = await toOllamaEmbedRequest(
         modelName,
--- a/js/plugins/ollama/src/index.ts
+++ b/js/plugins/ollama/src/index.ts
@@ -84,7 +84,7 @@ function initializer(serverAddress: string, params: OllamaPluginParams = {}) {
     }
   }
 
-  if (params?.embedders && params.serverAddress) {
+  if (params?.embedders) {
     for (const embedder of params.embedders) {
       actions.push(
         defineOllamaEmbedder({
--- a/js/plugins/ollama/src/index.ts
+++ b/js/plugins/ollama/src/index.ts
@@ -34,7 +34,6 @@ import {
   type GenerateRequest,
   type GenerateResponseData,
   type MessageData,
-  type ModelInfo,
   type ToolDefinition,
 } from 'genkit/model';
 import {
@@ -43,6 +42,11 @@ import {
   type GenkitPluginV2,
   type ResolvableAction,
 } from 'genkit/plugin';
+import {
+  ANY_JSON_SCHEMA,
+  DEFAULT_OLLAMA_SERVER_ADDRESS,
+  GENERIC_MODEL_INFO,
+} from './constants.js';
 import { defineOllamaEmbedder } from './embeddings.js';
 import type {
   ApiType,
@@ -67,23 +71,6 @@ export type OllamaPlugin = {
   ): ModelReference<typeof OllamaConfigSchema>;
   embedder(name: string, config?: Record<string, any>): EmbedderReference;
 };
-
-const ANY_JSON_SCHEMA: Record<string, any> = {
-  $schema: 'http://json-schema.org/draft-07/schema#',
-};
-
-const GENERIC_MODEL_INFO = {
-  supports: {
-    multiturn: true,
-    media: true,
-    tools: true,
-    toolChoice: true,
-    systemRole: true,
-    constrained: 'all',
-  },
-} as ModelInfo;
-
-const DEFAULT_OLLAMA_SERVER_ADDRESS = 'http://localhost:11434';
 
 async function listActions(
   serverAddress: string,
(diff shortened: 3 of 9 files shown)


js/plugins/ollama/src/index.ts
```diff
@@ -82,159 +86,23 @@ const GENERIC_MODEL_INFO = {
...
       supports: {
-        multiturn: !model.type || model.type === 'chat',
+        multiturn: !modelDef.type || modelDef.type === 'chat',
         systemRole: true,
-        tools: model.supports?.tools,
+        tools: modelDef.supports?.tools,
       },
     },
-    async (input, streamingCallback) => {
+    async (input, opts: any) => {
```
cabljac: why `any` ?

js/plugins/ollama/src/index.ts
```diff
@@ -336,6 +203,223 @@ function defineOllamaModel(
...
+        'Content-Type': 'application/json',
+        ...(await getHeaders(serverAddress, requestHeaders)),
+      },
+    });
+  } catch (e) {
+    throw new Error(`Make sure the Ollama server is running.`, {
+      cause: e,
+    });
+  }
+  const modelResponse = JSON.parse(await res.text()) as ListLocalModelsResponse;
```
cabljac: we might be able to just do `await res.json()` here

js/plugins/ollama/src/index.ts
```diff
@@ -250,20 +118,20 @@ function defineOllamaModel(
...
-      const type = model.type ?? 'chat';
+      const type = modelDef.type ?? 'chat';
       const request = toOllamaRequest(
-        model.name,
+        modelDef.name,
         input,
         options,
         type,
-        !!streamingCallback
+        !!opts.sendChunk
```
cabljac: is it clear where sendChunk comes from? Is this defined in a genkit type?

js/plugins/ollama/src/index.ts
```diff
@@ -336,6 +203,223 @@ function defineOllamaModel(
   );
 }
 
+async function createOllamaEmbedder(
```
cabljac: what happens to:

https://github.com/firebase/genkit/blob/main/js/plugins/ollama/src/embeddings.ts

I'd expect changes there?

js/plugins/ollama/src/index.ts
```diff
@@ -82,159 +86,23 @@ const GENERIC_MODEL_INFO = {
...
-    .describe(
-      GenerationCommonConfigDescriptions.topP + ' The default value is 0.9.'
-    )
-    .optional(),
-});
-
-function defineOllamaModel(
-  ai: Genkit,
-  model: ModelDefinition,
+async function createOllamaModel(
```
cabljac: does it definitely need to be async?

js/plugins/ollama/tests/model_test.ts
```diff
@@ -79,24 +85,12 @@ describe('ollama models', () => {
...
-        name: 'get_current_weather',
-        description: 'gets weather',
-        inputSchema: z.object({ format: z.string(), location: z.string() }),
-      },
-      async () => {
-        return MAGIC_WORD;
-      }
-    );
-
+  it('should successfully return basic response', async () => {
```
cabljac, line 185: are tools no longer supported in v2 or something? seems weird to change this test?

js/plugins/ollama/src/index.ts
HassanBahati: this has been update

js/plugins/ollama/tests/model_test.ts
HassanBahati: tools are still supported. though the approach to handling basic responses versus tool responses is a bit different. so added an extra test for basic responses

js/plugins/ollama/src/index.ts
HassanBahati: not necessarily. updated it

js/plugins/ollama/src/embeddings.ts
```diff
@@ -76,12 +79,12 @@ export function defineOllamaEmbedder(
       },
     },
     async (input, config) => {
-      const serverAddress = config?.serverAddress || options.serverAddress;
```
HassanBahati, line 79: in the v2 plugin system, the config parameter in the embedder action handler doesn't seem to have a serverAddress property

cabljac: your change looks correct to me, in the new API i think config is `runtime` config. they probably shouldnt pass the address in every embed call

js/plugins/ollama/src/embeddings.ts
```diff
@@ -76,12 +79,12 @@ export function defineOllamaEmbedder(
...
+      const serverAddress = options.serverAddress || 'http://localhost:11434';
```
CorieW: Where does that address come from? Is there no better alternatives than having this?

js/plugins/ollama/src/embeddings.ts
```diff
@@ -76,12 +79,12 @@ export function defineOllamaEmbedder(
       },
     },
     async (input, config) => {
```
CorieW: Rename `input` to `request`, as I believe this better reflects the params intentions here.

js/plugins/ollama/src/index.ts
```diff
@@ -82,161 +85,25 @@ const GENERIC_MODEL_INFO = {
 
 const DEFAULT_OLLAMA_SERVER_ADDRESS = 'http://localhost:11434';
 
-async function initializer(
```
CorieW, line 85: Perhaps keep this code in the same position, just so the PR is not as large.

js/plugins/ollama/src/embeddings.ts
cabljac: We're no longer getting serverAddress from config, because (as far as i understand) it doesn't exist on that interface any more in v2 API.

Can we mention this in the PR description

js/plugins/ollama/tests/embeddings_test.ts
```diff
@@ -41,50 +41,56 @@ global.fetch = async (input: RequestInfo | URL, options?: RequestInit) => {
...
-  const options: OllamaPluginParams = {
-    models: [{ name: 'test-model' }],
-    serverAddress: 'http://localhost:3000',
-  };
+const options: OllamaPluginParams = {
+  models: [{ name: 'test-model' }],
+  serverAddress: 'http://localhost:3000',
+};
+
+// TODO: also have tests that do not need initializing genkit
```
cabljac: @HassanBahati

js/plugins/ollama/tests/embeddings_test.ts
```diff
@@ -41,50 +41,56 @@ global.fetch = async (input: RequestInfo | URL, options?: RequestInit) => {
...
-          serverAddress: 'http://localhost:3000',
-        }),
-      ],
+      plugins: [ollama(options)],
     });
   });
 
-  it('should successfully return embeddings', async () => {
-    const embedder = defineOllamaEmbedder(ai, {
+  it.only('should successfully return embeddings', async () => {
```
cabljac: (TODO: remove only)

js/plugins/ollama/tests/model_test.ts
```diff
@@ -79,6 +84,14 @@ describe('ollama models', () => {
...
 
+  it('should successfully return basic response', async () => {
+    const result = await ai.generate({
+      model: 'ollama/test-model',
+      prompt: 'Hello',
+    });
+    assert.ok(result.text === 'The weather is sunny');
+  });
+
   it('should successfully return tool call response', async () => {
```
cabljac, line 193: let's figure out why this test isn't passing if it wasn't, instead of changing the test

js/plugins/ollama/src/index.ts
```diff
@@ -329,13 +253,69 @@ function defineOllamaModel(
...
+function ollamaPlugin(params?: OllamaPluginParams): GenkitPluginV2 {
+  if (!params) {
+    params = {};
+  }
+  if (!params.serverAddress) {
+    params.serverAddress = DEFAULT_OLLAMA_SERVER_ADDRESS;
+  }
+  const serverAddress = params.serverAddress;
+
+  return genkitPluginV2({
```
CorieW: Improve readability of this code by extracting the init and resolve anonymous functions into their own functions outside of the `ollamaPlugin` function.

js/plugins/ollama/src/embeddings.ts
```diff
@@ -59,11 +61,13 @@ async function toOllamaEmbedRequest(
...
-  return ai.defineEmbedder(
+export function defineOllamaEmbedder({
+  name,
+  modelName,
+  dimensions,
+  options,
+}: DefineOllamaEmbeddingParams): EmbedderAction<any> {
+  return embedder(
     {
       name: `ollama/${name}`,
```
CorieW, line 75: replace prefixes in `name` with `namespace` property (this goes for all embedders, models, etc...)

cabljac: I think this is just for references

CorieW: nvm, thought this was an embedder ref

js/plugins/ollama/package.json
```diff
@@ -18,7 +18,7 @@
     "build:clean": "rimraf ./lib",
     "build": "npm-run-all build:clean check compile",
     "build:watch": "tsup-node --watch",
-    "test": "find tests -name '*_test.ts' ! -name '*_live_test.ts' -exec node --import tsx --test {} +",
+    "test": "find tests -name 'model_test.ts' ! -name '*_live_test.ts' -exec node --import tsx --test {} +",
```
dackers86: Checking if we meant to switch off the glob testing for this package?

js/plugins/ollama/src/index.ts
cabljac: hmm could do, arguable whether it'd be much more readable, they're not very long functions. There are other functions i'd extract out of index.ts before this to be honest, it's quite long and full of functions

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,60 @@ export type OllamaPlugin = {
...
-      {
-        name: actionName,
-      },
-      serverAddress,
-      requestHeaders
-    );
-  }
+      return actions;
+    },
+    async resolve(actionType, actionName) {
```
dackers86: Is there a reason this is called actionName and not name?

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,60 @@ export type OllamaPlugin = {
   embedder(name: string, config?: Record<string, any>): EmbedderReference;
 };
 
-const ANY_JSON_SCHEMA: Record<string, any> = {
-  $schema: 'http://json-schema.org/draft-07/schema#',
-};
+function ollamaPlugin(params?: OllamaPluginParams): GenkitPluginV2 {
+  if (!params) {
+    params = {};
+  }
```
dackers86:
```suggestion
function ollamaPlugin(params?: OllamaPluginParams = {}): GenkitPluginV2 {
```

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,60 @@ export type OllamaPlugin = {
...
+  if (!params.serverAddress) {
+    params.serverAddress = DEFAULT_OLLAMA_SERVER_ADDRESS;
+  }
+  const serverAddress = params.serverAddress;
```
dackers86:
```suggestion
  const serverAddress = params.serverAddress ||  DEFAULT_OLLAMA_SERVER_ADDRESS;
```

js/plugins/ollama/package.json
cabljac:
```suggestion
    "test": "find tests -name '*_test.ts' ! -name '*_live_test.ts' -exec node --import tsx --test {} +",
```

js/plugins/ollama/src/index.ts
cabljac: We should factor these out into helpers (like in original)

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,60 @@ export type OllamaPlugin = {
...
-    constrained: 'all',
-  },
-} as ModelInfo;
+  return genkitPluginV2({
+    name: 'ollama',
+    init() {
+      const actions: ResolvableAction[] = [];
 
-const DEFAULT_OLLAMA_SERVER_ADDRESS = 'http://localhost:11434';
+      if (params?.models) {
```
dackers86: Could this be extracted into a helper function?

```js
function getModelActions(params: OllamaPluginParams, serverAddress: string): ResolvableAction[] {
  /** Extract variables **/
  const { models, requestHeaders } = params;

  /** If no models, return empty array; **/
  if (!models || !models.length) return [];
  
 /** Return Ollama models **/
  return models.map(m => createOllamaModel(m, serverAddress, requestHeaders));
}
```

```js
init() {
  return [
    ...getModelActions(params, serverAddress),
    ...getEmbedderActions(params, serverAddress)
  ];
}
```

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,60 @@ export type OllamaPlugin = {
...
+      if (actionType === 'model') {
+        return await createOllamaModel(
+          {
+            name: actionName,
+          },
+          serverAddress,
+          params?.requestHeaders
+        );
+      }
+      return undefined;
```
cabljac: We probably should handle the case where someone tries to resolve an embedder?

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,60 @@ export type OllamaPlugin = {
...
+    },
```
dackers86: Is there a reason why we removed the resolveAction function? The original resolveActionWe can return early if undefined. Also should we return undefined here.

```suggestion
async resolve(actionType, actionName) {
  /** If no model actions, return undefined **/
  if (actionType !== 'model') return undefined;
  
  return await createOllamaModel(
    {
      name: actionName,
    },
    serverAddress,
    params?.requestHeaders
  );
}
```

js/plugins/ollama/src/embeddings.ts
```diff
@@ -75,13 +79,14 @@ export function defineOllamaEmbedder(
         },
       },
     },
-    async (input, config) => {
-      const serverAddress = config?.serverAddress || options.serverAddress;
+    async (request, config) => {
+      const serverAddress =
+        options.serverAddress || DEFAULT_OLLAMA_SERVER_ADDRESS;
```
cabljac: need to double check this, make sure we're not breaking anything by removing config.serverAddress

js/plugins/ollama/src/index.ts
cabljac: instead of just returning undefined

js/plugins/ollama/src/index.ts
```diff
@@ -131,39 +133,13 @@ async function listActions(
       ?.filter((m) => m.model && !m.model.includes('embed'))
       .map((m) =>
         modelActionMetadata({
-          name: `ollama/${m.model}`,
+          name: m.model,
```
cabljac: double check on prefixes

js/plugins/ollama/src/index.ts
cabljac: I believe this follows the compat-oai example here, though "name" is used in migration guide, so we're not sure which is preferred

https://github.com/firebase/genkit/blob/main/js/plugins/compat-oai/src/index.ts

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,57 @@ export type OllamaPlugin = {
...
+  }
+
+  if (params?.embedders && params.serverAddress) {
+    for (const embedder of params.embedders) {
+      actions.push(
+        defineOllamaEmbedder({
+          name: embedder.name,
+          modelName: embedder.name,
+          dimensions: embedder.dimensions,
+          options: { ...params, serverAddress },
```
CorieW: do we need `serverAddress` here? It wasn't there before by the looks of it

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,57 @@ export type OllamaPlugin = {
...
-      requestHeaders
-    );
+interface ResolveActionOptions {
+  params: OllamaPluginParams;
+  actionType: string;
+  actionName: string;
+  serverAddress: string;
+}
+
+function resolveAction({
```
CorieW, line 103: Can't we keep the params as they were instead of defining `ResolveActionOptions`?

js/plugins/ollama/src/index.ts
```diff
@@ -218,26 +214,25 @@ export const OllamaConfigSchema = GenerationCommonConfigSchema.extend({
...
+  modelDef: ModelDefinition,
   serverAddress: string,
   requestHeaders?: RequestHeaders
 ) {
-  return ai.defineModel(
+  return model<typeof OllamaConfigSchema>(
     {
-      name: `ollama/${model.name}`,
-      label: `Ollama - ${model.name}`,
+      name: modelDef.name,
```
CorieW: May need prefix

js/plugins/ollama/src/index.ts
```diff
@@ -218,26 +214,25 @@ export const OllamaConfigSchema = GenerationCommonConfigSchema.extend({
...
+        tools: modelDef.supports?.tools,
       },
     },
-    async (input, streamingCallback) => {
+    async (request, opts) => {
       const { topP, topK, stopSequences, maxOutputTokens, ...rest } =
-        input.config as any;
-      const options: Record<string, any> = { ...rest };
+        request.config || {};
+      const options = { ...rest };
```
CorieW, line 229: Type of `Record<string, any>` was removed, was this intentional?

js/plugins/ollama/tests/list_test.ts
```diff
@@ -0,0 +1,142 @@
+/**
+ * Copyright 2024 Google LLC
```
dackers86:
```suggestion
 * Copyright 2025 Google LLC
```

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +72,57 @@ export type OllamaPlugin = {
...
-  if (actionType === 'model') {
-    defineOllamaModel(
-      ai,
-      {
-        name: actionName,
-      },
-      serverAddress,
-      requestHeaders
-    );
+interface ResolveActionOptions {
```
dackers86: Does this belong on this types file?

js/plugins/ollama/src/index.ts
cabljac: yeah I believe it comes through typed now, but will double check

js/plugins/ollama/src/index.ts
cabljac: It could do, was balancing keeping it next to the function. wdyt?

js/plugins/ollama/src/index.ts
cabljac: We could do, just having 4 unnamed params was bugging me, I thought this would be cleaner

js/plugins/ollama/src/index.ts
cabljac: Yeah now we've added the config schema, it infers its type.

js/plugins/ollama/src/index.ts
cabljac: I think i'll revert making this type, see discussion below

js/plugins/ollama/src/index.ts
cabljac: I'm torn. I think it's nicer having an interface because three of the args are strings, stops someone putting it in the wrong order

cabljac (approved):
LGTM!

js/plugins/ollama/src/index.ts
```diff
@@ -65,59 +73,50 @@ export type OllamaPlugin = {
...
-  );
+  if (params?.models) {
+    for (const model of params.models) {
+      actions.push(
+        defineOllamaModel(model, serverAddress, params.requestHeaders)
+      );
+    }
+  }
+
+  if (params?.embedders && params.serverAddress) {
```
ifielker: Why are you adding the requirement for params.serverAddress to be defined before you define the embedders? You have a default serverAddress...?

cabljac: ah good catch, will push a fix.

MichaelDoyle:
@ifielker @HassanBahati @cabljac what's the current status here? still needed?

PR #3548, opened 2025-09-11
fix(go): fixes to AIM telemetry

huangjeff5:
Addresses feedback here: https://docs.google.com/spreadsheets/d/1yYuyBLzA6S9blsGv05C-YKJSqyZsfMjh8IX7SW4RbTY/edit?gid=0#gid=0

Mainly:
1) Move the generate span generation to the ai.Generate() method - the Generation Action already gets a span from the action framework so there's no need to add it there, and ai.Generate() DOES need the extra span since it's not an action. This is the cause of most of the weirdness
2) Add support for GCLOUD_PROJECT (same as genkit ts)
3) Make firebase plugin match googlecloud plugin in terms of options (pass thru directly)
4) Handle credential detection automatically
5) Make sure exact same logs as genkit ts are being emitted by ensuring that logic for which telemetry module to call is exactly the same. Also update the user engagement tests to reflect reality; since user engagement spans have type="userEngagement"

Tested manually, some screenshots:
unnamed
Screenshot 2025-09-11 at 11 25 19 AM

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -340,13 +340,7 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
 				}
 			}
 
-			toolCount := 0
-			for _, part := range resp.Message.Content {
-				if part.IsToolRequest() {
-					toolCount++
-				}
-			}
-			if toolCount == 0 || opts.ReturnToolRequests {
+			if len(resp.ToolRequests()) > 0 || opts.ReturnToolRequests {
 				return resp, nil
 			}
 
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -316,52 +316,63 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
 
 	fn := core.ChainMiddleware(mw...)(m.Generate)
 
-	currentTurn := 0
-	for {
-		resp, err := fn(ctx, req, cb)
-		if err != nil {
-			return nil, err
+	// Inline recursive helper function that captures variables from parent scope.
+	var generate func(context.Context, *ModelRequest, int) (*ModelResponse, error)
+
+	generate = func(ctx context.Context, req *ModelRequest, currentTurn int) (*ModelResponse, error) {
+		spanMetadata := &tracing.SpanMetadata{
+			Name:    "generate",
+			Type:    "util",
+			Subtype: "util",
 		}
 
-		if formatHandler != nil {
-			resp.Message, err = formatHandler.ParseMessage(resp.Message)
+		return tracing.RunInNewSpan(ctx, spanMetadata, req, func(ctx context.Context, req *ModelRequest) (*ModelResponse, error) {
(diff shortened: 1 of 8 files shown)


go/ai/generate.go
```diff
@@ -508,7 +500,15 @@ func Generate(ctx context.Context, r api.Registry, opts ...GenerateOption) (*Mod
 	}
 	actionOpts.Messages = processedMessages
 
-	return GenerateWithRequest(ctx, r, actionOpts, genOpts.Middleware, genOpts.Stream)
+	spanMetadata := &tracing.SpanMetadata{
```
apascal07: This doesn't seem right. `GenerateUtilAction` is only used by the Dev UI, whereas `GenerateWithRequest` is the central generate logic for all paths, including code. So this will show up in the traces even for calls outside of the Dev UI. Is this actually what we want?

go/plugins/googlecloud/googlecloud.go
```diff
@@ -128,18 +143,22 @@ func initializeTelemetry(opts *GoogleCloudTelemetryOptions) {
 			logInputAndOutput: !opts.DisableLoggingInputAndOutput, // Default true, disable if flag set
 			projectId:         projectID,
 		}
+
+		// Note: Ideally we should use SimpleSpanProcessor in dev mode for immediate export
+		// However I noticed there's a strange interaction effect with Dev UI that causes listActions
```
apascal07, line 148: Does this cause a delay in Dev UI trace rendering?

huangjeff5: It doesn't because we have writeTelemetryImmediate => which creates a simple span exporter

go/ai/generate.go
```diff
@@ -324,52 +316,63 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
...
-		}
+			toolCount := 0
+			for _, part := range resp.Message.Content {
+				if part.IsToolRequest() {
+					toolCount++
+				}
+			}
+			if toolCount == 0 || opts.ReturnToolRequests {
+				return resp, nil
+			}
```
apascal07:
```suggestion
			if len(resp.ToolRequests()) > 0 || opts.ReturnToolRequests {
				return resp, nil
			}
```

go/ai/generate.go
```diff
@@ -340,13 +340,7 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
...
 			}
 
-			toolCount := 0
-			for _, part := range resp.Message.Content {
-				if part.IsToolRequest() {
-					toolCount++
-				}
-			}
-			if toolCount == 0 || opts.ReturnToolRequests {
+			if len(resp.ToolRequests()) > 0 || opts.ReturnToolRequests {
```
apascal07:
```suggestion
			if len(resp.ToolRequests()) == 0 || opts.ReturnToolRequests {
```

Whoops, replaced it with the wrong logic.

PR #3555, opened 2025-09-12
fix(go/reflection): fixed trace info capturing in reflection api

pavelgj:
Checklist (if applicable):

### DIFF
--- a/go/core/api/action.go
+++ b/go/core/api/action.go
@@ -21,8 +21,8 @@ import (
 	"encoding/json"
 )
 
-type ActionRunJSONResult struct {
-	Result  json.RawMessage
+type ActionRunResult[T any] struct {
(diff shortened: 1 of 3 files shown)


go/core/api/action.go
```diff
@@ -21,13 +21,21 @@ import (
 	"encoding/json"
 )
 
+type ActionRunJSONResult struct {
+	Result  json.RawMessage
+	TraceId string
+	SpanId  string
+}
```
apascal07:
```suggestion
type ActionRunResult[T any] struct {
	Result  T
	TraceId string
	SpanId  string
}
```

Then use this directly in `action.go` instead of redefining another type?

pavelgj: Done.

PR #3563, opened 2025-09-12
fix(go/plugins/compat_oai): allow custom baseURL in Anthropic plugin

hugoaguirre:
`ANTHROPIC_BASE_URL` (if set) will take precedence against the default base URL.

Checklist (if applicable):

### DIFF
--- a/go/plugins/compat_oai/anthropic/anthropic.go
+++ b/go/plugins/compat_oai/anthropic/anthropic.go
@@ -95,12 +95,11 @@ func (a *Anthropic) Name() string {
 }
 
 func (a *Anthropic) Init(ctx context.Context) []api.Action {
-	// Set the base URL
 	url := os.Getenv("ANTHROPIC_BASE_URL")
 	if url == "" {
 		url = baseURL
 	}
-	a.Opts = append(a.Opts, option.WithBaseURL(url))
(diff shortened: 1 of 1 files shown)


go/plugins/compat_oai/anthropic/anthropic.go
```diff
@@ -95,7 +96,11 @@ func (a *Anthropic) Name() string {
 
 func (a *Anthropic) Init(ctx context.Context) []api.Action {
 	// Set the base URL
-	a.Opts = append(a.Opts, option.WithBaseURL(baseURL))
+	url := os.Getenv("ANTHROPIC_BASE_URL")
+	if url == "" {
+		url = baseURL
+	}
+	a.Opts = append(a.Opts, option.WithBaseURL(url))
```
apascal07: This is fine but we still don't support passing `option.WithBaseURL()` in `Opts`. We were always appending our own and overriding what the user would have passed in. We should fix that.

apascal07:
```suggestion
	url := os.Getenv("ANTHROPIC_BASE_URL")
	if url == "" {
		url = baseURL
	}
	a.Opts = append([]option.RequestOption{option.WithBaseURL(url)}, a.Opts...)
```

apascal07 (approved):

PR #3571, opened 2025-09-14
feat(js/fows): added `schemaMode` option on flows (and other actions) to control schema behavour

pavelgj:
Right now we only validate, but if the schema has defaults or transforms those don't get applied. Ex:

```ts
      const testFlow = defineFlow(
        registry,
        {
          name: 'testFlow',
          inputSchema: z.object({
            foo: z.string().default('default foo'),
          }),
          outputSchema: z.object({
            input: z
              .object({
                foo: z.string().optional(),
              })
              .optional(),
            bar: z.string().transform((val) => `${val}-transformed`),
          }),
          schemaMode: 'parse',
        },
        async (input) => {
          return { input, bar: 'bar' };
        }
      );

      const result = await testFlow({} as any);

      assert.deepStrictEqual(result, {
        bar: 'bar-transformed',
        input: {
          foo: 'default foo',
        },
      });
```

also can set `schemaMode: 'none'`, this might be useful in production mode to improve performance in case of large inputs.

Checklist (if applicable):

### DIFF
--- a/js/core/src/action.ts
+++ b/js/core/src/action.ts
@@ -157,6 +157,8 @@ export type Action<
   ): StreamingResponse<O, S>;
 };
 
+export type ActionSchemaMode = 'none' | 'validate' | 'parse';
+
 /**
  * Action factory params.
  */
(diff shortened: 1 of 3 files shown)


MichaelDoyle:
@pavelgj you still aiming to get this in? or close/revisit later?

PR #3582, opened 2025-09-16
chore(js): mark runWithStreamingCallback as deprecated

10kc-emilyk:
All relevant uses were removed in #3243 . This is a documentation change so that library users can see in typescript that this method is deprecated. Note that this only updates the jsdoc. `util.deprecate` could also be used to wrap the function and provide a runtime deprecation message.

Checklist (if applicable):
- [ ] _N/A_ Tested (manually, unit tested, etc.)
- [ ] _N/A_ Docs updated (updated docs or a docs bug required)


MichaelDoyle:
Thanks @10kc-emilyk! 

@pavelgj @ifielker WDYT?

PR #3595, opened 2025-09-17
fix(cli/init-ai-tools): remove Gemini CLI extension installation

ssbushi:
Will reintroduce extension install once fixed for user-level configuration.

Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/genkit-tools/cli/src/commands/init-ai-tools/ai-tools/gemini.ts
+++ b/genkit-tools/cli/src/commands/init-ai-tools/ai-tools/gemini.ts
@@ -16,48 +16,29 @@
 
 import { Runtime } from '@genkit-ai/tools-common/manager';
 import { logger } from '@genkit-ai/tools-common/utils';
-import { select } from '@inquirer/prompts';
 import { existsSync, readFileSync } from 'fs';
-import { mkdir, writeFile } from 'fs/promises';
-import path from 'path';
+import { lstat, mkdir, readlink, symlink, writeFile } from 'fs/promises';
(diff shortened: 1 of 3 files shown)


genkit-tools/cli/src/commands/init-ai-tools/ai-tools/gemini.ts
```diff
@@ -140,48 +107,47 @@ async function installInMdFile(runtime: Runtime): Promise<AIToolConfigResult> {
...
   try {
-    const { updated } = await initOrReplaceFile(
-      extensionPath,
-      JSON.stringify(GENKIT_EXTENSION_CONFIG, null, 2)
-    );
-    extensionUpdated = updated;
-    if (extensionUpdated) {
-      logger.info(
-        `Genkit extension for Gemini CLI initialized at ${extensionPath}`
+    const stats = await lstat(GENKIT_MD_SYMLINK_PATH);
```
ssbushi: Gemini CLI only works well when the `contextFileName` is in the current directory. I'm using a symlink approach to keep `GENKIT.md` centrally located, consistent with where other tools look for.

LMK if this is undesired for any reason.

pavelgj: Would it work on Windows?

ssbushi: Good call. Updated to copy if on windows.

PR #3598, opened 2025-09-17
feat(js/core): Add dynamic-action-provider support

ifielker:
Checklist (if applicable):

### DIFF
--- a/js/ai/src/tool.ts
+++ b/js/ai/src/tool.ts
@@ -218,7 +218,8 @@ export async function lookupToolByName(
   const tool =
     (await registry.lookupAction(name)) ||
     (await registry.lookupAction(`/tool/${name}`)) ||
-    (await registry.lookupAction(`/prompt/${name}`));
+    (await registry.lookupAction(`/prompt/${name}`)) ||
+    (await registry.lookupAction(`/dynamic-action-provider/${name}`));
   if (!tool) {
     throw new Error(`Tool ${name} not found`);
   }
--- a/js/core/src/dynamic-action-provider.ts
+++ b/js/core/src/dynamic-action-provider.ts
@@ -1,0 +1,197 @@
+/**
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type * as z from 'zod';
+import { Action, ActionMetadata, defineAction } from './action.js';
+import { ActionType, Registry } from './registry.js';
+
+type DapValue = {
+  [K in ActionType]?: Action<z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>[];
+};
+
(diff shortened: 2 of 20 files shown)


js/plugins/mcp/src/index.ts
```diff
@@ -62,11 +62,52 @@ export interface McpServerOptions {
...
+ *
+ * @param options Configuration for the MCP Client Host, including the definitions of MCP servers to connect to.
+ * @returns A new instance of GenkitMcpHost.
+ */
+export function defineMcpHost(ai: Genkit, options: McpHostOptionsWithName) {
+  const mcpHost = new GenkitMcpHost(options);
+  ai.defineDynamicActionProvider(options.name, async () => ({
+    tool: await mcpHost.getActiveTools(ai),
+    resource: await mcpHost.getActiveResources(ai),
+  }));
```
ssbushi: Not a strong opinion but looks like the delta between the old and new APIs is only these 4 lines.

Do we need to deprecate the old API instead of providing dynamic actions as an add on?

```
createMcpHost({...}).withDynamicActions({ai})
```

This is not consistent with our `defineBlah({ai, ...})` APIs but it will avoid deprecating the old one?

ifielker: It requires the ai parameter and it also makes the name field of the options a required field. It's a breaking change to createMcpHost.

js/ai/src/tool.ts
```diff
@@ -218,7 +218,8 @@ export async function lookupToolByName(
   const tool =
     (await registry.lookupAction(name)) ||
     (await registry.lookupAction(`/tool/${name}`)) ||
-    (await registry.lookupAction(`/prompt/${name}`));
+    (await registry.lookupAction(`/prompt/${name}`)) ||
+    (await registry.lookupAction(`/dynamic-action-provider/${name}`));
```
pavelgj, line 222: would we allow specifying tool names that are only resolvable through DAP without using the DAP ref syntax?

ifielker: No. I don't think that's a good idea. You might have a /dynamic-action-provider/myDap1:tool/fs/read_file that is configured to read stuff from directory A and another one /dynamic-action-provider/myDap2:tool/fs/read_file that is configured to read stuff from directory B. If you just say /tool/fs/read_file it's ambiguous (and we don't know where to find it).

js/core/src/dynamic-action-provider.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  invalidate() {
+    this.value = undefined;
+  }
+}
+
+export interface DynamicRegistry {
+  __cache: SimpleCache;
+  invalidateCache(): void;
```
pavelgj, line 86: To the DAP consumer cache is an implementation detail. DAP API is read-only. Instead, I think DapFn should take invalidateCache callback and if DAP implementer cares about cache management, they have access:

```ts
export type DapFn = (opts: { invalidateCache: () => void }) => Promise<DapValue>;
```

ifielker: Discussed offline. We are leaving it as is.

js/core/src/dynamic-action-provider.ts
```diff
@@ -0,0 +1,197 @@
...
+  getAction(
+    actionType: string,
+    actionName: string
+  ): Promise<Action<z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>>;
+  listActions(
+    actionType: string,
+    actionName: string
+  ): Promise<Action<z.ZodTypeAny, z.ZodTypeAny, z.ZodTypeAny>[]>;
```
pavelgj: does this need to return `Action` or would `ActionMetadata` be sufficient?

ifielker: I can change it to metadata.

pavelgj (approved):

PR #3600, opened 2025-09-18
fix(go): prevent swallowing metrics errors

huangjeff5:


go/plugins/googlecloud/googlecloud.go
```diff
@@ -241,25 +248,22 @@ func setupGCPLogger(projectID string, level slog.Leveler, credentials *google.Cr
...
-		slog.Warn("Switched to stderr logging due to Google Cloud logging failure", "error", err)
-		slog.Error("Unable to send logs to Google Cloud", "error", err)
+		slog.SetDefault(stderrLogger)
+		stderrLogger.Warn("Unable to send logs to Google Cloud", "error", err)
 		if loggingDenied(err) {
 			showLoggingInstructionsOnce.Do(func() {
 				fmt.Fprint(os.Stderr, loggingDeniedHelpText(projectID))
 			})
 		}
+		/*
```
schnecle, line 261: Should we remove this commented out part?

huangjeff5: I added a comment, ideally we do re-init the logger but I'm not sure what the solution is right now.

PR #3627, opened 2025-09-22
Fix: Propagate Generic Type to `GenerateStreamResponse.stream`

### LINKED ISSUE #3603: [JS] Incorrect Generic Type for generateStream's Stream Response
**Describe the bug**
There is a type inference issue with the stream property on the GenerateStreamResponse interface. The generateStream method is defined as:

```ts
generateStream<O extends z.ZodTypeAny = z.ZodTypeAny, CustomOptions extends z.ZodTypeAny = typeof GenerationCommonConfigSchema>(parts: GenerateOptions<O, CustomOptions> | PromiseLike<GenerateOptions<O, CustomOptions>>): GenerateStreamResponse<z.infer<O>>;
```

It correctly returns a GenerateStreamResponse<z.infer<O>>. However, the definition of GenerateStreamResponse is:

```ts
interface GenerateStreamResponse<O extends z.ZodTypeAny = z.ZodTypeAny> {
    get stream(): AsyncIterable<GenerateResponseChunk>; // Lacks the generic <O>
    get response(): Promise<GenerateResponse<O>>;
}
```

The stream getter returns an `AsyncIterable<GenerateResponseChunk>`, but the `GenerateResponseChunk` itself is not correctly typed with the generic `O`. This results in the stream's chunks always being of type `GenerateResponseChunk<unknown>`, forcing users to perform type assertions and losing the benefit of the schema provided.

**To Reproduce**

The issue can be observed in the typings when using the generateStream method with a Zod schema.

```ts
import { Genkit } from 'genkit';
import { z } from 'zod';

const JokeSchema = z.object({
  joke: z.string(),
  punchline: z.string(),
});

async function getJokeStream() {
  // Assume genkit is configured
  const genkit = new Genkit(); 
  
  const streamResponse = await genkit.generateStream({
    prompt: 'Tell me a short, funny joke.',
    output: {
      schema: JokeSchema,
    },
  });

  // Problem: `chunk` is incorrectly typed as `GenerateResponseChunk<unknown>`
  // instead of `GenerateResponseChunk<{ joke: string; punchline: string; }>`
  for await (const chunk of streamResponse.stream) {
    // Accessing chunk properties lacks type safety.
    console.log(chunk.text); 
  }
}
```

**Expected behavior**

The stream property on GenerateStreamResponse should propagate the generic type O to GenerateResponseChunk. This would correctly type the chunks within the async iterable, providing proper type safety and autocompletion based on the Zod schema passed to the generateStream method.

The resulting type for the stream should be AsyncIterable<GenerateResponseChunk<z.infer<O>>>.

**Runtime**

OS: Windows 11 Home

Version: 24H2

**Node version**

v22.18.0

**Additional context**

This issue can be resolved with a minor change to the GenerateStreamResponse interface.

Current Definition:
```ts
interface GenerateStreamResponse<O extends z.ZodTypeAny = z.ZodTypeAny> {
    get stream(): AsyncIterable<GenerateResponseChunk>;
    get response(): Promise<GenerateResponse<O>>;
}
```

Proposed Fix:
```ts
interface GenerateStreamResponse<O extends z.ZodTypeAny = z.ZodTypeAny> {
    get stream(): AsyncIterable<GenerateResponseChunk<O>>; // Propagate the <O> generic
    get response(): Promise<GenerateResponse<O>>;
}
```

Dinesh-Gautam:
Fixes: #3603 
### Context  
The `GenerateStreamResponse` interface did not propagate its generic `O` type to the `stream` property.  
As a result, the chunks yielded by `stream` were always typed as `GenerateResponseChunk<unknown>`,  
forcing users to manually assert types and losing schema-based type inference.

### What Was Changed  
Updated the `GenerateStreamResponse` interface to propagate the generic `O` type to the `stream` property.

### After Changes
Screenshot 2025-09-22 211707

Checklist (if applicable):


parkcoool:
I've encountered the same type inference issue with the stream response. It would be great if this could be merged soon as it's blocking proper TypeScript support for streaming.

Dinesh-Gautam:
@apascal07 quick follow-up on this PR when you have a moment. Happy to adjust anything if needed. Thanks!

js/ai/src/generate.ts
```diff
@@ -162,7 +162,7 @@ export interface GenerateOptions<
   /** Maximum number of tool call iterations that can be performed in a single generate call (default 5). */
   maxTurns?: number;
   /** When provided, models supporting streaming will call the provided callback with chunks as generation progresses. */
-  onChunk?: StreamingCallback<GenerateResponseChunk>;
+  onChunk?: StreamingCallback<GenerateResponseChunk<z.infer<O>>>;
```
pavelgj, line 165: that was actually intentional, as streamed chunks might be unpredictable partials of the output type technically not conforming to the interface.

Dinesh-Gautam: Thanks for the clarification, my motivation for the change was mainly to avoid losing the schema info, since the current unknown typing makes it bit harder to use structured outputs even when models emit meaningful partial chunks.
If strict correctness of partial chunks is a concern, Im happy to adjust. Maybe something like:
`GenerateResponseChunk<Partial<z.infer<O>>>`
If keeping the current untyped behavior is preferred, no problem, I can close the PR. Otherwise, I’m happy to update it in this direction.

MichaelDoyle: @pavelgj WDYT?

Dinesh-Gautam:
There is not a lot of community support for this, so I think we can close this.

MichaelDoyle:
Thanks @Dinesh-Gautam. Hopefully we'll see other contributions from you in the future. Either way, don't hesitate to send us any feedback.

PR #3630, opened 2025-09-23
fix(go/ai): enable use of media URIs in prompts

hugoaguirre:
Previously, `.prompt` files were not supporting `gs://` and `http(s):` URIs

Checklist (if applicable):

### DIFF
--- a/go/samples/prompts/main.go
+++ b/go/samples/prompts/main.go
@@ -39,17 +39,17 @@ func main() {
 		genkit.WithPromptDir("prompts"),
 	)
 
-	//	SimplePrompt(ctx, g)
+	SimplePrompt(ctx, g)
 	PromptWithMultiMessage(ctx, g)
-	// PromptWithInput(ctx, g)
-	// PromptWithOutputType(ctx, g)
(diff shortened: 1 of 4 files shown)


go/samples/prompts/main.go
```diff
@@ -206,7 +206,10 @@ func PromptWithMultiMessage(ctx context.Context, g *genkit.Genkit) {
 	if prompt == nil {
 		log.Fatal("empty prompt")
 	}
-	resp, err := prompt.Execute(ctx)
+	resp, err := prompt.Execute(ctx,
+		ai.WithModelName("googleai/gemini-2.5-pro"),
+		ai.WithInput(map[string]any{"videoUrl": "https://www.youtube.com/watch?v=K-hY0E6cGfo video/mp4"}),
```
apascal07, line 211: Is providing the content type like this really standard?

hugoaguirre: That's how it got implemented in dotprompt package for both JS and Go. The parser splits the URI in two, the right side is always the `contentType`

krishnaKapadia (approved):

PR #3631, opened 2025-09-23
fix(go/internal): discover actions and plugins in parent registry

hugoaguirre:
The child registry was not considering the registered parents' actions and plugins

Checklist (if applicable):


huangjeff5:
May be worth adding a comment that the code is recursive (not immediately obvious)

apascal07 (approved):

PR #3638, opened 2025-09-25
fix(go): add Go MCP PR explainer example and safeguard Gemini

huangjeff5:
Adding a Go example with official Github MCP server. This demo shows that you can easily build an agent that leverages domain-specific tools through an MCP server. 

Also fixed Gemini plugin issues along the way:
1. MCP server returns tool input JSON schema where minItems, maxItems, minimum, maximum => are already numbers, but gemini plugin expects strings in the JSON schema and casts to int64/float64. Current plugin implementation is brittle and panics, we handle all possible cases explicitly and use the most precise type.
2. Related issue was happening to "required", "propertyOrdering", "enum" fields. gemini plugin expects []any, but they are actually typed as []string. Handle both cases explicitly.


go/plugins/googlegenai/gemini.go
```diff
@@ -598,6 +650,12 @@ func toGeminiToolChoice(toolChoice ai.ToolChoice, tools []*ai.ToolDefinition) (*
 // translateCandidate translates from a genai.GenerateContentResponse to an ai.ModelResponse.
 func translateCandidate(cand *genai.Candidate) *ai.ModelResponse {
 	m := &ai.ModelResponse{}
+	// Guard against nil candidate or nil content to avoid panics
```
hugoaguirre: If I remember correctly, the direction we should follow in this situation was to return an error saying that no valid candidates were found? cc @apascal07

huangjeff5:
Removed nil panic handling; handled in https://github.com/firebase/genkit/pull/3512/files#diff-096e7aec5bf004683d663cc97ae7d7794730aa60f225c3c5d2e4519ae38a81e2

go/plugins/googlegenai/gemini.go
huangjeff5: Gotcha, I didn't know that you were touching that part as well. Decided to revert this logic for now so we don't run into conflicts.

Issue #3641, opened 2025-09-25
JS Express Testapp Routes Do Not Register As Documented


Zozman:
When running the [Express testapp](https://github.com/firebase/genkit/tree/main/js/testapps/express) for testing the `@genkit-ai/express` middleware, some of the documented `curl` commands do not work.  For example, the `url http://localhost:8080/jokeFlow?stream=true -d '{"data": "banana"}' -H "content-type: application/json" -H "authorization: open sesame"` command documented [here](https://github.com/firebase/genkit/blob/main/js/testapps/express/src/index.ts#L99) does not work because the `f.name` value does not equal the name of the flow.  It equals `actionFn` instead.

In addition, other `curl` commands such as `curl http://localhost:8080/jokeHandler?stream=true -d '{"data": "banana"}' -H "content-type: application/json"` do not return a `UserFacxingError` as expected and instead return an Internal Service Error due to it having issues with the `context` object and the server throwing `TypeError: Cannot read properties of undefined (reading 'username')` errors.

PR #3642, opened 2025-09-25
fix(js/testapps/express): Fixed express testapp endpoints

Zozman:
As mentioned in #3641, currently the endpoints [in the express testapp](https://github.com/firebase/genkit/blob/main/js/testapps/express/src/index.ts) do not work and therefore the documented `curl` commands do not work as documented in the file's comments.  Therefore this PR makes the following changes to fix this:

- Repairs the `context.auth.username` check inside the `jokeFlow` to only perform the check if the object exists (meaning it ran through an endpoint that uses the `auth` `contextProvider` and thus doesn't crash if accessed by an endpoint not using the `contextProvider`).
- Updates the registration of the `POST /jokeFlow` endpoint to not rely on iterating through `ai.flows` because the `f.name` attribute previously being used does not equal the flow's name anymore so simplifying the logic to ensure it's registered and the `curl` command works.
- Moved the port of the express server to `8080` because port `5000` is [used by Mac OS X Control Center](https://stackoverflow.com/questions/72369320/why-always-something-is-running-at-port-5000-on-my-mac) so moved it to make development easier across OSes.

Tested by running the server and executing all the provided `curl` commands in the comments and calling `curl http://localhost:8080/jokeFlow?stream=true -d '{"data": "banana"}' -H "content-type: application/json" -H "authorization: hi"` to simulate an authenticated call that fails to ensure that the `PERMISSION_DENIED` error is properly returned in this usecase.

Checklist (if applicable):


MichaelDoyle:
@Zozman thanks!  I don't think we can take this as is (see comments). Can you clarify what is going wrong?

js/testapps/express/src/index.ts
```diff
@@ -41,13 +41,13 @@ const ai = genkit({
 export const jokeFlow = ai.defineFlow(
   { name: 'jokeFlow', inputSchema: z.string(), outputSchema: z.string() },
   async (subject, { context, sendChunk }) => {
-    if (context!.auth!.username != 'Ali Baba') {
-      throw new UserFacingError('PERMISSION_DENIED', context!.auth!.username!);
+    if (context?.auth?.username && context.auth.username !== 'Ali Baba') {
```
MichaelDoyle, line 44: Changing this way will allow the auth to be bypassed entirely. We could do:

```ts
if (context?.auth?.username !== 'Ali Baba') {
  throw new UserFacingError('PERMISSION_DENIED', context?.auth?.username || 'Anonymous');
}
```

js/testapps/express/src/index.ts
```diff
@@ -95,18 +95,16 @@ const acls: Record<string, string> = {
...
-  app.post(
-    `/${f.name}`,
-    expressHandler(f, { contextProvider: auth(acls[f.name]) })
-  );
-});
+// curl http://localhost:8080/jokeFlow?stream=true -d '{"data": "banana"}' -H "content-type: application/json" -H "authorization: open sesame"
+app.post(
+  '/jokeFlow',
+  expressHandler(jokeFlow, { contextProvider: auth(acls['jokeFlow']) })
+);
```
MichaelDoyle, line 102: I think this behavior was correct - can you explain?

js/testapps/express/src/index.ts
```diff
@@ -142,7 +140,7 @@ app.get('/jokeStream', async (req: Request, res: Response) => {
   res.end();
 });
 
-const port = process.env.PORT || 5000;
+const port = process.env.PORT || 8080;
```
MichaelDoyle, line 143: 8080 is also common and/or will conflict. Let's leave as is. It's easy enough to provide an alternative port.

PR #3647, opened 2025-09-26
feat: add support for express 5.x in @genkit-ai/express

### LINKED ISSUE #3281: [JS] Add support for express 5.x in genkit express plugin
Currently genkit express plugin depends on express 4.x. The latest version is express 5.x, which is pretty new, but something genkit should probably support eventually.

The best solution probably is to set the peerDependency to something like:

```
// Support both versions
{
  "peerDependencies": {
    "express": "^4.21.1 || ^5.0.0"
  }
}
```

We need to test it but at a glance I see that all the express features we're using are well-supported by express 5.x.

gioboa:
Close #3281

Creating a new Angular SSR project with the CLI installs express version 5.1.0.

Screenshot 2025-09-26 at 17 48 49

To use `@genkit-ai/express` without errors during the npm installation, expand the peer dependency is needed.

Checklist (if applicable):

### DIFF
--- a/js/plugins/express/package.json
+++ b/js/plugins/express/package.json
@@ -33,7 +33,7 @@
   },
   "peerDependencies": {
     "genkit": "workspace:^",
-    "express": "^4.21.1"
+    "express": "^4.21.1 || ^5.0.0"
   },
   "devDependencies": {
     "get-port": "^5.1.0",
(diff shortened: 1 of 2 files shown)


js/plugins/express/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "peerDependencies": {
     "genkit": "workspace:^",
-    "express": "^4.21.1"
+    "express": "^4.21.1 || ^5.0.0"
```
gioboa, line 36: already fixed by #3770

PR #3650, opened 2025-09-28
fix(go): Use ResolveActionFor to support dynamic plugins

eric642:

Checklist (if applicable):

### DIFF
--- a/go/ai/evaluator.go
+++ b/go/ai/evaluator.go
@@ -291,7 +291,7 @@ func DefineBatchEvaluator(r api.Registry, name string, opts *EvaluatorOptions, f
 // LookupEvaluator looks up an [Evaluator] registered by [DefineEvaluator].
 // It returns nil if the evaluator was not defined.
 func LookupEvaluator(r api.Registry, name string) Evaluator {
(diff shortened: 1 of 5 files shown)


eric642:
@apascal07 please review it. Enable all action types to support dynamic plugins

apascal07:
What is your use case? We use resolve for models and embedders because they can be provided by plugins dynamically. The other types are user-defined mostly so it doesn't make sense for them to call resolve instead of look up.

eric642:
I have a set of stateful tools, and I hope they can be re-instantiated when used by different agents. For example: agent as tool can be used for ReActAgent. In addition, our company has other prompt management platforms. We hope to be able to obtain the latest version of the prompt in real time through dynamic plugins.

eric642:
ResolveActionFor first performs a lookup; only if the lookup fails does it proceed to search for dynamic plugins. Therefore, I believe that using ResolveActionFor does not alter the program's behavior, while still enabling dynamic plugin functionality, and it appears to present no significant issues.

apascal07 (approved):

PR #3658, opened 2025-09-30
feat: real time telemetry updates

huangjeff5:


js/core/src/action.ts
```diff
@@ -344,6 +344,13 @@ export function action<
         metadata.name = actionName;
         metadata.input = input;
 
+        // Emit an initial chunk with just the traceId
+        try {
+          if (options?.onChunk) {
```
pavelgj, line 349: don't think this approach will work... for one, what if its non-streaming call.
I think `action.run` needs another param (in `ActionRunOptions`), a callback of some sort, that will be called with the trace id when action starts. Then in the reflection api you can send the trace id in the headers.

PR #3661, opened 2025-09-30
fix(go/internal): refine JSON markdown detection

hugoaguirre:
Checklist (if applicable):

### DIFF
--- a/go/internal/base/json_test.go
+++ b/go/internal/base/json_test.go
@@ -41,7 +41,7 @@ func TestExtractJSONFromMarkdown(t *testing.T) {
 		{
 			desc: "simple markdown",
 			in:   "```foo bar```",
-			want: "foo bar",
+			want: "```foo bar```",
 		},
 		{
 			desc: "json markdown",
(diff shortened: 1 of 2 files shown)


go/internal/base/json_test.go
```diff
@@ -58,6 +63,21 @@ func TestExtractJSONFromMarkdown(t *testing.T) {
...
 		},
+		{
+			desc: "yaml markdown",
+			in:   "```yaml\nkey: 1\nanother-key: 2```",
+			want: "```yaml\nkey: 1\nanother-key: 2```",
+		},
+		{
+			desc: "yaml + json markdown",
+			in:   "```yaml\nkey: 1\nanother-key: 2``` ```json\n{\"a\": 1}\n```",
+			want: "\n{\"a\": 1}\n",
```
hugoaguirre: @apascal07 we're keeping the original string with the original new line characters.

Is it okay if we keep them or should we trim them?

apascal07: We should trim them down.

apascal07 (approved):

PR #3671, opened 2025-10-03
fix(go/plugins/googlegenai): Update gemini models

edwrdc:
Checklist (if applicable):


edwrdc:
@hugoaguirre Hi, can you take a look at this. Thanks.

hugoaguirre:
Hi @edwrdc, thanks for the contribution. Actually, there's no longer need of doing this for recent models. 

You can type the name of the model you want to use and it should be available for use.

Either you set it using WithDefaultModel during Genkit init or using .WithModelName() helper function in the Generate() call.

edwrdc:
Ah, cool. Thanks for the heads up! I'll go ahead and close the PR then.

PR #3679, opened 2025-10-06
feat(js/plugins/compat-oai): Add reasoning_content support to OpenAI Compatible API plugin

neokn:
This commit adds support for handling reasoning_content in both streaming and non-streaming for OpenAI Compatible API Chat Completions responses, allowing models that support reasoning_content to properly return their reasoning traces.

CHANGELOG:
  - [x] Add reasoning_content handling in fromOpenAIChoice
  - [x] Add reasoning_content handling in fromOpenAIChunkChoice
  - [x] Add test cases for reasoning_content in both functions

[OpenAI](https://platform.openai.com/docs/api-reference/chat/object) it self not support `reasoning_content` but 
- [xAI](https://docs.x.ai/docs/guides/reasoning)
- [DeepSeek](https://api-docs.deepseek.com/guides/reasoning_model)
- [LiteLLM](https://docs.litellm.ai/docs/reasoning_content)

support `reasoning_content ` when using reasoning model

Checklist (if applicable):

### DIFF
--- a/js/plugins/compat-oai/src/model.ts
+++ b/js/plugins/compat-oai/src/model.ts
@@ -277,19 +277,31 @@ export function fromOpenAIChoice(
   const toolRequestParts = choice.message.tool_calls?.map((toolCall) =>
     fromOpenAIToolCall(toolCall, choice)
   );
+
+  // Build content array based on what's present in the message
+  let content: Part[] = [];
+
+  if (toolRequestParts) {
+    content = toolRequestParts as ToolRequestPart[];
+  } else {
+    // Handle reasoning_content if present
+    if ('reasoning_content' in choice.message && choice.message.reasoning_content) {
+      content.push({ reasoning: choice.message.reasoning_content as string });
+    }
+
+    // Handle regular content
+    content.push(
+      jsonMode
+        ? { data: JSON.parse(choice.message.content!) }
+        : { text: choice.message.content! }
+    );
+  }
+
   return {
     finishReason: finishReasonMap[choice.finish_reason] || 'other',
     message: {
       role: 'model',
-      content: toolRequestParts
-        ? // Note: Not sure why I have to cast here exactly.
-          // Otherwise it thinks toolRequest must be 'undefined' if provided
-          (toolRequestParts as ToolRequestPart[])
-        : [
-            jsonMode
-              ? { data: JSON.parse(choice.message.content!) }
-              : { text: choice.message.content! },
-          ],
+      content,
     },
   };
 }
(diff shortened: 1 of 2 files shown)


neokn:
Hi @pavelgj, just wanted to kindly check if you’ve had a chance to look at this PR.
Please let me know if there’s anything I can update or clarify. Thanks a lot!

pavelgj:
@ssbushi could you please take a look?

js/plugins/compat-oai/src/model.ts
```diff
@@ -383,9 +414,19 @@ export function toOpenAIRequestBody(
...
-      type: 'json_object',
-    };
+    if (request.output?.schema) {
+      body.response_format = {
+        type: 'json_schema',
+        json_schema: {
+          name: 'output',
+          schema: request.output!.schema,
+        },
+      };
```
ssbushi, line 424: I could not find API reference for `json_schema` for DeepSeek. Are you sure this works?

http://api-docs.deepseek.com/api/create-chat-completion#request

ssbushi (commented):
Thanks for the PR. Looks good overall, I just need a bit of clarification on the json schema change. :)

js/plugins/compat-oai/src/model.ts
neokn: hi @ssbushi thanks for your review, yes, DeepSeek didn't support json_schema
so there still has fallback for json_object
when integrate to DeepSeek, user can set the output.format without JsonSchema
e.g. `ai.generate({output: {format: 'json'}});`

then it will fallback to 
```
body.response_format = {
    type: 'json_object',
};
```

I'm follow these two document for implement `json_schema` part
[LiteLLM](https://docs.litellm.ai/docs/completion/json_mode)
[OpenAI](https://platform.openai.com/docs/api-reference/chat/create)
[xAI](https://docs.x.ai/docs/api-reference#chat-completions)

neokn: Hi @ssbushi just want to check is there still has any concern in this commit?

ssbushi: Thanks for waiting. Since you say that DeepSeek does not support json_schema, I think we can do one extra check/coercion for deepseek in the requestBuilder:

https://github.com/neokn/genkit/blob/8cad6af81e2dbcfbad468b7ca464241825b36867/js/plugins/compat-oai/src/deepseek/deepseek.ts#L31

Basically do: 
```
if `request.output?.format`is set to `json`:
    if  `request.output?.schema`  is also set:
        LOG that this is unsupported for DeepSeek,
        Clear `request.output?.schema` 
```
This will force DeepSeek to always go to the fallback you described

This will prevent users from accidentally defining both `json` format and proving a custom json schema for DeepSeek. WDYT?

The PR looks good otherwise, so I am approving this.

neokn: Thanks, this is a good point, I’ll send another pull request for this.

neokn: Hi @ssbushi , may I know when will be merge for this PR?

ssbushi: Hey, thanks for pinging. I just merged this PR. You can proceed with the follow-up PR for deepseek. Thanks!

PR #3686, opened 2025-10-08
fix(go/plugins/googlegenai): use one `FunctionDeclaration` for all tools

hugoaguirre:
Checklist (if applicable):

### DIFF
--- a/go/plugins/googlegenai/googleai_live_test.go
+++ b/go/plugins/googlegenai/googleai_live_test.go
@@ -202,7 +202,8 @@ func TestGoogleAILive(t *testing.T) {
 		_, err := genkit.Generate(ctx, g,
 			ai.WithConfig(&genai.GenerateContentConfig{
 				Tools: []*genai.Tool{
-					{GoogleSearch: &genai.GoogleSearch{}}, {GoogleSearch: &genai.GoogleSearch{}},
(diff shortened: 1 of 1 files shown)


go/plugins/googlegenai/googleai_live_test.go
```diff
@@ -191,7 +197,42 @@ func TestGoogleAILive(t *testing.T) {
 			t.Errorf("got %q, expecting it to contain %q", out, want)
 		}
 	})
-
+	t.Run("api side tools", func(t *testing.T) {
+		m := googlegenai.GoogleAIModel(g, "gemini-2.5-flash")
+		_, err := genkit.Generate(ctx, g,
+			ai.WithConfig(&genai.GenerateContentConfig{
+				Tools: []*genai.Tool{
+					{GoogleSearch: &genai.GoogleSearch{}}, {GoogleSearch: &genai.GoogleSearch{}},
```
apascal07: Is this intentionally duplicated?

hugoaguirre: Yes. I wanted to use multiple API-side tools for the same request. Thinking about it, maybe I should use `CodeExecution` instead.

apascal07: It doesn't really make sense to do the same one twice so yeah maybe change it to code execution or something else.

PR #3688, opened 2025-10-08
feat(go/ai): add `ModelResponse.Media()` helper

apascal07:


go/ai/generate.go
```diff
@@ -774,6 +774,19 @@ func (mr *ModelResponse) Interrupts() []*Part {
...
 }
 
+// Media returns the media content of the [ModelResponse] as a string.
+func (mr *ModelResponse) Media() string {
+	if mr.Message == nil {
+		return ""
+	}
+	for _, part := range mr.Message.Content {
+		if part.IsMedia() {
+			return part.Text
```
pavelgj, line 784: I was wondering about this... Typically media is a object with url and optional contentType. In go it seems to be just the text field. What was the reasoning?

apascal07: It was like this from the beginning, I hadn't considered otherwise. I do find it strange.

PR #3691, opened 2025-10-09
feat(js/plugins/vertexai/rerankers): Migrated to V2 plus new config options

ifielker:
Checklist (if applicable):

### DIFF
--- a/js/plugins/vertexai/src/rerankers/index.ts
+++ b/js/plugins/vertexai/src/rerankers/index.ts
@@ -14,5 +14,8 @@
  * limitations under the License.
  */
 
-export { type PluginOptions, vertexAIRerankers } from './legacy/index.js';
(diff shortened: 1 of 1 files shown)


js/plugins/vertexai/src/rerankers/index.ts
```diff
@@ -14,31 +14,5 @@
...
-      location,
-      authClient,
-      rerankOptions: options.rerankers.map((o) =>
-        typeof o === 'string' ? { model: o } : o
-      ),
-    });
-  });
-}
+export * from './legacy/index.js';
+export * from './v2/index.js';
```
pavelgj: similarly here, list exported symbols

ifielker: right, thanks

PR #3706, opened 2025-10-13
fix(js/samples/js-coffee-shop): update model so that the sample works

yesudeep:
ERROR:

```
[GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent: [404 Not Found] models/gemini-1.5-flash is not found for API version v1beta, or is not supported for generateContent. Call ListModels to see the list of available models and their supported methods.
Error: [GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent: [404 Not Found] models/gemini-1.5-flash is not found for API version v1beta, or is not supported for generateContent. Call ListModels to see the list of available models and their supported methods.
...
```

CHANGELOG:
- [ ] Fixes the use of the model so the sample works.

### DIFF
--- a/samples/js-coffee-shop/src/index.ts
+++ b/samples/js-coffee-shop/src/index.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { gemini15Flash, googleAI } from '@genkit-ai/googleai';
+import { googleAI } from '@genkit-ai/google-genai';
 import { genkit, z } from 'genkit';
 
 const ai = genkit({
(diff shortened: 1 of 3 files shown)


samples/js-coffee-shop/src/index.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { gemini15Flash, googleAI } from '@genkit-ai/googleai';
+import { googleAI } from '@genkit-ai/googleai';
```
pavelgj: Ideally this should be updated to the @genkit-ai/google-genai plugin

yesudeep: done

PR #3735, opened 2025-10-21
feat(genkit-tools): support trace uploads via tools router

MichaelDoyle:
Will be used by Dev UI (and other? tools) for team collaboration and for customer/community support.

Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/server/router.ts
+++ b/genkit-tools/common/src/server/router.ts
@@ -23,6 +23,7 @@ import {
 } from '../eval';
 import type { RuntimeManager } from '../manager/manager';
 import { GenkitToolsError, type RuntimeInfo } from '../manager/types';
+import { TraceDataSchema } from '../types';
 import type { Action } from '../types/action';
 import * as apis from '../types/apis';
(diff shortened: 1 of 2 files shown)


genkit-tools/common/src/server/router.ts
```diff
@@ -164,6 +165,14 @@ export const TOOLS_SERVER_ROUTER = (manager: RuntimeManager) =>
         return manager.getTrace(input);
       }),
 
+    /** Adds a trace to the trace store */
+    addTrace: loggedProcedure
+      .input(TraceDataSchema)
```
ssbushi, line 170: Looks like this is specific for Genkit traces? The schema does not say anything Genkit specific but recent conversations with Pavel has led me to believe that there is no universal format here....

This is fine, wondering if we should surface this requirement to the user somewhere.

MichaelDoyle: @ssbushi yes that's right. these are genkit traces. Are you thinking just a comment for now? Or something more (like a specific TS type)?

PR #3749, opened 2025-10-27
refactor(plugins/googleai): migrate to v2

### LINKED ISSUE #3443: refactor: migrate googleai plugin to V2 Plugin API
This is a sub-issue to track migration of the `googleai` js plugin to V2.

CorieW:
Resolves https://github.com/firebase/genkit/issues/3443

[Plugin migration guide](https://docs.google.com/document/d/1s1YKwhqdrGN44IG7lSAwltxM3Bqh9p6jTcri8SNaBWk/edit?usp=sharing)

**Notable Changes**
- Migrates from v1 to v2 API.
 
**Testing**
  - [x] Checked that the same number of modes and embedders load.

image

  - [x] Checked that the models load with the same names/prefixes on the UI.

image

  - [x] Checked that the embedders load with the same names/prefixes on the UI.

image

  - [x] Checked that at least one Gemini model works consistent with pre-migration on the UI.

image

  - [x] Checked that at least one Imagen model works consistent with pre-migration on the UI.

image

  - [ ] Checked that at least one Veo model works consistent with pre-migration on the UI.
[**Can't see any load in, even before migration.**](https://gyazo.com/b5b51427e6a471a970e0eae96436d81a)

  - [x] Checked that at least one Gemma model works consistent with pre-migration on the UI.

image

  - [x] Checked that at least one embedder works consistent with pre-migration on the UI.

image

  - [x] Using a model object, checked using a flow works consistent with pre-migration.

image

  - [x] Using an embedder object, checked using a flow works consistent with pre-migration.

image

  - [x] Using a model string (`'pluginName/modelName'`), checked using a flow works consistent with pre-migration.

image

  - [x] Using an embedder string (`'pluginName/embedderName'`), checked using a flow works consistent with pre-migration.

image

  - [x] Using a resolved model object, checked using a flow works.

image

  - [x] Using a resolved embedder object, checked using a flow works.

image

- [x] Using a model string, checked using a streaming flow works consistent with pre-migration.

image

- [x] Using a model object, checked using a streaming flow works consistent with pre-migration.

image

- [x] Using a resolved model object, checked using a streaming flow works.

image

  - [x] Tested error response from model.
image

  - [x] Tested error response from embedder.

image

  - [x] Tested model media input.

image

  - [x] Tested model tool calling.

image

  - [x] Tested `.list`

image

### DIFF
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -42,8 +42,8 @@ import { GenkitError, z, type JSONSchema } from 'genkit';
 import {
   GenerationCommonConfigDescriptions,
   GenerationCommonConfigSchema,
-  getBasicUsageStats,
   modelRef as createModelRef,
+  getBasicUsageStats,
   type CandidateData,
   type MediaPart,
   type MessageData,
@@ -57,11 +57,11 @@ import {
   type ToolResponsePart,
 } from 'genkit/model';
 import { downloadRequestMedia } from 'genkit/model/middleware';
+import { model } from 'genkit/plugin';
 import { runInNewSpan } from 'genkit/tracing';
 import { getApiKeyFromEnvVar, getGenkitClientHeader } from './common';
 import { handleCacheIfNeeded } from './context-caching';
 import { extractCacheConfig } from './context-caching/utils';
-import { model } from 'genkit/plugin';
 
 // Extra type guard to keep the compiler happy and avoid a cast to any. The
 // legacy Gemini SDK is no longer maintained, and doesn't have updated types.
@@ -1205,7 +1205,10 @@ export function defineGoogleAIModel({
       configSchema: GeminiConfigSchema,
       use: middleware,
     },
-    async (request, { streamingRequested, sendChunk, abortSignal, registry }) => {
+    async (
+      request,
+      { streamingRequested, sendChunk, abortSignal, registry }
+    ) => {
       const options: RequestOptions = { apiClient: getGenkitClientHeader() };
       if (apiVersion) {
         options.apiVersion = apiVersion;
@@ -1254,12 +1257,11 @@ export function defineGoogleAIModel({
 
       if (codeExecutionFromConfig) {
         tools.push({
-          codeExecution:
-            requestConfig.codeExecution ?
-              requestConfig.codeExecution === true
-                ? {}
-                : requestConfig.codeExecution
-              : {},
+          codeExecution: requestConfig.codeExecution
+            ? requestConfig.codeExecution === true
+              ? {}
+              : requestConfig.codeExecution
+            : {},
         });
       }
 
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -38,12 +38,12 @@ import {
   type ToolConfig,
   type UsageMetadata,
 } from '@google/generative-ai';
-import { GenkitError, z, type Genkit, type JSONSchema } from 'genkit';
+import { GenkitError, z, type JSONSchema } from 'genkit';
 import {
   GenerationCommonConfigDescriptions,
   GenerationCommonConfigSchema,
   getBasicUsageStats,
-  modelRef,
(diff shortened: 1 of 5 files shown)


js/plugins/googleai/src/index.ts
```diff
@@ -125,29 +131,31 @@ async function initializer(ai: Genkit, options?: PluginOptions) {
...
-      defineGoogleAIModel({
-        ai,
-        name,
-        apiKey: options?.apiKey,
-        apiVersion: undefined,
-        baseUrl: options?.baseUrl,
-        debugTraces: options?.experimental_debugTraces,
-      })
+      actions.push(
+        defineGoogleAIModel({
```
cabljac, line 148: does the original defineGoogleAIModel overwrite different version, same name? or do we get both versions registered?

Can apiVersions include both, and then do we end up with different behaviour in v2 now, where we have both actions?

Need to confirm what's going on here so we don't change expected behaviour i think

js/plugins/googleai/src/index.ts
```diff
@@ -125,29 +131,31 @@ async function initializer(ai: Genkit, options?: PluginOptions) {
...
-      defineGoogleAIModel({
-        ai,
-        name,
-        apiKey: options?.apiKey,
-        apiVersion: 'v1beta',
-        baseUrl: options?.baseUrl,
-        debugTraces: options?.experimental_debugTraces,
-      })
+      actions.push(
+        defineGoogleAIModel({
```
cabljac, line 135: Q: does this always return an action, or can it return `undefined` or something?

do we need to do some kind of filtering here maybe?

js/plugins/googleai/src/index.ts
```diff
@@ -125,29 +131,31 @@ async function initializer(ai: Genkit, options?: PluginOptions) {
...
+          apiKey: options?.apiKey,
+          apiVersion: 'v1beta',
+          baseUrl: options?.baseUrl,
+          debugTraces: options?.experimental_debugTraces,
+        })
+      )
     );
   }
   if (apiVersions.includes('v1')) {
     Object.keys(SUPPORTED_GEMINI_MODELS).forEach((name) =>
```
cabljac: on this occasion i'd actually prefer a map and spread now maybe:

```
actions.push(
  ...Object.keys(SUPPORTED_GEMINI_MODELS).map((name) =>
    defineGoogleAIModel({ ... })
  )
);
```

forEach is clean for the previous code where we're just doing side effects, but now we're actually interested in the actions returned.

js/plugins/googleai/src/gemini.ts
```diff
@@ -1428,9 +1430,9 @@ export function defineGoogleAIModel({
 
       // If debugTraces is enable, we wrap the actual model call with a span, add raw
       // API params as for input.
-      return debugTraces
+      return debugTraces && registry
```
cabljac: Is this definitely correct? Without checking out code it's hard to verify that you can destructure registry like this.

google-genai is using some experimental flag, see:

https://github.com/firebase/genkit/blob/main/js/plugins/google-genai/src/googleai/gemini.ts#L657-L661

js/plugins/googleai/src/index.ts
CorieW: It returns `ModelAction`, but can throw an error if API key not set

js/plugins/googleai/src/index.ts
CorieW: I agree.

js/plugins/googleai/src/gemini.ts
CorieW: I'd go with what's in google-genai, as I've just realised the current `runInNewSpan` function where you pass in the registry is deprecated.

js/plugins/googleai/src/index.ts
CorieW: Worth double checking.

MichaelDoyle:
I'm assume we don't need this, with the advent of the google genai plugin. If not, let me know / feel free to reopen.

PR #3751, opened 2025-10-27
fix(go/plugins/mcp): Retrieve MCP server output schema

pfernandom:
Currently,  tool definitions retrieved from an MCP server using `GentkitMCPClient` method `GetActiveTools`  contain only an empty map at `tool.Definition().OutputSchema`, even if the server explicitly declares an output schema for the given tool.

e.g. For the following MCP response:

```
{
  "tools": [
    {
      "name": "embed",
      "description": "Embed a text into a vector",
      "inputSchema": {
        "properties": { "text": { "type": "string" } },
        "required": ["text"],
        "type": "object"
      },
      "outputSchema": {
        "properties": {
          "result": { "items": { "type": "number" }, "type": "array" }
        },
        "required": ["result"],
        "type": "object",
        "x-fastmcp-wrap-result": true
      },
      "_meta": { "_fastmcp": { "tags": [] } }
    }
  ]
}
```
The following data is included in the tool definition:
Screenshot 2025-10-27 at 2 16 19 PM

This is in part because the current version for `github.com/mark3labs/mcp-go` doesn't include the `OutputSchema` attribute in the struct used to deserialize the message:

Screenshot 2025-10-27 at 2 19 51 PM

This PR introduces the following changes:
- Update `github.com/mark3labs/mcp-go` to the latest version, which does include `OutputSchema` in the `Tool` struct.
- Update the genkit MCP functions to actually retrieve the schema and use it during tool retrieval.

Checklist (if applicable):

### DIFF
--- a/go/ai/tools.go
+++ b/go/ai/tools.go
@@ -213,6 +213,20 @@ func NewToolWithInputSchema[Out any](name, description string, inputSchema map[s
 	return NewTool(name, description, fn, WithInputSchema(inputSchema))
 }
 
+// ToolSchema is a struct that contains the input and output schemas for a tool.
+type ToolSchema struct {
+	Input  map[string]any
+	Output map[string]any
+}
+
+// NewToolWithOutputSchema creates a new [Tool] with a custom output schema. It can be passed directly to [Generate].
+func NewToolWithSchema[In, Out any](name, description string, schema ToolSchema, fn ToolFunc[In, Out]) Tool {
+	metadata, wrappedFn := wrapToolFunc(name, description, fn)
+	metadata["dynamic"] = true
+	toolAction := core.NewStructuredAction(name, api.ActionTypeTool, metadata, schema.Input, schema.Output, wrappedFn)
+	return &tool{Action: toolAction}
+}
+
 // DefineMultipartTool creates a new multipart [Tool] and registers it.
 // Multipart tools can return both output data and additional content parts (like media).
 // Use [WithInputSchema] to provide a custom JSON schema instead of inferring from the type parameter.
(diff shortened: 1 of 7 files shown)


MichaelDoyle:
@apascal07 can you take a peek?

adesinah (changes requested):
Hi @pfernandom thanks for working on this.

The output schema is now exposed through the tool definition, which is the right direction, but I don’t think this is ready to merge yet.

The main blocker is that the new TestToolOutputSchema currently fails when the tool is actually called. Genkit
validates the outer MultipartToolResponse, while the MCP output schema describes the inner StructuredContent.
The special case added to ValidateValue does not help because validation receives the multipart wrapper rather
than a direct *mcp.CallToolResult.

Running go test ./plugins/mcp fails with:

 invalid output from action "/tool/test_getWeather":
 - (root): Weather is required
 - (root): Temperature is required
I also found a few related issues:

An MCP tool without an output schema produces {"type":""} rather than an empty schema. Because that map is
non-empty, schema-less tools are sent through the new schema-aware path and can be validated against an
invalid schema.

The exact schema from the PR description is not fully preserved. For example, x-fastmcp-wrap-result is
dropped, as would other unsupported JSON Schema keywords.

NewToolWithSchema creates a legacy tool action, whereas the other current tool constructors use tool.v2. That
means adding an output schema unexpectedly changes the action type.

I think the implementation needs to keep the schema exposed by Tool.Definition() separate from the schema used
to validate Genkit’s outer multipart response. It would also be better to avoid adding MCP-specific handling to
the generic validation package.

Once the runtime validation and schema-less cases are addressed, the existing test should pass and this will be
much closer to the intended behavior.

adesinah:
@pfernandom  I have addressed the required changes mentioned above here #5928 - it also includes your initial commits. So I will be closing this PR.

PR #3756, opened 2025-10-29
feat(js/plugins/google-genai): Vertexai veo - support reference images…

ifielker:
Also: vertexai: support location config for all models.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/common/utils.ts
+++ b/js/plugins/google-genai/src/common/utils.ts
@@ -56,6 +56,27 @@ export function extractErrMsg(e: unknown): string {
 }
 
 /**
+ * Custom replacer function for JSON.stringify to truncate long string fields.
+ * Truncates strings to the first 100 and last 10 characters
+ * if the original string is longer than 110 characters.
+ *
+ * @param key The key of the property being stringified.
+ * @param value The value of the property being stringified.
+ * @return The transformed value, or the original value if no transformation is needed.
+ */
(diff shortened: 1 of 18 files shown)


js/plugins/google-genai/src/common/utils.ts
```diff
@@ -55,6 +55,27 @@ export function extractErrMsg(e: unknown): string {
...
+/**
+ * Custom replacer function for JSON.stringify to truncate long string fields.
+ * Truncates strings to the first 100 and last 10 characters
+ * if the original string is longer than 110 characters.
+ *
+ * @param key The key of the property being stringified.
+ * @param value The value of the property being stringified.
+ * @return The transformed value, or the original value if no transformation is needed.
+ */
+export function stringTruncator(key: string, value: unknown): unknown {
```
pavelgj, line 67: unused?

ifielker: It's REALLY handy for debugging anything with video or image data URLs. I didn't want to have to write it fresh or keep it somewhere else. I figured other people might like the use it to debug stuff too because 37000 pages of base64 data is not helpful :) I could comment it out. You use it like JSON.stringify(myObject, stringTruncator, 2);
I could have commented it out, but then I'd have to find it and uncomment it, and then remember to re-comment it out etc. Seems nice to just be able to drop a debug log without that.

pavelgj (approved):

PR #3766, opened 2025-10-31
feat(py): add embedder reference to Python

MengqinShen:
Unit Test and Integration tests don't pass, still under debugging


MengqinShen:
I am closing this as this deals the same issue as PR #3835 but the changes didn't meet all the requirements. PR-3835 is the more correct version.

PR #3770, opened 2025-11-03
feat(js): durable streaming for flows (experimental)

pavelgj:
Genkit flows can be slow operations that stream intermediate progress. In scenarios with unreliable network conditions or client-side interruptions (e.g., page reloads), clients can lose their connection to the stream. Without a durability mechanism, the stream is lost, and the client cannot retrieve the in-progress or final results.

This PR introduces the concept of a `StreamManager` which is responsible for persisting the state of a stream. A `StreamManager` can be provided when defining a flow handler.

Three `StreamManager` implementations are included:

*   `InMemoryStreamManager`: Stores stream state in memory. This is useful for development and testing but is not truly durable as it will lose state if the process restarts.
*   `FirestoreStreamManager`: A durable `StreamManager` that uses Firebase Firestore to store stream state.
*   `RtdbStreamManager`: A durable `StreamManager` that uses the Firebase Realtime Database to store stream state.

To make a flow's streams durable, pass a `streamManager` instance to the handler. This is supported in both the Express (`expressHandler`) and Next.js (`appRoute`) plugins. The new `durable-streaming` sample app demonstrates this for Express:

```typescript
// From js/testapps/durable-streaming/src/index.ts

const fApp = initializeApp();
export const firestore = new FirestoreStreamManager({
  firebaseApp: fApp,
  db: getFirestore(fApp),
  collection: 'streamy',
});

// ...

app.post(
  '/streamyFirestore',
  expressHandler(streamy, { streamManager: firestore })
);
```

The sample app includes a `streamyThrowy` flow that intentionally throws an error mid-stream to demonstrate how a durable stream manager can preserve the stream's state up to the point of failure.

A client can then connect to a durable stream and resume it if necessary:

```typescript
// Start a new stream
const result = streamFlow({
  url: `http://localhost:8080/myFlowDurable`,
  input: 'tell me a long story',
});
const streamId = await result.streamId; // Save this ID

// ... later, reconnect if needed ...
const reconnectedResult = streamFlow({
  url: `http://localhost:8080/myFlowDurable`,
  streamId: streamId,
});
```

**Beta Feature:** This is to be released as a beta feature.

Checklist (if applicable):

### DIFF
--- a/js/plugins/firebase/tests/rtdb_test.ts
+++ b/js/plugins/firebase/tests/rtdb_test.ts
@@ -100,8 +100,8 @@ describe('RtdbStreamManager', () => {
 
     const db = getDatabase(app);
     const streamRef = db.ref(`genkit-streams/${streamId}`);
-    streamRef.set({ metadata: { createdAt: Date.now() } }).then(() => {
-      streamManager.subscribe(streamId, {
+    streamRef.set({ metadata: { createdAt: Date.now() } }).then(async () => {
+      await streamManager.subscribe(streamId, {
         onChunk: (chunk) => {
           chunks.push(chunk);
         },


js/core/src/streaming.ts
ssbushi, line 1: Some JSDocs in this file would help for creating custom stream managers?

js/plugins/firebase/src/stream-manager/firestore.ts
```diff
@@ -0,0 +1,155 @@
...
+        }
+      }
+      lastIndex = stream.length - 1;
+    });
+    resetTimeout();
+    return { unsubscribe };
+  }
+}
+
+export function firestoreStreamManager(
```
ssbushi: nit: should this be a verb (createBlah)?

js/plugins/firebase/src/stream-manager/firestore.ts
```diff
@@ -0,0 +1,155 @@
...
+          callbacks.onDone?.(event.output);
+          unsubscribe();
+        } else if (event.type === 'error') {
+          clearTimeout(timeoutId);
+          callbacks.onError?.(event.err);
+          unsubscribe();
+        }
+      }
+      lastIndex = stream.length - 1;
+    });
```
ssbushi, line 153: q: How does this work? chunks are being forwarded within the unsubscribe method, while IIUC they should be forwarded till user unsubscribes?

js/plugins/firebase/src/stream-manager/rtdb.ts
```diff
@@ -0,0 +1,152 @@
...
+    const unsubscribe = () => {
+      streamDataRef.off('child_added', handleEvent);
+    };
+    streamDataRef.on('child_added', handleEvent);
+    resetTimeout();
+    return { unsubscribe };
+  }
+}
+
+export function rtdbStreamManager(
```
ssbushi: nit: same here (verb, instead of noun)

js/core/src/streaming.ts
apascal07: +1

js/plugins/firebase/src/stream-manager/firestore.ts
```diff
@@ -0,0 +1,155 @@
...
+  firebaseApp?: App;
+  db?: Firestore;
+  collection: string;
+  timeout?: number;
+}
+
+class FirestoreActionStream<S, O> implements ActionStreamInput<S, O> {
+  private streamDoc;
+
+  constructor(streamDoc) {
```
ssbushi: Use `DocumentReference` type?

js/plugins/firebase/src/stream-manager/firestore.ts
```diff
@@ -0,0 +1,155 @@
...
+      stack: err.stack,
+      ...err,
+    };
+    await this.update({
+      stream: FieldValue.arrayUnion({ type: 'error', err: serializableError }),
+    });
+  }
+}
+
+export class FirestoreStreamManager implements StreamManager {
```
apascal07, line 84: Is the 1MiB limit per collection a concern?

js/plugins/firebase/src/stream-manager/firestore.ts
```diff
@@ -0,0 +1,155 @@
...
+  private async update(data: any): Promise<void> {
+    await this.streamDoc.update({
+      ...data,
+      updatedAt: FieldValue.serverTimestamp(),
+    });
+  }
+
+  async write(chunk: S): Promise<void> {
+    await this.update({
+      stream: FieldValue.arrayUnion({ type: 'chunk', chunk }),
```
apascal07: Is `arrayUnion` an append or is it a union? Will repeated chunks be deduped?

js/core/src/async.ts
```diff
@@ -126,3 +126,28 @@ export function lazy<T>(fn: () => T | PromiseLike<T>): PromiseLike<T> {
...
+export class AsyncTaskQueue {
+  private last: Promise<any> = Promise.resolve();
+
+  /**
+   * Adds a task to the queue.
+   * The task will be executed when its turn comes up in the queue.
+   * @param task A function that returns a value or a PromiseLike.
+   */
+  enqueue(task: () => any | PromiseLike<any>) {
+    this.last = this.last.then(() => lazy(task)).then((res) => res);
```
apascal07:
```suggestion
    this.last = this.last.catch(() => {}).then(() => lazy(task)).then((res) => res);
```

Without this you will break the chain if one task fails so the user stream will stop even if just the durable part is what had an issue.

js/plugins/firebase/src/stream-manager/rtdb.ts
pavelgj: got rid of the helper, `new RtdbStreamManager(opts)` should be fine

js/plugins/firebase/src/stream-manager/firestore.ts
pavelgj: got rid of the helper, `new FirestoreStreamManager(opts)` should be fine

js/plugins/firebase/src/stream-manager/firestore.ts
pavelgj: we subscribe to document updates using `streamDoc.onSnapshot` which returns a function which unsubscribes from the firestore update listener. on each update, we iterate over chunks in the stream and call the callback, then we store lastIndex and next time the callback is stored we start from lastIndex... once we get done or error we unsubscribe

js/plugins/firebase/src/stream-manager/firestore.ts
pavelgj: yeah... mentioned in the readme. We can technically try to chunk them... future problem.

js/plugins/firebase/src/stream-manager/firestore.ts
pavelgj: good point. Added uuid to make sure they are always unique.

js/plugins/firebase/src/stream-manager/firestore.ts
pavelgj: done

js/core/src/streaming.ts
pavelgj: done

js/core/src/async.ts
pavelgj: hmm... yeah... but made it configurable.

PR #3774, opened 2025-11-07
feat(js/ai): implemented `retry` and `fallback` middleware

pavelgj:
```ts
import { retry } from 'genkit/model/middleware';

  model: googleAI.model('gemini-2.5-pro'),
  prompt: 'You are a helpful AI assistant named Walt, say hello',
  use: [
    retry({
      maxRetries: 2,
      initialDelayMs: 1000,
      backoffFactor: 2,
    }),
  ],
});
```

```ts
import { fallback } from 'genkit/model/middleware';

  model: googleAI.model('gemini-2.5-pro'),
  prompt: 'You are a helpful AI assistant named Walt, say hello',
  use: [
    fallback(ai, {
      models: [googleAI.model('gemini-2.5-flash')],
      statuses: ['RESOURCE_EXHAUSTED'],
    }),
  ],
});
```

Checklist (if applicable):

### DIFF
--- a/js/ai/src/model/middleware.ts
+++ b/js/ai/src/model/middleware.ts
@@ -291,11 +291,13 @@ let __setTimeout: (
  * FOR TESTING ONLY.
  * @internal
  */
-export function __setRetryTimeout(
-  impl: (callback: (...args: any[]) => void, ms?: number) => NodeJS.Timeout
-) {
-  __setTimeout = impl;
-}
+export const TEST_ONLY = {
+  setRetryTimeout(
+    impl: (callback: (...args: any[]) => void, ms?: number) => NodeJS.Timeout
+  ) {
+    __setTimeout = impl;
+  },
+};
 
 const DEFAULT_RETRY_STATUSES: StatusName[] = [
   'UNAVAILABLE',
--- a/js/ai/tests/model/middleware_test.ts
+++ b/js/ai/tests/model/middleware_test.ts
@@ -30,7 +30,7 @@ import {
 } from '../../src/model.js';
 import {
   CONTEXT_PREFACE,
-  __setRetryTimeout,
+  TEST_ONLY,
   augmentWithContext,
   fallback,
   retry,
@@ -40,6 +40,8 @@ import {
   type AugmentWithContextOptions,
 } from '../../src/model/middleware.js';
 import { defineProgrammableModel } from '../helpers.js';
+
+const { setRetryTimeout } = TEST_ONLY;
 
 initNodeFeatures();
 
@@ -188,7 +190,7 @@ describe('retry', () => {
       };
     };
 
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       callback();
       return 0 as any;
     });
@@ -219,7 +221,7 @@ describe('retry', () => {
       };
     };
 
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       callback();
       return 0 as any;
     });
@@ -242,7 +244,7 @@ describe('retry', () => {
       throw new GenkitError({ status: 'UNAVAILABLE', message: 'test' });
     };
 
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       callback();
       return 0 as any;
     });
@@ -267,7 +269,7 @@ describe('retry', () => {
       throw new Error('test error');
     };
 
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       callback();
       return 0 as any;
     });
@@ -335,7 +337,7 @@ describe('retry', () => {
     };
 
     let totalDelay = 0;
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       totalDelay += ms!;
       callback();
       return 0 as any;
@@ -369,7 +371,7 @@ describe('retry', () => {
     };
 
     let totalDelay = 0;
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       totalDelay += ms!;
       callback();
       return 0 as any;
@@ -410,7 +412,7 @@ describe('retry', () => {
     };
 
     let totalDelay = 0;
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       totalDelay += ms!;
       callback();
       return 0 as any;
@@ -451,7 +453,7 @@ describe('retry', () => {
     };
 
     let totalDelay = 0;
-    __setRetryTimeout((callback, ms) => {
+    setRetryTimeout((callback, ms) => {
       totalDelay += ms!;
       callback();
       return 0 as any;


js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,171 @@ export function augmentWithContext(
...
+   * @default ['UNAVAILABLE', 'DEADLINE_EXCEEDED', 'RESOURCE_EXHAUSTED', 'ABORTED', 'INTERNAL']
+   */
+  statuses?: StatusName[];
+  /**
+   * The initial delay between retries, in milliseconds.
+   * @default 0
+   */
+  initialDelayMs?: number;
+  /**
+   * The factor by which the delay increases after each retry (exponential backoff).
```
ifielker, line 268: seems odd that "exponential" backoff defaults to linear... maybe it should be 2 as a default so it is actually exponential...

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,171 @@ export function augmentWithContext(
...
+   */
+  maxRetries?: number;
+  /**
+   * An array of `StatusName` values that should trigger a retry.
+   * @default ['UNAVAILABLE', 'DEADLINE_EXCEEDED', 'RESOURCE_EXHAUSTED', 'ABORTED', 'INTERNAL']
+   */
+  statuses?: StatusName[];
+  /**
+   * The initial delay between retries, in milliseconds.
+   * @default 0
```
ifielker: I don't think this should be 0. Maybe 1000 (1 second)? Because if we leave it as everything default, it will retry 3 times immediately with no backoff and no delay which is not as helpful as it could be. Granted a working query might take a while, but for error like resource exhausted it will be very quick and 3 immediate retries won't help.

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,171 @@ export function augmentWithContext(
   };
 }
 
+/**
+ * Options for the `retry` middleware.
+ */
+export interface RetryOptions {
```
ifielker, line 246: I've seen nice retries where a bit of jitter (a random amount 0-1000 ms) is added to prevent synchronization of requests (e.g. if some server event causes synchronization of failed responses to multiple clients, then all those client will retry at exactly the same time if they use default values).

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,171 @@ export function augmentWithContext(
...
+  /**
+   * The maximum number of times to retry a failed request.
+   * @default 3
+   */
+  maxRetries?: number;
```
ifielker, line 251: I've also seen both maxDelay and maxRetries separately. Where after your delay gets to a certain maxDelay value (say 60 seconds) it stops doubling (or being multiplied by backoffFactor if it's not 2) and any retries after that are 60 seconds long. 
And this is separate from the maximum number of times you can retry.

js/ai/tests/model/middleware_test.ts
```diff
@@ -140,6 +142,303 @@ describe('validateSupport', () => {
   });
 });
 
+describe('retry', () => {
+  let registry: Registry;
```
ifielker, line 149: You'll probably need a fake clock here to properly test your initial backoff and subsequent exponential backoff. 
google-genai/tests/googleai/client_test.ts has an example: let clock: sinon.SinonFakeTimers;

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,171 @@ export function augmentWithContext(
...
+ * ```
+ */
+export function retry(options: RetryOptions = {}): ModelMiddleware {
+  const {
+    maxRetries = 3,
+    statuses = DEFAULT_RETRY_STATUSES,
+    initialDelayMs = 0,
+    backoffFactor = 1,
+    onError,
+  } = options;
```
ifielker, line 346: We should probably do some validation of values so we are not handing people a foot-gun (or accidental DOS-attacker). Especially maxRetries. If someone decides that maxRetries = 100000 is good and backoffFactor defaults to 1 and initialDelayMs defaults to 0; Then that's a bunch of useless and unnecessary traffic...
I think we need to make sure that either maxRetries is small or the combination of initial delay and backoff reaches say 60s within a smallish number of iterations?  We definitely need to have a reasonable maximum for maxRetries. e.g. not POSITIVE_INFINITY or anywhere near 1.7976931348623157e+308.

ssbushi (approved):

js/plugins/google-genai/src/googleai/client.ts
```diff
@@ -352,12 +353,31 @@ async function makeRequest(
...
+          break;
+      }
+      throw new GenkitError({
+        status,
+        message: `Error fetching from ${url}: [${response.status} ${response.statusText}] ${errorMessage}`,
+      });
     }
     return response;
   } catch (e: unknown) {
+    if (e instanceof GenkitError) {
```
ifielker, line 380: I think we still want this after the console.error line so we can still see errors. We might want to add a console.log(`retrying ${i} of ${maxRetries}`) somewhere in the retry as well?

js/plugins/google-genai/src/vertexai/client.ts
```diff
@@ -369,12 +370,31 @@ async function makeRequest(
...
+          break;
+      }
+      throw new GenkitError({
+        status,
+        message: `Error fetching from ${url}: [${response.status} ${response.statusText}] ${errorMessage}`,
+      });
     }
     return response;
   } catch (e: unknown) {
+    if (e instanceof GenkitError) {
```
ifielker, line 397: After the console.error line

js/ai/src/model/middleware.ts
pavelgj: there's no backoff by default... so it's 1.

js/ai/src/model/middleware.ts
pavelgj: I believe no delay is the correct behaviour by default...

js/ai/src/model/middleware.ts
pavelgj: feature request... you're welcome to implement. :)

js/ai/src/model/middleware.ts
pavelgj: feature request... you're welcome to implement. :)

js/plugins/google-genai/src/vertexai/client.ts
pavelgj: there should be no `console.error`... if anything, maybe, `logger.error` maybe. But it's up to whoever is upstream to log the error.

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,202 @@ export function augmentWithContext(
...
+  /**
+   * The maximum number of times to retry a failed request.
+   * @default 3
+   */
+  maxRetries?: number;
+  /**
+   * An array of `StatusName` values that should trigger a retry.
+   * @default ['UNAVAILABLE', 'DEADLINE_EXCEEDED', 'RESOURCE_EXHAUSTED', 'ABORTED', 'INTERNAL']
+   */
+  statuses?: StatusName[];
```
ifielker, line 256: This should be a subset of StatusName maybe RetryableStatusName that only includes things that should be retried. (i.e. not Permission Denied or 404 etc.)

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,202 @@ export function augmentWithContext(
...
+   * @default 60000
+   */
+  maxDelayMs?: number;
+  /**
+   * The factor by which the delay increases after each retry (exponential backoff).
+   * @default 2
+   */
+  backoffFactor?: number;
+  /**
+   * Whether to disable jitter on the delay.
```
ifielker: Maybe also a quick explanation of what jitter is.

js/ai/tests/model/middleware_test.ts
pavelgj: kept it simple

js/ai/src/model/middleware.ts
pavelgj: added maxDelay

js/ai/src/model/middleware.ts
pavelgj: added jitter

js/ai/src/model/middleware.ts
pavelgj: added default delay

js/ai/src/model/middleware.ts
pavelgj: changed it to 2

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,202 @@ export function augmentWithContext(
...
+ *       backoffFactor: 2,
+ *     }),
+ *   ],
+ * });
+ * ```
+ */
+export function retry(options: RetryOptions = {}): ModelMiddleware {
+  const {
+    maxRetries = 3,
+    statuses = DEFAULT_RETRY_STATUSES,
```
ifielker, line 340: Also need to validate the statuses as actually being retryable. No sense in retrying permission denied or not found etc.

js/ai/src/model/middleware.ts
pavelgj: backoffFactor is now 2, we have max delay and default delay is 1000

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,202 @@ export function augmentWithContext(
...
+            }
+          } else {
+            shouldRetry = true;
+          }
+
+          if (shouldRetry) {
+            onError?.(error, i + 1);
+            let delay = currentDelay;
+            if (!noJitter) {
+              delay = delay * (1 + Math.random());
```
ifielker: delay = delay + (1000 * Math.random()); then you get 0 - 1000 ms of extra delay, rather than 1x - 2x the amount of delay which could be much more than maxDelay if currentDelay is already maxDelay... A 1 second jitter is different than a random up to extra minute delay...

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,202 @@ export function augmentWithContext(
...
+export interface FallbackOptions {
+  /**
+   * An array of models to try in order.
+   */
+  models: ModelArgument[];
+  /**
+   * An array of `StatusName` values that should trigger a fallback.
+   * @default ['UNAVAILABLE', 'DEADLINE_EXCEEDED', 'RESOURCE_EXHAUSTED', 'ABORTED', 'INTERNAL']
+   */
+  statuses?: StatusName[];
```
ifielker, line 397: We might want to include different statuses here. Not Found, permission denied etc. make sense here if we are trying a different model with maybe different permissions or if we couldn't find the first one e.g. deprecated model

js/ai/src/model/middleware.ts
```diff
@@ -235,6 +240,202 @@ export function augmentWithContext(
...
+  onError?: (error: Error, attempt: number) => void;
+}
+
+let __setTimeout: (
+  callback: (...args: any[]) => void,
+  ms?: number
+) => NodeJS.Timeout = setTimeout;
+
+/**
+ * FOR TESTING ONLY.
```
ifielker, line 291: there's a nice pattern for this:

========== Source File ==============
// No export here
function setRetryTimeout( ... ) {...}

export const TEST_ONLY = {
  setRetryTimeout,
};

=========== Test File =============

import { TEST_ONLY } from ...;

const { setRetryTimeout } = TEST_ONLY;

js/ai/src/model/middleware.ts
ifielker: It's great that the defaults are good now, but we still need to stop bad combinations here...
There needs to be something stopping people from saying initialDelay = 0; maxRetries = POSITIVE_INFINITY; (Statuses = ['PERMISSION_DENIED'])

js/ai/tests/model/middleware_test.ts
```diff
@@ -140,6 +143,481 @@ describe('validateSupport', () => {
...
+    });
+
+    const result = await generate(registry, {
+      model: 'programmableModel',
+      prompt: 'test',
+      use: [
+        retry({
+          maxRetries: 2,
+          initialDelayMs: 50,
+          noJitter: false, // do hitter
```
ifielker: s/hitter/jitter

js/plugins/google-genai/src/vertexai/client.ts
ifielker: Please make this the same as for googleai.

js/ai/src/model/middleware.ts
pavelgj: it's up to the user.

js/ai/src/model/middleware.ts
pavelgj: No, up to the user.

js/ai/src/model/middleware.ts
pavelgj: Not sure we need to... there's an obvious malicious intent in that combination... and putting it in a `while(true)` loop will be much more efficient.

js/plugins/google-genai/src/vertexai/client.ts
pavelgj: Done.

js/ai/src/model/middleware.ts
pavelgj: changed jitter to value between 0 and 2^n seconds, where n is the retry attempt number

js/ai/src/model/middleware.ts
pavelgj: Done.

js/ai/src/model/middleware.ts
pavelgj: Done.

js/ai/src/model/middleware.ts
ifielker: I think a well meaning but uninformed person might think the best option is to retry as quickly as possible until it succeeds... Maybe at least enforce initialDelay > 1 (so it's not 0 * backoff or an increasingly smaller fraction of backoff) and maxRetries <= 25? Or some "reasonable" limit?

pavelgj: why is 25 reasonable but 26 isn't? why not 53? why not 87238763? I don't want to pull random numbers out of the air. We will add a limit if we there's a good reason to add one.

pavelgj: I'm not completely disagreeing with you, but this feels solidly as a P2 problem and solution requires a set of decisions that I currently don't have enough information about so hardcoding some limitations feels arbitrary, so I'm choosing to not make a decision on this right now. This is definitely not a blocker.

PR #3776, opened 2025-11-07
fix(js): switched google-genai from responseSchema to responseJsonSchema field

pavelgj:
To switch back to the old/partial schema:

```ts
const ai = genkit({
  plugins: [
    googleAI({ legacyResponseSchema: true }),
    vertexAI({ legacyResponseSchema: true }),
  ]
})
```

Also enabled poper constained generation for OpenAI models.
Also enabled recursive schemas.

Checklist (if applicable):

### DIFF
--- a/js/testapps/basic-gemini/src/index.ts
+++ b/js/testapps/basic-gemini/src/index.ts
@@ -14,13 +14,12 @@
  * limitations under the License.
  */
 
-import { googleAI, vertexAI } from '@genkit-ai/google-genai';
+import { googleAI } from '@genkit-ai/google-genai';
 import * as fs from 'fs';
 import {
   genkit,
   z,
   type MediaPart,
-  type ModelReference,
   type Operation,
   type Part,
   type StreamingCallback,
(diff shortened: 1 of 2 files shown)


js/plugins/google-genai/src/common/types.ts
```diff
@@ -960,6 +960,46 @@ export declare interface GenerationConfig {
...
+   * - `properties`
+   * - `additionalProperties`
+   * - `required`
+   *
+   * The non-standard `propertyOrdering` property may also be set.
+   *
+   * Cyclic references are unrolled to a limited degree and, as such, may only
+   * be used within non-required properties. (Nullable properties are not
+   * sufficient.) If `$ref` is set on a sub-schema, no other properties, except
+   * for than those starting as a `$`, may be set.
```
ifielker, line 1000: s/for than those/those/

s/starting as a/starting with a/

js/plugins/google-genai/src/common/types.ts
```diff
@@ -960,6 +960,46 @@ export declare interface GenerationConfig {
...
+   */
+  responseJsonSchema?: Record<string, any>;
```
ifielker, line 1002: Can it be Record<string, unknown>? instead?

js/testapps/basic-gemini/src/index.ts
```diff
@@ -14,29 +14,62 @@
  * limitations under the License.
  */
 
-import { googleAI } from '@genkit-ai/google-genai';
+import { googleAI, vertexAI } from '@genkit-ai/google-genai';
```
ifielker: Last time I tried putting vertexAI and googleAI samples in the same file, you told me I needed googleAI to be in one file and vertexAI in a different file so you could easily point to samples without confusion. So now there are two. Please use index-vertexai.ts for the vertex samples. You can swap which one gets used for genkit:dev in package.json. (And don't just delete the other one because vertex has stuff that gemini doesn't support and vice versa). If you want to change your mind and put everything in one file, then please include the index-vertexai stuff here too (and deduplicate it). (Some flows will then be defined or not based on the environment variable if they are only for vertex or only for googleai)

js/testapps/basic-gemini/src/index.ts
```diff
@@ -14,29 +14,62 @@
...
+const nanoBanana: ModelReference<any> = useVertex
+  ? vertexAI.model('gemini-2.5-flash-image')
+  : googleAI.model('gemini-2.5-flash-image');
+
+const imagen: ModelReference<any> = useVertex
+  ? vertexAI.model('imagen-3.0-generate-002')
+  : googleAI.model('imagen-3.0-generate-002');
+
+const geminiTTS: ModelReference<any> = useVertex
+  ? vertexAI.model('gemini-2.5-flash-preview-tts')
```
ifielker: This model doesn't exist as far as I know? TTS is only for gemini

js/plugins/google-genai/src/common/types.ts
pavelgj: Not my words, copy pasted from gemini api docs. If you feel strongly please use "Add a suggestion" button.

js/plugins/google-genai/src/common/types.ts
pavelgj: no, schema is `Record<string, any>` throughout.

js/testapps/basic-gemini/src/index.ts
pavelgj: didn't see index-vertexai.ts... undone these changes

Issue #3711, opened 2025-10-14
[Go] Genkit ignores dotprompt-defined roles


janishorsts:
**Describe the bug**

When the Genkit renders the dotprompt prompt, it always uses the "user" role and ignores the dotprompt-defined roles.

**To Reproduce**

Use the PromptWithMultiMessage from `go/samples/prompts` to reproduce.

**Expected behavior**

The dotprompt-defined roles are retained.

**Screenshots**

The following screenshot of debugging the `go/samples/prompts` sample shows that the system prompt `You are a great pirate and an AI assistant.` has been added as a user message after rendering.

<img width="2304" height="1296" alt="Image" src="https://github.com/user-attachments/assets/35d44f4d-3c1b-4b28-bcb2-7354776277a4" />

**Runtime (please complete the following information):**
 - OS: MacOS
 - Version 26

**Go version**

```shell
go version go1.25.2 darwin/arm64
```

**Additional context**

Relates to #3710

janishorsts:
The implementation fails when a partial containing a dotprompt role is used in a dotprompt. Otherwise, without partials, it works.

For example, this fails ❌

```handlebars
{{>pirate}}

<<<dotprompt:role:user>>>
Say hello
```

Where `pirate` is a partial

```handlebars
<<<dotprompt:role:system>>>
You are a pirate

hugoaguirre:
Hey @janishorsts thanks for the heads up. We will investigate this.

PR #3780, opened 2025-11-09
fix(go/ai): genkit ignores dotprompt-defined roles

josearomeroj:
Fixed Genkit's prompt load to respect role definitions from dotprompt files instead of forcing all messages to use the "user" role.

#3711 

Checklist (if applicable):

### DIFF
--- a/go/ai/prompt_test.go
+++ b/go/ai/prompt_test.go
@@ -1158,7 +1158,6 @@ model: test/chat
 description: A test prompt
 ---
 <<<dotprompt:role:system>>>
-
 You are a pirate!
 
 <<<dotprompt:role:user>>>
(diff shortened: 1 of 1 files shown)


go/ai/prompt_test.go
```diff
@@ -1148,3 +1148,65 @@ Hello!
...
+	mockPromptContent := `---
+model: test/chat
+description: A test prompt
+---
+<<<dotprompt:role:system>>>
+
+You are a pirate!
+
+<<<dotprompt:role:user>>>
+Hello!
```
hugoaguirre, line 1164: The test case is failing due to the string mismatch.

`dotprompt.ToMessages` trims all Unicode space chars except new lines, this means that the system prompt will look like this: `\nYou are a pirate\n` and the user message like this: `Hello!\n`

I think that, for the sake of the test case, it would be fine to trim these new lines.

josearomeroj: Hi @hugoaguirre, I have changed the test case to trim the newlines.

eric642:
@hugoaguirre @rutujaD1999 @apascal07 My actual verification did not solve this problem, and this submission introduced a new bug. The modified code has caused the shared variable to behave in such a way that when multiple threads access the same prompt simultaneously, the message remains unchanged indefinitely.
```
if len(nonSystemMessages) > 0 {
	promptOpts = append(promptOpts, WithMessages(nonSystemMessages...))
} 
```

apascal07:
@rutujaD1999 please take a look.

PR #3789, opened 2025-11-12
fix(genkit-tools/devui): handle unknown span status

### LINKED ISSUE #3788: [Tooling] Auto-instrumented spans (e.g. Firestore) render as errors instead of ‘Unknown’ when status is UNSET
**Describe the bug**  
In the Genkit Developer UI, auto-instrumented spans (for example, Firestore RPCs that come from `@opentelemetry/auto-instrumentations-node`) always render with the red error icon even when the RPC succeeds. These spans don’t set `genkit:state`, so the UI falls back to `span.status.code`, but the fallback currently treats every value other than `1` (`SpanStatusCode.OK`) as an error. Firestore RPC spans emit `status.code = 0` (`UNSET`), so they are misclassified as failures even though the operations succeed.

**To Reproduce**
1. Create any Genkit project (the public `npm create genkit-app` starter works) and enable Firestore auto instrumentation by using the Google Cloud telemetry plugin.
2. Define a flow that wraps a Firestore call so it shows up in the Dev UI traces (from [Defining AI workflows](https://genkit.dev/docs/develop/flows)). For example:
   ```ts
   import { z } from 'genkit';
   import { getFirestore } from 'firebase-admin/firestore';
   import { ai } from './genkit'; // wherever you set up genkit()

   export const firestoreProbe = ai.defineFlow(
     {
       name: 'firestoreProbe',
       inputSchema: z.object({ path: z.string() }),
       outputSchema: z.object({ exists: z.boolean() }),
     },
     async ({ path }) => {
       const snap = await getFirestore().doc(path).get();
       return { exists: snap.exists };
     }
   );
   ```
3. Launch the Dev UI with your usual dev server, e.g.
   ```bash
   npx genkit start -- npm run dev
   ```
4. Invoke the `firestoreProbe` flow from the Run tab (or programmatically). In the trace viewer the Firestore spans (`DocumentReference.get`, `BatchGetDocuments`, `Transaction.run`, etc.) show the red error icon even though the RPCs succeed.

**Expected behavior**  
Only spans with `status.code = 2` (`SpanStatusCode.ERROR`) should be displayed as failures. Spans with `status.code = 0` (`UNSET`) should match the production Genkit Monitoring dashboard and display the neutral “Unknown” state instead of an error.

**Screenshots**  
1. Current State: Dev UI shows Firestore spans marked with red error failure icons.

  Image

2. Production Dashboard (in Firebase Console): Genkit Monitoring displays the neutral question-mark indicator for the unknown-status spans.  

  Image

3. Proposed State: – Dev UI updated to render a gray question-mark chip for `genkit:state = unknown`.

  Image

**Desktop**
- OS: macOS 15.6.1
- Browser: Chrome 142.0.7444.60
- Genkit CLI/UI assets: 1.22.0 (`~/.genkit/assets/1.22.0/ui/browser`)

**Mobile**  
N/A

**Additional context**  
- Production Genkit Monitoring already shows these spans as “Unknown” when the status code is `UNSET`, so the inconsistency is limited to the Dev UI bundle that ships with the CLI.  
- The OpenTelemetry spec states that instrumentation libraries “Unset represents a span that completed without an error. Ok represents when a developer explicitly marks a span as successful. In most cases, it is not necessary to explicitly mark a span as Ok." (from https://opentelemetry.io/docs/concepts/signals/traces/#span-status), so Firestore RPC spans staying UNSET is expected, not a misconfiguration in our flow code, I think.
- The Dev UI JavaScript bundle currently does:
  ```ts
  span.status?.code !== 1 ? 'error' : 'success';
  ```
  Updating the fallback to treat only `code === 2` as an error (and map `0` to “Unknown”) would make the Dev UI consistent with the backend dashboard and avoid flagging successful Firestore operations as failures.

---

## Source references

These files document how span status codes are produced so reviewers can confirm the analysis without inspecting the minified UI bundle:

1. **OpenTelemetry status enum** – `@opentelemetry/api` defines `SpanStatusCode.UNSET = 0`, `SpanStatusCode.OK = 1`, and `SpanStatusCode.ERROR = 2`. See `node_modules/@opentelemetry/api/build/src/trace/status.js`.
2. **Genkit Google Cloud telemetry plugin** – the Dev UI receives spans that come from the `GcpOpenTelemetry` exporter, which enables `getNodeAutoInstrumentations(...)` and therefore emits spans for Firestore RPCs without setting `genkit:state`. Source: `js/plugins/google-cloud/src/gcpOpenTelemetry.ts` (around lines 149‑214 in the current main branch).

eddyleelin:
## Summary
- Fixes #3788
- Added `backfillSpanStates` so runtimes copy OpenTelemetry status codes into `genkit:state` for spans that were emitted via auto-instrumentation.
- Added Jest coverage for the helper and wired it into `RuntimeManager#getTrace` so Dev UI receives consistent state metadata.

⚠️⚠️⚠️ **IMPORTANT UI note:** the Dev UI source isn’t part of this repository; only the compiled assets are shipped. The question-mark icon for the `unknown` status will need to be added in Google’s private Dev UI codebase so the frontend mirrors production Monitoring. This PR should NOT be submitted by itself until the Dev UI can support the Unknown status. Otherwise, it will look like this (UI fails to show any traces starting at the first unknown status trace):

  image

For my local patch testing, I updated the Dev UI to manually edit ~/.genkit/assets/1.22.0/ui/browser/main-IC7V75ML.js to add `unknown:{class:"muted",icon:"help",label:"Unknown"}` to the status map so the question‑mark chip renders:

```py
from pathlib import Path

bundle = Path.home() / ".genkit/assets/1.22.0/ui/browser/main-IC7V75ML.js"
needle = ',icon:"hourglass_bottom",label:"Interrupted"},error:{class:"error",icon:"error",label:"Failed"}'
unknown = 'unknown:{class:"muted",icon:"help",label:"Unknown"}'

text = bundle.read_text()
if unknown not in text and needle in text:
    bundle.write_text(text.replace(needle, f"{needle},{unknown}", 1))
```

## Checklist

 Docs updated (updated docs or a docs bug required) - No docs need to be updated, as far as I know of.
- [x]  PR title is following https://www.conventionalcommits.org/en/v1.0.0/
- [x]  Tested (manually, unit tested, etc.) - Unit tested and manually tested:
  
  image

- [x] Docs: no changes required (behavior matches production Monitoring, no new user-facing API).

## Testing
- Installed the patched `@genkit-ai/tools-common` and `genkit-cli` tarballs and verified Firestore spans now appear with the neutral status in the Dev UI.

### DIFF
--- a/genkit-tools/common/src/utils/spanState.ts
+++ b/genkit-tools/common/src/utils/spanState.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2024 Google LLC
+ * Copyright 2025 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/genkit-tools/common/tests/utils/spanState_test.ts
+++ b/genkit-tools/common/tests/utils/spanState_test.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2024 Google LLC
+ * Copyright 2025 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


pavelgj:
Thanks for the PR. The bug was on the dev ui side and the fix was submitted and will go out in the next release (expecting today-tomorrow). I'm going to close this PR.

PR #3806, opened 2025-11-14
feat(py): Add Evaluator Metrics (ANSWER_RELEVANCY, FAITHFULNESS, MALICIOUSNESS)

hendrixmar:

Checklist (if applicable):

### DIFF
--- a/py/packages/genkit/src/genkit/blocks/model.py
+++ b/py/packages/genkit/src/genkit/blocks/model.py
@@ -48,9 +48,9 @@ from genkit.core.typing import (
     GenerateResponseChunk,
     GenerationUsage,
     Message,
+    ModelInfo,
     Part,
     ToolRequestPart,
-    ModelInfo,
 )
 
 # type ModelFn = Callable[[GenerateRequest], GenerateResponse]


py/packages/genkit/src/genkit/blocks/evaluator.py
```diff
@@ -39,3 +46,79 @@
 # User-provided batch evaluator function that evaluates an EvaluationRequest
 # type BatchEvaluatorFn[T] = Callable[[EvalRequest, T], list[EvalFnResponse]]
 BatchEvaluatorFn = Callable[[EvalRequest, T], list[EvalFnResponse]]
+
+
+
+def new_evaluator(
```
ssbushi: I haven;t been keeping up with Python changes, but why do we need this function? We already have a `define_evaluator` method here: https://github.com/firebase/genkit/blob/1adf3db8e5cb4a0788c22ccdda738e95bada1e73/py/packages/genkit/src/genkit/ai/_registry.py#L281

hendrixmar: You are right. I noticed it a week before. The part that is missing are the metrics types defined in the [GenkitEvaluator plugin](https://github.com/firebase/genkit/blob/d94b4a104ff87bf9b4af85e746cfbd5d4c464b16/py/plugins/evaluators/src/genkit/plugins/evaluators/plugin_api.py#L57)

MengqinShen:
@hendrixmar you might need to run './py/bin/generate_schema_typing' locally and then commit the changes to push the changes. Do you use pre-commit? It can prevent this failing by running checks locally first.

PR #3808, opened 2025-11-15
feat(js/ai): added support for tools that can return multipart responses

pavelgj:
```ts
      const t = ai.defineTool(
        { name: 'multipart-tool', description: 'description', multipart: true },
        async () => ({
          output: 'tool called',
          content: [{ media: { url: '...'} }],
        })
      );
```

Checklist (if applicable):

### DIFF
--- a/py/packages/genkit/src/genkit/core/typing.py
+++ b/py/packages/genkit/src/genkit/core/typing.py
@@ -823,14 +823,6 @@ class Part(
     )
 
 
-class MultipartToolResponse(BaseModel):
-    """Model for multiparttoolresponse data."""
-
-    model_config = ConfigDict(extra='forbid', populate_by_name=True)
-    output: Any | None = None
-    content: list[Part] | None = None
-
-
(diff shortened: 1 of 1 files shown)


pavelgj:
PR train: https://github.com/firebase/genkit/pull/3807

js/ai/src/tool.ts
```diff
@@ -470,8 +504,73 @@ export function dynamicTool<I extends z.ZodTypeAny, O extends z.ZodTypeAny>(
...
+      return interrupt();
+    }
+    return {
+      output: await fn(input, ctx),
+    };
+  });
+}
+
+export const MultipartToolResponseSchema = z.object({
+  output: z.any().optional(),
```
apascal07: This doesn't need to be hard typed?

pavelgj: No, it doesn't, but that reminded me --- this needs to be in the "shared" place... moved.

PR #3827, opened 2025-11-19
fix(go/ai): update `ModelResponseChunk.Index` JSON tag field

hugoaguirre:
Checklist (if applicable):

### DIFF
--- a/go/internal/cmd/jsonschemagen/jsonschemagen.go
+++ b/go/internal/cmd/jsonschemagen/jsonschemagen.go
@@ -43,11 +43,11 @@ var (
 	noFormat   = flag.Bool("nofmt", false, "do not format output")
 	configFile = flag.String("config", "", "config filename")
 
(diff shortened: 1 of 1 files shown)


go/ai/gen.go
```diff
@@ -271,7 +271,7 @@ type ModelResponseChunk struct {
 	Aggregated bool    `json:"aggregated,omitempty"`
 	Content    []*Part `json:"content,omitempty"`
 	Custom     any     `json:"custom,omitempty"`
-	Index      int     `json:"index,omitempty"`
+	Index      int     `json:"index"`
```
apascal07, line 274: Is this just so it's set to 0 if it's not set rather than be completely missing? Can't we set it in `GenerateWithRequest` if we see that it's missing?

hugoaguirre: Yeah, I tried setting it explicitly to zero but the `omitempty` tag causes it to be ignored if it has its zero value.

apascal07: Ah true, good look.

PR #3834, opened 2025-11-20
docs: use codewiki instead of deepwiki

hugoaguirre:
There's no badge available at the expected URL 😞


yesudeep (commented):
What do you think about keeping both? Deepwiki will also index our code if we retain this link?

Issue #3694, opened 2025-10-10
[PY] Please migrate to the release version of dotpromptz


yesudeep:
**Describe the bug**
Currently, the Python Genkit library is in alpha and uses the 
git repository URL for dotprompt for its integration:

**To Reproduce**
See: https://github.com/firebase/genkit/blob/main/py/pyproject.toml#L85

**Expected behavior**
The expected behavior is to use the production release version
of dotprompt from pypi.

**Screenshots**
N/A

**Runtime (please complete the following information):**
 - OS: Linux, MacOS
 - Genkit Version: Python alpha

** Python version
```
❯ python3 --version
Python 3.13.7
```

**Additional context**
N/A

hilariie:
#3928 dropped the git source for dotpromptz and switched the Python workspace to the released package from PyPI

PR #3838, opened 2025-11-22
chore(deps): use dotpromptz version 0.1.3 from pypi

MengqinShen:
For issue #3694 
CHAGELOG:
1. update pyproject.toml to use dotpromtz latest version from pypi 
2. update uv.lock


MengqinShen:
With the changes above, using "uv pip list | grep dotpromptz" shows the dotpromptz version, which doesn't match that in pyproject.toml. Needs to investigate a bit more to verify the changes are correct. 

Using Python 3.12.12 environment at: /Users/mq.shen/Documents/genkit/.venv
dotpromptz                          0.1.0

MengqinShen:
Close it as a new PR #3928 completed the same task

PR #3843, opened 2025-11-24
Create devcontainer.json

Diamond111990:

Checklist (if applicable):


MichaelDoyle:
Hey @Diamond111990 thanks! I don't think we're strictly _opposed_ to adding a dev container, but "as is" the basic container does not add much value in terms of setting up a ready-to-go dev environment.

Given we're in the process of breaking up the repo, I am going to close for now - but we can revisit later.

PR #3845, opened 2025-11-24
feat(js/plugins/google-genai): added support for tool request streaming for Gemini 3

pavelgj:
```ts
    const { response, stream } = ai.generateStream({
      model: vertexAI.model('gemini-3-pro-preview'),
      config: {
        functionCallingConfig: {
          streamFunctionCallArguments: true,
        },
      },
      tools: [...],
    });

      if (chunk.toolRequest?.partial) {
        console.log('partial'); 
      }  else {
        console.log('final'); 
      }
    }
```

Checklist (if applicable):


js/plugins/google-genai/src/common/converters.ts
```diff
@@ -354,10 +358,9 @@ function maybeAddThoughtSignature(geminiPart: GeminiPart, part: Part): Part {
 }
 
 function fromGeminiThought(part: GeminiPart): Part {
-  return {
+  return maybeAddThoughtSignature(part, {
     reasoning: part.text || '',
-    metadata: { thoughtSignature: part.thoughtSignature },
```
ifielker, line 359: I would say either a) leave it as is to make it clear that we are expecting a part.thoughtSignature as part of the thought Part. Or b) If you want to use maybe... then Error if it doesn't have a thought Signature. Otherwise it looks like a regular text Part which may or may not have a thought signature. I think this one is different because for all the others it's an optional thing, and for this one, it's required.

js/plugins/google-genai/src/common/converters.ts
```diff
@@ -400,34 +403,147 @@ function fromGeminiFileData(part: GeminiPart): Part {
...
+    } else if (partialArg.stringValue !== undefined) {
+      value = partialArg.stringValue;
+    }
+    if (value === undefined) {
+      continue;
+    }
+
+    let current: any = target;
+    const path = JSONPath.toPathArray(partialArg.jsonPath);
+    for (let i = 1; i < path.length - 1; i++) {
```
ifielker, line 438: Leave a comment about why we are starting with 1 and not 0.

js/plugins/google-genai/src/common/converters.ts
```diff
@@ -400,34 +403,147 @@ function fromGeminiFileData(part: GeminiPart): Part {
   });
 }
 
-function fromGeminiFunctionCall(part: GeminiPart, ref: string): Part {
+export function applyStreamingJsonPath(target: object, geminiPart: GeminiPart) {
```
ifielker: Please leave some comments describing what this does, it's not obvious from the title.

js/plugins/google-genai/src/common/converters.ts
```diff
@@ -400,34 +403,147 @@ function fromGeminiFileData(part: GeminiPart): Part {
...
+
+  // we try to find if there's a previous partial tool request part.
+  const prevPart = previousChunks?.at(-1)?.message.content?.at(-1);
+  const prevPartialToolRequestPart =
+    prevPart?.toolRequest && prevPart?.toolRequest.partial
+      ? prevPart
+      : undefined;
+
+  // if the current functionCall has partialArgs, we try to apply the diff to the
+  // potentilly including the previous partial part.
```
ifielker: potentilly? sp?

js/plugins/google-genai/src/common/converters.ts
```diff
@@ -400,34 +403,147 @@ function fromGeminiFileData(part: GeminiPart): Part {
...
+  if (part.functionCall.partialArgs) {
+    const newInput = prevPartialToolRequestPart?.toolRequest?.input
+      ? JSON.parse(JSON.stringify(prevPartialToolRequestPart.toolRequest.input))
+      : {};
+    applyStreamingJsonPath(newInput, part);
+    req.input = newInput;
+  }
+
+  // If there's a previous partial part, we copy some fields over, because the
+  // API wil not return these.
```
ifielker: s/wil/will/

js/plugins/google-genai/src/common/types.ts
```diff
@@ -296,18 +296,44 @@ export declare interface GenerativeContentBlob {
...
+   * `function_call` and return the response with the matching `id`.
+   */
+  id?: string;
   /** The name of the function specified in FunctionDeclaration.name. */
-  name: string;
+  name?: string;
   /** The arguments to pass to the function. */
-  args: object;
+  args?: object;
+  /** Optional. The partial argument value of the function call. If provided, represents the arguments/fields that are streamed incrementally. This field is not supported in Gemini API. */
```
ifielker: Instead of "This field is not support in Gemini API" at the end, we usually just start with "Vertex AI only."

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -416,7 +417,58 @@ async function getResponsePromise(
...
+
+  // If we're here, it's a regular, non-streaming tool call.
+  newPart.functionCall = part.functionCall;
+  return {
+    shouldContinue: false,
+    newActivePartialToolRequest: activePartialToolRequest,
+  };
+}
+
+export function aggregateResponses(
```
ifielker: If we are only exporting this for testing then please use the `export TEST_ONLY = { aggregateResponses }` pattern.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -424,7 +425,7 @@ const KNOWN_MODELS = {
   ...KNOWN_GEMMA_MODELS,
 };
 
-export function model(
+export function googleaiModelRef(
```
ifielker: why the rename? This was consistent with all the others. Please don't change it for no reason just to make it not match. Part of the niceness of this package is that everything is consistently named.

js/plugins/google-genai/src/googleai/index.ts
```diff
@@ -164,7 +164,7 @@ export const googleAI = googleAIPlugin as GoogleAIPlugin;
     return imagen.model(name, config);
   }
   // gemma, tts, gemini and unknown model families.
-  return gemini.model(name, config);
+  return gemini.googleaiModelRef(name, config);
```
ifielker: See this was nice before. googleAI.model delegates to veo.model or imagen.model or gemini.model. And now it's weird gemini.googleaiModelRef for only one of them. 

This nice pattern is throughout googleai plugin, vertexai plugin and modelgarden plugin. Please keep the nice pattern and leave it as gemini.model.

js/plugins/google-genai/src/vertexai/gemini.ts
```diff
@@ -252,6 +253,12 @@ export const GeminiConfigSchema = GenerationCommonConfigSchema.extend({
     .object({
       mode: z.enum(['MODE_UNSPECIFIED', 'AUTO', 'ANY', 'NONE']).optional(),
       allowedFunctionNames: z.array(z.string()).optional(),
+      /**
+       * When set to true, arguments of a single function call will be streamed out in
+       * multiple parts/contents/responses. Partial parameter results will be returned in the
+       * [FunctionCall.partial_args] field. This field is not supported in Gemini API.
```
ifielker, line 259: This is specific to vertexAI. We are not expecting anything here to be supported in another plugin (googleAI or otherwise). That was only for stuff in the common folder.

js/plugins/google-genai/src/vertexai/gemini.ts
```diff
@@ -399,7 +406,7 @@ export function isGeminiModelName(value?: string): value is GeminiModelName {
   return !!value?.startsWith('gemini-') && !value.includes('embedding');
 }
 
-export function model(
+export function vertexModelRef(
```
ifielker: no. See above.

js/plugins/google-genai/src/vertexai/index.ts
```diff
@@ -167,7 +167,7 @@ export const vertexAI = vertexAIPlugin as VertexAIPlugin;
     return veo.model(name, config);
   }
   // gemini and unknown model families
-  return gemini.model(name, config);
+  return gemini.vertexModelRef(name, config);
```
ifielker: vertexAI.model delegates to imagen.model, lyria.model, veo.model and now gemini.vertexModelRefWierdnessNonStandardEwwwww. :) Please put it back to just gemini.model :)

js/testapps/basic-gemini/src/index-vertexai.ts
```diff
@@ -352,7 +350,7 @@ ai.defineFlow(
         temperature: 2, // we want creativity
       },
       output: { schema: RpgCharacterSchema },
-      prompt: `Generate an RPC character called ${name}`,
+      prompt: `Generate sample data`,
```
ifielker: Where are the streaming function call samples?

js/plugins/google-genai/src/common/converters.ts
pavelgj: thoughtSignature is not a special feature of thinking parts. It's is required for other parts as well, so thoughts do not require special handling. We definitely should not be throwing errors when the API doesn't return something... if they don't return it, then it likely not meant to be returned.

js/plugins/google-genai/src/googleai/gemini.ts
pavelgj: The name of this function is very confusing/ambiguous -- it's not returning a "model", it's returning a "model ref", which is very important to be clear about. I assume this was copy-pasted from old implementation and back then it was exported to the end user, model refs were created like this:

```ts
import { model } from '@genkit-ai/googleai';

await generate({
  model: model('gemini-1.5-flash')
})
```

that was before we switched  to

```ts
import { googleAI } from '@genkit-ai/googleai';

await generate({
  model: googleAI.model('gemini-1.5-flash')
})
```

This function is no longer exported, so there's no reason to keep the naming "end user friendly". Clarity (that it's a ref, not model) is more important. I considered naming it `modelRef` but there's already `modelRef` imported from `'genkit/model'`.

js/plugins/google-genai/src/googleai/index.ts
pavelgj: see my comment above.

js/plugins/google-genai/src/vertexai/gemini.ts
pavelgj: WDYM? We 100% expect this to be implemented for googleAI.

js/plugins/google-genai/src/vertexai/gemini.ts
pavelgj: see my comment above.

js/plugins/google-genai/src/vertexai/index.ts
pavelgj: see my comment above.

image

js/testapps/basic-gemini/src/index-vertexai.ts
pavelgj: forgot to undo. Done.

js/plugins/google-genai/src/common/utils.ts
pavelgj: Done.

js/plugins/google-genai/src/common/types.ts
pavelgj: I removed "Vertex AI only." text. We expect this to be supported for googleAI at some point.

js/plugins/google-genai/src/common/converters.ts
pavelgj: Done.

js/plugins/google-genai/src/common/converters.ts
pavelgj: Done.

js/plugins/google-genai/src/googleai/gemini.ts
pavelgj: reverted.

PR #3858, opened 2025-11-27
feat(go/ai): make supportsConstrained function work with Model interface

eric642:
Refactor supportsConstrained to accept Model interface instead of *model type, enabling constrained output support for custom model implementations.


MichaelDoyle:
@apascal07 PTAL

Issue #3352, opened 2025-07-30
[JS] Upgrade to OpenTelemetry 2.0 sdk


MichaelDoyle:
See also: https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md

marcelfolaron:
I ran a quick trial locally but was blocked by `@google-cloud/opentelemetry-resource-util` requiring opentelemetry 1.x 
Particularly because of the change to the [resources api](https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md#-opentelemetryresources-api-changes)

Team seems to be aware [Issue 790](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/790) but I don't see any plans.

Haven't tried installing with `--legacy-peer-deps` but I am generally not a big fan of that anyways. 

Might want to call this blocked for now.

10kc-emilyk:
Looks like [`@google-cloud/opentelemetry-resource-util`](https://www.npmjs.com/package/@google-cloud/opentelemetry-resource-util)`@3.0.0` was released on Sept. 9. Does this unblock you from upgrading to opentelemetry v2? Thanks!

PR #3863, opened 2025-11-27
feat(tracing): upgrade to Otel 2.0+

MichaelDoyle:
BREAKING CHANGE: upgrade to Otel 2.0+

### Key Changes:

1. Core Telemetry Fixes (`js/core`)
   - **Initialization Race Condition**: Updated `enableTelemetry` in `tracing.ts` to correctly store and await the actual initialization promise. This prevents spans from being created before the SDK is fully ready.
   - **Improved Test Reliability**: Added `await flushTracing()` to `beforeEach` and relevant test cases in `action_test.ts` and `flow_test.ts`. This ensures all asynchronously emitted spans are captured before assertions are made.
   - **Version-Agnostic Span Processing**: Implemented a `MultiSpanProcessor` in `node-telemetry-provider.ts` to reliably handle multiple span processors across different OpenTelemetry versions.
   - **SDK Re-initialization**: Refined the `enableTelemetry` logic to properly shut down existing SDK instances before re-initializing, avoiding duplicate instrumentation and resource leaks.

2. OpenTelemetry 2.0 Compatibility (`js/plugins/google-cloud`)
   - **Traces Test**: Updated `traces_test.ts` to use `parentSpanContext?.spanId` instead of the deprecated `parentSpanId` property, which was removed in OTel 2.0.
   - **Metrics Test**: 
       - Updated `metrics_test.ts` to use the `InstrumentType` enum for metric type comparisons.
       - Addressed the removal of the `type` property from the `MetricDescriptor` interface.

Checklist (if applicable):

### AIM

image

### Dev UI

image


sakshamgoyal-01:
hey sorry for another ping. any plans to work on this again? it's throwing up a lot of security warnings for the old package versions and our CI is blocking us from using it because of that.

#3352 seems related but no updates on there either.

MichaelDoyle:
I see. Thanks for clarifying.

In that case, I’d recommend opening an issue describing the specific
problems you are seeing, so that we can prioritize it with the right
context.

Otel 2.0 is unfortunately going to require a larger refactoring that we are
in the planning stages for.

PR #3885, opened 2025-12-03
feat(go): Add cancelAction and early trace ID to headers in go reflection server

huangjeff5:
- Refactors runAction/streamAction to return headers early (including traceID)
- Added cancelAction endpoint to reflection server

### DIFF
--- a/go/core/schemas.config
+++ b/go/core/schemas.config
@@ -28,6 +28,7 @@ TimeEventAnnotation omit
 TraceData omit
 SpanStartEvent omit
 SpanEndEvent omit
+SpanEventBase omit
 TraceEvent omit
 
 GenerationCommonConfig.maxOutputTokens type int


go/core/api/action.go
```diff
@@ -27,6 +27,10 @@ type ActionRunResult[T any] struct {
 	SpanId  string
 }
 
+// TelemetryCallback is called when telemetry information becomes available.
+// It receives the trace ID and span ID as soon as the span is created.
+type TelemetryCallback func(traceID, spanID string)
```
apascal07:
```suggestion
type telemetryCallback = func(traceID, spanID string)
```

Does this need to be exported?

go/core/api/action.go
```diff
@@ -35,7 +39,8 @@ type Action interface {
 	// RunJSON runs the action with the given JSON input and streaming callback and returns the output as JSON.
 	RunJSON(ctx context.Context, input json.RawMessage, cb func(context.Context, json.RawMessage) error) (json.RawMessage, error)
 	// RunJSONWithTelemetry runs the action with the given JSON input and streaming callback and returns the output as JSON along with telemetry info.
-	RunJSONWithTelemetry(ctx context.Context, input json.RawMessage, cb func(context.Context, json.RawMessage) error) (*ActionRunResult[json.RawMessage], error)
+	// The telemetryCb callback, if provided, is called as soon as the trace span is created with the trace ID and span ID.
+	RunJSONWithTelemetry(ctx context.Context, input json.RawMessage, cb func(context.Context, json.RawMessage) error, telemetryCb TelemetryCallback) (*ActionRunResult[json.RawMessage], error)
```
apascal07: This is a breaking change...

apascal07: If I understand correctly, this is only used for reflection API? Can we pull that from the context instead so that this isn't a breaking change?

go/core/api/action.go
apascal07: Even unexported, this probably doesn't belong in `api`. This directory is specifically for API contracts between modules and this is an internal type.

go/genkit/reflection.go
```diff
@@ -334,35 +420,127 @@ func handleRunAction(g *Genkit) func(w http.ResponseWriter, r *http.Request) err
...
 		if err != nil {
-			if stream {
-				refErr := core.ToReflectionError(err)
-				refErr.Details.TraceID = &resp.Telemetry.TraceID
-				reflectErr, err := json.Marshal(refErr)
-				if err != nil {
-					return err
+			// Check if context was cancelled
+			if errors.Is(err, context.Canceled) {
+				errorResponse := map[string]interface{}{
```
apascal07: I think we have a struct type for this IIRC, slightly preferred to fill that and then marshal into JSON.

go/internal/cmd/jsonschemagen/jsonschema.go
```diff
@@ -33,7 +33,7 @@ type Schema struct {
 	Description          string                   `json:"description,omitempty"`
 	Properties           map[string]*Schema       `json:"properties,omitempty"`
 	AdditionalProperties *Schema                  `json:"additionalProperties,omitempty"`
-	Const                bool                     `json:"const,omitempty"`
+	Const                any                      `json:"const,omitempty"`
```
apascal07, line 36: Curious, what happened here?

go/genkit/reflection.go
```diff
@@ -334,35 +420,127 @@ func handleRunAction(g *Genkit) func(w http.ResponseWriter, r *http.Request) err
 			json.Unmarshal(body.Context, &contextMap)
 		}
 
-		resp, err := runAction(ctx, g, body.Key, body.Input, body.TelemetryLabels, cb, contextMap)
+		// Run the action with telemetry callback
+		resp, err := runAction(actionCtx, g, body.Key, body.Input, body.TelemetryLabels, cb, contextMap, telemetryCb)
+
+		// Clean up active action if we have a trace ID
+		if traceID != "" {
```
apascal07:
```suggestion
		if callbackTraceID != "" {
```

Was a bit hard to follow at a glance, I think this naming might clear it up a bit.

apascal07: Although now that I think about it, is there a reason why we shouldn't just be going off only `resp.Telemetry.TraceID` at this point? It should be filled even in the error case.

go/genkit/reflection.go
```diff
@@ -334,35 +420,127 @@ func handleRunAction(g *Genkit) func(w http.ResponseWriter, r *http.Request) err
...
+			}
 			return writeJSON(ctx, w, errorResponse)
 		}
 
-		return writeJSON(ctx, w, resp)
+		// Success case
+		if stream {
+			// For streaming, write the final chunk with result and telemetry
+			// This matches JS: response.write(JSON.stringify({result, telemetry}))
+			finalResponse := map[string]interface{}{
```
apascal07: Same with this I think.

go/internal/cmd/jsonschemagen/jsonschema.go
huangjeff5: https://github.com/firebase/genkit/pull/3801 updates genkit-schema.json
const now can take string values, rather than bool values
So we need to update jsonschema.go to be able to generate updated schemas

go/core/api/action.go
huangjeff5: Done, moved to go/core/tracing/tracing.go

go/genkit/reflection.go
huangjeff5: Good point, updated to simplify conditional

go/core/tracing/tracing.go
```diff
@@ -173,10 +173,12 @@ type SpanMetadata struct {
 
 // RunInNewSpan runs f on input in a new span with the provided metadata.
 // The metadata contains all span configuration including name, type, labels, etc.
+// If telemetryCb is provided, it will be called with the trace ID and span ID as soon as the span is created.
 func RunInNewSpan[I, O any](
 	ctx context.Context,
 	metadata *SpanMetadata,
 	input I,
+	telemetryCb func(traceID, spanID string),
```
apascal07: Can we revert this arg, unexport `TelemetryCb` so that it remains an internal detail (`WithTelemetryCb` is fine to expose, no harm), and just called it on the context here? Seems cleaner and more localized.

go/core/tracing/tracing.go
```diff
@@ -371,6 +379,20 @@ func (sm *spanMetadata) attributes() []attribute.KeyValue {
 // spanMetaKey is for storing spanMetadatas in a context.
 var spanMetaKey = base.NewContextKey[*spanMetadata]()
 
+// telemetryCbKey is the context key for telemetry callbacks.
+var telemetryCbKey = base.NewContextKey[func(traceID, spanID string)]()
+
+// WithTelemetryCb returns a context with the telemetry callback attached.
+// Used by the reflection server to pass callbacks to actions.
+func WithTelemetryCb(ctx context.Context, cb func(traceID, spanID string)) context.Context {
```
apascal07:
```suggestion
func WithTelemetryCallback(ctx context.Context, cb func(traceID, spanID string)) context.Context {
```

go/core/tracing/tracing.go
```diff
@@ -371,6 +379,20 @@ func (sm *spanMetadata) attributes() []attribute.KeyValue {
...
+	return telemetryCbKey.NewContext(ctx, cb)
+}
+
+// TelemetryCb retrieves the telemetry callback from context, or nil if not set.
+func TelemetryCb(ctx context.Context) func(traceID, spanID string) {
```
apascal07:
```suggestion
func telemetryCallback(ctx context.Context) func(traceID, spanID string) {
```

PR #3891, opened 2025-12-03
feat(genkit-tools/mcp): Add `genkit:init` command with support for runtime tools

ssbushi:
Checklist (if applicable):

### DIFF
--- a/genkit-tools/cli/src/mcp/prompts/init.ts
+++ b/genkit-tools/cli/src/mcp/prompts/init.ts
@@ -15,32 +15,18 @@
  */
 
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp';
+import { z } from 'zod';
 
-export function defineInitPrompt(server: McpServer) {
-  server.registerPrompt(
-    'genkit:init',
-    {
-      title: 'Initialize Genkit',
-      description: 'Initializes a new Genkit project',
-    },
-    () => ({
-      messages: [
-        {
-          role: 'user',
-          content: {
-            type: 'text',
-            text: `You are a Genkit expert. Help the user initialize a Genkit project.
-
-Follow these rules based on the user's environment and preference:
-
+const GENERAL_GUIDELINES = `
 ## General Guidelines
 - ALWAYS refer to documentation when available. Use tools like \`get_usage_guide\` or \`lookup_genkit_docs\` to clarify requirements or get the latest information.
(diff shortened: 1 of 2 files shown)


genkit-tools/cli/src/mcp/prompts/init.ts
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp';
+
+export function defineInitPrompt(server: McpServer) {
+  server.registerPrompt(
+    'genkit:init',
+    {
+      title: 'Initialize Genkit',
+      description: 'Initializes a new Genkit project',
```
pavelgj: I wonder if this should take `lang` parameter (can be optional), so we can return Js or Go specific init instructions.

Also (P2), perhaps we could support `model` options as well... for the model provider. We can default to gemini... but attempt to inject instructions for other providers.

genkit-tools/cli/src/mcp/runtime.ts
```diff
@@ -0,0 +1,99 @@
...
+export function defineRuntimeTools(
+  server: McpServer,
+  manager: McpRuntimeManager
+) {
+  server.registerTool(
+    'start_runtime',
+    {
+      title: 'Starts a Genkit runtime process',
+      description:
+        'Use this to start a Genkit runtime process. This will be picked up by the `genkit start` command to power the Dev UI features like model and flow playgrounds. The inputSchema for this tool matches the function prototype for `NodeJS.child_process.spawn`',
```
pavelgj: This might need a couple of examples... like `{command: 'go', args: ['run', 'main.go']}`

genkit-tools/cli/src/mcp/prompts/init.ts
ssbushi: Added the `lang` parameter.

I think we don't need to go into details about providers etc for the init command, its purpose should be to setup the project and all other modification should be done via the usage guide and docs. I was contemplating whether to include Gemini in here too, but I guess it is a good default path to support our own models. 

Instead of explicitly mentioning model providers, perhaps we can include guidance to check with the user if they want to use Gemini or while initializing, and if they decide not to, we can defer these changes to post-init, using the docs. 

WDYT?

ssbushi: Will address as follow-up

PR #3905, opened 2025-12-07
feat(go): refactored formatters + added support for formatting streams

apascal07:
Originally, formatters were implemented as something that modifies the model response message in place which was different from how it was implemented in JS. This was done in error as this solution was not capable of handling non-JSON formats (e.g. `ModelResponse.Output()` assumed that the content was JSON and since there was no formatter attached to the type, it was impossible to hook into the format-specific logic).

In this PR, we introduce a new interface called `StreamingFormatHandler` which, despite the name, is the general format handler interface. It implements `ParseOutput(*Message)` and `ParseChunk(*ModelResponseChunk)` which are used when the user calls `ModelResponse.Output()`, `ModelResponseChunk.Output()`, or the convenience function `OutputFrom[T any]()` without modifying content in place.

Checklist (if applicable):

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -783,23 +783,24 @@ func (mr *ModelResponse) Output(v any) error {
 		return errors.New("no content in response")
 	}
 
-	if mr.formatHandler != nil {
-		output, err := mr.formatHandler.ParseOutput(mr.Message)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


pavelgj:
@gemini-code-assist

/gemini review

PR #3907, opened 2025-12-08
fix(vertexai): use correct API URL for global location

### LINKED ISSUE #3651: [JS] Vertex AI plugin uses incorrect URL for global location in Genkit
**Describe the bug**
When using the Vertex AI plugin in Genkit and specifying location: 'global', the request URL becomes `https://global-aiplatform.googleapis.com`, which results in a 404 error. The correct URL should be `https://aiplatform.googleapis.com`.

It seems that the root cause is in this line of code:
https://github.com/firebase/genkit/blob/d25fc31b7b52717cc8021e4b04d7e0a206535dff/js/plugins/vertexai/src/list-models.ts#L41-L42

Here, when location is global, it should use `https://aiplatform.googleapis.com` instead of `https://global-aiplatform.googleapis.com`.

For reference, the official Google library handles this correctly:
https://github.com/googleapis/js-genai/blob/beef4e38c8df0c41957a3b3394aa75298c646ee2/src/_api_client.ts#L185C1-L196C4
```
private baseUrlFromProjectLocation(): string {
    if (
      this.clientOptions.project &&
      this.clientOptions.location &&
      this.clientOptions.location !== 'global'
    ) {
      // Regional endpoint
      return `https://${this.clientOptions.location}-aiplatform.googleapis.com/`;
    }
    // Global endpoint (covers 'global' location and API key usage)
    return `https://aiplatform.googleapis.com/`;
}
```
**To Reproduce**
Steps to reproduce the behavior. 
```
import { vertexAI } from '@genkit-ai/google-genai';

const ai = genkit({
  plugins: [
    vertexAI({ location: 'global' }),
  ],
});
```
When attempting to list or use models with location: 'global', the request fails with a 404 error.

**Expected behavior**
When `location` is set to 'global', the request should go to `https://aiplatform.googleapis.com` instead of `https://global-aiplatform.googleapis.com`, avoiding the 404 error.

**Screenshots**

 - OS: MacOS
 - Version 14.3

** Node version
 -  v23.1.0

janplancraft:
Fixes the Vertex AI plugin to use https://aiplatform.googleapis.com instead of https://global-aiplatform.googleapis.com when location: 'global' is specified.

Fixes #3651

Changes:
- list-models.ts - Model listing endpoint
- predict.ts - Predict endpoint
- upsert_datapoints.ts - Vector search upsert endpoint
- evaluator_factory.ts - Evaluation endpoint
- model_garden.ts (legacy) - Model garden template URL
- llama.ts (v2) - Llama model template URL

### DIFF
--- a/js/plugins/vertexai/src/evaluation/evaluator_factory.ts
+++ b/js/plugins/vertexai/src/evaluation/evaluator_factory.ts
@@ -82,7 +82,10 @@ export class EvaluatorFactory {
 
         metadata.input = request;
         const client = await this.auth.getClient();
-        const url = `https://${this.location}-aiplatform.googleapis.com/v1beta1/${locationName}:evaluateInstances`;
+        const baseUrl = this.location === 'global'
+          ? 'https://aiplatform.googleapis.com'
+          : `https://${this.location}-aiplatform.googleapis.com`;
+        const url = `${baseUrl}/v1beta1/${locationName}:evaluateInstances`;
         const response = await client.request({
           url,
           method: 'POST',
--- a/js/plugins/vertexai/src/list-models.ts
+++ b/js/plugins/vertexai/src/list-models.ts
@@ -38,8 +38,11 @@ export async function listModels(
 ): Promise<Model[]> {
   const fetch = (await import('node-fetch')).default;
   const accessToken = await authClient.getAccessToken();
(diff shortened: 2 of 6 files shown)


janplancraft:
We have migrated to `@genkit-ai/google-genai` and with that the issue is resolved.

PR #3910, opened 2025-12-09
refactor(go/plugins/compat_oai): normalize plugin

hugoaguirre:
Normalize the `compat_oai` plugin to match the rest of the plugins architecture.

- Removed `ModelGenerator` and all its attached functions
- Removed `assert` use from live tests
- Removed the _builder_ pattern from the plugin.
- Use translation functions as used in the rest of the plugins

Checklist (if applicable):


hugoaguirre:
The following issue (https://github.com/firebase/genkit/issues/4683) is addressed in this refactor.

MichaelDoyle:
@apascal07 is this something still needed?

PR #3921, opened 2025-12-11
fix(js/plugins/googleai,vertexai): support thoughtSignature in fromGeminiPart

### LINKED ISSUE #3920: [JS] Unsupported GeminiPart type error with Gemini 3 Pro
**Describe the bug**

Gemini 3 Pro (gemini-3-pro-preview) employs dynamic thinking by default, which cannot be disabled. Consequently, the model API returns response parts containing the thoughtSignature property to represent this thinking process.

However, the current fromGeminiPart implementation in @genkit-ai/googleai and @genkit-ai/vertexai does not strictly account for parts that contain thoughtSignature without the thought property. This leads to a crash when consuming responses from Gemini 3 Pro.

Error Message
```
Error: Unsupported GeminiPart type: {"thoughtSignature":"..."}
    at fromGeminiPart (gemini.ts:1026:9)
    ...
```

**To Reproduce**
1. Initialize Genkit with @genkit-ai/vertexai or @genkit-ai/googleai.
2. Generate content using the gemini-3-pro-preview model.

```
import { vertexAI } from '@genkit-ai/vertexai';

const ai = genkit({
  plugins: [vertexAI()],
});

await ai.generate({
  model: 'vertexai/gemini-3-pro-preview', 
  prompt: "How does AI work?",
  config: {
    thinkingConfig: {
      thinkingLevel: "low",
    }
  },
});
```

**Expected behavior**

The plugin should recognize the thoughtSignature property in GeminiPart and map it correctly using the existing fromThought (or fromGeminiThought) handler, ensuring the model response is processed without error.

- OS: MacOS
- Node version: v22.17.0

** Node version
 -  run `node --version` at paste here

**Additional context**

Per official documentation, Gemini 3 Pro uses thinking by default. The SDK needs to support the thoughtSignature part type which accompanies this behavior.

yeonjaee:
Fixes #3920

**Problem**

When using the Gemini 3 Pro model (e.g., `gemini-3-pro-preview`) which enables thinking by default, the API returns response parts containing a `thoughtSignature` property. The current fromGeminiPart implementation in both `googleai` and `vertexai` plugins throws an `Unsupported GeminiPart type` error because it does not recognize parts that contain `thoughtSignature` without the `thought` property.

**Solution**

This PR updates the fromGeminiPart logic in both plugins to correctly identify and handle `thoughtSignature` parts by checking for both `'thought' in part` and `'thoughtSignature' in part`. This allows these parts to be processed by the existing fromThought / fromGeminiThought functions, which already support the `thoughtSignature` property.

**Checklist:**
- [x] Tested (Manually verified with `gemini-3-pro-preview` - error resolved, responses with `thoughtSignature` now parse correctly)
- [ ] Docs updated (N/A - internal fix, no API changes)

### DIFF
--- a/js/plugins/googleai/src/gemini.ts
+++ b/js/plugins/googleai/src/gemini.ts
@@ -1016,7 +1016,7 @@ function fromGeminiPart(
   jsonMode: boolean,
   ref: string
 ): Part {
-  if ('thought' in part) return fromThought(part as any);
+  if ('thought' in part || 'thoughtSignature' in part) return fromThought(part as any);
   if (typeof part.text === 'string') return { text: part.text };
   if (part.inlineData) return fromInlineData(part);
   if (part.functionCall) return fromFunctionCall(part, ref);
--- a/js/plugins/vertexai/src/gemini.ts
+++ b/js/plugins/vertexai/src/gemini.ts
@@ -888,7 +888,7 @@ function fromGeminiPart(
   jsonMode: boolean,
   ref?: string
 ): Part {
-  if ('thought' in part) return fromGeminiThought(part as any);
+  if ('thought' in part || 'thoughtSignature' in part) return fromGeminiThought(part as any);
   if (typeof part.text === 'string') return { text: part.text };
   if (part.inlineData) return fromGeminiInlineDataPart(part);
   if (part.fileData) return fromGeminiFileDataPart(part);


MichaelDoyle:
Thanks! I’m not totally surprised, but this does seem a little bit curious. Do you have a trace where this happened?

yeonjaee:
> Thanks! I’m not totally surprised, but this does seem a little bit curious. Do you have a trace where this happened?

Thanks for looking into this!

The legacy plugins (@genkit-ai/googleai, @genkit-ai/vertexai) currently have a known issue with the global endpoint (see #3651), so I had to patch things locally to build and test.

The issue manifests specifically in the `generate trace path`, and i think providing an exact trace isn’t appropriate here since reproducing it requires additional code changes to address that separate global-endpoint issue.

---

I understand and agree with the project’s direction toward `@genkit-ai/google-genai` as the intended replacement, and I appreciate the clarification from the maintainer.

That said, while the legacy plugins are labeled as such, they are still being released and actively used. In that context, I think closing issues solely by pointing to the replacement, rather than adding guardrails or safer failure modes for existing users of the legacy plugins, can be rough from a developer experience standpoint.

Since the maintainer has closed the issue #3920 and confirmed that `@genkit-ai/google-genai` fully resolves this, I’m fine with this PR being closed as well — whether by me or by the maintainers — and I’ll go ahead and close it on my side. That said, if there’s still interest in adding guardrails for legacy plugin users in the future, I’d be happy to revisit this.

MichaelDoyle:
@yeonjaee sorry for the rough experience. Is there something that would prevent you from switching over to the new plugin? Or more of an issue in communication?

I'm on the Firebase/Google team as well. I'll relay this to the others.

yeonjaee:
@MichaelDoyle 
Thanks for the follow-up, and I appreciate you'll relay this internally.

There’s nothing fundamentally preventing me from switching — and I’m not opposed to the direction. My hesitation is simply based on my own experience that switching plugins isn’t always a trivial response to a bug, especially when reproducing or validating behavior across integrations.

So for me it’s less about communication, and more about the friction I experienced when trying to diagnose the issue.

Thanks again for taking this seriously.

MichaelDoyle:
Of course. I understand completely.

The new plugin has been out for a while now, so it’s definitely on us if
folks didn’t get the message to switch much earlier. That’s really what I
mean by communication.

The SDK that underpins the googleai and vertexai plugins is no longer being
updated, so it won’t be possible to fix forward for issues like this now or
in the future.

yeonjaee:
@MichaelDoyle 

I really appreciate the transparency around the underlying SDK situation, and it helps put the context together.
Thanks again for taking the time to explain this and for the thoughtful responses throughout the thread.

MichaelDoyle:
We've since been more forthcoming on the deprecation plan. Closing out.

PR #3927, opened 2025-12-11
feat(js/core): DAP expansion: Dynamic Actions are now expanded in the…

ifielker:
… action list.

Checklist (if applicable):

### DIFF
--- a/js/core/src/registry.ts
+++ b/js/core/src/registry.ts
@@ -384,11 +384,17 @@ export class Registry {
     for (const [key, action] of Object.entries(await this.listActions())) {
       resolvableActions[key] = action.__action;
       if (isDynamicActionProvider(action)) {
-        // Include the dynamic actions
-        const dapPrefix = `${action.__action.actionType}/${action.__action.name}`;
-        const dapMetadataRecord =
-          await action.getActionMetadataRecord(dapPrefix);
-        resolvableActions = { ...resolvableActions, ...dapMetadataRecord };
+        try {
+          // Include the dynamic actions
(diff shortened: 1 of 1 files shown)


js/core/src/registry.ts
```diff
@@ -380,9 +380,16 @@ export class Registry {
...
       })
     );
-    // Also add actions that are already registered.
+    // Also add actions that are already registered, and expand DAP actions
     for (const [key, action] of Object.entries(await this.listActions())) {
       resolvableActions[key] = action.__action;
+      if (isDynamicActionProvider(action)) {
+        // Include the dynamic actions
+        const dapPrefix = `${action.__action.actionType}/${action.__action.name}`;
+        const dapMetadataRecord =
```
pavelgj: Will every time the dev ui call listActions result in a trace for dap call? Let's make sure that it doesn't.

ifielker: Good call! Done.

pavelgj: wrap this in try-catch and log the error, but not rethrow. We don't want dev-ui to fail if DAP is failing.

pavelgj (approved):

js/core/src/registry.ts
ifielker: But we throw errors above when we load the models dynamically... why is this different when we load the DAP actions dynamically?

pavelgj: we don't throw on list, but we do throw on plugin initialize

ifielker: I saw the throw above and missed the try... oops.
Thanks! Done.

PR #3946, opened 2025-12-16
feat(js/plugins/compat-oai): allow passing api key at runtime instead of config time

pavelgj:
```ts
const ai = genkit({
  plugins: [openAI({apiKey: false})],
});


await ai.generate({
  prompt: 'tell me why...',
  model: openAI.model('gpt-4o'),
  config: {
    apiKey: '...'
  }
});
```

Checklist (if applicable):

### DIFF
--- a/js/plugins/compat-oai/tests/compat_oai_test.ts
+++ b/js/plugins/compat-oai/tests/compat_oai_test.ts
@@ -1479,7 +1479,7 @@ describe('openAIModelRunner', () => {
     };
     const runner = openAIModelRunner(
       'gpt-4o',
-      openaiClient as unknown as OpenAI,
+      openaiClient as unknown as OpenAI
     );
     await runner({ messages: [] });
     expect(openaiClient.chat.completions.create).toHaveBeenCalledWith(


js/plugins/compat-oai/tests/compat_oai_test.ts
ssbushi, line 1: test for request scoped api key?

pavelgj:
> What is the point of supporting per-request API keys?

Folks build apps where user provides the API key. The key is dynamically fetched from a DB.

js/plugins/compat-oai/tests/compat_oai_test.ts
pavelgj: Done.

js/plugins/compat-oai/tests/compat_oai_test.ts
```diff
@@ -1574,6 +1585,146 @@ describe('openAIModelRunner', () => {
...
+              finish_reason: 'stop',
+            },
+          ],
+        },
+      });
+
+      const defaultClient = new OpenAI({ apiKey: 'default-key' });
+      const runner = openAIModelRunner('gpt-4o', defaultClient, undefined, {
+        name: 'openai',
+        apiKey: 'test',
```
ssbushi: Looks like `PluginOptions.apiKey ` is not used as a fallback in this implementation of `maybeCreateRequestScopedOpenAIClient`?

This may be confusing since `apiKey` is configurable in PluginOptions too....

We can either:

1. Respect the `apiKey` from plugin options (request scoped -> plugin options -> default Client API key)
2. Or we can omit API key being configurable in `openAIModelRunner`'s PluginOptions arg.

ssbushi (approved):
LGTM, save for 1 small comment.

js/plugins/compat-oai/tests/compat_oai_test.ts
pavelgj: I know it's a little confusing, but it's necessary... the idea is that pluginOptions are actually originally used to create defaultClient, and those pluginOptions may already contain an apiKey, but in addition to the api key they can contain other critical options, like baseUrl, timeout, etc. which we need in maybeCreateRequestScopedOpenAIClient in case we want to create a request scoped client.

I went ahead and change the `openAIModelRunner` signature to:

`pluginOptions?: Omit<PluginOptions, 'apiKey'>` 

as we care about PluginOptions, just not the apiKey as the api key is provided at runtime.

There are actually cleaner ways to do this all, but the problem is that `client` is part of the PluginOptions initializer/resolver/listActions signature, so trying to keep it backwards compatible creates a little bit of a challenge.

PR #3953, opened 2025-12-16
fix(js/plugins/anthropic): budgetTokens not working in UI

CorieW:
Before:
image

After:
image
image
image


js/plugins/anthropic/src/types.ts
```diff
@@ -93,15 +93,26 @@ export type AnthropicBaseConfigSchemaType = typeof AnthropicBaseConfigSchema;
 export const ThinkingConfigSchema = z
   .object({
     enabled: z.boolean().optional(),
-    budgetTokens: z.number().int().min(1_024).optional(),
+    budgetTokens: z.number().min(1_024).optional(),
   })
   .superRefine((value, ctx) => {
```
pavelgj, line 98: I'm not sure if we apply these anywhere... (we may not be calling .parse for this schema)

PR #3954, opened 2025-12-16
fix(js/plugins/anthropic):  use beta models list method in beta runner 

cabljac:
This PR makes sure the beta runner uses the correct `list` method to list models from the API.

manually tested:

image


pavelgj:
@gemini-code-assist

PR #3959, opened 2025-12-16
fix: handle stream errors in runAction to prevent circular JSON crash

huangjeff5:
If the runtime returns an error stream, that was causing tools server to crash with "Converting circular structure to JSON" crash error. This makes sense because you can't call JSON.parse on a stream, it's not serializable due to circular references. This PR ensures that the error passed to httpErrorHandler is serializable.

### DIFF
--- a/genkit-tools/common/src/manager/manager.ts
+++ b/genkit-tools/common/src/manager/manager.ts
@@ -228,7 +228,9 @@ export class RuntimeManager {
             responseType: 'stream',
           }
         )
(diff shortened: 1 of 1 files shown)


MichaelDoyle:
Related? https://github.com/genkit-ai/genkit/pull/4005

MichaelDoyle:
@huangjeff5 do we still want this? cc @ifielker 

Also, please see if this is/isn't related to https://github.com/genkit-ai/genkit/pull/4005

PR #3971, opened 2025-12-17
feat(py): add sample test for directory and file prompt loading

MengqinShen:
CHANGELOG:
1. Add prompt files to test variant, name and extension
2. Create a test file for prompt loading test
3. Add GoogleAI as plugin
4. Update root uv.lock and pyproject.toml


py/samples/prompt_demo/pyproject.toml
```diff
@@ -0,0 +1,55 @@
+# Copyright 2025 Google LLC
```
yesudeep, line 1: need to update the root uv.lock file

MengqinShen: done

PR #3988, opened 2025-12-19
fix:(js/core/schema): Allow disabling runtime schema compilation

ssbushi:
image

Code:

```
import { googleAI } from "@genkit-ai/google-genai";
import { genkit, disableSchemaCodeGeneration } from "genkit";

const ai = genkit({
  plugins: [googleAI({ apiKey: "..." })],
  model: googleAI.model("gemini-2.5-flash", {
    temperature: 0.8,
  }),
});

disableSchemaCodeGeneration();

export interface Env {}

export default {
  async fetch(
    request: Request,
    env: Env,
    ctx: ExecutionContext
  ): Promise<Response> {
      prompt: `Tell me a joke about banana`,
    });
    return new Response(JSON.stringify({ message: text }), {
      headers: {
        "content-type": "application/json;charset=UTF-8",
      },
    });
  },
};
```

Fixes: https://github.com/firebase/genkit/issues/3968

Checklist (if applicable):

### DIFF
--- a/js/core/src/schema.ts
+++ b/js/core/src/schema.ts
@@ -17,7 +17,7 @@
 import { Validator } from '@cfworker/json-schema';
 import Ajv, { type ErrorObject, type JSONSchemaType } from 'ajv';
 import addFormats from 'ajv-formats';
-import { z, type ZodIssue } from 'zod';
+import { z } from 'zod';
 import zodToJsonSchema from 'zod-to-json-schema';
 import { GenkitError } from './error.js';
 import { logger } from './logging.js';
@@ -37,6 +37,11 @@ export function disableSchemaCodeGeneration() {
     "It looks like you're trying to disable schema code generation. Please ensure that the '@cfworker/json-schema' package is installed: `npm i --save @cfworker/json-schema`"
   );
   global[SCHEMA_VALIDATION_MODE] = 'interpret';
+}
+
+/** Visible for testing */
+export function resetSchemaCodeGeneration() {
+  global[SCHEMA_VALIDATION_MODE] = undefined;
 }
 
 export { z }; // provide a consistent zod to use throughout genkit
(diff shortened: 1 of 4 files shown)


js/core/package.json
```diff
@@ -26,25 +26,26 @@
   "author": "genkit",
   "license": "Apache-2.0",
   "dependencies": {
+    "@cfworker/json-schema": "^4.1.1",
```
pavelgj: Let's move it to optional dependencies... we document that if someone wants to use disableSchemaCodeGeneration then they must `npm i @cfworker/json-schema`

js/core/src/schema.ts
```diff
@@ -14,15 +14,27 @@
...
 import addFormats from 'ajv-formats';
-import { z } from 'zod';
+import { z, type ZodIssue } from 'zod';
 import zodToJsonSchema from 'zod-to-json-schema';
 import { GenkitError } from './error.js';
 import type { Registry } from './registry.js';
 const ajv = new Ajv();
 addFormats(ajv);
 
+let validationMode: 'compile' | 'interpret' = 'compile';
```
pavelgj: "global" vars like this are unreliable when used with code bundlers (webpack, turbopack, etc). Use `global`.

Ex: https://github.com/firebase/genkit/blob/b0ba327d7970a038c9f9216aba1a5c0c7304b7a1/js/core/src/async-context.ts#L25

js/core/tests/schema_test.ts
```diff
@@ -162,3 +166,39 @@ describe('toJsonSchema', () => {
     );
   });
 });
+
+describe('disableSchemaCodeGeneration()', () => {
+  it('should validate using cfworker validator', () => {
+    const compileMock = mock.method(Ajv.prototype, 'compile');
+
+    disableSchemaCodeGeneration();
```
pavelgj, line 173: this does not get reset... running this test out of order will impact other tests.

js/plugins/google-cloud/tests/traces_test.ts
```diff
@@ -105,8 +105,8 @@ describe('GoogleCloudTracing', () => {
 
   it('sub actions are contained within flows', async () => {
     const testFlow = createFlow(ai, 'testFlow', async () => {
-      return await ai.run('subAction', async () => {
```
pavelgj, line 108: why? does it break?

js/core/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../tsconfig.json",
   "include": ["src"],
   "compilerOptions": {
+    "skipLibCheck": true,
```
pavelgj: is this necessary?

pavelgj (approved):

PR #3992, opened 2025-12-20
feat(py): loading prompts automatically

MengqinShen:
CHANGELOG:
1. Update main function of the prompt demo sample, all prompts are loaded automatically 
2. Update prompt demo : there are two flows can be tested

### DIFF
--- a/py/samples/prompt_demo/prompts/nested/nested_hello.prompt
+++ b/py/samples/prompt_demo/prompts/nested/nested_hello.prompt
@@ -1,5 +1,5 @@
 ---
-model: googleai/gemini-2.0-flash
+model: googleai/gemini-2.5-flash
 input:
   schema:
     name: string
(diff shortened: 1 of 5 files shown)


py/samples/prompt_demo/prompts/nested/nested_hello.prompt
```diff
@@ -1,5 +1,5 @@
 ---
-model: googleai/gemini-1.5-flash
+model: googleai/gemini-2.0-flash
```
yesudeep: can we please update the model to 2.5-flash for now and go ahead?

py/samples/prompt_demo/src/prompt_demo.py
```diff
@@ -20,22 +20,27 @@
...
 
 logger = structlog.get_logger(__name__)
 
 
-# Initialize with GoogleAI plugin
-ai = Genkit(plugins=[GoogleAI()])
+current_dir = Path(__file__).resolve().parent
+prompts_path = current_dir.parent / 'prompts'
 
+ai = Genkit(plugins=[GoogleAI()], model='googleai/gemini-2.0-flash', prompt_dir=prompts_path)
```
yesudeep: this should be the default right? "<current_dir>/prompts"

PR #3998, opened 2025-12-22
fix(prompt): update role syntax in multi-msg prompt for consistency

sahdev77:
 - [linking to an issue that includes more details](https://github.com/firebase/genkit/issues/3710)

Checklist (if applicable):

### DIFF
--- a/go/samples/prompts/prompts/multi-msg.prompt
+++ b/go/samples/prompts/prompts/multi-msg.prompt
@@ -6,11 +6,11 @@ input:
 output:
   summary: string
 ---
-{{ role "system" }}
+<<<dotprompt:role:system>>>
 
 You are a great AI assistant that summarizes videos talking as a pirate
 
-{{ role "user" }}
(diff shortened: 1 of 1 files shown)


go/samples/prompts/prompts/multi-msg.prompt
```diff
@@ -6,11 +6,11 @@ input:
 output:
   summary: string
 ---
-{{ role "system" }}
+<<<dotprompt:role:system>>>
 
 You are a great AI assistant that summarizes videos talking as a pirate
 
-{{ role "user" }}
```
MichaelDoyle, line 13: This is not correct; `{{role: "user"}}` is indeed the correct/expected syntax. There are bug(s) in message rendering that make it appear that `{{role: "user"}}` is not supported.

MichaelDoyle:
Hey @sahdev77 thanks for the PR. Should not be necessary. We've corrected the rendering issue(s) in https://github.com/firebase/genkit/pull/4035.

PR #3975, opened 2025-12-18
feat(go/plugins/googlegenai): add multi-part tools support

hugoaguirre:

PR #4006, opened 2025-12-24
fix(go/plugins/googlegenai): include `thoughtSignature` in part metadata

hugoaguirre:
Checklist (if applicable):


hugoaguirre:
Added `signature` to the `Thought` parts. Gemini documentation does not state otherwise.
Since we are manipulating Gemini parts internally, we should preserve `Thought` parts as they are.

mquinteiro:
Hi @hugoaguirre  looks like the validation is stalled in a bad formatted go file

hugoaguirre:
@mquinteiro thanks for the heads up. It seems Gemini suggestions will cause format issues.
The issue is now a test case failure that is being corrected in #3975

apascal07 (approved):

PR #4027, opened 2025-12-30
feat(genkit-tools/mcp): MCP fix for Antigravity

ssbushi:
- Accept projectRoot, if in AGY

Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/js/plugins/compat-oai/src/audio.ts
+++ b/js/plugins/compat-oai/src/audio.ts
@@ -30,7 +30,7 @@ import type {
   TranscriptionCreateParams,
 } from 'openai/resources/audio/index.mjs';
 import { PluginOptions } from './index.js';
-import { maybeCreateRequestScopedOpenAIClient } from './utils.js';
+import { maybeCreateRequestScopedOpenAIClient, toModelName } from './utils.js';
 
 export type SpeechRequestBuilder = (
   req: GenerateRequest,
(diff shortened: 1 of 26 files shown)


genkit-tools/cli/src/commands/mcp.ts
```diff
@@ -19,23 +19,38 @@ import {
...
     if (options.debug) {
-      debugToFile();
+      debugToFile(
+        typeof options.debug === 'string' ? options.debug : undefined
+      );
     }
-    await startMcpServer(options.projectRoot ?? (await findProjectRoot()));
+    await startMcpServer({
+      projectRoot: options.projectRoot ?? (await findProjectRoot()),
+      ide: options.ide,
```
pavelgj: I feel like past this point we should not be passing the ide, but instead `isRequireExplicitProjectRoot`

genkit-tools/cli/src/mcp/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+      await this.manager.stop();
+    }
+    const devManager = await startDevProcessManager(
+      projectRoot,
+      command,
+      args,
+      {
+        nonInteractive: true,
+        healthCheck: timeout !== 0,
+        timeout,
```
ssbushi, line 98: TODO: set cwd if in agy

genkit-tools/cli/src/commands/mcp.ts
ssbushi: I thought so too... I'd prefer making this the command arg as well, since making it AGY specific is not necessary...

PR #4028, opened 2026-01-02
feat(go/plugins/ollama): add runtime options and thinking

simakmak:
## Behavioral changes
The Ollama plugin now supports receiving text content alongside tool calls. If a model returns both a text response and a tool request, both will be preserved in the message content.

## Description
* Introduces `ThinkOption` type with constructors `ThinkEnabled(bool)` and `ThinkEffort(string)` to support thinking/reasoning for Ollama models (e.g. `deepseek-r1`) and GPT-OSS models (effort levels: `"low"`, `"medium"`, `"high"`)
* Introduces `GenerateContentConfig` with:
  * `Think *ThinkOption` for controlling reasoning behavior
  * Runtime options: `Seed`, `Temperature`, `TopK`, `TopP`, `MinP`, `Stop`, `NumCtx`, `NumPredict`
  * Ollama-specific: `KeepAlive`
* Adds `Ptr[T]` generic helper for initializing pointer fields
* Adds `ApplyOptions` to map `GenerateContentConfig`, `map[string]any`, and `ai.GenerationCommonConfig` into the Ollama chat request
* Parses the dedicated `thinking` JSON field from Ollama responses into `ai.ReasoningPart`
* Falls back to regex-based `<think>`/`<thinking>` tag extraction from content text when thinking is enabled but the model lacks native thinking field support (non-streaming only)
* Streams `ReasoningPart` chunks for models that use Ollama's native `thinking` field
* Preserves text content alongside tool calls (previously only one or the other was kept)
* Switches response stream parsing from `bufio.Scanner` to `json.Decoder` for correctness
* Registers `ConfigSchema` on model metadata via `core.InferSchemaMap`
* Adds safe type assertion in embedder model fallback
* Adds README with usage examples for chat, tools, thinking, multimodal, and embeddings
* Adds unit tests for `ApplyOptions`, `parseThinking`, `translateChatResponse`, and JSON marshaling

**Example usage:**

```go
genkit.Generate(ctx, g,
    ai.WithModel(m),
    ai.WithPrompt("What is heavier, 1kg of steel or 1kg of feathers?"),
    ai.WithConfig(&ollama.GenerateContentConfig{
        Think:       ollama.ThinkEnabled(true),
        Temperature: ollama.Ptr(0.2),
    }),
)

// Access reasoning and final answer separately
fmt.Println("Reasoning:", resp.Reasoning())
fmt.Println("Answer:", resp.Text())
```

```go
// GPT-OSS effort levels
ai.WithConfig(&ollama.GenerateContentConfig{
    Think: ollama.ThinkEffort("high"),
})
```

---

## Checklist

* [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
* [x] Tested (manually, unit tests)
* [x] Docs updated (README added with runtime options / thinking mode documentation)


go/plugins/ollama/ollama.go
```diff
@@ -369,6 +397,55 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+	cfg any,
+) {
+	if cfg == nil {
+		return
+	}
+
+	switch cfg := cfg.(type) {
+	case GenerateContentConfig:
+		// Thinking
+		if cfg.Think != "" {
```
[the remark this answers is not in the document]
simakmak: Corrected as per comment

go/plugins/ollama/ollama.go
```diff
@@ -142,6 +163,8 @@ type ollamaChatRequest struct {
 	Stream   bool             `json:"stream"`
 	Format   string           `json:"format,omitempty"`
 	Tools    []ollamaTool     `json:"tools,omitempty"`
+	Think    any              `json:"think,omitempty"`
+	Options  map[string]any   `json:"options,omitempty"`
```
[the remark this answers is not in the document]
simakmak: Corrected as per comment

go/plugins/ollama/ollama.go
```diff
@@ -369,6 +397,55 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+		return
+	}
+
+	switch cfg := cfg.(type) {
+	case GenerateContentConfig:
+		// Thinking
+		if cfg.Think != "" {
+			req.Think = cfg.Think
+		}
+
```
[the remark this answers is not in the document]
simakmak: Corrected as per comment

hugoaguirre:
Hi @simakmak, thanks for your contribution. The issue in the Go tests is already addressed in a different PR. I'll put my review in a moment

go/plugins/ollama/ollama.go
```diff
@@ -369,6 +399,73 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+func applyGenerateConfigToOllama(
+	req *ollamaChatRequest,
+	cfg any,
+) {
+	if cfg == nil {
+		return
+	}
+
+	switch cfg := cfg.(type) {
+	case GenerateContentConfig:
```
hugoaguirre: Good addition. Can you consider `*GenerateContentConfig` and `map[string]any` as valid config structs as well? 
The later one is basically to add forward compatibility for options that might not be defined in the struct. 
You can find references in `plugins/googlegenai/gemini.go` on how to handle these scenarios.

Could you also add the proper UTs for these new code changes?

go/plugins/ollama/ollama.go
```diff
@@ -263,12 +289,16 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+
 		chatReq := ollamaChatRequest{
 			Messages: messages,
 			Model:    g.model.Name,
 			Stream:   stream,
 			Images:   images,
 		}
+
+		applyGenerateConfigToOllama(&chatReq, input.Config)
+		fmt.Printf("TEST: %+v\n", chatReq)
```
hugoaguirre:
```suggestion
```

simakmak:
@hugoaguirre, thank you for your comments. I made edits and added tests. I would also be grateful if you could guide me to examples where I can see the generation of options via Dotprompt, since I am new to your project, but genkit hooked me and I would like to add a little more support for ollama.  I also wanted to know about the documentation, do I need to add it and refer to this commit after the merge, or how to do it correctly?

hugoaguirre:
> I made edits and added tests. I would also be grateful if you could guide me to examples where I can see the generation of options via Dotprompt, since I am new to your project, but genkit hooked me and I would like to add a little more support for ollama

Hey, thanks for the interest in the continuous contributions! 

As per the documentation changes, you can include them in the plugin's `README.md`. Also, we have the [docsite](https://github.com/genkit-ai/docsite) repo open for docs contributions.

In regards of Dotprompt. What kind of examples do you want to see? There's a specific samples folder that explains multiple ways of using it in your generate calls. Samples can be found in the `go/samples/prompts/*` folder. 

Let me know how can I help you :-)

go/plugins/ollama/model.go
```diff
@@ -0,0 +1,165 @@
+package ollama
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/firebase/genkit/go/ai"
+)
+
+var allowedOllamaOptions = map[string]string{
```
hugoaguirre: I don't think having this "allow-list" in the plugin is a good idea since we could run into the scenario: 
If the Ollama API enables a new parameter that is not present in this list, users will not be able to use it due to this constraint.

Let's only keep the check for top-level parameters since they should change very rarely and keep the opts.

So, the allow-list could look like this:

```go
var topLevelOpts = map[string]struct{}{
  "think": {},
  "keep_alive": {},
  // rest of top-level opts ...
}
```

We can delete `allowedOllamaOptions` and do the filtering against `topLevelOpts` something like this
```go
func (o *ollamaChatRequest) applyMapAny(m map[string]any) error {
  if len(m) == 0 {
    return nil
  }
  if o.Options == nil {
    o.Options = make(map[string]any)
  }

  for k, v := range m {
    // check if it's a known top-level param
    if _, isTopLevel := topLevelOptions[k]; isTopLevel {
        // switch/case to handle top level settings without returning errors
    }
    // this is to enable forward compat for new params
    o.Options[k] = v
  }
}
```

go/plugins/ollama/model.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+func (o *ollamaChatRequest) applyGenerationCommonConfig(cfg *ai.GenerationCommonConfig) error {
+	opts := map[string]any{}
+	b, err := json.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(b, &opts)
```
hugoaguirre: I think having a manual mapping would be better since the JSON tags for `ai.GenerationCommonConfig` are camel case and Ollama requires snake case tags.  For example, `topK` and `topP`, Ollama API expects  `top_k` and `top_P` respectively, also for the special case for `maxOutputTokens` and `num_predict`.

go/plugins/ollama/model.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return err
+	}
+
+	if len(opts) > 0 {
+		o.Options = opts
+	}
+	return nil
+}
+func (o *ollamaChatRequest) applyMapAny(m map[string]any) error {
```
hugoaguirre, line 153: See my comment above for the suggested refactor

hugoaguirre:
Hi @simakmak seems you need to pull the latest changes in main and make sure to run `go fmt` to pass the formatting check

apascal07:
@hugoaguirre Have you tested out the plugin?

hugoaguirre:
> @hugoaguirre Have you tested out the plugin?

Not yet. I lost these updates in my notifications. I'll update the PR with my findings.

hugoaguirre:
Hey guys, this one got me thinking.

There are some models in Ollama that returns thinking in 3 different ways:
1. As part of the response message
2. As part of the response message but between `<think>` tags
3. As part of a `thinking` field for latest Ollama versions

I've added support for all three scenarios.
1 and 2 were locally tested using `magistral` and `deepseek-r1:1.5b`  respectively and the thinking/reasoning content was properly set.

For #3, due HW constraints and limitations, I was not able to install an run a larger model but I've added test cases when a response includes the `thinking` field in the response message.

This PR is ready to go!

Thanks @simakmak 
@apascal07

go/plugins/ollama/ollama.go
```diff
@@ -284,6 +288,7 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
 	if err != nil {
 		return nil, err
 	}
+	fmt.Printf("Ollama Request Payload: %s\n", string(payloadBytes))
```
apascal07: Remove this and other debug statements please.

go/plugins/ollama/ollama.go
```diff
@@ -119,29 +119,27 @@ type ollamaMessage struct {
...
-	Images   []string         `json:"images,omitempty"`
-	Model    string           `json:"model"`
-	Stream   bool             `json:"stream"`
-	Format   string           `json:"format,omitempty"`
-	Tools    []ollamaTool     `json:"tools,omitempty"`
+type GenerateContentConfig struct {
+	// Thinking mode:
+	// ollama: true | false
+	// gpt-oss: "low" | "medium" | "high"
+	Think any
```
apascal07: Add JSON tags to all these fields otherwise requests from Dev UI won't work correctly.

go/plugins/ollama/ollama.go
```diff
@@ -576,3 +609,15 @@ func concatImages(input *ai.ModelRequest, roleFilter []ai.Role) ([]string, error
 	}
 	return images, nil
 }
+
+// parseThinking extracts the thinking content from the response string.
+func parseThinking(content string) (string, string) {
```
apascal07, line 698: Can there be multiple <think> blocks?

hugoaguirre: Yes. I did a further research here and it depends on how the model gets configured in the system prompt. The standard way is to have one block of thinking but it could be more (like a chain-of-thoughts). 

I've updated the code to match this behavior.

apascal07:
`Think (Ollama: true/false; GPT-OSS: "low" | "medium" | "high")`

Can we do better than this? We should avoid "any" types if at all possible. Generally, we only have "any" when the burden of maintaining generic type params outweighs the compile time validation. In this case, it seems like a cop out.

apascal07:
Also we parse "thinking" blocks regardless of whether `Think` is enabled which is heavy handed. It's possible for a model to return its own blocks as part of a text response, which would be hijacked here. We should on trigger this regex parsing when thinking is enabled.

PR #4044, opened 2026-01-05
feat(genkit-tools): add  `/listValues` endpoint to expose defaultModel

ssbushi:
Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/js/genkit/src/genkit.ts
+++ b/js/genkit/src/genkit.ts
@@ -1105,7 +1105,5 @@ function toModelRef(
   const modelAction = modelArg as ModelAction;
   return modelRef({
     name: modelAction.__action.name,
-    configSchema: modelAction.__configSchema,
-    info: modelAction.__action.metadata?.model,
   });
 }


genkit-tools/common/src/types/apis.ts
```diff
@@ -80,6 +80,22 @@ export const ListActionsRequestSchema = z
...
 
+export const ListValuesRequestSchema = z.object({
+  runtimeId: z
+    .string()
+    .optional()
+    .describe(
+      'ID of the Genkit runtime to run the action on. Typically $pid-$port.'
+    ),
+  type: z
+    .enum(['defaultModel'])
```
ssbushi: Restricted to `defaultModel` only.

js/genkit/src/genkit.ts
```diff
@@ -953,7 +954,7 @@ export class Genkit implements HasRegistry {
       this.registry.registerValue(
         'defaultModel',
         'defaultModel',
-        this.options.model
+        modelName(this.options.model)
```
ssbushi: Realized that having the conversion done at registry time is more convenient, since all types needed are available here. WDYT?

genkit-tools/common/src/manager/manager.ts
```diff
@@ -198,6 +198,32 @@ export class RuntimeManager {
...
+          : 'No runtimes found. Make sure your app is running using `genkit start -- ...`. See getting started documentation.'
+      );
+    }
+    const response = await axios
+      .get(`${runtime.reflectionServerUrl}/api/values`, {
+        params: {
+          type: input.type,
+        },
+      })
+      .catch((err) => this.httpErrorHandler(err, 'Error listing values.'));
```
pavelgj: backwards compatibility consideration -- Go and python won't have this endpoint implemented, so will be getting 404. Should 404 be handled here and an empty record returned, or bubble up the error all the way to the Dev UI and let it deal with backwards compatibility?

I'm leaning towards handle 404 here...

ssbushi: Done.

genkit-tools/common/src/types/apis.ts
pavelgj: Let's not encode this in the api definition. The restriction is temporary. Just make it a string.

js/core/src/reflection.ts
```diff
@@ -151,6 +151,22 @@ export class ReflectionServer {
...
 
+    server.get('/api/values', async (req, response, next) => {
+      logger.debug('Fetching values.');
+      try {
+        const type = req.query.type;
+        if (!type) {
+          response.status(400).send('Query parameter "type" is required.');
+          return;
+        }
+        const values = await this.registry.listValues(type as string);
```
pavelgj, line 170: don't you need to limit this to defaultModel only? and prerhaps normalize the model to model ref structure?

js/genkit/src/genkit.ts
```diff
@@ -1086,3 +1087,19 @@ let disableReflectionApi = false;
 export function __disableReflectionApi() {
   disableReflectionApi = true;
 }
+
+/** Helper method to map ModelArgument to string */
+function modelName(
+  modelArg: ModelArgument<any> | undefined
+): string | undefined {
```
pavelgj: normalize to model ref. The model ref can contain config, which you are losing when normalizing to string.

pavelgj (changes requested):
a few more comments

js/genkit/src/genkit.ts
```diff
@@ -1086,3 +1088,24 @@ let disableReflectionApi = false;
...
+    return modelRef({ name: modelArg });
+  }
+  if ((modelArg as ModelReference<any>).name) {
+    return modelArg as ModelReference<any>;
+  }
+  const modelAction = modelArg as ModelAction;
+  return modelRef({
+    name: modelAction.__action.name,
+    configSchema: modelAction.__configSchema,
+    info: modelAction.__action.metadata?.model,
```
pavelgj: not sure if these are necessary... I want to say drop 'em.

PR #4046, opened 2026-01-06
refactor(js/plugins/anthropic):  extract out some shared converters

cabljac:
This PR restructures code a bit to extract out some shared logic, using structural types to keep them agnostic wrt the SDK namespaces.

### DIFF
--- a/js/plugins/anthropic/src/runner/stable.ts
+++ b/js/plugins/anthropic/src/runner/stable.ts
@@ -95,7 +95,7 @@ export class Runner extends BaseRunner<RunnerTypes> {
       const signature = this.getThinkingSignature(part);
       if (!signature) {
         throw new Error(
-          'Anthropic thinking parts require a signature when sending back to the API. Preserve the `custom.anthropicThinking.signature` value from the original response.'
+          'Anthropic thinking parts require a signature when sending back to the API. Preserve the `metadata.thoughtSignature` value from the original response.'
         );
       }
       return {
--- a/js/plugins/anthropic/src/runner/beta.ts
+++ b/js/plugins/anthropic/src/runner/beta.ts
@@ -159,7 +159,7 @@ export class BetaRunner extends BaseRunner<BetaRunnerTypes> {
       const signature = this.getThinkingSignature(part);
       if (!signature) {
         throw new Error(
-          'Anthropic thinking parts require a signature when sending back to the API. Preserve the `custom.anthropicThinking.signature` value from the original response.'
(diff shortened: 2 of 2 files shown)


js/plugins/anthropic/src/runner/stable.ts
```diff
@@ -412,47 +407,30 @@ export class Runner extends BaseRunner<RunnerTypes> {
...
+        return redactedThinkingBlockToPart(contentBlock);
+
+      case 'server_tool_use':
+        return serverToolUseBlockToPart(contentBlock);
+
+      case 'web_search_tool_result':
+        return webSearchToolResultBlockToPart(contentBlock);
 
       default: {
+        // Exhaustive check (uncomment when all types are handled):
```
cabljac, line 424: maybe just remove this comment, keep it flexible

js/plugins/anthropic/tests/beta_runner_test.ts
```diff
@@ -775,6 +775,15 @@ describe('BetaRunner', () => {
       },
     });
 
+    // Unknown block types log a warning and return empty text
+    // (Exhaustive check is commented out for forward compatibility)
```
cabljac: remove comment, see above

js/plugins/anthropic/src/runner/converters/shared.ts
```diff
@@ -0,0 +1,120 @@
...
+export function thinkingBlockToPart(block: {
+  thinking: string;
+  signature?: string;
+}): Part {
+  if (block.signature !== undefined) {
+    return {
+      reasoning: block.thinking,
+      custom: {
+        [ANTHROPIC_THINKING_CUSTOM_KEY]: { signature: block.signature },
+      },
```
pavelgj: thought signatures should go into metadata. Ex: https://github.com/firebase/genkit/blob/a209ac17f0d16a45e3539f9878bbeac04e5805ec/js/plugins/google-genai/src/common/converters.ts#L362

Also, a part can't be both reasoning and custom... the semantics is "oneOf" -- a part can have only one of the fields set.

js/plugins/anthropic/src/runner/converters/stable.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Converts a server_tool_use block to a Genkit Part.
+ * In the stable API, name is always present.
+ */
+export function serverToolUseBlockToPart(block: {
+  id: string;
+  name: string;
+  input: unknown;
+}): Part {
+  return {
+    text: `[Anthropic server tool ${block.name}] input: ${JSON.stringify(block.input)}`,
```
pavelgj, line 33: same here -- can't be both `text` and `custom`...

PR #4048, opened 2026-01-06
feat(py): add define_resource for mcp plugin

MengqinShen:
CHANGELOG:
- [ ] Added 'Resource' action kind 
- [ ] Added a define_resource method to the Genkit class
- [ ] Created a resource management module - genkit/blocks/resource.py
- [ ] Added test files test_resource.py and resource_test.py

### DIFF
--- a/py/packages/genkit/src/genkit/ai/_registry.py
+++ b/py/packages/genkit/src/genkit/ai/_registry.py
@@ -43,7 +43,7 @@ import traceback
 import uuid
 from collections.abc import AsyncIterator, Callable
 from functools import wraps
-from typing import Any, Callable, Type, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Callable, Type
 
 if TYPE_CHECKING:
     from genkit.blocks.resource import ResourceFn, ResourceOptions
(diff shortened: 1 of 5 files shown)


py/packages/genkit/src/genkit/ai/_registry.py
```diff
@@ -42,7 +43,10 @@
 import uuid
 from collections.abc import AsyncIterator, Callable
 from functools import wraps
-from typing import Any, Type
+from typing import TYPE_CHECKING, Any, Callable, Type
+
+if TYPE_CHECKING:
```
yesudeep, line 48: what does TYPE_CHECKING do?

py/packages/genkit/tests/genkit/ai/test_resource.py
```diff
@@ -0,0 +1,241 @@
+# Copyright 2025 Google LLC
```
yesudeep, line 1: 2026?

py/packages/genkit/src/genkit/ai/_registry.py
MengqinShen: YPE_CHECKING from the typing module is used to avoid circular imports. It is is True for static type checkers (like Mypy, VS Code's Pylance, or PyCharm) but False at runtime.

PR #4050, opened 2026-01-06
fix(js/core): Dap Action updates

ifielker:
New requirements that the DAP action takes no input and returns an ActionMetadata array are now satisfied.

Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/types/action.ts
+++ b/genkit-tools/common/src/types/action.ts
@@ -60,28 +60,7 @@ export type Action = z.infer<typeof ActionSchema>;
 
 export const ActionMetadataSchema = z
   .object({
-    actionType: z
-      .enum([
-        'custom',
-        'dynamic-action-provider',
-        'embedder',
-        'evaluator',
-        'executable-prompt',
-        'flow',
-        'indexer',
-        'model',
-        'background-model',
-        'check-operation',
-        'cancel-operation',
-        'prompt',
-        'reranker',
-        'retriever',
-        'tool',
-        'tool.v2',
-        'util',
-        'resource',
-      ])
-      .optional(),
+    actionType: z.string().optional(),
     name: z.string(),
     description: z.string().optional(),
     inputSchema: z.unknown().optional(),
--- a/js/core/src/action.ts
+++ b/js/core/src/action.ts
@@ -63,28 +63,7 @@ export interface ActionMetadata<
 }
 
 export const ActionMetadataSchema = z.object({
-  actionType: z
-    .enum([
-      'custom',
-      'dynamic-action-provider',
-      'embedder',
-      'evaluator',
-      'executable-prompt',
-      'flow',
-      'indexer',
-      'model',
-      'background-model',
-      'check-operation',
-      'cancel-operation',
-      'prompt',
-      'reranker',
-      'retriever',
-      'tool',
-      'tool.v2',
-      'util',
-      'resource',
-    ])
-    .optional(),
+  actionType: z.string().optional(),
   name: z.string(),
   description: z.string().optional(),
   inputSchema: z.unknown().optional(),


js/core/src/dynamic-action-provider.ts
```diff
@@ -142,12 +147,10 @@ export type DapMetadata = {
...
-    metadata[key] = value[key].map((a) => {
-      return a.__action;
-    });
+function transformDapValue(value: DapValue): ActionMetadata[] {
+  const metadata: ActionMetadata[] = [];
+  for (const typeKey of Object.keys(value)) {
+    metadata.push(...value[typeKey].map((a) => a.__action));
   }
   return metadata;
 }
```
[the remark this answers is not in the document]
ifielker: done.

js/core/src/dynamic-action-provider.ts
```diff
@@ -167,20 +168,13 @@ export function defineDynamicActionProvider(
     registry,
     {
       ...cfg,
+      inputSchema: z.any(),
+      outputSchema: z.any(),
```
pavelgj: Please define `ActionMetadataSchema` and make this `z.array(ActionMetadataSchema)`

We probably want to have a shared (across languages) definition of `ActionMetadataSchema`, so make sure it's defined in https://github.com/firebase/genkit/blob/2b015cd240748fb9cb3ed91a6587dafd0bd145ea/genkit-tools/common/src/types/action.ts#L42 as well so that it's shows up in https://github.com/firebase/genkit/blob/main/genkit-tools/genkit-schema.json

js/core/src/dynamic-action-provider.ts
```diff
@@ -167,20 +168,13 @@ export function defineDynamicActionProvider(
     registry,
     {
       ...cfg,
+      inputSchema: z.any(),
```
pavelgj: `z.void()` doesn't work?

js/core/src/dynamic-action-provider.ts
```diff
@@ -66,12 +66,17 @@ class SimpleCache {
...
+            // we are not allowed pass it in).
+            // We are not allowed to pass in either the DapValue we just got or
+            // the cache because the action input MUST be void
+            // and the output MUST be an array of ActionMetadata.
+            // And we can't construct the DapValue (which contains full Actions)
+            // from just the ActionMetadata.
+            // So unfortunately we need to fetch from the DAP twice, and there's
+            // a small chance that what gets traced is not what gets cached/used.
+            await this.dap.run();
           }
```
pavelgj: why not avoid duplicate `dapFn` call by doing something like:

```ts
if (params?.skipTrace) {
  this.value = await this.dapFn();
} else {
  this.value = await this.dap.run();
}
```

ifielker: dapFn returns a DapValue like `{ tools: [<tool1Action>, <tool2Action>], resources: [<resource1Action>, <resource2Action>]}`. dap.run() gets a new DapValue, flattens it and maps the actions to their metadata only. like this: `[<tool1MetadataOnly>,<tool2MetadataOnly>,<resource1MetadataOnly>,<resource2MetadataOnly>]`. I can't get the actions back from just the metadata. And I can't access the cache from the action, and I can't pass in either the DapValue or the cache to the action... so yeah. And we can't just save the metadata because it's an ActionProvider, not an ActionMetdataProvider. There's a getAction which needs to return the whole action.

ifielker: Unless you optionally let me pass in the cache? like if it's there, save it to cache, but if not then, ignore it. So whatever is calling the dap.run() is not required to pass anything, but it can? Or is that "breaking the contract"?

pavelgj: I'm not sure I understand how cache is relevant to my question... Let me rephrase:

Right now you are doing:

```ts
this.value = await this.dapFn();
if (!params?.skipTrace) {
  // this calls dapFn again, but you ignore the output.
  await this.dap.run();
}
```

This would do the same thing without calling `dapFn` twice:

```ts
if (params?.skipTrace) {
  this.value = await this.dapFn();
} else {
  this.value = await this.dap.run();
}
```

Cache is not involved in that logic as far as I can see, no?

ifielker: Did you not read my response? this.value is type DapValue. this.dap.run() does NOT return a DapValue and we cannot construct a DapValue from the array of Metadata. so `this.value = await this.dap.run();` is Type mismatch. It will not build. Maybe this will be more clear:
```
if (params?.skipTrace) {
  this.value = await this.dapFn();  // returns DapValue - a struct containing Actions
} else {
  this.value = await this.dap.run();  // TYPE MISMATCH: returns ActionMetadata[].
}
```
so unless there is some magic:
```
  const metadataArray = await this.dap.run();
  this.value = magicallyReconstructTheActionFromMetadataAlone(metadataArray);
```
Then we can't do this.

pavelgj: I see, I missed the "metadata only" part and then cache confused me further.

You can easily access the cache from inside the action with a simple closure:

```ts
export function defineDynamicActionProvider(...): DynamicActionProviderAction {
  //...
  const cache = new SimpleCache(...);
  const a = defineAction(
    registry,
    { /* ... */ },
    async (_options) => {
      // access `cache` here
    }
  );
  implementDap(a as DynamicActionProviderAction, cfg, fn, cache);
  return a as DynamicActionProviderAction;
}

```

ifielker: Well, it's not quite that simple, but sure.

js/core/src/dynamic-action-provider.ts
ifielker: done

js/core/src/dynamic-action-provider.ts
ifielker: done

genkit-tools/common/src/types/action.ts
```diff
@@ -58,6 +58,43 @@ export const ActionSchema = z
 
 export type Action = z.infer<typeof ActionSchema>;
 
+export const ActionMetadataSchema = z
+  .object({
+    actionType: z
```
pavelgj: Let's make this a simple string... enum might be too restrictive.

ifielker: done

PR #4051, opened 2026-01-06
feat(py): Implemented deepseek plugin

zarinn3pal:
Implemented Deepseek plugin with tests and samples


py/plugins/deepseek/conftest.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 # Copyright 2025 Google LLC
```
yesudeep, line 2: It looks like this file is only importing pytest? What does it do and if it is required can we document it?

py/samples/deepseek-hello/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+1. Setup environment and install dependencies:
+```bash
+uv venv
+source .venv/bin/activate
+
+uv sync
+```
+
+2. Set DeepSeek API key:
```
yesudeep: Let's add a link to documentation/console where one can obtain this key too?

py/plugins/deepseek/pyproject.toml
```diff
@@ -0,0 +1,52 @@
+# Copyright 2025 Google LLC
```
MengqinShen: 2026?

MengqinShen:
@zarinn3pal I can not run the sample successfully on my end. I shared the error message with you via chat. On top of that, please refer to model plugin in JS SDK to have the same model Config and Tool in the picture I shared with you.

MengqinShen:
@zarinn3pal please also add run.sh file to be consistent with other sample flows.

py/samples/deepseek-hello/src/deepseek_hello.py
MengqinShen: Please rename this file as main.py and add run.sh file to be consistent with other samples.

PR #4082, opened 2026-01-09
fix(genkit-tools/mcp): Fix MCP server failing to spin up runtime

ssbushi:
- Add explicit runtime health check
- Add optional file logging to MCP CLI command
- Use non-interactive mode and specify project root for ProcessManager.
- Tests for changes.

Checklist (if applicable):

### DIFF
--- a/genkit-tools/cli/src/utils/manager-utils.ts
+++ b/genkit-tools/cli/src/utils/manager-utils.ts
@@ -129,6 +129,9 @@ export async function waitForRuntime(
           resolve();
         }
       });
+      if (manager.listRuntimes().length > 0) {
+        resolve();
+      }
     });
 
     const timeoutPromise = new Promise<void>((_, reject) => {
--- a/genkit-tools/cli/tests/commands/start_test.ts
+++ b/genkit-tools/cli/tests/commands/start_test.ts
@@ -79,10 +79,18 @@ describe('start command', () => {
   });
 
   it('should start manager only when no args are provided', async () => {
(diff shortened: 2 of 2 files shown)


js/core/src/reflection.ts
```diff
@@ -103,7 +103,10 @@ export class ReflectionServer {
   }
 
   get runtimeId() {
-    return `${process.pid}${this.port !== null ? `-${this.port}` : ''}`;
+    return (
+      process.env.GENKIT_RUNTIME_ID ??
```
pavelgj: I wonder if this is correct... actually not sure if it was correct before.
the issue might come up in case of multiple genkit instances. If GENKIT_RUNTIME_ID is set, they all will have the same runtimeId, which is wrong.
I wonder if GENKIT_RUNTIME_ID can be used as prefix? So, the reflection API is allowed to append suffixes (like port number) to make ids unique, and then for your purposes you can check for prefix... WDYT?

ssbushi: If this is supposed to be an `ID`, why should we expect that all of them will get the same ID?

I get what you're saying and I feel like if we should either:

- Rename this `GENKIT_RUNTIME_ID_PREFIX` or something, if this is going to be a common env var, or
- Enforce that `GENKIT_RUNTIME_ID` is unique and does not already exists. 

You're right, the previous logic was not correct.

ssbushi: Updated based on our discussions.

pavelgj (approved):

PR #4103, opened 2026-01-11
fix(go): nested render and generate spans under the dotprompt span

### LINKED ISSUE #4063: [Go] Bring prompt tracing/observability up to parity with TS
To bring Go up to parity, prompts should be represented by a span (named after the prompt), with 2 child spans, 'render' and 'generate'. 

- For 'render', the "input" is the prompt input (i.e. variable bindings) and the "output" is the GenerateRequest. i.e. should match the "input" to the "generate" call.
- 'generate' is the same as always, just nested under the prompt

Image

apascal07:
Fixes #4063.

Checklist (if applicable):

### DIFF
--- a/go/ai/evaluator.go
+++ b/go/ai/evaluator.go
@@ -196,12 +196,7 @@ func NewEvaluator(name string, opts *EvaluatorOptions, fn EvaluatorFunc) Evaluat
 				if datapoint.TestCaseId == "" {
 					datapoint.TestCaseId = uuid.New().String()
 				}
-				spanMetadata := &tracing.SpanMetadata{
-					Name:    fmt.Sprintf("TestCase %s", datapoint.TestCaseId),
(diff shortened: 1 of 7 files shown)


MichaelDoyle:
Running a flow, with a prompt looks good to me on the Dev UI side. 

image

For running prompts directly in the prompt runner, it seems to create a "double" generate (i.e. nested generate).

image

There is one other peculiarity re: the HTTP spans I noticed, which I am not sure is related to this PR, but mentioning anyway.

1. Notice the double HTTP span above.
2. The HTTP span(s) _appear_ to come from the Firebase telemetry plugin, but I do not think the `menu` sample explicitly enables/adds it.

    image

MichaelDoyle (changes requested):
@apascal07 where did we leave off on this one?

PR #4116, opened 2026-01-13
fix(js): Schema compatibility for cfworker env

ssbushi:
Partial Fix for https://github.com/firebase/genkit/issues/3968

Cfworker restrictions on:
- Schema validation
- Express APIs in the runtime (without export default?)
- Virtual FS which affects runtime discovery
- Prompt templates (handlebars) wont work either.... we need an alternate. Workaround would be compiling the prompts in the user app.

https://developers.cloudflare.com/workers/runtime-apis/nodejs/

Checklist (if applicable):

### DIFF
--- a/js/core/tests/config_test.ts
+++ b/js/core/tests/config_test.ts
@@ -18,8 +18,8 @@ import * as assert from 'assert';
 import { afterEach, describe, it } from 'node:test';
 import {
   getGenkitRuntimeConfig,
-  setGenkitRuntimeConfig,
   resetGenkitRuntimeConfig,
+  setGenkitRuntimeConfig,
 } from '../src/config.js';
 
 describe('config', () => {


js/core/tests/schema_test.ts
```diff
@@ -162,3 +165,87 @@ describe('toJsonSchema', () => {
...
+        },
+      }
+    );
+
+    assert.strictEqual(result.valid, false);
+    const errorAtFoo = result.errors?.find((e) => e.path === 'foo');
+    assert.ok(errorAtFoo, 'Should have error at foo');
+    assert.strictEqual(compileMock.mock.callCount(), 0);
+    compileMock.mock.restore();
+    resetSchemaCodeGeneration();
```
pavelgj: you might want to move this to afterEach, otherwise this cleanup may not execute if some assertions fail.

js/core/package.json
```diff
@@ -29,22 +29,21 @@
...
     "@opentelemetry/sdk-node": "^0.52.0",
     "@opentelemetry/sdk-trace-base": "~1.25.0",
-    "@opentelemetry/exporter-jaeger": "^1.25.0",
     "@types/json-schema": "^7.0.15",
     "ajv": "^8.12.0",
     "ajv-formats": "^3.0.1",
     "async-mutex": "^0.5.0",
-    "body-parser": "^1.20.3",
     "cors": "^2.8.5",
+    "dotprompt": "^1.1.1",
```
pavelgj, line 41: is this necessary? core should not depend on dotprompt

js/core/src/reflection.ts
```diff
@@ -128,6 +128,13 @@ export class ReflectionServer {
    * The server will be registered to be shut down on process exit.
    */
   async start() {
+    if (global[SCHEMA_VALIDATION_MODE] === 'interpret') {
```
pavelgj: I don't like that we're triggering this behaviour off of SCHEMA_VALIDATION_MODE. This has nothing to do with schema validation, right. How about instead of `disableSchemaCodeGeneration()` we do something like:

```js
setGenkitRuntimeConfig({
  jsonSchemaMode: 'interpret',
  sandboxedRuntime: true, // disables reflection v1 API
});
```

PR #4183, opened 2026-01-16
fix(py): updated prompt sample to match JS SDK

MengqinShen:
CHANGELOG:
1. Removed unrelated flows (simplePrompt, threeGreetings, etc.) to focus on the prompt file demonstration.
2. Added new flows to match JS SDK. 
3. Added public schema registration API.


py/samples/prompt_demo/src/main.py
```diff
@@ -32,95 +33,141 @@
...
+    title: str = Field(..., description='recipe title')
+    ingredients: list[Ingredient]
+    steps: list[str] = Field(..., description='the steps required to complete the recipe')
 
 
-@ai.flow(name='simpleTemplate')
-async def simple_template(input: str = ''):
-    name = 'Fred'
-    return await ai.generate(prompt=f'You are a helpful AI assistant named Walt. Say hello to {name}.')
+if hasattr(ai.registry.dotprompt, '_schemas'):
```
yesudeep: Internal properties shouldn't be accessed by public code. Can we do this another way?

py/samples/prompt_demo/prompts/recipe.robot.prompt
```diff
@@ -0,0 +1,17 @@
...
+    title: string, recipe title
+    ingredients(array): 
+      name: string
+      quantity: string
+    steps(array, the steps required to complete the recipe): string
+---
+
+You are a robot chef famous for making creative recipes that robots love to eat. Robots love things like motor oil, RAM, bolts, and uranium.
+
+Generate a recipe for {{food}}.
```
yesudeep: let's ensure all files end with an EOL character. it can cause issues with some unix tools

py/samples/prompt_demo/src/main.py
```diff
@@ -32,95 +33,141 @@
...
+    result = Recipe.model_validate(response.output)
+    await logger.ainfo(f'chef_flow result: {result}')
+    return result
+
+
+@ai.flow(name='robot_chef_flow')
+async def robot_chef_flow(input: ChefInput) -> Recipe:
+    await logger.ainfo(f'robot_chef_flow called with input: {input}')
+    recipe_prompt = await get_sticky_prompt('recipe', variant='robot')
+    recipe_prompt._output_format = 'json'
```
yesudeep: see comment about using internal properties

py/samples/prompt_demo/src/main.py
MengqinShen: Revised.

py/samples/prompt_demo/src/main.py
MengqinShen: Added public schema registration API to match JS SDK.

PR #4184, opened 2026-01-16
RFC: [Go] `BidiAction`, `BidiFlow`, `BidiModel`, and `Agent`

apascal07:


docs/go-bidi-design.md
```diff
@@ -0,0 +1,645 @@
+# Genkit Go Bidirectional Streaming Features - Design Document
+
+## Overview
+
+This document describes the design for bidirectional streaming features in Genkit Go. The implementation introduces three new primitives:
+
+1. **BidiAction** - Core primitive for bidirectional operations
+2. **BidiFlow** - BidiAction with observability, intended for user definition
+3. **SessionFlow** - Stateful, multi-turn agent interactions with automatic persistence and turn semantics
```
pavelgj: we said we want to call it just `Agent`

docs/go-bidi-design.md
```diff
@@ -0,0 +1,645 @@
...
+```go
+// Artifact represents a named collection of parts produced during a session.
+// Examples: generated files, images, code snippets, etc.
+type Artifact struct {
+    Name  string     `json:"name"`
+    Parts []*ai.Part `json:"parts"`
+}
+
+// SessionFlowOutput wraps the output with session info for persistence.
+type SessionFlowOutput[State, Out any] struct {
```
pavelgj: Let's define these as common schemas (in `genkit-tools/common/src/types`) standardized across runtimes.

apascal07: I tend to agree but we haven't been doing the core concepts (action, etc) as generated types because they have type parameters and they're just "core" to the runtime and sometimes implemented differently. But `Artifact` I agree.

apascal07: In Go, we only generate types for the `ai` package. Not for any technical reason but that's just been the use case since they tend to be POD structs.

docs/go-bidi-design.md
apascal07: Clean. Done.

docs/go-bidi-design.md
pavelgj: I mean, we literally must define these types in `genkit-tools/common/src/types` because we will be building Dev UI agent playground. So, not only we need zod types for the Dev UI, we also need those types to be the same across languages (which generation ensures). So, I guess I don't have a strong opinion if you want to maintain these types by hand in Go, up to you, but they need to be treated as "source of truth is in genkit-tools/common".

xavidop (commented):
Overall, while this design introduces useful primitives for bidirectional flows and persistence, it does not yet align with the agentic patterns seen in modern frameworks like LangChain or LlamaIndex. The current approach feels closer to a single, persistent agent rather than a true agent system where agents can call, delegate to, or coordinate with other agents.

A key limitation is the lack of clear multi-agent orchestration. Modern agent frameworks treat agents as composable units that can collaborate, delegate tasks, or be orchestrated in workflows. 

Additionally, the API surface feels too low-level for end users. Concepts like bidirectional flows and actions require developers to manage execution mechanics explicitly. Many of these concerns should be handled by the framework itself, exposing a cleaner and more minimal interface that is easier to read, reason about, and scale over time.

This contrasts with existing patterns in Genkit JS and ADK, where agent definitions are more declarative and focused on intent rather than wiring. For reference, Genkit JS chat agents provide a much simpler mental model: https://genkit.dev/docs/tutorials/chat-with-pdf/#6-implement-the-ui, and ADK’s Go agent definitions show a clearer agent-first abstraction: https://google.github.io/adk-docs/get-started/go/#define-the-agent-code.

xavidop:
related to bidi flows:

That RFC is heading in a good direction, but I wonder if we can simplify the API further. Introducing defineBidiFlow adds another concept that users need to learn and distinguish from defineFlow, which may increase cognitive load without providing a clearly different mental model.

Would it be possible to support this functionality by extending defineFlow with a configuration option instead? For example, using defineFlow with a specific config to indicate bidirectional or streaming behavior would keep the API surface smaller and make it easier for users to understand what’s going on without having to learn a parallel set of abstractions.

From a usability perspective, having a single defineFlow entry point feels cleaner and more consistent.

apascal07:
> related to bidi flows:
> 
> That RFC is heading in a good direction, but I wonder if we can simplify the API further. Introducing defineBidiFlow adds another concept that users need to learn and distinguish from defineFlow, which may increase cognitive load without providing a clearly different mental model.
> 
> Would it be possible to support this functionality by extending defineFlow with a configuration option instead? For example, using defineFlow with a specific config to indicate bidirectional or streaming behavior would keep the API surface smaller and make it easier for users to understand what’s going on without having to learn a parallel set of abstractions.
> 
> From a usability perspective, having a single defineFlow entry point feels cleaner and more consistent.

There's not much we can do given the large differences in both flow inputs but also return type. They're distinctly different. Go is especially not flexible enough to handle things like this.

apascal07:
> Overall, while this design introduces useful primitives for bidirectional flows and persistence, it does not yet align with the agentic patterns seen in modern frameworks like LangChain or LlamaIndex. The current approach feels closer to a single, persistent agent rather than a true agent system where agents can call, delegate to, or coordinate with other agents.
> 
> A key limitation is the lack of clear multi-agent orchestration. Modern agent frameworks treat agents as composable units that can collaborate, delegate tasks, or be orchestrated in workflows.

What would you like to see here?

> Additionally, the API surface feels too low-level for end users. Concepts like bidirectional flows and actions require developers to manage execution mechanics explicitly. Many of these concerns should be handled by the framework itself, exposing a cleaner and more minimal interface that is easier to read, reason about, and scale over time.
> 
> This contrasts with existing patterns in Genkit JS and ADK, where agent definitions are more declarative and focused on intent rather than wiring. For reference, Genkit JS chat agents provide a much simpler mental model: https://genkit.dev/docs/tutorials/chat-with-pdf/#6-implement-the-ui, and ADK’s Go agent definitions show a clearer agent-first abstraction: https://google.github.io/adk-docs/get-started/go/#define-the-agent-code.

ADK's agent concept is closer (or almost identical) to our concept of a prompt. It doesn't provide meaningful control beyond the generation and tool calling, which we do automatically as part of either prompt or generate APIs. Agent here is more about the automatic state management, turn-based streaming chats, revertible snapshots, and more while still giving you control over being able to do anything you'd like during its lifecycle. It is meant to be lower level than other "agent" abstractions given that those map to our "prompt" abstraction.

apascal07:
@xavidop

xavidop:
> What would you like to see here?
> 
> >

@apascal07 

Thanks for the clarification, that helps, and I think we’re generally aligned on the intent of this abstraction being lower-level. My main point isn’t about whether this primitive should exist, but about how easy it is to build more agentic patterns on top of it.

With the current pattern, it’s somewhat hard to express agents where the agent itself decides which sub-agent or capability should run next (for example, supervisor, router, or planner-style agents). While this is definitely possible, it tends to require a fair amount of additional orchestration code, which can make these patterns less straightforward to implement and reason about.

One of the things I really appreciate about Genkit is that it’s clean, easy to use, and largely self-explainable. From experience working on teams that build AI agents in production, I’ve often seen that when users need to write a lot of glue code to express common patterns, they either look at other frameworks or end up building their own abstractions on top.

What I’d love to see here is a slightly cleaner way to express common agentic patterns where delegation and routing are easier to model. Even if this remains a low-level abstraction, having some framework-supported patterns or helpers could reduce boilerplate and make these use cases more approachable.

Overall, I think this is a useful direction, and with a bit more support for agent-driven orchestration, it could fit very nicely with Genkit’s philosophy and developer experience.

PR #4210, opened 2026-01-21
RFC: [JS] Bidirectional Actions, Flows, and Models

pavelgj:
Introduces bidirectional streaming capabilities to Genkit Actions, Flows, and Models. This allows these primitives to accept a continuous input stream and an initialization payload, in addition to producing an output stream.

### DIFF
--- a/docs/rfc-bidi-actions-flows-models.md
+++ b/docs/rfc-bidi-actions-flows-models.md
@@ -1,0 +1,201 @@
+# RFC: Bidirectional Actions, Flows, and Models
+
+## Summary
+
+Introduces bidirectional streaming capabilities to Genkit Actions, Flows, and Models. This allows these primitives to accept a continuous input stream and an initialization payload, in addition to producing an output stream.
+
+## Motivation
+
+Core primitives need to support advanced interaction patterns beyond simple request/response:
+- **Real-time interactions**: Voice or text chat where input and output happen simultaneously (e.g., Gemini Live, OpenAI Realtime).
+- **Session management**: Long-running flows that maintain state across inputs.
+- **Tool interruptions**: Human-in-the-loop scenarios where execution pauses for input.
+
+## Design
+
+### 1. Bidi Action Primitive
+
+The `Action` interface is extended to support bidirectional communication. This is non-breaking; existing actions remain compatible.
+
+- **`inputStream`**: An `AsyncIterable` of input chunks.
+- **`init`**: An optional payload for setup context available before streaming begins.
+
+#### Definition (`defineBidiAction`)
+
+```typescript
+import { defineBidiAction } from '@genkit-ai/core';
+import { z } from 'zod';
+
+export const bidiChat = defineBidiAction(
+  {
+    name: 'bidiChat',
+    actionType: 'custom',
+    inputSchema: z.string(),   // Schema for items in inputStream
+    outputSchema: z.string(),  // Schema for the final return value
+    streamSchema: z.string(),  // Schema for items in output stream
+    initSchema: z.object({     // Schema for initialization data
+      userId: z.string() 
+    }), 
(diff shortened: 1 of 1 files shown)


pavelgj:
@gemini-code-assist

docs/rfc-bidi-actions-flows-models.md
```diff
@@ -0,0 +1,201 @@
...
+  console.log(chunk.content);
+}
+```
+
+### 4. Execution API (`streamBidi`)
+
+Actions and Flows expose a `streamBidi` method that returns a `BidiStreamingResponse`.
+
+```typescript
+interface BidiStreamingResponse<O, S, I> {
```
apascal07, line 153:
```suggestion
interface BidiStreamingResponse<I, O, S> {
```

docs/rfc-bidi-actions-flows-models.md
```diff
@@ -0,0 +1,201 @@
...
+```typescript
+export const myRealtimeModel = defineBidiModel(
+  { name: 'myRealtimeModel' },
+  async function* ({ inputStream, init }) {
+    // 1. Establish session using configuration from init
+    const session = await upstreamApi.connect({
+      model: 'my-model',
+      config: init?.config,
+      systemPrompt: init?.messages?.find(m => m.role === 'system'),
+      tools: init?.tools,
```
apascal07, line 105: Tool requests are also streamed out, unless I'm missing something, we have no way of handling them automatically, the client will have to handle these?

xavidop (commented):
This RFC is heading in a good direction, but I wonder if we can simplify the API further. Introducing defineBidiFlow adds another concept that users need to learn and distinguish from defineFlow, which may increase cognitive load without providing a clearly different mental model.

Would it be possible to support this functionality by extending defineFlow with a configuration option instead? For example, using defineFlow with a specific config to indicate bidirectional or streaming behavior would keep the API surface smaller and make it easier for users to understand what’s going on without having to learn a parallel set of abstractions.

From a usability perspective, having a single defineFlow entry point feels cleaner and more consistent.

docs/rfc-bidi-actions-flows-models.md
```diff
@@ -0,0 +1,201 @@
...
+
+export const flow = ai.defineBidiFlow(
+  {
+    name: 'chatFlow',
+    inputSchema: z.string(),
+    streamSchema: z.string(),
+    initSchema: z.object({ topic: z.string() }),
+  },
+  async function* ({ inputStream, init }) {
+    yield `Welcome to ${init.topic}`;
```
abdallahshaban557, line 73: would this run at the beginning of the flow basically everytime? @pavelgj

docs/rfc-bidi-actions-flows-models.md
```diff
@@ -0,0 +1,201 @@
...
+    
+    for await (const msg of inputStream) {
```
abdallahshaban557, line 75: just to make sure I understand this example - if there is no "return" - then this stream will continue to be open?

docs/rfc-bidi-actions-flows-models.md
pavelgj: Yes.

docs/rfc-bidi-actions-flows-models.md
pavelgj: Yes, in this example the output stream will continue to be open while the input stream is open. Once the input stream closes, this function will end, the flow will return void/undefined and the output stream will close.

docs/rfc-bidi-actions-flows-models.md
pavelgj: that's fine. Tool requests will be handled out of the output stream and responses injected back into the input stream.

PR #4212, opened 2026-01-21
RFC: [JS] Agent Primitive

pavelgj:
Introduces `defineAgent`, a high-level abstraction built on top of Bidi Actions designed to simplify the creation of stateful, multi-turn agents. It unifies state management, allowing both client-side state handling and server-side persistence via pluggable stores.

Relates RFCs:
 - https://github.com/firebase/genkit/pull/4210

### DIFF
--- a/docs/rfc-agent-primitive.md
+++ b/docs/rfc-agent-primitive.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Introduces `defineAgent` (renamed from `defineSessionFlow`), a high-level abstraction built on top of Bidi Actions designed to simplify the creation of stateful, multi-turn agents. It unifies state management, allowing both client-side state handling and server-side persistence via pluggable stores.
+Introduces `defineAgent`, a high-level abstraction built on top of Bidi Actions designed to simplify the creation of stateful, multi-turn agents. It unifies state management, allowing both client-side state handling and server-side persistence via pluggable stores.
 
 `defineAgent` would replace the current Chat API as there is significant overlap and Agent primitive is more flexible.
 
@@ -128,8 +128,6 @@ export const persistentAgent = ai.defineAgent(
     let messages = init?.messages ?? [];
 
     for await (const input of inputStream) {
-      if (!input) break;
-
       const response = await ai.generate({
         messages: [...messages, input],
         model: googleAI.model('gemini-2.5-flash'),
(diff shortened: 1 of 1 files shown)


xavidop:
Added the same commment in the GO RFC:

Overall, while this design introduces useful primitives for bidirectional flows and persistence, it does not yet align with the agentic patterns seen in modern frameworks like LangChain or LlamaIndex. The current approach feels closer to a single, persistent agent rather than a true agent system where agents can call, delegate to, or coordinate with other agents.

A key limitation is the lack of clear multi-agent orchestration. Modern agent frameworks treat agents as composable units that can collaborate, delegate tasks, or be orchestrated in workflows.

Additionally, the API surface feels too low-level for end users. Concepts like bidirectional flows and actions require developers to manage execution mechanics explicitly. Many of these concerns should be handled by the framework itself, exposing a cleaner and more minimal interface that is easier to read, reason about, and scale over time.

This contrasts with existing patterns in Genkit JS and ADK, where agent definitions are more declarative and focused on intent rather than wiring. For reference, Genkit JS chat agents provide a much simpler mental model: https://genkit.dev/docs/tutorials/chat-with-pdf/#6-implement-the-ui, and ADK’s Go agent definitions show a clearer agent-first abstraction: https://google.github.io/adk-docs/get-started/go/#define-the-agent-code.

docs/rfc-agent-primitive.md
```diff
@@ -0,0 +1,221 @@
...
+  messages: z.array(MessageSchema).optional(),
+  state: z.any().optional(),
+  artifacts: z.array(AgentArtifactSchema).optional(),
+});
+```
+
+#### Stream Schema (`AgentStreamSchema`)
+```typescript
+const AgentStreamSchema = z.object({
+  sessionId: z.string(),
```
apascal07: Does this need to be here?

docs/rfc-agent-primitive.md
```diff
@@ -0,0 +1,221 @@
...
+  chunk: GenerateResponseChunkSchema.optional(), // Token generation
```
apascal07: Is `GenerateResponseChunkSchema` the right type here? We don't want to be streaming out what the model is streaming out, you should have control. It should be a `Stream` type.

pavelgj:
> Added the same commment in the GO RFC:
> 
> Overall, while this design introduces useful primitives for bidirectional flows and persistence, it does not yet align with the agentic patterns seen in modern frameworks like LangChain or LlamaIndex. The current approach feels closer to a single, persistent agent rather than a true agent system where agents can call, delegate to, or coordinate with other agents.
> 
> A key limitation is the lack of clear multi-agent orchestration. Modern agent frameworks treat agents as composable units that can collaborate, delegate tasks, or be orchestrated in workflows.
> 
> Additionally, the API surface feels too low-level for end users. Concepts like bidirectional flows and actions require developers to manage execution mechanics explicitly. Many of these concerns should be handled by the framework itself, exposing a cleaner and more minimal interface that is easier to read, reason about, and scale over time.
> 
> This contrasts with existing patterns in Genkit JS and ADK, where agent definitions are more declarative and focused on intent rather than wiring. For reference, Genkit JS chat agents provide a much simpler mental model: https://genkit.dev/docs/tutorials/chat-with-pdf/#6-implement-the-ui, and ADK’s Go agent definitions show a clearer agent-first abstraction: https://google.github.io/adk-docs/get-started/go/#define-the-agent-code.

This RFC is for one step in our overall vision and yeah, we need to find a way to communicate the overall vision as well. I mostly agree with you -- this is not an agent framework, this does not include multi-agent orchestration.

If you look at other "agent frameworks", they usually make opinionated stances on how agent orchestration must be done. They have hardcoded prompts and magical tools that get injected into the model call to facilitate "handoffs" or subagent calls. This is actually very inconsistent with how we've been doing things in Genkit. We've been avoiding hardcoded prompts or doing any unexpected/surprising LLM call alterations. The Chat API is no exception -- it does not inject any hardcoded prompts (it injects prompts that developer provides).

This RFC is all about agent abstraction. One agent. That agent might have sub-agents, but this RFC does not concern itself with that. It's a simple, but extremely flexible abstraction that can support agents of any complexity. At the same time enabling Dev UI tooling for those agents.

The next step, after we have a flexible way to represent one agent (or multiple independent agents), we can design agent orchestration layer on top of the agent abstraction, but the current thinking is that the orchestration layer, by nature, is too opinionated, so it will not be in the core Genkit framework. It will most likely be a plugin, or even multiple plugins, separate package(s). It will allow users/community to build custom orchestration layers while still taking advantage of genkit tooling that already understands agent building blocks.

xavidop:
> > Added the same commment in the GO RFC:
> > Overall, while this design introduces useful primitives for bidirectional flows and persistence, it does not yet align with the agentic patterns seen in modern frameworks like LangChain or LlamaIndex. The current approach feels closer to a single, persistent agent rather than a true agent system where agents can call, delegate to, or coordinate with other agents.
> > A key limitation is the lack of clear multi-agent orchestration. Modern agent frameworks treat agents as composable units that can collaborate, delegate tasks, or be orchestrated in workflows.
> > Additionally, the API surface feels too low-level for end users. Concepts like bidirectional flows and actions require developers to manage execution mechanics explicitly. Many of these concerns should be handled by the framework itself, exposing a cleaner and more minimal interface that is easier to read, reason about, and scale over time.
> > This contrasts with existing patterns in Genkit JS and ADK, where agent definitions are more declarative and focused on intent rather than wiring. For reference, Genkit JS chat agents provide a much simpler mental model: https://genkit.dev/docs/tutorials/chat-with-pdf/#6-implement-the-ui, and ADK’s Go agent definitions show a clearer agent-first abstraction: https://google.github.io/adk-docs/get-started/go/#define-the-agent-code.
> 
> This RFC is for one step in our overall vision and yeah, we need to find a way to communicate the overall vision as well. I mostly agree with you -- this is not an agent framework, this does not include multi-agent orchestration.
> 
> If you look at other "agent frameworks", they usually make opinionated stances on how agent orchestration must be done. They have hardcoded prompts and magical tools that get injected into the model call to facilitate "handoffs" or subagent calls. This is actually very inconsistent with how we've been doing things in Genkit. We've been avoiding hardcoded prompts or doing any unexpected/surprising LLM call alterations. The Chat API is no exception -- it does not inject any hardcoded prompts (it injects prompts that developer provides).
> 
> This RFC is all about agent abstraction. One agent. That agent might have sub-agents, but this RFC does not concern itself with that. It's a simple, but extremely flexible abstraction that can support agents of any complexity. At the same time enabling Dev UI tooling for those agents.
> 
> The next step, after we have a flexible way to represent one agent (or multiple independent agents), we can design agent orchestration layer on top of the agent abstraction, but the current thinking is that the orchestration layer, by nature, is too opinionated, so it will not be in the core Genkit framework. It will most likely be a plugin, or even even multiple plugins, separate package(s). It will allow users/community to build custom orchestration layers while still taking advantage of genkit tooling that already understands agent building blocks.

Thank you so much @pavelgj, for putting all this together. Now it is clear to me. One step at a time, 100% agreed. I can help you with that design, I have been building bots for a while.

PR #4219, opened 2026-01-22
fix(js/core): Stop if reflection server fails initialization

ssbushi:
Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/js/core/src/reflection.ts
+++ b/js/core/src/reflection.ts
@@ -423,7 +423,6 @@ export class ReflectionServer {
         } catch (err) {
           logger.error(`Failed to stop server gracefully: ${err}`);
         }
-        process.exit(1);
       }
     });
   }


js/core/src/reflection.ts
```diff
@@ -418,6 +418,12 @@ export class ReflectionServer {
         await this.writeRuntimeFile();
       } catch (e) {
         logger.error(`Error initializing plugins: ${e}`);
+        try {
+          await this.stop();
+        } catch (err) {
+          logger.error(`Failed to stop server gracefully: ${err}`);
+        }
+        process.exit(1);
```
pavelgj: remove this

pavelgj:
```suggestion
```

PR #4235, opened 2026-01-22
feat(cli): add streaming and reasoning to dev:test-model cmd

hugoaguirre:
Added stream test cases:
 - Multi-turn conformance
 - Tool request
 - Structured output

Added `reasoning` validator: 
```shell
cd js/testapps/basic-gemini
genkit dev:test-model --from-file=../../plugins/google-genai/tests/model-tests-tts.yaml  -- tsx src/index.ts
```

In `model-tests-tts.yaml` you can specify how to enable `reasoning` for your model.
For Gemini models:
```yaml
- model: googleai/gemini-3-pro-preview
supports:
  - reasoning
 tests:
   - name: Reasoning conformance
     input:
       messages:
          - role: user
              content:
                - text: A banana farmer harvest 10 bananas but he eats 3 and sells 4. How many bananas are remaining?
        config:
          thinkingConfig:
            thinkingBudget: 1024
            includeThoughts: true
```

### DIFF
--- a/genkit-tools/cli/src/commands/dev-test-model.ts
+++ b/genkit-tools/cli/src/commands/dev-test-model.ts
@@ -495,7 +495,6 @@ async function runTest(
     const modelKey = model.startsWith('/') ? model : `/model/${model}`;
     const shouldStream = !!testCase.stream;
 
-    let chunks = 0;
     const collectedChunks: any[] = [];
 
     const actionResponse = await manager.runAction(
@@ -506,12 +505,11 @@ async function runTest(
       shouldStream
         ? (chunk) => {
             collectedChunks.push(chunk);
-            chunks++;
           }
         : undefined
     );
 
-    if (shouldStream && chunks === 0) {
+    if (shouldStream && collectedChunks.length === 0) {
       throw new Error('Streaming requested but no chunks received.');
     }
     const response = GenerateResponseSchema.parse(actionResponse.result);


genkit-tools/cli/src/commands/dev-test-model.ts
```diff
@@ -357,18 +386,56 @@ async function waitForRuntime(manager: RuntimeManager) {
...
-    const actionResponse = await manager.runAction({
-      key: modelKey,
-      input: testCase.input,
-    });
 
+    // Media output models usually do not support streaming
+    const isMediaOutput = testCase.validators.some((v) =>
+      v.startsWith('valid-media')
+    );
+    const shouldStream = stream && !isMediaOutput;
```
pavelgj: I think this should be an explicit flag on the test...

hugoaguirre: I initially opted for the global `--stream` flag as a way to quickly verify across all the existing core test cases without having to duplicate each one of them. I was thinking stream mode more like a toggle to the existing tests. 

I see your point about explicit control. I'm okay to rework this with your suggestion.

If we are keeping the per-test flag, should I go and duplicate the core test cases for this?

genkit-tools/cli/src/commands/dev-test-model.ts
```diff
@@ -343,6 +356,22 @@ const TEST_CASES: Record<string, TestCase> = {
...
+          content: [
+            {
+              text: 'A banana farmer harvest 10 bananas but eats 3 and sells 4 of them, how many bananas are remaining?',
+            },
+          ],
+        },
+      ],
+    },
+    validators: ['reasoning'],
+  },
```
pavelgj: remove this -- this test data won't actually trigger reasoning and you're not using it, you're only using the validator, which is fine.

genkit-tools/cli/src/commands/dev-test-model.ts
```diff
@@ -359,23 +489,54 @@ async function runTest(
...
     const response = GenerateResponseSchema.parse(actionResponse.result);
+    if (shouldStream && streamedText) {
+      const message = response.message || response.candidates?.[0]?.message;
+      if (message?.content) {
+        const textPart = message.content.find((p) => 'text' in p);
+        if (textPart) {
+          (textPart as { text: string }).text = streamedText;
+        }
+      }
+    }
```
pavelgj: what is this doing? it's not necessary, is it?

hugoaguirre: Indeed, it needs to be removed. Initially I had a bug in my tests (where the last chunk was truncated from the final response) and I've added this logic to accumulate the streamed text for further checks.

genkit-tools/cli/src/commands/dev-test-model.ts
```diff
@@ -363,19 +493,35 @@ async function runTest(
   try {
     // Adjust model name if needed (e.g. /model/ prefix)
     const modelKey = model.startsWith('/') ? model : `/model/${model}`;
-    const actionResponse = await manager.runAction({
-      key: modelKey,
-      input: testCase.input,
-    });
+    const shouldStream = !!testCase.stream;
 
+    let chunks = 0;
```
pavelgj: super nit: this feels unnecessary, replace with `collectedChunks.length`

Issue #4262, opened 2026-01-25
[JS] Type overload shadowing for OpenAI models with specific suffixes


7hokerz:
**Describe the bug**
The generic GPT model overload (gpt-${string}) is placed before the specific STT/TTS/Image model overloads.

**To Reproduce**
1. Define a transcription model using the name gpt-4o-transcribe (or any name starting with gpt- and ending with -transcribe).
```
import { openAI } from '@genkit-ai/compat-oai/openai';

const whisper = openAI.model('gpt-4o-transcribe', {
  response_format: 'json'
});
```
2. Attempt to pass a transcription-specific config (e.g. `response_format` or `language`).

**Expected behavior**
The type system should correctly infer `TranscriptionConfigSchema`.

**Screenshots**
- In case of `whisper-1`
<img width="731" height="98" alt="Image" src="https://github.com/user-attachments/assets/7897a203-97c4-4e53-b8fa-72e4c7b14ddd" />

- In case of `gpt-4o-transcribe`
<img width="407" height="103" alt="Image" src="https://github.com/user-attachments/assets/dee16a1b-4faf-4bd0-ab2f-7848642e3bd1" />

**Runtime (please complete the following information):**
 - OS: windows 10 22H2

** Node version
 -  v24.13.0

**Additional context**
To solve this problem, Reorder the function overloads in type OpenAIPlugin so that specific models (Image, TTS, STT) are defined before the generic GPT models.

PR #4263, opened 2026-01-25
fix(js/plugin/compat-oai): Fix type overload shadowing bug for OpenAI models with specific suffixes

7hokerz:
 - Reordering the function overloads in index.ts
 - Fixed this issue: #4262

Checklist (if applicable):
- [V] PR title is following https://www.conventionalcommits.org/en/v1.0.0/
- [V] Tested (manually, unit tested, etc.)
- [V] Docs updated (updated docs or a docs bug required)


xavidop:
good catch!

PR #4295, opened 2026-01-27
feat(cli,js): implemented reflection API v2 in CLI and JS

pavelgj:
it's behind `--experimental-reflection-v2` flag.

RFC: https://github.com/firebase/genkit/pull/4211

Checklist (if applicable):

### DIFF
--- a/genkit-tools/common/src/manager/manager-v2.ts
+++ b/genkit-tools/common/src/manager/manager-v2.ts
@@ -16,6 +16,7 @@
 
 import EventEmitter from 'events';
 import getPort, { makeRange } from 'get-port';
+import path from 'path';
 import { WebSocket, WebSocketServer } from 'ws';
 import {
   Action,
@@ -190,7 +191,7 @@ export class RuntimeManagerV2 extends BaseRuntimeManager {
       reflectionApiSpecVersion: params.reflectionApiSpecVersion,
       reflectionServerUrl: `ws://localhost:${this.port}`, // Virtual URL for compatibility
       timestamp: new Date().toISOString(),
-      projectName: params.name || 'Unknown', // Or derive from other means if needed
+      projectName: path.basename(this.projectRoot), // Or derive from other means if needed
     };
 
     this.runtimes.set(runtimeInfo.id, { ws, info: runtimeInfo });


huangjeff5:
BTW we updated core/typing.py => _core/_typing.py to signal internal/private. Think you need to merge and regenerate the types.

MichaelDoyle (approved):
LGTM for genkit-tools

js/core/src/reflection-v2.ts
```diff
@@ -0,0 +1,399 @@
...
+
+interface JsonRpcResponse {
+  jsonrpc: '2.0';
+  result?: any;
+  error?: {
+    code: number;
+    message: string;
+    data?: any;
+  };
+  id: number | string;
```
apascal07: Did we want to keep this as union?

pavelgj: Good catch. Looks like I missed these. Should be `string` for simplicitly.

apascal07:
Is there any handling of reconnections, like if CLI restarts or machine goes to sleep? Maybe just have some simple fixed number back-off retries?

genkit-tools/common/src/types/reflection.ts
```diff
@@ -0,0 +1,116 @@
...
+ * ReflectionEndInputStreamSchema is the payload for the 'endInputStream' method.
+ */
+export const ReflectionEndInputStreamParamsSchema = z.object({
+  requestId: z.string(),
+});
+
+/**
+ * ReflectionListActionsResponseSchema is the result for the 'listActions' method.
+ */
+export const ReflectionListActionsResponseSchema = z.record(
```
apascal07: Don't feel strongly about it but do we want this to be a straight map instead of an object holding a map?

js/core/src/reflection-v2.ts
```diff
@@ -0,0 +1,394 @@
...
+  private register() {
+    const params = ReflectionRegisterParamsSchema.parse({
+      id: process.env.GENKIT_RUNTIME_ID || this.runtimeId,
+      pid: process.pid,
+      name: this.options.name || this.runtimeId,
+      genkitVersion: GENKIT_VERSION,
+      reflectionApiSpecVersion: GENKIT_REFLECTION_API_SPEC_VERSION,
+      envs: this.options.configuredEnvs,
+    });
+    this.sendNotification('register', params);
```
apascal07: What if registration fails? Should this be a request instead that can return an error?

js/core/src/reflection-v2.ts
```diff
@@ -0,0 +1,394 @@
...
+  }
+
+  private register() {
+    const params = ReflectionRegisterParamsSchema.parse({
+      id: process.env.GENKIT_RUNTIME_ID || this.runtimeId,
+      pid: process.pid,
+      name: this.options.name || this.runtimeId,
+      genkitVersion: GENKIT_VERSION,
+      reflectionApiSpecVersion: GENKIT_REFLECTION_API_SPEC_VERSION,
+      envs: this.options.configuredEnvs,
```
apascal07, line 237: This is not part of the schema it seems.

genkit-tools/common/src/types/reflection.ts
pavelgj: Good call! Done.

pavelgj:
> Is there any handling of reconnections, like if CLI restarts or machine goes to sleep? Maybe just have some simple fixed number back-off retries?

Interesting question... I was thinking connection disruption should not be a thing... but I guess it's possible. Added reconnect.

js/core/src/reflection-v2.ts
pavelgj: Fixed.

js/core/src/reflection-v2.ts
pavelgj: good call. combined register and configure into request/response combo.

PR #4302, opened 2026-01-27
fix(go/plugins/mcp): capture server errors

hugoaguirre:
- Refactored `serverRef.Error` to be an `error` instead of a `string`
- `NewGenkitMCPClient` now returns an error if the connection or init fails.
- `connect` captures any transport error detected

Checklist (if applicable):


hugoaguirre:
To be done in: https://github.com/firebase/genkit/pull/4388

PR #4314, opened 2026-01-28
docs(go): Add `GEMINI.md` instructions

hugoaguirre:
Checklist (if applicable):

### DIFF
--- a/go/GEMINI.md
+++ b/go/GEMINI.md
@@ -1,5 +1,15 @@
 # Go Development Guidelines
 
+## Agent Instructions
+
+As an AI developer working on this codebase, you MUST adhere to these operational protocols:
+
+* **Registry-First Thinking**: Genkit is built around a central registry. Always use `genkit.DefineX` (e.g., `DefineFlow`, `DefineTool`) for any component that should be discoverable by tools or the Dev UI.
+* **Context is Mandatory**: Every Genkit function and tool execution requires a `context.Context`. Pass it through faithfully; never use `context.Background()` inside an implementation unless it is the entry point.
+* **Schema Strictness**: When defining input/output types for Flows or Tools, use clear, JSON-tagged Go structs. The model uses these tags and types to understand the interface.
+* **Search Before Implementing**: Many common utilities exist in `internal/base` or `ai/`. Check there before implementing new JSON parsing or string manipulation logic.
+* **Idiomatic Concurrency**: Prefer Go's native concurrency (goroutines/channels) but be mindful of the `context` lifecycle within long-running background tasks.
+
 ## Architecture & Patterns
 
 * **Package Structure**:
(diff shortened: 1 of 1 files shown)


go/GEMINI.md
```diff
@@ -0,0 +1,161 @@
...
+* **Production Ready**: The objective is to produce production-grade code.
+* **Shift Left**: Employ a "shift left" strategy—catch errors early.
+* **Strict Typing**: Go is statically typed. Do not use `interface{}` (or `any`) unless absolutely necessary and documented.
+* **No Warning Suppression**: Avoid ignoring linter warnings unless there is a compelling, documented reason.
+* Group imports: standard library first, then third-party, then internal. `goimports` handles this automatically.
+
+## Generated Files & Data Model
+
+* **Do Not Edit Generated Files**: Files generated by tools (like protobufs or strict-type generators) should not be modified directly.
+* **Do Not Edit gen.go**: `go/ai/gen.go` is an auto-generated file. **DO NOT MODIFY IT DIRECTLY.**
```
apascal07, line 106: It'd be good to indicate here how to regenerate it.

go/GEMINI.md
```diff
@@ -0,0 +1,161 @@
...
+* **Generator/Sanitizer**: Any necessary transformations to the core types must be applied to the generator script or the schema sanitizer.
+* **Canonical Parity**: The data model MUST be identical to the JSON schema defined in the JavaScript (canonical) implementation.
+
+## API & Behavior Parity
+
+* **JS Canonical Conformance**: The Go implementation MUST be identical in API structure and runtime behavior to the JavaScript (canonical) implementation.
```
apascal07: This isn't true and it's too complicated to explain exactly what must match and what should be idiomatic. Better to leave it out.

go/GEMINI.md
hugoaguirre: Good point. Adding it

go/GEMINI.md
```diff
@@ -0,0 +1,169 @@
...
+* **Syntax**:
+  * Use standard Go formatting (`gofmt`).
+  * Use idioms like `if err != nil` for error handling.
+  * Prefer short variable names for short scopes (e.g., `i` for index, `ctx` for context).
+* **Interfaces**: Define interfaces where they are used (consumer-side), not where they are implemented (producer-side).
+* **Concurrency**: Use channels and goroutines for concurrency. Avoid shared mutable state where possible.
+* **Comments**:
+  * Use proper punctuation.
+  * Start function comments with the function name.
+  * Use `// TODO: Fix this later.` format for stubs.
```
yesudeep: Do you think it is better to have `// TODO(issue id): Fix this later` in the code so that we create an issue to track and the TODO links to that issue?

go/GEMINI.md
```diff
@@ -0,0 +1,169 @@
...
+      ```go
+      if diff := cmp.Diff(want, got); diff != "" {
+          t.Errorf("mismatch (-want +got):\n%s", diff)
+      }
+      ```
+
+* **Scope**: Write comprehensive unit tests.
+* **Execution**: Run via `go test ./...`.
+* **Porting**: Maintain 1:1 logic parity accurately if porting tests. Do not invent behavior.
+* **Fixes**: Fix underlying code issues rather than special-casing tests.
```
yesudeep, line 210: Perhaps also mention running https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize on the code

go/GEMINI.md
yesudeep: In Python, we've made this the rule so that issueless TODOs don't get into code and there is always an issue associated with one to track.

yesudeep (approved):

PR #4321, opened 2026-01-29
refactor(py): move session/chat into blocks and remove public API

huangjeff5:
Consolidate session/chat under blocks, drop ai-level session/chat surfaces, and
introduce shared interfaces to avoid type-only import cycles.

### DIFF
--- a/py/packages/genkit/src/genkit/blocks/interfaces.py
+++ b/py/packages/genkit/src/genkit/blocks/interfaces.py
@@ -18,7 +18,7 @@
 
 from __future__ import annotations
 
-from typing import Generic, Protocol, TypedDict, TypeVar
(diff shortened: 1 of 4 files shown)


py/packages/genkit/src/genkit/blocks/interfaces.py
```diff
@@ -0,0 +1,110 @@
...
+        if self.content_type is not None:
+            result['content_type'] = self.content_type
+        if self.instructions is not None:
+            result['instructions'] = self.instructions
+        if self.constrained is not None:
+            result['constrained'] = self.constrained
+        return result
+
+
+class ExecutablePromptLike(Protocol):
```
huangjeff5: Remove this usage

yesudeep (approved):

PR #4329, opened 2026-01-30
fix(go): Remove raw bytes from the debug messages

traut:
The fix adjust debug messages in `ActionDef.runWithTelemetry` function: it removes JSON bytes from the debug messages and accompanying JSON marshaling code. Debug messages with raw bytes produce unreadable noisy logs and introduce the risk of unintended data exposure through the logs.


hugoaguirre (approved):

traut:
could we get this in, please?

PR #4434, opened 2026-02-04
fix(py): fixed broken sample and lint errors of multi-server

MengqinShen:
Updated main.py to:

1. Pass list(range(...)) for ports (fixing the TypeError without modifying library code).
2. Write the runtime file on startup (fixing the Dev UI connection).
3. Initialize Genkit and share the registry (fixing "No registered actions").

### DIFF
--- a/py/pyproject.toml
+++ b/py/pyproject.toml
@@ -131,6 +131,7 @@ default-groups = ["dev", "lint"]
 
 [tool.uv.sources]
 # Samples (alphabetical by package name from pyproject.toml)
+amazon-bedrock-hello              = { workspace = true }
 anthropic-hello                   = { workspace = true }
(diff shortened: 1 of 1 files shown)


yesudeep (approved):

py/pyproject.toml
```diff
@@ -132,10 +132,13 @@ default-groups = ["dev", "lint"]
 
 [tool.uv.sources]
 # Samples (alphabetical by package name from pyproject.toml)
-amazon-bedrock-hello              = { workspace = true }
 anthropic-hello                   = { workspace = true }
+aws-hello                         = { workspace = true }
```
yesudeep: these plugins and samples have been consolidated. can you please keep the version of the pyproject.toml file from `main`?

PR #4453, opened 2026-02-05
feat(genkit-tools/cli): Add docs discovery tools to the CLI

ssbushi:
Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/py/uv.lock
+++ b/py/uv.lock
@@ -4415,7 +4415,7 @@ wheels = [
 
 [[package]]
 name = "litestar"
-version = "2.19.0"
+version = "2.20.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "anyio" },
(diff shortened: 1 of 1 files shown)


genkit-tools/cli/src/commands/docs.ts
```diff
@@ -0,0 +1,107 @@
...
+      );
+      results.forEach((r) => {
+        logger.info(`${clc.bold(r.doc.title)}`);
+        logger.info(`  Path: ${r.file}`);
+        if (r.doc.description) {
+          logger.info(`  ${clc.italic(r.doc.description)}`);
+        }
+        logger.info('');
+      });
+} catch (e: unknown) {
```
pavelgj: indentation is borken...

genkit-tools/cli/src/utils/docs.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export function searchDocs(
+  documents: Record<string, Doc>,
+  query: string,
+  lang: string
+) {
+  const terms = query
+    .toLowerCase()
+    .split(/\s+/)
+    .filter((t) => t.length > 2); // Filter out short words to reduce noise
```
pavelgj: might want to trim `s` at the end (naively normalize plurals to singular)
also, maybe filter out `the`...

ssbushi: I'm not sure about the naive singularization (I'm afraid we might mutate things we dont intend to). The current logic seems to be performing well, but I will implement the stop word removal.

PR #4464, opened 2026-02-06
feat(go/ai): added `DefineMiddleware` (Middleware V2)

### LINKED ISSUE #4466: RFC: Middleware V2
Doc(s): https://github.com/firebase/genkit/blob/ap/go-middleware-design/docs/go-middleware-design.md
PR(s): https://github.com/firebase/genkit/pull/4464 (Go)

apascal07:
Adds a `Middleware` interface for wrapping generation, model calls, and tool execution with composable hooks, plus the registration and plugin plumbing needed to expose middleware to the Dev UI and cross-runtime tooling. Built-in implementations (retry, fallback, tool approval, filesystem, skills) ship in a follow-up PR.

The core mental model: **a middleware is a config struct with two methods**. `Name()` returns its stable registered identifier and `New(ctx)` produces a per-call bundle of hook functions.

### Attaching middleware to a Generate call

Middleware is attached per call with `ai.WithUse`. The chain applies outer-to-inner, so `ai.WithUse(A, B)` expands to `A { B { model } }`:

```go
response, _ := genkit.Generate(ctx, g,
    ai.WithPrompt("Explain quantum computing."),
    ai.WithUse(
        Logger{Prefix: "[trace]"},
        SomeOtherMiddleware{},
    ),
)
```

No plugin registration is required for pure-Go use. `WithUse` calls the config's `New` directly on the local fast path. Registration is what makes middleware visible to the Dev UI and callable from other runtimes.

### Defining your own middleware

Implement `ai.Middleware` by providing two methods on a config struct. `Name()` returns the registered identifier (a stable constant read from the zero value of the type). `New(ctx)` returns an `*ai.Hooks` value with any of `WrapGenerate`, `WrapModel`, `WrapTool`, and `Tools` populated. Nil fields pass through.

```go
type Logger struct {
    Prefix string `json:"prefix,omitempty"`
}

func (Logger) Name() string { return "mine/logger" }

func (l Logger) New(ctx context.Context) (*ai.Hooks, error) {
    return &ai.Hooks{
        WrapModel: func(ctx context.Context, p *ai.ModelParams, next ai.ModelNext) (*ai.ModelResponse, error) {
            start := time.Now()
            resp, err := next(ctx, p)
            log.Printf("%s model call took %s", l.Prefix, time.Since(start))
            return resp, err
        },
    }, nil
}
```

Per-call state is closure-captured inside `New` (which runs once per `Generate` invocation). Plugin-level state belongs on unexported fields of the config struct; the plugin's `Middlewares()` method sets those fields on a prototype that value-copies across JSON-dispatched calls to preserve them.

### Ad-hoc inline middleware

For one-off middleware that does not need Dev UI visibility or a named type, use `ai.MiddlewareFunc`:

```go
ai.WithUse(ai.MiddlewareFunc(func(ctx context.Context) (*ai.Hooks, error) {
    return &ai.Hooks{
        WrapModel: func(ctx context.Context, p *ai.ModelParams, next ai.ModelNext) (*ai.ModelResponse, error) {
            log.Printf("model call")
            return next(ctx, p)
        },
    }, nil
}))
```

### Plugin-provided middleware

Plugins advertise middleware by implementing `MiddlewarePlugin`. `Middlewares` is called during `genkit.Init`, and the returned descriptors become visible to the Dev UI and available for lookup by name. Any unexported fields on the prototype (like `client` below) are preserved across Dev UI and cross-runtime invocations via value-copy in the descriptor's build closure.

```go
type MyPlugin struct{
    client *http.Client
}

func (p *MyPlugin) Name() string                          { return "myplugin" }
func (p *MyPlugin) Init(ctx context.Context) []api.Action { return nil }

func (p *MyPlugin) Middlewares(ctx context.Context) ([]*ai.MiddlewareDesc, error) {
    return []*ai.MiddlewareDesc{
        ai.NewMiddleware("description of my middleware", Logger{client: p.client}),
    }, nil
}
```

---

### `ai.Middleware` interface

```go
// A middleware is identified by a stable name and produces a fresh hooks
// bundle for each Generate() call.
type Middleware interface {
    Name() string
    New(ctx context.Context) (*Hooks, error)
}
```

### `ai.Hooks` bundle

```go
type Hooks struct {
    Tools        []Tool
    WrapGenerate func(ctx context.Context, params *GenerateParams, next GenerateNext) (*ModelResponse, error)
    WrapModel    func(ctx context.Context, params *ModelParams, next ModelNext) (*ModelResponse, error)
    WrapTool     func(ctx context.Context, params *ToolParams, next ToolNext) (*MultipartToolResponse, error)
}
```

- `WrapGenerate` wraps each iteration of the tool loop (invoked `N+1` times for a generate with `N` tool turns).
- `WrapModel` wraps each model API call.
- `WrapTool` wraps each tool execution and may run concurrently for parallel tool calls.
- `Tools` contributes additional tools to register for the generation.

### Hook parameters

```go
// GenerateParams holds params for the WrapGenerate hook.
type GenerateParams struct {
    Options      *GenerateActionOptions // original options passed to Generate
    Request      *ModelRequest          // current model request with accumulated messages
    Iteration    int                    // zero-indexed tool-loop iteration
    MessageIndex int                    // next index for streamed response chunks
    Callback     ModelStreamCallback    // streaming callback, nil if not streaming
}

// ModelParams holds params for the WrapModel hook.
type ModelParams struct {
    Request  *ModelRequest
    Callback ModelStreamCallback
}

// ToolParams holds params for the WrapTool hook.
type ToolParams struct {
    Request *ToolRequest
    Tool    Tool
}

// Chain continuation functions invoked by hooks to pass control to the next layer.
type GenerateNext = func(ctx context.Context, params *GenerateParams) (*ModelResponse, error)
type ModelNext    = func(ctx context.Context, params *ModelParams) (*ModelResponse, error)
type ToolNext     = func(ctx context.Context, params *ToolParams) (*MultipartToolResponse, error)
```

### Registration

```go
// NewMiddleware constructs a descriptor without registering it. The prototype
// supplies both the registered name (via its Name() method) and any plugin-
// level state that should flow into JSON-dispatched invocations via unexported
// fields preserved by value-copy.
func NewMiddleware[M Middleware](description string, prototype M) *MiddlewareDesc

// DefineMiddleware creates and registers a middleware descriptor.
func DefineMiddleware[M Middleware](r api.Registry, description string, prototype M) *MiddlewareDesc

// LookupMiddleware returns the registered middleware descriptor with the given name.
func LookupMiddleware(r api.Registry, name string) *MiddlewareDesc
```

Matching conveniences exist in the `genkit` package:

```go
func genkit.DefineMiddleware[M ai.Middleware](g *Genkit, description string, prototype M) *ai.MiddlewareDesc
func genkit.LookupMiddleware(g *Genkit, name string) *ai.MiddlewareDesc
```

Application code typically uses `genkit.DefineMiddleware` to register middleware it owns directly. Plugin authors typically use `ai.NewMiddleware` and return descriptors from `MiddlewarePlugin.Middlewares()`; `genkit.Init` registers them during plugin setup.

### Inline adapter

```go
// MiddlewareFunc adapts a per-call factory closure to the Middleware interface
// for ad-hoc inline middleware, without a registered descriptor or plugin wiring.
type MiddlewareFunc func(ctx context.Context) (*Hooks, error)
```

### Attaching to a call

```go
// WithUse attaches middleware to a Generate call. Accepts either a middleware
// config struct or a MiddlewareFunc adapter. The chain applies outer-to-inner.
func WithUse(middleware ...Middleware) CommonGenOption
```

### Plugin integration

```go
// MiddlewarePlugin is implemented by plugins that provide middleware.
// Middlewares() is called during genkit.Init; returned descriptors become
// visible to the Dev UI and are reachable via the reflection API.
type MiddlewarePlugin interface {
    Middlewares(ctx context.Context) ([]*MiddlewareDesc, error)
}
```

---

## Design notes

- **Config struct is the middleware.** The config value the user passes to `WithUse` is the same type whose `New` method the runtime invokes. Plugin-level state rides on unexported fields and is preserved across JSON dispatch by value-copy inside the descriptor's build closure. Per-call state lives in closures captured by `New`.
- **No factory parameter, no prototype copy by hand.** `NewMiddleware` captures the typed prototype in a closure that unmarshals JSON on a value copy (preserving unexported fields, sharing pointers). No `BaseMiddleware` embedding or field-by-field copy required.
- **Hook chains are built once per `GenerateWithRequest`** and reused across every tool-loop iteration instead of being rebuilt per turn.
- **Local calls skip the registry entirely.** `WithUse(Retry{...})` invokes `Retry.New(ctx)` directly on the fast path; the registry is consulted only for JSON-dispatched invocations from the Dev UI or cross-runtime callers.

## Other changes

- The existing `ai.ModelMiddleware` / `ai.WithMiddleware` API is preserved and marked deprecated. New code should prefer `ai.Middleware` / `ai.WithUse`, which adds `WrapGenerate` and `WrapTool` hooks and supports dynamically injected tools via `Hooks.Tools`.
- `MultipartToolResponse.Content` now flows through the resume path (previously dropped in `handleResumedToolRequest`).
- `WrapTool` returns `*MultipartToolResponse` so metadata and content round-trip without an out-of-band capture.
- Duplicate middleware-contributed tool names are rejected explicitly in `GenerateWithRequest` rather than panicking during registry registration.
- `NewToolInterruptError` is exported so `WrapTool` hooks can interrupt tools without constructing a `ToolContext`.
- Schema updates in `genkit-tools` and `py/packages/genkit` keep the middleware reference type consistent across the JS, Go, and Python runtimes.
- Ships without any built-in implementations. See the follow-up PR for `Retry`, `Fallback`, `ToolApproval`, `Filesystem`, and `Skills`.


huangjeff5 (approved):

genkit-tools/common/src/types/index.ts
pavelgj, line 1: please make sure genkit-schema.json file gets updated with new schemas

pavelgj: nevermind, I'm blind...

genkit-tools/common/src/types/middleware.ts
```diff
@@ -0,0 +1,36 @@
...
+import { z } from 'zod';
+
+/** Descriptor for a registered middleware, returned by reflection API. */
+export const MiddlewareDescSchema = z.object({
+  /** Unique name of the middleware. */
+  name: z.string(),
+  /** Human-readable description of what the middleware does. */
+  description: z.string().optional(),
+  /** JSON Schema for the middleware's configuration. */
+  configSchema: z.record(z.any()).nullish(),
```
pavelgj, line 25:
```suggestion
  configSchema: z.record(z.any()).nullish(),
  /** User defined metadata for the middleware. */
  metadata: z.record(z.any()).optional(),
```

pavelgj (approved):

MichaelDoyle (approved):
LGTM from a `genkit-tools` perspective.

apascal07:
@pavelgj Minor but pretty great improvements to the middleware abstraction in case you want to review it again. The PR description covers it all.

go/ai/middleware_test.go
```diff
@@ -0,0 +1,576 @@
...
+	})
+
+	_, err := Generate(testCtx, r,
+		WithModelName("test/toolLoop"),
+		WithPrompt("use it"),
+		WithTools(ToolName("myTool")),
+		WithUse(tracker),
+	)
+	assertNoError(t, err)
+	if got := atomic.LoadInt32(&iters); got < 2 {
```
pavelgj, line 531: it would be good to assert on the correct order as well...  ex: in the "tracker" middleware you could wrap generate and tool as well, and create an ordered "log", then assert that the hook order was correct: `["generate", "tool", "generate",  "model"]`

pavelgj: actually, this is a completely different test case... would be good to add one for tool restarts.... but also more generally, from experience, this is a good assertion pattern for middleware

apascal07: Done.

pavelgj (approved):

PR #4495, opened 2026-02-07
fix(py/core): prevent infinite recursion in create_prompt_from_file()

### LINKED ISSUE #4491: bug(py): Infinite recursion in create_prompt_from_file() when loading .variant.prompt files
## Bug Description

Loading any file-based prompt (most visibly `.variant.prompt` files) causes infinite recursion in `create_prompt_from_file()`, resulting in `RecursionError: maximum recursion depth exceeded`.

**Error message:**
```
Failed to load lazy action recipe.robot: maximum recursion depth exceeded
```

**Previously filed on google/dotprompt (closed -- not a dotprompt bug):**
https://github.com/google/dotprompt/issues/523

## Root Cause

`create_prompt_from_file()` in `py/packages/genkit/src/genkit/blocks/prompt.py` (line ~2194) calls `registry.resolve_action_by_key()` with **its own action key** before setting `_cached_prompt`. This triggers `_trigger_lazy_loading()` in `registry.py`, which re-invokes `create_prompt_from_file()`. Since `_cached_prompt` is still `None`, the memoization guard doesn't fire, and the function re-enters fully -- infinite recursion.

### Recursion cycle

```
create_prompt_from_file()                         # invocation 1
  _cached_prompt is None → proceed
  executable_prompt = ExecutablePrompt(...)
  registry.resolve_action_by_key(own_key)         # line ~2194 ← SELF-REFERENCE
    → _trigger_lazy_loading(same_action)           # registry.py:377
      → async_factory()                            # registry.py:350
        → create_prompt_from_file()                # invocation 2
          _cached_prompt is STILL None → proceed
          registry.resolve_action_by_key(own_key)  # line ~2194
            → _trigger_lazy_loading(same_action)
              → async_factory()
                → create_prompt_from_file()        # invocation 3
                  → ... (repeats ~65-100x until RecursionError)
```

There is also a second self-referential `resolve_action_by_key` call at line ~2200 for the `EXECUTABLE_PROMPT` action, which has the same `async_factory`.

### Why it manifests most visibly with variants

The bug affects **all** file-based prompts, but `RecursionError` is caught by `except Exception` in `_trigger_lazy_loading()` and swallowed as a warning log. Prompts may appear to work despite the bug. Variants were where the failure was first noticed because both the base and variant go through this loop during initial loading.

## Cross-Runtime Impact

- **Python**: Affected.
- **JavaScript/TypeScript**: NOT affected. JS uses a `lazy()` wrapper with single-evaluation semantics and doesn't have the self-referential `resolve_action_by_key()` pattern.
- **google/dotprompt library**: NOT affected. The dotprompt library only handles parsing/rendering. The bug is in Genkit's registry-based lazy loading layer.

## Reproduction

Place these files in a `prompts/` directory:

### `prompts/recipe.prompt` (base prompt)

```yaml
---
model: googleai/gemini-3-flash-preview
input:
  schema:
    food: string
    ingredients?(array): string
output:
  schema: Recipe
  format: json
---

You are a chef famous for making creative recipes that can be prepared in 45 minutes or less.

Generate a recipe for {{food}}.

{{#if ingredients}}
Make sure to include the following ingredients:
{{list ingredients}}
{{/if}}
```

### `prompts/recipe.robot.prompt` (variant -- triggers visible failure)

```yaml
---
model: googleai/gemini-3-flash-preview
input:
  schema:
    food: string
  default:
    food: binary burgers
output:
  schema:
    title: string, recipe title
    ingredients(array):
      name: string
      quantity: string
    steps(array, the steps required to complete the recipe): string
---

You are a robot chef famous for making creative recipes that robots love to eat. Robots love things like motor oil, RAM, bolts, and uranium.

Generate a recipe for {{food}}.
```

### Python entry point

```python
import asyncio
from genkit.ai import Genkit
from genkit.plugins.google_genai import GoogleAI

ai = Genkit(plugins=[GoogleAI()], model='googleai/gemini-3-flash-preview')

async def main():
    # This triggers the infinite recursion
    recipe_robot = await ai.prompt('recipe', variant='robot')
    result = await recipe_robot(input={'food': 'pizza'})
    print(result.text)

if __name__ == '__main__':
    ai.run_main(main())
```

### Fix 1: Set `_cached_prompt` before self-referential calls (simplest)

In `py/packages/genkit/src/genkit/blocks/prompt.py`, inside `create_prompt_from_file()`, move the `_cached_prompt = executable_prompt` assignment to **before** the `resolve_action_by_key` calls:

```python
# Set cache BEFORE resolving actions to break self-recursion
_cached_prompt = executable_prompt

# Now these calls won't recurse because _cached_prompt is already set
prompt_action = await registry.resolve_action_by_key(prompt_lookup_key)
exec_prompt_action = await registry.resolve_action_by_key(exec_prompt_lookup_key)
```

### Fix 2: Add re-entrancy guard in `_trigger_lazy_loading()` (robust)

In `py/packages/genkit/src/genkit/core/registry.py`, add a re-entrancy guard set:

```python
# Add to Registry.__init__:
self._loading_actions: set[str] = set()

async def _trigger_lazy_loading(self, action: Action | None) -> Action | None:
    if action is None:
        return None
    async_factory = getattr(action, '_async_factory', None)
    if async_factory is not None and action.metadata.get('lazy'):
        action_id = f'{action.kind}/{action.name}'
        if action_id in self._loading_actions:
            return action  # Already being loaded, skip
        self._loading_actions.add(action_id)
        try:
            await async_factory()
        except Exception as e:
            logger.warning(f'Failed to load lazy action {action.name}: {e}')
        finally:
            self._loading_actions.discard(action_id)
    return action
```

### Files to change

- `py/packages/genkit/src/genkit/blocks/prompt.py` (line ~2194-2214) -- move `_cached_prompt` assignment earlier
- `py/packages/genkit/src/genkit/core/registry.py` (line ~338-353) -- add `_loading_actions` re-entrancy guard

## Workaround

Remove `.variant.prompt` files to avoid triggering the visible failure. The underlying self-recursion still occurs for all file-based prompts but is swallowed as a warning.

yesudeep:
## Description

Fixes #4491 — `RecursionError` when loading file-based prompts (especially `.variant.prompt` files).

### Root Cause

`create_prompt_from_file()` calls `resolve_action_by_key()`, which triggers `_trigger_lazy_loading()`, which calls the same factory again — infinite recursion.

### Fix

**Registry re-entrancy guard** (`registry.py`): Added a `_loading_actions` set to `_trigger_lazy_loading()` that tracks which actions are currently being loaded. If a factory tries to resolve its own action key during initialization, the guard short-circuits and returns the action immediately — breaking the cycle.

### What changed (updated per review)

- **`registry.py`**: Added `_loading_actions` re-entrancy guard to `_trigger_lazy_loading()`
- **`prompt.py`**: No changes (reverted per @huangjeff5 review — the registry guard alone is sufficient)
- **Tests**: Two regression tests validate both the re-entrancy guard and the variant prompt loading

### Testing

Both regression tests pass with only the registry fix:
- `test_variant_prompt_loading_does_not_recurse`
- `test_trigger_lazy_loading_reentrant_guard`

### DIFF
--- a/py/packages/genkit/src/genkit/blocks/prompt.py
+++ b/py/packages/genkit/src/genkit/blocks/prompt.py
@@ -2184,13 +2184,6 @@ def load_prompt(registry: Registry, path: Path, filename: str, prefix: str = '',
             _ns=ns,  # Store namespace for action lookup
         )
 
-        # Cache the result BEFORE resolving actions to break self-recursion.
-        # resolve_action_by_key() triggers _trigger_lazy_loading(), which calls
-        # this factory again.  Without the early cache assignment the memoization
-        # guard (_cached_prompt is not None) never fires and we recurse until
-        # RecursionError.  See https://github.com/firebase/genkit/issues/4491
-        _cached_prompt = executable_prompt
-
         # Store reference to PROMPT action on the ExecutablePrompt
         # Actions are already registered at this point (lazy loading happens after registration)
         definition_key = registry_definition_key(name, variant, ns)
@@ -2217,6 +2210,8 @@ def load_prompt(registry: Registry, path: Path, filename: str, prefix: str = '',
                 if output_json_schema:
                     action.output_schema = output_json_schema
 
+        # Cache the result for subsequent calls
+        _cached_prompt = executable_prompt
         return executable_prompt
 
     # Store the async factory in a way that can be accessed later


huangjeff5:
Hey, The fix in genkit/core/registry.py looks good.

I think we should revert the changes to genkit/blocks/prompt.py. Since the registry fix already handles the recursion, the extra caching logic there isn't needed.

yesudeep:
@huangjeff5 done

PR #4500, opened 2026-02-08
fix(py): add dropped_attributes/events/links to RedactedSpan for OTLP export

### LINKED ISSUE #4493: bug(py): RedactedSpan crashes OTLP export with AttributeError on dropped_attributes
## Description

The `RedactedSpan` wrapper class in `genkit.core.trace.adjusting_exporter` causes an
`AttributeError` when the OTLP trace encoder tries to serialize spans for export to
backends like AWS X-Ray.

## Error

```
AttributeError: 'TimeAdjustedSpan' object has no attribute '_attributes'
```

Full traceback:
```
opentelemetry/.../trace_encoder/__init__.py:130 in _encode_span
    dropped_attributes_count=sdk_span.dropped_attributes,

opentelemetry/sdk/trace/__init__.py:412 in dropped_attributes
    if isinstance(self._attributes, BoundedAttributes):
                  ^^^^^^^^^^^^^^^^
AttributeError: 'TimeAdjustedSpan' object has no attribute '_attributes'
```

## Root Cause

`RedactedSpan` (and its subclass `TimeAdjustedSpan`) wraps a `ReadableSpan` without
calling `super().__init__()`. This is intentional to avoid duplicating span state.
However, three properties in the base `ReadableSpan` class access private instance
variables that are only set by `ReadableSpan.__init__()`:

- `dropped_attributes` → accesses `self._attributes`
- `dropped_events` → accesses `self._events`
- `dropped_links` → accesses `self._links`

Since `RedactedSpan.__init__()` never calls `super().__init__()`, these private
fields don't exist, and the OTLP encoder crashes when it tries to serialize
span metadata.

## Impact

Any telemetry plugin that uses `AdjustingTraceExporter` (AWS Bedrock, Google Cloud,
etc.) will crash at span export time, silently dropping all trace data.

## Fix

Add `dropped_attributes`, `dropped_events`, and `dropped_links` property overrides
to `RedactedSpan` that delegate to `self._span` (the wrapped span).

## Files

- `py/packages/genkit/src/genkit/core/trace/adjusting_exporter.py`
- `py/packages/genkit/tests/genkit/core/trace/adjusting_exporter_test.py`

kaigritun:
## Description

This PR fixes the `AttributeError: 'TimeAdjustedSpan' object has no attribute '_attributes'` crash that occurs when exporting traces via OTLP to backends like AWS X-Ray.

## Root Cause

`RedactedSpan` wraps a `ReadableSpan` without calling `super().__init__()` (intentionally, to avoid duplicating span state). However, three properties in the base `ReadableSpan` class access private instance variables that are only set by `ReadableSpan.__init__()`:

- `dropped_attributes` → accesses `self._attributes`
- `dropped_events` → accesses `self._events`
- `dropped_links` → accesses `self._links`

When the OTLP encoder tries to serialize span metadata, it crashes because these private fields don't exist.

## Fix

Added property overrides to `RedactedSpan` that delegate to the wrapped span:

```python
@property
@override
def dropped_attributes(self) -> int:
    return self._span.dropped_attributes

@property
@override
def dropped_events(self) -> int:
    return self._span.dropped_events

@property
@override
def dropped_links(self) -> int:
    return self._span.dropped_links
```

## Testing

Added 4 new tests to verify the delegation works correctly:
- `test_redacted_span_exposes_dropped_attributes`
- `test_redacted_span_exposes_dropped_events`
- `test_redacted_span_exposes_dropped_links`
- `test_redacted_span_dropped_properties_default_to_zero`

All 18 tests pass.

## Impact

This enables any telemetry plugin that uses `AdjustingTraceExporter` (AWS Bedrock, Google Cloud, etc.) to properly export traces via OTLP without crashing.

Fixes #4493

### DIFF
--- a/py/packages/genkit/src/genkit/core/trace/adjusting_exporter.py
+++ b/py/packages/genkit/src/genkit/core/trace/adjusting_exporter.py
@@ -182,6 +182,24 @@ class RedactedSpan(ReadableSpan):
         """Return the instrumentation scope."""
         return self._span.instrumentation_scope
 
+    @property
+    @override
+    def dropped_attributes(self) -> int:
+        """Return the number of dropped attributes."""
+        return self._span.dropped_attributes
+
+    @property
+    @override
+    def dropped_events(self) -> int:
+        """Return the number of dropped events."""
+        return self._span.dropped_events
+
+    @property
+    @override
+    def dropped_links(self) -> int:
+        """Return the number of dropped links."""
+        return self._span.dropped_links
+
 
 class AdjustingTraceExporter(SpanExporter):
     """Adjusts spans before exporting for PII redaction and enhancement.
--- a/py/packages/genkit/tests/genkit/core/trace/adjusting_exporter_test.py
+++ b/py/packages/genkit/tests/genkit/core/trace/adjusting_exporter_test.py
@@ -394,3 +394,98 @@ def test_error_handler_called_on_export_error() -> None:
 
     assert len(errors) == 1
     assert str(errors[0]) == 'Export failed'
+
+
+# =============================================================================
+# RedactedSpan Dropped Properties Tests (OTLP Compatibility)
+# =============================================================================
+
+
+def test_redacted_span_exposes_dropped_attributes() -> None:
+    """Test that RedactedSpan delegates dropped_attributes to wrapped span.
+
+    This is critical for OTLP export compatibility. The OTLP trace encoder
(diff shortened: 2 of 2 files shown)


yesudeep:
https://github.com/firebase/genkit/pull/4494 should have fixed this. Let me know and we can close this PR.

kaigritun:
Thanks @yesudeep! Just checked - #4494 has the same fix. Closing this as redundant. Appreciate the heads up!

PR #4576, opened 2026-02-11
fix(py): fix Dev UI error handling for invalid api key

MengqinShen:
This PR improves the quality and clarity of error messages returned by the Genkit Python SDK's reflection API, ensuring consistency with the JavaScript SDK and providing a clean experience in the Dev UI.
Summary of Changes
* Reflection API Format Alignment: Updated   reflection.py to wrap error responses in an   error field. This matches the canonical (JS) SDK behavior and prevents JavaScript "undefined" errors in the Dev UI when processing action failures.
* Clean Error Serialization: Modified GenkitError.to_serializable in   error.py to use the original_message instead of a verbose repr(cause). This ensures that technical details like full JSON response payloads are not part of the primary error message shown to users.
* Action Wrapper Transparency: Updated the action tracing wrappers in   _action.py to preserve original   GenkitError exceptions. Previously, these were caught and wrapped in a generic "Error while running action..." message, which obscured the actual underlying root cause (e.g., "API key not valid").

image

### DIFF
--- a/py/packages/genkit/src/genkit/core/error.py
+++ b/py/packages/genkit/src/genkit/core/error.py
@@ -213,7 +213,7 @@ class GenkitError(Exception):
         return GenkitReflectionApiErrorWireFormat(
             details=GenkitReflectionApiDetailsWireFormat(**self.details) if self.details else None,
             code=StatusCodes[self.status].value,
-            message=self.original_message,
+            message=f'{self.original_message}: {repr(self.cause)}' if self.cause else self.original_message,
         )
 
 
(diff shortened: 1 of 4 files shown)


py/packages/genkit/src/genkit/core/error.py
```diff
@@ -214,7 +213,7 @@ def to_serializable(self) -> GenkitReflectionApiErrorWireFormat:
         return GenkitReflectionApiErrorWireFormat(
             details=GenkitReflectionApiDetailsWireFormat(**self.details) if self.details else None,
             code=StatusCodes[self.status].value,
-            message=repr(self.cause) if self.cause else self.original_message,
+            message=f'{self.original_message}: {repr(self.cause)}' if self.cause else self.original_message,
```
jeffdh5: Good improvement to include the `original_message` alongside the cause. Previously, `repr(self.cause)` completely replaced the user-friendly message, so the Dev UI would show something like `ClientError('...<full JSON response>...')` with no context.

Nit: note the asymmetry with `to_callable_serializable()` (line ~206) which still uses the old `repr(self.cause) if self.cause else self.original_message` pattern. Was this intentional because callable errors follow a different contract, or should both methods be updated for consistency?

jeffdh5 (changes requested):
PR mostly looks good. Think we should continue returning the full traceback ideally - or at least some helpful guiding information.

PR #4616, opened 2026-02-12
fix(py): add logger to live test in flow testing tool

MengqinShen:


MengqinShen:
merged into https://github.com/firebase/genkit/pull/4648

PR #4654, opened 2026-02-13
fix: Add Firebase dashboard resource attributes for Python telemetry visibility

zarinn3pal:
This PR adds dashboard resource attributes so that AIM dashboard gets populated


huangjeff5:
Hmm this doesn't seem right. I was able to get some stuff to show in AIM dashboard just by fixing the logging pattern.

Issue #4652, opened 2026-02-13
[JS] Prompts always appends user message on resume, breaking interrupt resume


Ehesp:
Given the scenario; I've got a Genkit prompt which generates a poem. The prompt has a tool, which handles interruption (asking the user to review the poem).

I invoke this like so:

```ts
const response = await poemPrompt(input, {
  messages: input.resumable?.messages ?? [],
  resume: input.resumable ? {
    respond: input.resumable?.responses,
  } : undefined,
});
```

This works fine, and the tool is triggered. I collate the response(s), the messages which existed on the `response` at the time the interrupts happened.

I then run the prompt again with the `resumable` object, containing those messages and responses, as above.

I get an error:

```
GenkitError3: FAILED_PRECONDITION: Cannot 'resume' generation unless the previous message is a model message with at least one tool request.
```

Looking at the [code](https://github.com/firebase/genkit/blob/main/js/ai/src/prompt.ts#L546):

When resuming an interrupted generation by calling a defined prompt with existing history and the resume payload (e.g. `prompt(input, { messages: response.messages, resume: { respond: answers } }))`, the prompt pipeline still appends the template as a new user message in `renderUserPrompt()`.

That makes the last message in messages a user message instead of the model message that contains the tool request, so the resume check in generate() fails with "Cannot 'resume' generation unless the previous message is a model message with at least one tool request." The resume option is correctly forwarded into the final generate options, but the message list is already wrong.

I think the prompt layer should skip appending the user prompt when renderOptions.resume is set (e.g. in renderUserPrompt(), only push the user message when !renderOptions?.resume), but I'm not 100% sure.

**Runtime (please complete the following information):**
 - OS: MacOS
 - Version 26.2

** Node version
v24.8.0

**Additional context**

Note, I'm not using dotprompt files, I'm defining the [prompts in code](https://genkit.dev/docs/dotprompt/#defining-prompts-in-code) but not sure this makes any difference.

PR #4657, opened 2026-02-13
fix(js): do not append prompt when resuming generation (#4652)

BenInvertase:
Fixes https://github.com/firebase/genkit/issues/4652

When resuming an interrupted generation with `messages` and `resume`, the SDK was still appending the prompt as a new user message, so the last message was no longer the model message with a tool request and the resume precondition failed. This change skips appending the prompt when `resume` is set in both `generate.ts` (for raw `generate` calls) and `prompt.ts` (for `definePrompt`).

### DIFF
--- a/js/ai/src/generate.ts
+++ b/js/ai/src/generate.ts
@@ -190,6 +190,13 @@ export async function toGenerateRequest(
   registry: Registry,
   options: GenerateOptions
 ): Promise<GenerateRequest> {
+  if (options.prompt && options.resume) {
+    throw new GenkitError({
+      status: 'INVALID_ARGUMENT',
+      message:
+        'prompt is not supported when resume is set. The message history in messages is used instead.',
+    });
+  }
   const messages: MessageData[] = [];
   if (options.system) {
     messages.push({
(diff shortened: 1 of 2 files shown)


js/ai/src/generate.ts
```diff
@@ -343,7 +343,7 @@ function messagesFromOptions(options: GenerateOptions): MessageData[] {
   if (options.messages) {
     messages.push(...options.messages);
```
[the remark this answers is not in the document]
cabljac: this seems preexisting? not something introduced here.

js/ai/src/generate.ts
```diff
@@ -200,7 +200,7 @@ export async function toGenerateRequest(
   if (options.messages) {
     messages.push(...options.messages.map((m) => Message.parseData(m)));
   }
-  if (options.prompt) {
+  if (options.prompt && !options.resume) {
```
cabljac, line 210: (same as other comment in this file i think)

js/ai/src/generate.ts
```diff
@@ -343,7 +343,7 @@ function messagesFromOptions(options: GenerateOptions): MessageData[] {
   if (options.messages) {
     messages.push(...options.messages);
   }
-  if (options.prompt) {
+  if (options.prompt && !options.resume) {
```
cabljac, line 360: only thing i'm thinking is if a user explicitly does e.g

```typescript
ai.generate({
  prompt: 'count to 10',
  messages: [...],
  resume: { respond: [...] },
})
```
then the prompt is silently ignored, which might be confusing -

might be better to be defensive here and do:

```typescript
  if (options.prompt && options.resume) {
    throw new GenkitError({
      status: 'INVALID_ARGUMENT',
      message: 'prompt is not supported when resume is set. The message history in messages is used instead.',
    });
  }
```

BenInvertase:
@cabljac Thanks for the review. I've updated based on your suggestion.

PR #4681, opened 2026-02-15
fix(samples): fix ModuleNotFoundError for samples.shared and improve setup UX

yesudeep:
Root cause: run.sh scripts cd into sample subdirectories before launching
`uv run`, but py/ was never added to PYTHONPATH, so `from samples.shared
import ...` failed at runtime for all 38 samples that use shared code.

Fixes:
- _common.sh: export PYTHONPATH with py/ resolved from BASH_SOURCE[0]
- _common.sh: add check_setup() that prompts to run setup.sh when uv is
  missing or .venv is absent — called automatically by install_deps()
- _common.sh: add GENKIT_NO_BROWSER=1 support to genkit_start_with_browser
  for CI/headless usage
- setup.sh: improve key prompt UX — show current value on its own line
  with ✓, then prompt "New value (Enter=keep current)" instead of the
  confusing single-line format
- setup.sh: prompt to run `gcloud components update` when gcloud was not
  installed via a package manager (brew/apt/dnf)
- setup.sh: force-refresh .venv by removing it before uv sync to avoid
  stale cached environment issues
- pyproject.toml: add pythonpath=["."] to pytest config so samples.shared
  is importable during test runs

Tests:
- Add tests/smoke/sample_import_test.py (77 pass, 3 skip):
  - samples.shared imports cleanly and exports expected names
  - Every sample main.py is syntactically valid Python (parametrized)
  - Every `from samples.shared import X` resolves to a real export

ISSUE: https://github.com/firebase/genkit/issues/4674

### DIFF
--- a/py/pyproject.toml
+++ b/py/pyproject.toml
@@ -150,6 +150,7 @@ addopts = [
   "--ignore=tools/conform/tests",
   "--import-mode=importlib",
 ]
+pythonpath = ["."]  # For samples.shared imports
 asyncio_default_fixture_loop_scope = "session"
 norecursedirs = ["testapps", ".git", ".tox", ".nox", ".venv", "build", "dist"]
 python_files = ["*_test.py"]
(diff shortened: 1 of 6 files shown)


prashant4nov:
Screenshot 2026-02-15 at 7 53 10 AM
Screenshot 2026-02-15 at 7 53 17 AM
The error is gone. Thank you so much for the quick fix!

PR #4703, opened 2026-02-16
fix(go/plugins/anthropic): resolve models using aliases

hugoaguirre:
Providing a claude model using an alias without the `@YYYYMMDD` format will resolve into a valid model name

Checklist (if applicable):

### DIFF
--- a/go/plugins/anthropic/anthropic.go
+++ b/go/plugins/anthropic/anthropic.go
@@ -22,6 +22,7 @@ import (
 	"log/slog"
 	"os"
 	"reflect"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -211,6 +212,7 @@ func newModel(client anthropic.Client, name, apiModelName string, opts ai.ModelO
 }
 
 func resolveModelID(id string, availableModels []string) (string, bool) {
+	// First check for exact match
 	for _, m := range availableModels {
 		if m == id {
 			return m, true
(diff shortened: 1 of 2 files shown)


go/plugins/anthropic/anthropic.go
```diff
@@ -143,8 +163,27 @@ func (a *Anthropic) ResolveAction(atype api.ActionType, id string) api.Action {
...
+
+	models, err := listModels(ctx, &a.aclient)
+	if err != nil {
+		return nil, err
+	}
+
+	a.models = models
+	a.lastUpdated = time.Now()
+	return models, nil
+}
```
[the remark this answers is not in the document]
hugoaguirre: Having the cache to prevent redundant API calls when requesting new model aliases seems to be enough to enhance performance. The `singleflight` approach might seem a little bit overkill at this point.

go/plugins/anthropic/anthropic_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+package anthropic
+
+import (
+	"testing"
+)
+
+func TestResolveModelID(t *testing.T) {
+	availableModels := []string{
+		"claude-3-opus-20240229",
```
apascal07: What about the more recent models?

hugoaguirre: Yeah, I reused the ones we previously had in the codebase. Will update these.

Issue #4707, opened 2026-02-17
[JS] Transcription API returns 400 error


7hokerz:
**Describe the bug**
The STT (Speech-to-Text) model definition function incorrectly passes the full, namespaced model name (e.g., openai/whisper-1) directly to the OpenAI API.

**To Reproduce**
```ts
// STT sample
ai.defineFlow('transcribe', async () => {
  const audioFile = fs.readFileSync('jfk.flac');

  const { text } = await ai.generate({
    model: openAI.model('whisper-1'),
    prompt: [
      {
        media: {
          contentType: 'audio/flac',
          url: `data:audio/flac;base64,${audioFile.toString('base64')}`,
        },
      },
    ],
  });

  return text;
});
```

**Expected behavior**
Should return Transcription successfully.

**Screenshots**
- in Genkit UI
<img width="1035" height="511" alt="Image" src="https://github.com/user-attachments/assets/fd3e0702-7c26-4b09-afa7-b7c6129b1c21" />

**Runtime (please complete the following information):**
 - OS: windows 10 22H2

** Node version
 -  v24.13.0

**Additional context**
The TTS implementation correctly utilizes the toModelName utility to strip the openai/ prefix before interacting with the API. 
The STT implementation lacks this sanitization step.

PR #4708, opened 2026-02-17
fix(js/plugin/compat-oai): invalid model ID error in transcription models

7hokerz:
 - Applied the toModelName utility function to the STT model definition. (like TTS implementation)
 - Added an integration test flow for STT in `testapps/compat-oai/src/index.ts`.
 - Added `jfk.flac` as a test asset. (This audio file was sourced directly from the official Whisper GitHub repository without any modifications) [audio link](https://github.com/openai/whisper/tree/main/tests)
 - Implemented a test case to verify that the audio file is correctly transcribed using the `openai/whisper-1` model reference.
 - Fixed the issue: #4707 

Checklist (if applicable):


js/testapps/compat-oai/jfk.flac
ifielker, line 1: Instead of a political speech, let's go with a "hello world" audio file instead (since Genkit has nothing to do with politics and it can be sensitive and polarizing topic for some people). Something generated by a model is fine.

7hokerz: I changed the audio file! Could you please take a look? @ifielker

PR #4781, opened 2026-02-19
feat: Update trace filtering to support array-based eq/neq filters and numeric filters (gt, gte, lt, lte)

shrutip90:
While implementing filtering on dev UI, I was having issues doing neq filter for type along with the "Show only Genkit traces" since that needs a second neq filter. So I have modified the eq/neq filters to take arrays instead to check for multiple eq conditions that will be ORed and multiple neq conditions that will be ANDed.

Also, added numeric filters (gt, gte, lt, lte) that can be used to perform time range filtering on dev UI.

Related to https://github.com/firebase/genkit/issues/4439

Checklist (if applicable):
- [x] Tested (unit tested)

### DIFF
--- a/genkit-tools/telemetry-server/src/file-trace-store.ts
+++ b/genkit-tools/telemetry-server/src/file-trace-store.ts
@@ -417,10 +417,7 @@ export class Index {
             return undefined;
           }
         })
-        .filter((d) => {
-          if (!d) return false;
-          return isMatch(d, query?.filter);
-        })
+        .filter((d) => (!d ? false : isFilterMatch(d, query?.filter)))
         .reverse() as Record<string, string | number>[];
 
       fullData.push(...fileData);
@@ -456,7 +453,7 @@ export class Index {
   }
 }
 
-function isMatch(
+function isFilterMatch(
   d: Record<string, string | number>,
   filter?: TraceQueryFilter
 ): boolean {


genkit-tools/telemetry-server/src/file-trace-store.ts
```diff
@@ -420,21 +420,55 @@ export class Index {
...
+                return false;
+            }
+          }
+          if (query.filter.lte) {
+            for (const k of Object.keys(query.filter.lte)) {
+              const val = d[k];
+              if (typeof val !== 'number' || val > query.filter.lte[k])
+                return false;
+            }
           }
```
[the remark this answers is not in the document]
shrutip90: Done

genkit-tools/telemetry-server/src/file-trace-store.ts
```diff
@@ -420,21 +420,54 @@ export class Index {
...
           if (!d) return false;
           if (!query?.filter) return true;
-          if (
-            query.filter.eq &&
-            Object.keys(query.filter.eq).find(
-              (k) => d[k] !== query.filter!.eq![k]
-            )
-          ) {
-            return false;
+          const { eq, neq, gt, gte, lt, lte } = query.filter;
```
ssbushi: This can go into a helper method.

shrutip90: Done

genkit-tools/telemetry-server/src/file-trace-store.ts
```diff
@@ -419,24 +419,7 @@ export class Index {
...
-          if (
-            query.filter.neq &&
-            Object.keys(query.filter.neq).find(
-              (k) => d[k] === query.filter!.neq![k]
-            )
-          ) {
-            return false;
-          }
-          return true;
+          return isMatch(d, query?.filter);
```
ssbushi: could be simplified?

`.filter((d) => !d ? false :  isFilterMatch(d, query?.filter))`

genkit-tools/telemetry-server/src/file-trace-store.ts
```diff
@@ -473,6 +456,63 @@ export class Index {
   }
 }
 
+function isMatch(
```
ssbushi: nit: rename to `isFilterMatch`

genkit-tools/telemetry-server/src/file-trace-store.ts
shrutip90: Done, thanks.

genkit-tools/telemetry-server/src/file-trace-store.ts
shrutip90: Done.

PR #4785, opened 2026-02-20
chore(genkit-tools): update model and middleware types

MichaelDoyle:
- I'm "pulling forward" the middleware schema, so to speak
- Updated a couple small things missing in `js/ai/src/model.ts` and `js/ai/src/parts.ts` as well.


MichaelDoyle:
@huangjeff5 what do we need to do about the python errors? I don't think this should necessarily be causing an error. Either there is a problem in the translation step, or the rule needs to be relaxed?

genkit-tools/common/src/types/common.ts
```diff
@@ -0,0 +1,47 @@
...
+    configSchema?: Record<string, any> | null | undefined;
+    description?: string | undefined;
+  },
+  {
+    name: string;
+    metadata?: Record<string, any> | null | undefined;
+    configSchema?: Record<string, any> | null | undefined;
+    description?: string | undefined;
+  }
+>;
```
ssbushi: dumb q: why use z.ZodObject? Why not `z.object` ?

genkit-tools/common/src/types/common.ts
ssbushi, line 1: I thought everything under `common/` is common? Why do we need this separate file?

genkit-tools/common/src/types/model.ts
```diff
@@ -61,9 +57,38 @@ export {
...
 
 //
-// IMPORTANT: Keep this file in sync with genkit/ai/src/model.ts!
+// IMPORTANT: Keep this file in sync with genkit/ai/src/model-types.ts!
 //
 
+/**
+ * Zod schema of middleware reference.
+ */
+export const MiddlewareRefSchema: z.ZodTypeAny = z.object({
```
ssbushi:
```suggestion
export const MiddlewareRefSchema = z.object({
```

genkit-tools/common/src/types/model.ts
```diff
@@ -61,9 +57,38 @@ export {
...
+  config: z.any().optional(),
+});
+
+/**
+ * Middleware reference.
+ */
+export type MiddlewareRef = z.infer<typeof MiddlewareRefSchema>;
+
+/**
+ * Zod schema of an opration representing a model reference.
```
ssbushi, line 90:
```suggestion
 * Zod schema of an operation representing a model reference.
```

genkit-tools/common/src/types/common.ts
MichaelDoyle: ha - I wonder if this came from Gemini. This was "copied" from Pavel's branch, but it actually looks like:

```
/** Descriptor for a registered middleware, returned by reflection API. */
export const MiddlewareDescSchema = z.object({
  /** Unique name of the middleware. */
  name: z.string(),
  /** Human-readable description of what the middleware does. */
  description: z.string().optional(),
  /** JSON Schema for the middleware's configuration. */
  configSchema: z.record(z.any()).nullish(),
  /** User defined metadata for the middleware. */
  metadata: z.record(z.any()).nullish(),
});
export type MiddlewareDesc = z.infer<typeof MiddlewareDescSchema>;
```

Will clean up...

genkit-tools/common/src/types/model.ts
MichaelDoyle: This comes directly from:

https://github.com/firebase/genkit/compare/main...pj/tool_approval

Not sure. :)

huangjeff5:
> @huangjeff5 what do we need to do about the python errors? I don't think this should necessarily be causing an error. Either there is a problem in the translation step, or the rule needs to be relaxed?

Looking

genkit-tools/common/src/types/parts.ts
```diff
@@ -116,14 +120,14 @@ const ToolResponseSchemaBase = z.object({
  * Tool response part.
  */
 export type ToolResponse = z.infer<typeof ToolResponseSchemaBase> & {
-  content?: Part[];
+  content?: DocumentPart[];
```
pavelgj: why this change?

MichaelDoyle: There is a collision with multiple definitions of `Part` (`models.ts`, `parts.ts`) as both get exported, which causes an issue for the JSON schema generation.

Previously this was isolated to `documents.ts`, but after things got refactored out into `parts.ts` it became a broader issue.

ssbushi: This means we are downgrading from a generic part to only text/media parts, it this intentional?

genkit-tools/common/src/types/parts.ts
```diff
@@ -174,28 +181,17 @@ export const ResourcePartSchema = EmptyPartSchema.extend({
...
+// due to it clashing with similar schema in model.ts, and genkit-tools
+// exporting all types at root. We use a different name here and updated
+// coresponding the imports.
+export const DocumentPartSchema = z.union([TextPartSchema, MediaPartSchema]);
+export type DocumentPart = z.infer<typeof DocumentPartSchema>;
 
 export const MultipartToolResponseSchema = z.object({
   output: z.unknown().optional(),
-  content: z.array(PartSchema).optional(),
+  content: z.array(DocumentPartSchema).optional(),
```
ssbushi: Same here.

genkit-tools/common/src/types/parts.ts
MichaelDoyle: Not quite. Both Document (via `document.ts`) and Tools (via `parts.ts`) have `Part` as defined `TextPart  | MediaPart`. This was my/gemini's attempt to consolidate the two.

Probably it will be least confusing to just let each file redefine `Part` as it was doing before, but give them a different name in each place?

The issue we have is we do `export * filename` for every files types. So we have 3 definitions of Part. `document.ts`, `model.ts` and `parts.ts`, which will not work.

genkit-tools/common/src/types/parts.ts
MichaelDoyle: See above.

genkit-tools/common/src/types/parts.ts
MichaelDoyle: See also: https://github.com/firebase/genkit/blob/main/js/ai/src/parts.ts#L179-L185

genkit-tools/common/src/types/parts.ts
MichaelDoyle: See also: https://github.com/firebase/genkit/blob/main/js/ai/src/parts.ts#L179-L185

MichaelDoyle:
I broke this out into 3 PRs to make it easier to review and merge, since I touched things from 3 different areas.

PR #4786, opened 2026-02-20
feat(js/plugins/compat-oai): add translation adapter and add translation flow in testapps

7hokerz:
 - support for OpenAI's audio translation models 
 (Specially, requests utilizing `whisper-1-translate` are dynamically mapped to `whisper-1` before hitting the OpenAI API to distinguish between STT and translation workflows)
(I'm not sure if this is the best approach. I'm open to suggestions if you see any issues)
- Implemented the new feature by referencing existing logic and conventions to ensure code consistency

Checklist (if applicable):

### DIFF
--- a/js/plugins/compat-oai/tests/compat_oai_translate_test.ts
+++ b/js/plugins/compat-oai/tests/compat_oai_translate_test.ts
@@ -49,6 +49,7 @@ describe('toTranslationRequest', () => {
                 url: 'data:audio/wav;base64,aGVsbG8=',
               },
             },
+            { text: 'Translate this file' },
           ],
         },
       ],
@@ -59,7 +60,7 @@ describe('toTranslationRequest', () => {
     expect(actualOutput).toStrictEqual({
       model: 'whisper-1',
       file: expect.any(File),
-      prompt: '',
+      prompt: 'Translate this file',
       response_format: 'text',
     });
   });


ssbushi:
Thank you @7hokerz for adding this. I agree with the changes in the compat-oai plugin, but I'm unsure about the `whisper-1-translate` approach. 

On one hand, this is an "pseudo" model, which makes it confusing for the user to discover. Also, there are no more `*-translate` models currently offered by OpenAI, which makes the patterns implemented here forward looking and not relevant for today. 
On the other hand, I can see that this approach should work since we handle the model name override. 

One way to approach this to either:
 - Not handle the conversion based on the name, but have the user overload the config? (Not a clean approach IMO)
 - Have a new `translateModel` overload on the OpenAI plugin to specify this type of models (Not consistent with other plugins)
 - Have `model` somehow take an extra optional param to denote a specific type/flavor of a model on which to decide whether `whisper-1` is a transcription or translation model. 
 
 Curious what you think? @pavelgj @ifielker

7hokerz:
@ssbushi, I totally see your point, and I've actually been having the same concerns myself. It’s a tricky trade-off, and I agree it needs a better solution.

ssbushi:
> @ssbushi, I totally see your point, and I've actually been having the same concerns myself. It’s a tricky trade-off, and I agree it needs a better solution.

Yeah, I think the config based fix makes sense here, but for that to work we may need to do a bit of a refactor. Here's the plan:

1. Introduce a new "type" of model / config called `WhisperConfigSchema`, this can live in addition to the current transcription `model` helper. This will only cover the `whisper` model family (which I think only has one model in it?)
2. The `WhisperConfigSchema` can have a `translate?: z.boolean()` field which will denote whether the whisper model is to be used for translation or transcription (default). If set to true, we define it as a translation model using the compatOai helpers you wrote, otherwise, it should keep the current transcription behavior. 

This refactor means we will have to separate out the `whisper-*` parts of the model resovler / dynamic model definitions into this new logic, but it should not make a difference to the user. 

WDYT?

7hokerz:
@ssbushi, I'll give it a try based on your opinion.

xavidop (approved):

7hokerz:
@ssbushi, could you please take a look at this when you have a moment?

js/plugins/compat-oai/src/audio.ts
ssbushi, line 1: Sorry if I was unclear. Since `whisper` is specific to OpenAI, it does not make sense to provide defineWhisperModel and WhisperConfigSchema in this file. 

This file can remain as it was before (`define..TranslationModel` and `TranslationConfigSchema`). You can define defineWhisperModel and WhisperConfigSchema under `/openai` and that can leverage these compat functions under the hood.

ssbushi: Once you move this logic over there, please add tests to check that whisper-1 works with both translation and transcription endpoints. :)

ssbushi:
Thank you @7hokerz, added a quick comment.

7hokerz:
I've finished the refactoring as discussed and added the test codes for Whisper. @ssbushi, could you please take a look at this when you have a moment?

js/plugins/compat-oai/src/openai/whisper.ts
```diff
@@ -0,0 +1,147 @@
...
+    systemRole: false,
+    tools: false,
+  },
+};
+
+/**
+ * Config schema for Whisper models. Extends the transcription config with
+ * a `translate` flag that switches between transcription and translation APIs.
+ */
+export const WhisperConfigSchema = TranscriptionConfigSchema.extend({
```
ssbushi, line 49: Should this be a combination of TranscriptionConfigSchema AND TranslationConfigSchema?

We can use discriminated unions here... (off the `translate` field)
https://v3.zod.dev/?id=discriminated-unions

js/plugins/compat-oai/tests/compat_oai_translate_test.ts
```diff
@@ -0,0 +1,277 @@
...
+        },
+      ],
+      output: { format: 'text' },
+    } as GenerateRequest;
+
+    const actualOutput = toTranslationRequest('whisper-1', request);
+    expect(actualOutput).toStrictEqual({
+      model: 'whisper-1',
+      file: expect.any(File),
+      prompt: '',
```
ssbushi:
```suggestion
      prompt: 'Translate this file',
```

js/plugins/compat-oai/tests/compat_oai_translate_test.ts
```diff
@@ -0,0 +1,277 @@
...
+      messages: [
+        {
+          role: 'user',
+          content: [
+            {
+              media: {
+                contentType: 'audio/wav',
+                url: 'data:audio/wav;base64,aGVsbG8=',
+              },
+            },
```
ssbushi, line 51:
```suggestion
            },
            { text: 'Translate this file' }
```

Adding a simple prompt to check that the prompt is also passed along correctly. :)

ssbushi (approved):
Looks good. Approved pending changes. Thanks for taking the time to contribute @7hokerz !

js/plugins/compat-oai/src/openai/whisper.ts
ssbushi: If you try out discriminated unions, can you please verify it it works well with the Dev UI (Model Runner -> Select whisper-1 -> Check the Model config panel looks/works)

image

If it is unusable, let us just stick to simple union + extension (translate: boolean) for now. I will revisit this to see if it can be improved later. Thanks!

7hokerz: @ssbushi , I've tested discriminated unions.

```ts
const WhisperTranscriptionConfigSchema = TranscriptionConfigSchema.extend({
  translate: z.literal(false),
});

const WhisperTranslationConfigSchema = TranslationConfigSchema.extend({
  translate: z.literal(true),
});

/**
 * Config schema for Whisper models. Discriminated on `translate`:
 * - `translate: false` → Transcription API (supports language, chunking_strategy, etc.)
 * - `translate: true` → Translation API (supports response_format, temperature only)
 */
export const WhisperConfigSchema = z.discriminatedUnion('translate', [
  WhisperTranscriptionConfigSchema,
  WhisperTranslationConfigSchema,
]);
```

Unfortunately, the error occured in Genkit dev ui. 


And i applied your suggestions in the test code.

ssbushi: Makes sense. Let's leave it as it is for now. Thanks for verifying.

PR #4798, opened 2026-02-21
fix(go/plugins/googlegenai): fixed tool thought signature

apascal07:
Checklist (if applicable):


hugoaguirre (approved):

go/plugins/googlegenai/gemini.go
```diff
@@ -592,22 +592,36 @@ func toGeminiPart(p *ai.Part) (*genai.Part, error) {
...
 	}
 
 	return gp, nil
 }
+
+// metadataSignature extracts the thought signature from part metadata.
+// It handles both []byte (original value) and string (base64-encoded
+// after a JSON clone roundtrip).
+func metadataSignature(metadata map[string]any) []byte {
+	switch sig := metadata["signature"].(type) {
```
pavelgj, line 616: in js & dart we put it into `thoughtSignature`... but not sure if it matters. It's provider specific and we probably don't want/need to have any handling in the dev ui... so, probably ok...

apascal07: Won't change it now, it was already `signature`. I think `thoughtSignature` is used elsewhere but can't find it now... will review this in V2.

PR #4802, opened 2026-02-21
feat(js): add "fetch" plugin 

Ehesp:
While Genkit currently supports Express, the modern JavaScript ecosystem is rapidly consolidating around Web Standard APIs (Request, Response, Headers). This PR introduces a new plugin that allows Genkit flows to be exposed via the Fetch API standard, making Genkit natively compatible with virtually every modern JS runtime and framework.

By moving away from Node-specific HTTP abstractions, Genkit can now run seamlessly at the Edge, in Serverless environments, and in high-performance runtimes like Bun and Deno:

Runtimes:

- Bun
- Deno
- Cloudflare Workers (Workerd)
- Node (18+) via fetch
- Edge (Vercel, Netlify, Fastly)
- ... and probably more

Frameworks:

- Hono
- Elysia
- Astro
- Remix
- Nitro (+ frameworks built on-top of it)
- TanStack Start
- ... and probably more

 Usage (e.g. Hono):
 
```ts
const app = new Hono();

// Mount a single flow
app.all('/myFlow', (c) => handleFlow(c.req.raw, myFlow));

// Mount Genkit flows under /api - handleFlows routes by path
app.all('/api/*', (c) => handleFlows(c.req.raw, flows, '/api'));
```

This PR also includes a sample `js-hono`.

Checklist (if applicable):

TODO:

- [x] Not sure on the plugin name `web` - might be a bit too generic.

### DIFF
--- a/samples/package.json
+++ b/samples/package.json
@@ -11,7 +11,6 @@
     "build:js-prompts": "cd js-prompts && npm install && npm run build",
     "build:js-schoolAgent": "cd js-schoolAgent && npm install && npm run build",
     "build:js-gemini": "cd js-gemini && npm install && npm run build",
-    "build:js-hono": "cd js-hono && npm install && npm run build",
     "build:all-samples": "concurrently npm:build:js-*"
   },
   "pre-commit": [


js/plugins/web/README.md
```diff
@@ -0,0 +1,174 @@
+# Genkit Web Plugin
```
pavelgj: Overall this is pretty cool! I'm just not sure about the naming... just "web" might be misleading, like, first I saw this I thought it's meant to run in the browser. Perhaps `@genkit-ai/fetch`?

Ehesp: Yeah I agree, didn't feel right when I created it. Happy to go with fetch!

Ehesp:
Sample CI check failing due to the new package not being published (I think!)

js/plugins/fetch/src/index.ts
```diff
@@ -0,0 +1,523 @@
...
+  type z,
+} from 'genkit/beta';
+import {
+  getCallableJSON,
+  getHttpStatus,
+  type ContextProvider,
+  type RequestData,
+} from 'genkit/context';
+import { logger } from 'genkit/logging';
+import { getErrorMessage, getErrorStack } from './utils';
```
pavelgj:
```suggestion
import { getErrorMessage, getErrorStack } from './utils.js';
```

samples/js-hono/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "genkit:dev": "genkit start -- npx tsx --watch src/index.ts",
+    "build": "tsc",
+    "build:watch": "tsc --watch"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@genkit-ai/google-genai": "^1.17.0",
+    "@genkit-ai/fetch": "file:../../js/plugins/fetch",
```
pavelgj: this won't work... move this to js/testapps instead and use `workspace:*`

Ehesp: done!

PR #4812, opened 2026-02-23
chore(py): Simplify Dev UI integration and remove unused abstractions

huangjeff5:
## Problem

Previously, run_main() was running reflection server and running your code, both as sync tasks on the same event loop. Web frameworks typically own the event loop and provide their own hook. This means we (as genkit team) have to provide a custom implementation of that hook to support the framework. With FastAPI, we needed to provide a `genkit_lifespan()` hook. Genkit users also have to add that hook boilerplate into their code.

The reflection server now starts automatically in a background daemon thread when `GENKIT_ENV=dev` is set, right at `Genkit()` construction. The thread owns its own `asyncio` event loop so it never conflicts with FastAPI's, Flask's, or any other framework's loop. The registry is already protected by a `threading.RLock` so concurrent access from the background thread is safe.

In the case of a web app like FastAPI - Since both the Reflection Server and FastAPI have the ability to call a plugin from their own eventloop, there are model plugin bugs because the underlying model client errors out when executing from a different eventloop than the one that it was created in. Because of that, plugin authors need to create and cache a client on every running eventloop that executes the model client code. 

We provide a convenience method to make that easier. Here's how Plugin code can be executed on either loop. Plugin authors need to write something like this now:

`
from genkit.core._loop_local import _loop_local_client

class MyPlugin(Plugin):
  def __init__(self, **params): 
    self._client_params = params
    self._runtime_client = _loop_local_client(
        lambda: AsyncClient(**self._client_params)
    )
 ... (reuse self._runtime_client() later to access the SDK client
`

## Before:

**Scripts** needed `run_main()`:
```python
ai = Genkit(plugins=[GoogleAI()], model='googleai/gemini-2.0-flash')

@ai.flow()
async def my_flow(input: str) -> str:
    return (await ai.generate(prompt=input)).text

async def main():
    print(await my_flow('Hello'))

ai.run_main(main())  # starts reflection server + blocks for Dev UI
```

**FastAPI apps** needed an additional lifespan plugin — even for local development only:
```python
from genkit.plugins.fastapi import genkit_lifespan

ai = Genkit(...)
app = FastAPI(lifespan=genkit_lifespan(ai))  # only needed for Dev UI
```

## After

**FastAPI apps** — no lifespan needed:
```python
ai = Genkit(...)  # Dev UI starts here automatically
app = FastAPI()
```

## Future (Plugin Authors)

Re-export from loop-local client to genkit.plugin package (containing set of utilities and types commonly used by plugin authors)

```
from genkit.plugin import loop_local_client
```

**Scripts** — unchanged:
```python
ai = Genkit(...)

@ai.flow()
async def my_flow(input: str) -> str:
    return (await ai.generate(prompt=input)).text

async def main():
    print(await my_flow('Hello'))

ai.run_main(main())  # still needed to keep the process alive for Dev UI
```

## Changes

- `Genkit.__init__()`: starts the reflection server in a background daemon thread when `GENKIT_ENV=dev`
- `run_main()`: simplified — no longer starts the server itself, just runs the user coroutine then blocks on Ctrl+C to keep the daemon thread alive
- `genkit_lifespan()`: deprecated to a no-op with `DeprecationWarning` — existing code keeps working
- `web/manager/` directory removed — the internal abstraction that existed to run multiple ASGI servers on the same event loop, which was the previous approach to this problem
- `web-multi-server`, `web-endpoints-hello`, `web-short-n-long` samples removed
- Dead types removed: `ToolRequestLike`, `MatchableAction`, `ResourceFn` Protocol, `ASGIApp` Protocol
- Error wire format names simplified: `GenkitReflectionApiErrorWireFormat` → `ReflectionError`, `GenkitReflectionApiErrorWireFormat` => `RefectionErrorDetails`
-

### DIFF
--- a/py/packages/genkit/src/genkit/core/_loop_local.py
+++ b/py/packages/genkit/src/genkit/core/_loop_local.py
@@ -1,3 +1,19 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
 """Internal loop-local cache for async resources."""
 
 import asyncio


py/packages/genkit/src/genkit/blocks/resource.py
```diff
@@ -76,30 +76,9 @@ class ResourceOutput(BaseModel):
...
-# but we also want to support the strict Protocol for those who want it.
-# Note: Callable[..., T] is used for flexible args because accurate variable arg Union logic is complex/verbose.
-FlexibleResourceFn = ResourceFn | Callable[..., Awaitable[ResourcePayload] | ResourcePayload]
-
-
-class MatchableAction(Protocol):
-    """Protocol for actions that have a matches method."""
-
-    matches: Callable[[object], bool]
+FlexibleResourceFn = Callable[..., Awaitable[ResourcePayload] | ResourcePayload]
```
huangjeff5: Call this ResourceFn

py/packages/genkit/src/genkit/core/error.py
```diff
@@ -105,19 +105,19 @@
...
 
     model_config: ClassVar[ConfigDict] = ConfigDict(extra='allow', populate_by_name=True, alias_generator=to_camel)
 
     stack: str | None = None
     trace_id: str | None = None
 
 
-class GenkitReflectionApiErrorWireFormat(BaseModel):
-    """Wire format for HTTP errors."""
+class ErrorResponse(BaseModel):
```
huangjeff5: ReflectionErrorResponse maybe

huangjeff5: and ReflectionErrorDetails

PR #4818, opened 2026-02-24
chore(genkit-tools): update schema and regenerate types

MichaelDoyle:
I'm "pulling forward" the middleware schema, so to speak at the same time. I think that should be "OK". But let me know if not.

### DIFF
--- a/genkit-tools/common/src/types/model.ts
+++ b/genkit-tools/common/src/types/model.ts
@@ -20,6 +20,7 @@ import {
   DataPartSchema,
   MediaPartSchema,
   MultipartToolResponseSchema,
+  Part,
   PartSchema,
   ReasoningPartSchema,
   ResourcePartSchema,
@@ -30,7 +31,6 @@ import {
   type DataPart,
   type MediaPart,
   type MultipartToolResponse,
-  type Part,
   type ReasoningPart,
   type ResourcePart,
   type TextPart,
@@ -42,6 +42,7 @@ export {
   DataPartSchema,
   MediaPartSchema,
   MultipartToolResponseSchema,
+  Part,
   PartSchema,
   ReasoningPartSchema,
   ResourcePartSchema,
@@ -52,7 +53,6 @@ export {
   type DataPart,
   type MediaPart,
   type MultipartToolResponse,
-  type Part,
   type ReasoningPart,
   type ResourcePart,
   type TextPart,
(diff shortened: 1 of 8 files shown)


genkit-tools/genkit-schema.json
pavelgj, line 1: can you please regenerate this file...
I'm worried some of the changes are going to be breaking downstream to python, dart and go...

genkit-tools/common/src/types/model.ts
```diff
@@ -182,18 +239,65 @@ export const ToolDefinitionSchema = z.object({
...
+      .number()
+      .describe(GenerationCommonConfigDescriptions.topK)
+      .optional(),
+    topP: z
+      .number()
+      .describe(GenerationCommonConfigDescriptions.topP)
+      .optional(),
+    stopSequences: z
+      .array(z.string())
+      .max(5)
```
ssbushi: We should avoud these hard limitations in the common types and let the plugins override them if needed.

ssbushi:
LGTM from me

genkit-tools/common/src/types/model.ts
MichaelDoyle: This is getting really painful to keep in sync 😭

genkit-tools/common/src/types/model.ts
```diff
@@ -61,11 +57,53 @@ export {
...
+
+/**
+ * Middleware reference.
+ */
+export type MiddlewareRef = z.infer<typeof MiddlewareRefSchema>;
+
+/**
+ * Zod schema of an opration representing a model reference.
+ */
+export const ModelReferenceSchema = z.object({
```
apascal07, line 96: Curious, where is this used?

MichaelDoyle: Via reflection API to get the "default" model. We're going to use it here:

https://github.com/FirebasePrivate/genkit-ui/pull/1758

genkit-tools/common/src/types/model.ts
```diff
@@ -136,7 +178,7 @@ export const ModelInfoSchema = z.object({
       constrained: z.enum(['none', 'all', 'no-tools']).optional(),
       /** Model supports controlling tool choice, e.g. forced tool calling. */
       toolChoice: z.boolean().optional(),
-      /** Model supports long running operations. */
+      /** Model can perform long-running operations. */
       longRunning: z.boolean().optional(),
```
MichaelDoyle, line 182: I need to leave this until Python and Go can make adjustments.

PR #4924, opened 2026-03-14
feat(js/core): add overwrite option in definePrompt

7hokerz:
 - Added an overwrite flag to avoid displaying excessive log messages
 - Current Genkit implementation primarily focuses on static declarations during app initialization.
(In my personal backend development, I faced a problem with excessive logging during the dynamic prompt management process, especially when applying changes post-initialization.)

Checklist (if applicable):

### DIFF
--- a/js/ai/src/prompt.ts
+++ b/js/ai/src/prompt.ts
@@ -129,6 +129,7 @@ export interface PromptConfig<
   toolChoice?: ToolChoice;
   use?: ModelMiddleware[];
   context?: ActionContext;
+  overwrite?: boolean;
 }
 
 /**
@@ -238,7 +239,8 @@ export function definePrompt<
     registry,
     `${options.name}${options.variant ? `.${options.variant}` : ''}`,
     Promise.resolve(options),
-    options.metadata
+    options.metadata,
+    options.overwrite
   );
 }
 
@@ -250,7 +252,8 @@ function definePromptAsync<
   registry: Registry,
   name: string,
   optionsPromise: PromiseLike<PromptConfig<I, O, CustomOptions>>,
-  metadata?: Record<string, any>
+  metadata?: Record<string, any>,
+  overwrite?: boolean
 ): ExecutablePrompt<z.infer<I>, O, CustomOptions> {
   const promptCache = {} as PromptCache;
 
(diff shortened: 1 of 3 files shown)


MichaelDoyle:
@7hokerz can you speak more to the use case, and why you want/need to register the same prompt multiple times?

I wonder if we should just change the log level from `error` to `warn` or even `info`, and let you filter out the appropriate log level instead. cc @pavelgj

7hokerz:
@MichaelDoyle , Thanks for reply.
I am implementing a backend feature that allows prompts to be modified and applied in real-time without requiring a server restart in my company project. The execution flow is:
​A prompt is modified and saved to the database.
​definePrompt is immediately called to update the prompt definition in memory for the running application.

When definePrompt is executed to update an existing prompt, it generates an overwrite warning or log.
The overwrite logs are definitely helpful for preventing errors. However, I’ve noticed they can clutter the console when updates are intentional.

7hokerz:
@MichaelDoyle, @pavelgj, Could you please look over the comment I left?

MichaelDoyle:
@7hokerz if I understand the PR, we're adding an `overwrite` flag, however, it's only purpose is to flag whether or not a log is written; the registry behavior is unchanged?

If that's the case, I think this is a bit confusing and we're better off just adjusting the log level to `warn`, so you can more easily filter it out.

7hokerz:
@MichaelDoyle , Oh, I think your approach is simpler and more efficient than the current PR. However, my intention was to allow developers to explicitly determine whether the behavior is an error or not through the overwrite option.
I agree with changing it to 'warn,' but personally, I feel that if only this change is made, my intentions ultimately won't be incorporated into the bigger picture.

7hokerz:
@MichaelDoyle In that case, would it be best to actually add special logic regarding overwriting here?

ifielker:
It kind of sounds like you want an 'updatePrompt' would that be an accurate assumption?

7hokerz:
> It kind of sounds like you want an 'updatePrompt' would that be an accurate assumption?

@ifielker 
I think that method is good too.
However, if I add an updateprompt function, I'm not sure what logic to add here.
Regardless of the method, my goal was to reduce unnecessary log output and allow developers to explicitly select whether or not there are errors.

PR #4952, opened 2026-03-19
fix(py): Cache list actions endpoint and use correct embeddings model for google genai plugin

huangjeff5:
Noticed that Dev UI was slow because list actions endpoint takes ~3 seconds. Caching that endpoint in the plugin now. 

Also `text-embedding-004` is deprecated. Remove recommendations for that model.

### DIFF
--- a/py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
+++ b/py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
@@ -17,6 +17,7 @@
 """Google-Genai embedder model."""
 
 import sys
+import warnings
 from typing import Any, cast
 
 if sys.version_info < (3, 11):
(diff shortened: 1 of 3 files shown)


py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -94,10 +97,21 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+        # gemini-embedding-001 (3072 dims vs 768). Existing vectors are incompatible.
+        model = self._version
+        if str(model) == 'text-embedding-004':
+            warnings.warn(
+                'text-embedding-004 is deprecated (removed Jan 2026). '
+                'Using gemini-embedding-001 instead. Note: different dimensions (3072 vs 768).',
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            model = 'gemini-embedding-001'
```
huangjeff5, line 110: Don't rewire, just throw an error.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -65,7 +66,9 @@ class EmbeddingTaskType(StrEnum):
 
 def default_embedder_info(name: str) -> dict[str, Any]:
     """Returns default info for embedders given a name."""
-    return {'dimensions': 768, 'label': f'Google AI - {name}', 'supports': {'input': ['text']}}
+    # gemini-embedding-001 outputs 3072 dimensions; legacy models use 768
+    dimensions = 3072 if 'gemini-embedding' in name else 768
```
huangjeff5, line 70: Do a model mapping to make this more future proof

PR #4953, opened 2026-03-19
fix(py): Handle stream task cancellation without crashing callback

huangjeff5:
When the stream task is cancelled (e.g., client disconnects during streaming), the channel.closed done callback previously called result() on a cancelled future, which raised CancelledError and crashed the callback.

**Fix:** Add _propagate_closed_to_result that checks cancelled/exception/result before acting. When channel.closed is cancelled, cancel result_future instead of calling result(). Propagates cancellation cleanly to consumers.

### DIFF
--- a/py/packages/genkit/tests/genkit/core/action_test.py
+++ b/py/packages/genkit/tests/genkit/core/action_test.py
@@ -218,12 +218,9 @@ async def test_stream_cancellation_does_not_crash_callback() -> None:
     assert captured_task is not None
     captured_task.cancel()
 
-    # Give callbacks time to run
-    await asyncio.sleep(0.05)
-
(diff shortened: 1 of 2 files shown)


py/packages/genkit/tests/genkit/core/action_test.py
```diff
@@ -177,6 +179,53 @@ async def foo(
     assert chunks == ['1', '2']
 
 
+@pytest.mark.asyncio
+async def test_stream_cancellation_does_not_crash_callback() -> None:
+    """ACC-562: When stream task is cancelled, the channel.closed callback must not crash.
```
huangjeff5, line 184: Fix comment format

PR #4954, opened 2026-03-19
fix(py):  HttpErrorWireFormat JSON serialization in FastAPI/Flask

huangjeff5:
When validation errors or flow exceptions occur, FastAPI and Flask handlers called json.dumps(get_callable_json(ex)) directly. get_callable_json() returns HttpErrorWireFormat (Pydantic BaseModel), which json.dumps cannot serialize — TypeError was raised and clients received 500 instead of 422/400.

Fix is to make sure get_callable_json returns a dict that can be json.dump() immediately.

Also enabled the case where you want to expose a flow directly as an HTTP function.

```
@app.post('/chat', response_model=None)
@genkit_fastapi_handler(ai)
@ai.flow()
async def chat(prompt: str) -> str:
    response = await ai.generate(prompt=prompt)
    return response.text

```

Previously you had to write:
```
@app.post('/chat', response_model=None)
@genkit_fastapi_handler(ai)
async def chat():
    return my_flow

@ai.flow()
async def chat(prompt: str) -> str:
    response = await ai.generate(prompt=prompt)
    return response.text

```

### DIFF
--- a/py/plugins/fastapi/src/genkit/plugins/fastapi/handler.py
+++ b/py/plugins/fastapi/src/genkit/plugins/fastapi/handler.py
@@ -70,6 +70,7 @@ def genkit_fastapi_handler(
         async def chat():
             return my_flow
 
+
         @ai.flow()
         async def my_flow(prompt: str) -> str: ...
         ```
--- a/py/README.md
+++ b/py/README.md
@@ -205,6 +205,7 @@ class MyPlugin(Plugin):
             ActionMetadata(kind=ActionKind.MODEL, name='myplugin/my-model'),
         ]
 
+
 # Usage:
 ai = Genkit(plugins=[MyPlugin(api_key='...')])
 response = await ai.generate(model='myplugin/my-model', prompt='Hello!')
@@ -357,7 +358,7 @@ ai = Genkit(
     model=gemini_2_0_flash,
 )
 
-response = await ai.generate(prompt="Tell me a joke")
+response = await ai.generate(prompt='Tell me a joke')
 print(response.text)
 ```
 
--- a/py/packages/genkit/README.md
+++ b/py/packages/genkit/README.md
@@ -26,12 +26,14 @@ ai = Genkit(
     model='googleai/gemini-2.0-flash',
 )
 
+
 class RpgCharacter(BaseModel):
     """An RPG game character."""
 
     name: str = Field(description='name of the character')
     back_story: str = Field(description='back story')
     abilities: list[str] = Field(description='list of abilities (3-4)')
+
 
 @ai.flow()
 async def generate_character(name: str) -> RpgCharacter:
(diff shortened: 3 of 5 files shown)


py/plugins/fastapi/src/genkit/plugins/fastapi/handler.py
```diff
@@ -144,7 +144,7 @@ async def event_stream() -> AsyncIterator[str]:
                     ex = e.cause if isinstance(e, GenkitError) else e
                     return Response(
                         status_code=500,
-                        content=json.dumps(get_callable_json(ex), separators=(',', ':')),
+                        content=json.dumps(_to_dict(get_callable_json(ex)), separators=(',', ':')),
```
huangjeff5: Instead of adding _to_dict_ here
get_callable_json should just return a dict

py/plugins/fastapi/tests/fastapi_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+from fastapi.testclient import TestClient
+
+from genkit import ActionRunContext, Genkit
+from genkit.plugins.fastapi import genkit_fastapi_handler
+
+
+def create_app():
+    """Create a FastAPI application for testing."""
+    from fastapi import FastAPI
```
huangjeff5: move import to top

py/plugins/fastapi/src/genkit/plugins/fastapi/handler.py
```diff
@@ -133,7 +133,7 @@ async def event_stream() -> AsyncIterator[str]:
                         yield f'data: {json.dumps({"result": _to_dict(result)}, separators=(",", ":"))}\n\n'
                     except Exception as e:
                         ex = e.cause if isinstance(e, GenkitError) else e
-                        yield f'error: {json.dumps({"error": _to_dict(get_callable_json(ex))}, separators=(",", ":"))}'
+                        yield f'error: {json.dumps({"error": get_callable_json(ex)}, separators=(",", ":"))}'
```
huangjeff5: why do we have these formatters?

py/packages/genkit/tests/genkit/core/error_test.py
```diff
@@ -77,15 +76,17 @@ def test_get_http_status() -> None:
 def test_get_callable_json() -> None:
     genkit_error = GenkitError(status='DATA_LOSS', message='Oops')
```
huangjeff5: Is this a real status?

py/plugins/fastapi/src/genkit/plugins/fastapi/handler.py
```diff
@@ -50,42 +53,40 @@ def genkit_fastapi_handler(
...
+        ```
 
-
-        @app.post('/chat')
+    Example (wrapper when flow is defined later):
+        ```python
+        @app.post('/chat', response_model=None)
         @genkit_fastapi_handler(ai)
-        async def chat():
+        def chat():
```
huangjeff5: why not async?

PR #4956, opened 2026-03-19
fix(py): Cache list actions endpoint and use correct embeddings model for google genai plugin

huangjeff5:
1. Cache list_actions() so that Dev UI refresh loads quickly.
2. Update supported embedders to match current state of the world.

### DIFF
--- a/py/plugins/google-genai/test/models/googlegenai_embedder_test.py
+++ b/py/plugins/google-genai/test/models/googlegenai_embedder_test.py
@@ -32,7 +32,7 @@ from genkit.plugins.google_genai.models.embedder import (
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize('version', list(GeminiEmbeddingModels))
+@pytest.mark.parametrize('version', [x for x in GeminiEmbeddingModels])
 async def test_embedding(mocker: MockerFixture, version: GeminiEmbeddingModels) -> None:
     """Test the embedding method."""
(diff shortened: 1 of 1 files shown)


py/plugins/google-genai/src/genkit/plugins/google_genai/google.py
```diff
@@ -279,14 +283,13 @@ def _create_embedder_action(
...
+    label = f"{PLUGIN_DISPLAY_NAME[plugin_name]} - {clean_name}"
     action_metadata = embedder_action_metadata(
         name=name,
         options=EmbedderOptions(
-            label=embedder_info.get('label'),
-            supports=EmbedderSupports(input=embedder_info.get('supports', {}).get('input')),
-            dimensions=embedder_info.get('dimensions'),
+            label=label,
+            supports=EmbedderSupports(input=['text']),
+            dimensions=EMBEDDER_DIMENSIONS.get(clean_name),
```
pavelgj, line 292: Do we need this in the metadata? Embedders have default dimensions and would be great to not have to maintain them...

huangjeff5: Have to check but I think it is used to render Dev UI. I think what makes sense is calling the list models endpoint to get the available embedders and the expected dimensions, so we don't have to maintain it.

I'll track this as a followup.

PR #4957, opened 2026-03-19
chore(py): add back evaluator sample

huangjeff5:


py/samples/evaluators/src/main.py
```diff
@@ -0,0 +1,96 @@
...
+        test_case_id=datapoint.test_case_id or '',
+        evaluation=Score(score=found, details=Details(reasoning=f'URL found: {found}')),
+    )
+
+
+ai.define_evaluator(
+    name='byo/url',
+    display_name='URL Match',
+    definition='True if output contains a URL.',
+    is_billed=False,
```
huangjeff5: is_billed=true/false? why?

py/plugins/evaluators/src/genkit/plugins/evaluators/plugin.py
```diff
@@ -0,0 +1,231 @@
...
+# EvaluatorFn: (datapoint, options) -> EvalFnResponse
+EvaluatorFn = Callable[[BaseDataPoint, object | None], Coroutine[Any, Any, EvalFnResponse]]
+
+
+def _make_eval_stepper(metric_fn: EvaluatorFn) -> Callable[[EvalRequest], Coroutine[Any, Any, EvalResponse]]:
+    """Wrap a per-datapoint metric fn into an EvalRequest -> EvalResponse stepper."""
+
+    async def _stepper(req: EvalRequest) -> EvalResponse:
+        responses: list[EvalFnResponse] = []
+        for datapoint in req.dataset:
```
ssbushi: Does python have a `runInNewSpan` equivalent? I think each of these samples are wrapped in their own spans to dive deeper... or is that already part of the core SDK... ?

huangjeff5: Yeah there is. I was able to access that code path by refactoring the plugin to use to use ai.define_evaluator correctly and removed the custom code. I decided to change this code from a Plugin to a regular package with helpers, because there's no way to pass Genkit instance / Registry to the plugins in the new PluginV2 interface.

I see evaluators are still on the old interface, so not sure how we plan to address it there.

Here's the usage now:

```
ai = Genkit(plugins=[GoogleAI()])
register_genkit_evaluators(ai)
```

And then the plugin uses `ai.define_evaluator(...)` which hooks into the existing `runInNewSpan` wrapping logic.

py/samples/evaluators/src/main.py
```diff
@@ -0,0 +1,130 @@
...
+# Setup
+prompts_path = Path(__file__).resolve().parent.parent / 'prompts'
+ai = Genkit(
+    plugins=[GoogleAI()],
+    model='googleai/gemini-2.5-flash',
+    prompt_dir=prompts_path,
+)
+register_genkit_evaluators(ai)
+
+JUDGE_MODEL = os.getenv('JUDGE_MODEL', 'googleai/gemini-2.5-flash')
```
ssbushi: Prefer to use pro here.

Issue #4970, opened 2026-03-20
[Go] Gemini config IncludeServerSideToolInvocations not supported


Gerifield:
**Describe the bug**
There is a new tool config for Gemini called `includeServerSideToolInvocations` which allows to use the built-in tooling (like search) in combination with a defined tool list.
Right now if there is more than 0 tools set, the tool config will be fully overwritten.

**To Reproduce**
Setup the search tool to use, add the `IncludeServerSideToolInvocations` as true + define any tools, the config will be removed and the Gemini api will return an error:
```
Error 400, Message: Please enable tool_config.include_server_side_tool_invocations to use Built-in tools with Function calling., Status: INVALID_ARGUMENT
```

**Expected behavior**
It should simply forward this config and work without an issue.

**Screenshots**
- N/A

**Runtime (please complete the following information):**
 - OS: Linux, MacOS
 - Version Ubuntu 25.10, MacOS 26.3.1

**Go version**
 -  go version go1.26.1 darwin/arm64

**Additional context**
- N/A

Gerifield:
#4971 fixed it, thank you!

PR #4971, opened 2026-03-20
fix(go/plugins/googlegenai): add support to IncludeServerSideToolInvocations

Gerifield:
ISSUE: #4970

CHANGELOG:
  - [ ] Update genai toGeminiToolChoice function to not overwrite the tool config if it exists

Checklist (if applicable):


Gerifield:
Here is an example config how to test/trigger it:
```
ist := true
geminiSpecConfig := &genai.GenerateContentConfig{
		ToolConfig: &genai.ToolConfig{
			IncludeServerSideToolInvocations: &ist,
		},
		Tools: []*genai.Tool{
			{GoogleSearch: &genai.GoogleSearch{}},
		}
	}
```

Then it can be added to the Generate option via: `ai.WithConfig(geminiSpecConfig)`
If there are tools added with the `ai.WithTools` command, it should still work and do it's job.

PR #4976, opened 2026-03-23
feat(py): clean up public interrupt/respond api surface

huangjeff5:


py/packages/genkit/src/genkit/_ai/_prompt.py
```diff
@@ -113,6 +115,47 @@ class PromptGenerateOptions(TypedDict, total=False):
     metadata: dict[str, Any] | None
 
 
+def _prompt_opts_from_kwargs(
```
huangjeff5: Look into how we can simplify this opts => kwargs translation layer

huangjeff5: Right now every time we add a new field we have to map it correctly in multiple places

py/packages/genkit/tests/test_tool_interrupts.py
```diff
@@ -0,0 +1,405 @@
+# Copyright 2025 Google LLC
```
huangjeff5, line 1: Rewrite the tests to be more focused on the end user experience.

PR #4981, opened 2026-03-23
feat(js): Support Otel logging

shrutip90:
Related to https://github.com/genkit-ai/genkit/issues/4947

Sample logs generated: https://paste.googleplex.com/6529212040216576

Checklist (if applicable):
- [x] Tested (manually, unit tested)

### DIFF
--- a/genkit-tools/common/src/types/log.ts
+++ b/genkit-tools/common/src/types/log.ts
@@ -1,0 +1,66 @@
+/**
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { extendZodWithOpenApi } from '@asteasolutions/zod-to-openapi';
+import * as z from 'zod';
+
+import { InstrumentationLibrarySchema } from './trace';
+
+extendZodWithOpenApi(z);
+
+/**
+ * Zod schema for log record data.
+ */
+export const LogRecordSchema = z
+  .object({
+    logId: z.string(), // Server-generated if omitted
+    traceId: z.string().optional(),
+    spanId: z.string().optional(),
+    timestamp: z
+      .number()
+      .describe('log recorded time in milliseconds since the epoch'),
+    severityNumber: z.number().optional(),
+    severityText: z.string().optional(),
+    body: z.unknown().optional(), // Represents any value: string, number, boolean, array, or object
+    attributes: z.record(z.string(), z.unknown()).optional(),
+    instrumentationLibrary: InstrumentationLibrarySchema.optional(),
+  })
+  .openapi('LogRecordData');
+
+export type LogRecordData = z.infer<typeof LogRecordSchema>;
+
+/**
+ * Log query parameters.
+ */
+export const LogQuerySchema = z.object({
+  limit: z.coerce.number().optional(),
+  continuationToken: z.string().optional(),
+  traceId: z.string().optional(),
+  spanId: z.string().optional(),
+});
+
+export type LogQuery = z.infer<typeof LogQuerySchema>;
+
+export interface LogQueryResponse {
+  logs: LogRecordData[];
+  continuationToken?: string;
+}
+
+export interface LogStore {
+  init(): Promise<void>;
+  save(logs: LogRecordData[]): Promise<void>;
+  list(query?: LogQuery): Promise<LogQueryResponse>;
+}
--- a/genkit-tools/telemetry-server/src/file-log-store.ts
+++ b/genkit-tools/telemetry-server/src/file-log-store.ts
@@ -1,0 +1,263 @@
+/**
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  LogRecordData,
+  type LogQuery,
+  type LogQueryResponse,
+  type LogStore,
+} from '@genkit-ai/tools-common';
+import { logger } from '@genkit-ai/tools-common/utils';
+import { Mutex } from 'async-mutex';
+import fs from 'fs';
+import lockfile from 'lockfile';
+import path from 'path';
+
+function lockFile(file: string) {
+  return `${file}.lock`;
+}
+
+export class LocalFileLogStore implements LogStore {
+  private readonly storeRoot: string;
+  private readonly indexRoot: string;
+  private readonly index: LogIndex;
+  private mutex = new Mutex();
+
+  constructor(options: { storeRoot: string; indexRoot: string }) {
+    this.storeRoot = path.resolve(options.storeRoot, '.genkit/logs');
+    fs.mkdirSync(this.storeRoot, { recursive: true });
+    this.indexRoot = path.resolve(options.indexRoot, '.genkit/logs_idx');
+    fs.mkdirSync(this.indexRoot, { recursive: true });
+    logger.debug(
+      `[Telemetry Server] initialized local file log store at root: ${this.storeRoot}`
+    );
+    this.index = new LogIndex(this.indexRoot);
+  }
+
+  async init(): Promise<void> {
+    // Indexes and store are append-only.
+  }
+
+  private getCurrentLogFile(): string {
+    const now = new Date();
+    // format YYYY-MM-DD-HH
+    const pad = (n: number) => n.toString().padStart(2, '0');
+    const dateStr = `${now.getFullYear()}-${pad(now.getMonth() + 1)}-${pad(now.getDate())}-${pad(now.getHours())}`;
+    return path.resolve(this.storeRoot, `logs-${dateStr}.jsonl`);
+  }
+
+  async save(logs: LogRecordData[]): Promise<void> {
+    if (logs.length === 0) return;
+
+    await this.mutex.runExclusive(() => {
+      const logFile = this.getCurrentLogFile();
+      const relativeFileName = path.basename(logFile);
+
+      let currentOffset = 0;
+      if (fs.existsSync(logFile)) {
+        currentOffset = fs.statSync(logFile).size;
+      }
+
+      const indexEntries: LogIndexEntry[] = [];
+      let offsetTracker = currentOffset;
+
+      const linesToAppend = logs.map((log) => {
+        // Ensure log id exists if not provided
+        if (!log.logId && log.traceId && log.spanId) {
+          log.logId = `${log.traceId}-${log.spanId}-${log.timestamp}-${Math.random().toString(36).substring(7)}`;
+        } else if (!log.logId) {
+          log.logId = `${log.timestamp}-${Math.random().toString(36).substring(7)}`;
+        }
+        const line = JSON.stringify(log) + '\n';
+        const length = Buffer.byteLength(line, 'utf8');
+
+        indexEntries.push({
+          traceId: log.traceId,
+          spanId: log.spanId,
+          timestamp: log.timestamp,
+          severityText: log.severityText,
+          severityNumber: log.severityNumber,
+          file: relativeFileName,
+          offset: offsetTracker,
+          length: length,
+        });
+
+        offsetTracker += length;
+        return line;
+      });
+
+      fs.appendFileSync(logFile, linesToAppend.join(''));
+      this.index.add(indexEntries);
+    });
+  }
+
+  async list(query?: LogQuery): Promise<LogQueryResponse> {
+    const startFromIndex = query?.continuationToken
+      ? Number.parseInt(query.continuationToken)
+      : 0;
+    const limit = query?.limit ?? 100;
+
+    const searchResult = this.index.search({
+      limit,
+      startFromIndex,
+      traceId: query?.traceId,
+      spanId: query?.spanId,
+    });
+
+    const logs: LogRecordData[] = [];
+
+    for (const entry of searchResult.entries) {
+      const logFile = path.resolve(this.storeRoot, entry.file);
+      if (fs.existsSync(logFile)) {
+        const buffer = Buffer.alloc(entry.length);
+        const fd = fs.openSync(logFile, 'r');
+        fs.readSync(fd, buffer, 0, entry.length, entry.offset);
+        fs.closeSync(fd);
+        try {
+          logs.push(
+            JSON.parse(buffer.toString('utf8').trim()) as LogRecordData
+          );
+        } catch (e) {
+          logger.error(`Error parsing log at ${entry.file}:${entry.offset}`);
+        }
+      }
+    }
+
+    return {
+      logs,
+      continuationToken: searchResult.pageLastIndex
+        ? `${searchResult.pageLastIndex}`
+        : undefined,
+    };
+  }
+}
+
+export interface LogIndexEntry {
+  traceId?: string;
+  spanId?: string;
+  timestamp: number;
+  severityText?: string;
+  severityNumber?: number;
+  file: string;
+  offset: number;
+  length: number;
+}
+
+export interface LogIndexSearchResult {
+  pageLastIndex?: number;
+  entries: LogIndexEntry[];
+}
+
+export class LogIndex {
+  private currentIndexFile: string;
+
+  constructor(private indexRoot: string) {
+    this.currentIndexFile = path.resolve(
+      this.indexRoot,
+      this.newIndexFileName()
+    );
+    fs.mkdirSync(this.indexRoot, { recursive: true });
+  }
+
+  private newIndexFileName() {
+    return `idx_${(Date.now() + '').padStart(17, '0')}.jsonl`;
+  }
+
+  listIndexFiles() {
+    return fs.readdirSync(this.indexRoot).filter((f) => f.startsWith('idx_'));
+  }
+
+  add(entries: LogIndexEntry[]) {
+    if (entries.length === 0) return;
+    try {
+      lockfile.lockSync(lockFile(this.currentIndexFile));
+      const lines = entries.map((e) => JSON.stringify(e) + '\n').join('');
+      fs.appendFileSync(this.currentIndexFile, lines);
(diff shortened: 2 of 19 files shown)


genkit-tools/common/src/types/log.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export const LogRecordSchema = z
+  .object({
+    logId: z.string(), // Server-generated if omitted
+    traceId: z.string().optional(),
+    spanId: z.string().optional(),
+    timestamp: z.number().describe('log recorded time in milliseconds since the epoch'),
+    severityNumber: z.number().optional(),
+    severityText: z.string().optional(),
+    body: z.any().optional(), // Represents any value: string, number, boolean, array, or object
```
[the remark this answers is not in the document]
shrutip90: Done

genkit-tools/telemetry-server/src/file-log-store.ts
```diff
@@ -0,0 +1,238 @@
...
+    const result = {
+      entries: fullData.slice(query.startFromIndex, query.startFromIndex + query.limit),
+    } as LogIndexSearchResult;
+
+    if (query.startFromIndex + query.limit < fullData.length) {
+      result.pageLastIndex = query.startFromIndex + query.limit;
+    }
+
+    return result;
+  }
```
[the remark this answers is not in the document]
shrutip90: Done

genkit-tools/telemetry-server/src/index.ts
```diff
@@ -179,13 +184,77 @@ export async function startTelemetryServer(params: {
...
+          continuationToken: continuationToken
+            ? continuationToken.toString()
+            : undefined,
+          spanId,
+        })
+      );
+    } catch (e) {
+      next(e);
+    }
+  });
```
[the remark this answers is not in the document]
shrutip90: Ack

js/core/src/logging.ts
```diff
@@ -74,10 +121,40 @@ class Logger {
...
+
+      logs.getLogger('genkit-logger').emit({
+        severityNumber: SeverityNumber.ERROR,
+        severityText: 'ERROR',
+        body: msg,
+        attributes: metadata,
+        ...(activeContext ? { context: activeContext } : {})
+      });
+    } catch (e) {}
   }
```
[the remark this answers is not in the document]
shrutip90: Done

genkit-tools/telemetry-server/src/utils/otlp.ts
```diff
@@ -157,6 +185,63 @@ export function traceDataFromOtlp(otlpData: OtlpPayload): TraceData[] {
...
+  let body: any = undefined;
+  if (log.body) {
+    if (log.body.stringValue !== undefined) {
+      body = log.body.stringValue;
+    } else if (log.body.intValue !== undefined) {
+      body = log.body.intValue;
+    } else if (log.body.boolValue !== undefined) {
+      body = log.body.boolValue;
+    }
+  }
```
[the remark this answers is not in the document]
shrutip90: Done

js/core/src/tracing/exporter.ts
```diff
@@ -178,3 +179,97 @@ export class TraceServerExporter implements SpanExporter {
...
+            },
+            logRecords: [],
+          });
+        }
+
+        const attributes: any[] = [];
+        for (const [k, v] of Object.entries(log.attributes)) {
+          if (typeof v === 'string') attributes.push({ key: k, value: { stringValue: v } });
+          else if (typeof v === 'number') attributes.push({ key: k, value: { intValue: v } });
+          else if (typeof v === 'boolean') attributes.push({ key: k, value: { boolValue: v } });
```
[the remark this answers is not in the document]
shrutip90: Done

genkit-tools/common/src/types/log.ts
```diff
@@ -0,0 +1,66 @@
...
+    timestamp: z
+      .number()
+      .describe('log recorded time in milliseconds since the epoch'),
+    severityNumber: z.number().optional(),
+    severityText: z.string().optional(),
+    body: z.unknown().optional(), // Represents any value: string, number, boolean, array, or object
+    attributes: z.record(z.string(), z.unknown()).optional(),
+    instrumentationLibrary: InstrumentationLibrarySchema.optional(),
+  })
+  .openapi('LogRecordData');
```
MichaelDoyle, line 41: What is the source for `LogRecord`?

In particular `eventName` is missing, and I think in OTel 2.0 at least `instrumentationLibrary` has been supplanted by `instrumentationScope`.

That said, the Otel docs and various libraries also seem to be a bit out of sync with each other. :-/

See also: 

- https://opentelemetry.io/docs/specs/otel/logs/data-model/#log-and-event-record-definition
- [opentelemetry/proto/logs/v1/logs.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto)
- https://github.com/open-telemetry/opentelemetry-js/blob/b84271749257eed2eb08439288af115467ed350e/experimental/packages/api-logs/src/types/LogRecord.ts

If we can align on the current format from OTel, and do any translation internally (if needed), I think that's preferable.

Alternatively if this is low effort to change (assuming the log store is pretty "ephemeral") that is OK too.

genkit-tools/telemetry-server/src/utils/otlp.ts
```diff
@@ -70,8 +70,44 @@ interface OtlpResourceSpan {
...
-interface OtlpPayload {
-  resourceSpans: OtlpResourceSpan[];
+interface OtlpLogRecord {
+  timeUnixNano: string;
+  severityNumber?: number;
+  severityText?: string;
+  body?: OtlpValue;
+  attributes?: OtlpAttribute[];
+  traceId?: string;
+  spanId?: string;
```
MichaelDoyle, line 80: This doesn't seem to the fully aligned with the proto.

https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto

genkit-tools/telemetry-server/src/index.ts
```diff
@@ -179,13 +184,68 @@ export async function startTelemetryServer(params: {
...
+            : undefined,
+          traceId,
+        })
+      );
+    } catch (e) {
+      next(e);
+    }
+  });
+
+  api.get('/api/spans/:spanId/logs', async (request, response, next) => {
```
MichaelDoyle: is `spanId` globally unique?

Why not `/api/traces/:traceId/spans/:spanId/logs`?

genkit-tools/telemetry-server/src/index.ts
```diff
@@ -216,7 +290,10 @@ export async function startTelemetryServer(params: {
 
   api.post('/api/otlp', async (request, response) => {
```
MichaelDoyle, line 295: I am wondering if we should support `api/v1/logs` as a path as well. That's the default path for OTel collectors per the spec.

https://opentelemetry.io/docs/specs/otlp/#otlphttp-request

Over time we should probably add:

- `v1/traces`
- `v1/logs`
- `v1/metrics`

cc @pavelgj

js/core/src/tracing/exporter.ts
```diff
@@ -178,3 +182,103 @@ export class TraceServerExporter implements SpanExporter {
 function transformTime(time: HrTime) {
   return hrTimeToMilliseconds(time);
 }
+
+/**
+ * Exports collected OpenTelemetetry logs to the telemetry server.
+ */
+export class LogServerExporter implements LogRecordExporter {
```
MichaelDoyle, line 189: If we're using the OTLP endpoint, do we need our own exporter? Or is there an out of the box one that would work?

MichaelDoyle: e.g. https://www.npmjs.com/package/@opentelemetry/exporter-logs-otlp-http

js/core/src/logging.ts
```diff
@@ -74,10 +138,12 @@ class Logger {
 
   logStructured(msg: string, metadata: any) {
     getLogger().info(msg, metadata);
+    this._emitOtel('info', [], msg, metadata);
   }
 
   logStructuredError(msg: string, metadata: any) {
     getLogger().error(msg, metadata);
```
MichaelDoyle, line 80: Maybe we do it as part of "telemetry v2", but we probably need to revisit these. I think the `metadata` field is a bit of a carry-over from the gcp plugin (which uses a specific exporter (`winston`?) that is built around the idea of having a metadata field. I don't think this code does it, but we definitely don't want to "wrap" everything under the key metadata (which happens in GCP).

js/core/src/tracing/node-telemetry-provider.ts
```diff
@@ -66,6 +75,17 @@ async function enableTelemetry(
...
+
+  // Add LogRecordProcessors
+  const enableRealTimeTelemetry =
+    process.env.GENKIT_ENABLE_REALTIME_TELEMETRY === 'true';
+  const logExporter = new LogServerExporter();
+  const logProcessor: LogRecordProcessor =
+    isDevEnv() || enableRealTimeTelemetry
+      ? new SimpleLogRecordProcessor(logExporter)
+      : new BatchLogRecordProcessor(logExporter);
+  nodeOtelConfig.logRecordProcessor = logProcessor;
```
MichaelDoyle: Should we gate this around some kind of feature flag?

MichaelDoyle (commented):
Overall looking pretty good. Made a few comments we can all discuss.

genkit-tools/common/src/types/log.ts
MichaelDoyle: eventName in particular is important, because it allows us to know if a given log adheres to an otel semantic convention (i.e. structured log) or not.

js/core/src/tracing/node-telemetry-provider.ts
pavelgj: +1

shrutip90: Discussed offline with Pavel, and he asked to put this behind a flag too.

genkit-tools/telemetry-server/src/index.ts
shrutip90: spanId is not globally unique apparently. Will update this.

genkit-tools/telemetry-server/src/index.ts
pavelgj: I'm not sure about these conventions. If you could figure this out it'd be great. :)

js/core/src/tracing/exporter.ts
pavelgj: js runtime is not exporting via OTLP just yet... only Dart does. I'm not sure what we should do here... don't want to force the move to OTLP in this PR.

MichaelDoyle: agree

js/core/src/tracing/node-telemetry-provider.ts
shrutip90: Added a flag to gate this.

genkit-tools/telemetry-server/src/index.ts
shrutip90: Updated the API

js/core/src/logging.ts
```diff
@@ -47,25 +49,87 @@ function getLogger() {
 class Logger {
   readonly defaultLogger = _defaultLogger;
 
+  private _emitOtel(
```
pavelgj, line 53: should this be an no op when GENKIT_ENABLE_LOGS is not set?

pavelgj (approved):

js/core/src/tracing/node-telemetry-provider.ts
```diff
@@ -66,6 +75,19 @@ async function enableTelemetry(
     delete nodeOtelConfig.spanProcessor;
   }
   nodeOtelConfig.spanProcessors = processors;
+
+  // Add LogRecordProcessors
+  if (process.env.GENKIT_ENABLE_LOGS === 'true') {
```
MichaelDoyle:
```suggestion
  if (process.env.GENKIT_OTEL_ENABLE_LOGS === 'true') {
```

Or similar

genkit-tools/telemetry-server/src/file-log-store.ts
```diff
@@ -0,0 +1,262 @@
...
+    } catch (err) {
+      logger.error(
+        `Failed to lock log index file ${this.currentIndexFile}: ${err}`
+      );
+    } finally {
+      if (fs.existsSync(lockFile(this.currentIndexFile))) {
+        lockfile.unlockSync(lockFile(this.currentIndexFile));
+      }
+    }
+  }
```
MichaelDoyle: We might want to add some basic retry logic on the locking.

e.g. `lockfile.lockSync(lockFile(this.currentIndexFile), { retries: 10, retryWait: 100 });`

```suggestion
  add(entries: LogIndexEntry[]) {
    if (entries.length === 0) return;
    let lockAcquired = false;
    try {
      lockfile.lockSync(lockFile(this.currentIndexFile), { retries: 10, retryWait: 100 });
      lockAcquired = true;
      const lines = entries.map((e) => JSON.stringify(e) + '\n').join('');
      fs.appendFileSync(this.currentIndexFile, lines);
    } catch (err) {
      logger.error(
        `Failed to lock log index file ${this.currentIndexFile}: ${err}`
      );
      // Consider a retry mechanism or a persistent queue to avoid data loss.
    } finally {
      if (lockAcquired) {
        lockfile.unlockSync(lockFile(this.currentIndexFile));
      }
    }
  }
```

genkit-tools/telemetry-server/src/file-log-store.ts
```diff
@@ -0,0 +1,262 @@
...
+        fs.closeSync(fd);
+        try {
+          logs.push(
+            JSON.parse(buffer.toString('utf8').trim()) as LogRecordData
+          );
+        } catch (e) {
+          logger.error(`Error parsing log at ${entry.file}:${entry.offset}`);
+        }
+      }
+    }
```
MichaelDoyle: It seems like this re-opens the log file for every record, but its likely the logs will be in the same file(s). We can re-use the open file handles instead.

```ts
const fileHandles: Record<string, number> = {};
try {
  for (const entry of searchResult.entries) {
    const logFile = path.resolve(this.storeRoot, entry.file);
    if (!fileHandles[entry.file] && fs.existsSync(logFile)) {
      fileHandles[entry.file] = fs.openSync(logFile, 'r');
    }
    const fd = fileHandles[entry.file];
    if (fd !== undefined) {
      const buffer = Buffer.alloc(entry.length);
      fs.readSync(fd, buffer, 0, entry.length, entry.offset);
      logs.push(JSON.parse(buffer.toString('utf8').trim()));
    }
  }
} finally {
  Object.values(fileHandles).forEach(fd => fs.closeSync(fd));
}
```

js/core/src/tracing/exporter.ts
```diff
@@ -178,3 +182,103 @@ export class TraceServerExporter implements SpanExporter {
...
+        }
+
+        const attributes: any[] = [];
+        for (const [k, v] of Object.entries(log.attributes)) {
+          if (typeof v === 'string')
+            attributes.push({ key: k, value: { stringValue: v } });
+          else if (typeof v === 'number')
+            attributes.push({ key: k, value: { intValue: v } });
+          else if (typeof v === 'boolean')
+            attributes.push({ key: k, value: { boolValue: v } });
```
MichaelDoyle, line 237: This might be OK for now, but we should make a note to follow up and handle more complex types (arrays and maps).

https://opentelemetry.io/docs/specs/otel/common/#anyvalue

genkit-tools/common/src/types/log.ts
```diff
@@ -0,0 +1,66 @@
...
+    traceId: z.string().optional(),
+    spanId: z.string().optional(),
+    timestamp: z
+      .number()
+      .describe('log recorded time in milliseconds since the epoch'),
+    severityNumber: z.number().optional(),
+    severityText: z.string().optional(),
+    body: z.unknown().optional(), // Represents any value: string, number, boolean, array, or object
+    attributes: z.record(z.string(), z.unknown()).optional(),
+    instrumentationLibrary: InstrumentationLibrarySchema.optional(),
```
MichaelDoyle, line 39: @pavelgj do you know, is `instrumentationLibrary` a genkit-ism? Or was this previously a part of the otel spec? Wondering if we should move logs over to `instrumentationScope` (I think we want to do the same for tracing eventually).

js/core/src/logging.ts
```diff
@@ -47,25 +49,87 @@ function getLogger() {
...
+  }
+
   init(fn: any) {
     global[loggerKey] = fn;
   }
 
   info(...args: any) {
     // eslint-disable-next-line prefer-spread
     getLogger().info.apply(getLogger(), args);
+    this._emitOtel('info', args);
```
MichaelDoyle, line 126: `getLogger().info...` will check the log level to see if it should log. Then, `this._emitOtel` always fires. We also need emitOtel to check the log level to see if it should log, the way that `_defaultLogger` does.

MichaelDoyle (commented):
I don't expect it for this PR (or even before we get the UI work done), but we should make a note that we should support real-time. 

I think the main thing to resolve is the log level issue. The rest of the comments are around efficiency follow ups.

js/core/src/logging.ts
MichaelDoyle: I think it will no-op based on how otel works, with one exception - if the developer has already enabled otel logging. 

TL;DR I agree, we should at least no-op via the current flag, and when we productionize things, we need a formal way for developers to opt-out to avoid duplicate logging.

js/core/src/tracing/node-telemetry-provider.ts
shrutip90: Done

genkit-tools/telemetry-server/src/file-log-store.ts
shrutip90: Thanks, that makes sense and Gemini suggested async waiting since this is a server environment and blocking the event loop with lockSync and wait might not be ideal if contention is frequent. Converted LogIndex.add to an async method and used the asynchronous lockfile.lock with retries.

js/core/src/tracing/exporter.ts
shrutip90: Added a TODO

js/core/src/logging.ts
shrutip90: Done, thanks.

genkit-tools/telemetry-server/src/file-log-store.ts
shrutip90: Done

js/core/src/logging.ts
shrutip90: Added a check, thanks.

genkit-tools/common/src/types/log.ts
shrutip90: As discussed offline, instrumentationLibrary is deprecated and replaced by instrumentationScope, but leaving it as is since the information is hardcoded currently, and we can replace it in both trace data and logs data together later.

js/core/src/logging.ts
```diff
@@ -47,25 +49,96 @@ function getLogger() {
...
+      } else if (args.length === 1 && typeof args[0] === 'string') {
+        body = args[0];
+      } else {
+        const util = require('util');
+        body = util.format(...args);
+      }
+
+      let activeContext;
+      try {
+        const { context } = require('@opentelemetry/api');
```
MichaelDoyle: Why can't we use a standard import for this? :)

MichaelDoyle (approved):

js/core/src/logging.ts
shrutip90: Done

PR #5036, opened 2026-04-01
fix(js/genkit): Explicitly allow null types in promptDir

7hokerz:
 - Runtime already treats promptDir: null as an opt-out for automatic `./prompts` loading.
 - The problem is that while runtime logic seems to treat null values ​​specially and allow them, type does not.
 - This change aligns the type definition with existing behavior and adds coverage for the null opt-out case.

In `genkit.ts` `(972-978 line)`
```
private configure() {
...
if (this.options.promptDir !== null) {
      loadPromptFolder(
        this.registry,
        this.options.promptDir ?? './prompts',
        ''
      );
    }
...
```

Checklist (if applicable):

### DIFF
--- a/js/genkit/src/genkit.ts
+++ b/js/genkit/src/genkit.ts
@@ -360,7 +360,7 @@ export class Genkit implements HasRegistry {
       `${name}${options?.variant ? `.${options?.variant}` : ''}`,
       prompt(this.registry, name, {
         ...options,
-        dir: this.options.promptDir ?? './prompts',
+        dir: this.options.promptDir === undefined ? './prompts' : this.options.promptDir,
       })
     );
   }
--- a/js/genkit/src/genkit.ts
+++ b/js/genkit/src/genkit.ts
@@ -360,7 +360,10 @@ export class Genkit implements HasRegistry {
       `${name}${options?.variant ? `.${options?.variant}` : ''}`,
       prompt(this.registry, name, {
         ...options,
-        dir: this.options.promptDir === undefined ? './prompts' : this.options.promptDir,
+        dir:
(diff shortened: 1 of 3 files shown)


MichaelDoyle:
Looping in @ifielker and @pavelgj since they have more context than I do.

7hokerz:
@pavelgj I applied the format. Could you please run the CI again?

js/genkit/src/genkit.ts
```diff
@@ -360,7 +360,10 @@ export class Genkit implements HasRegistry {
       `${name}${options?.variant ? `.${options?.variant}` : ''}`,
       prompt(this.registry, name, {
         ...options,
-        dir: this.options.promptDir ?? './prompts',
+        dir:
+          this.options.promptDir === undefined
```
pavelgj, line 364: hmm... I believe `dir` is unused: https://github.com/genkit-ai/genkit/blob/d5c08d0fea31ef5944565eabd25742d0eb018cf5/js/ai/src/prompt.ts#L873
so ultimately this change doesn't matter... but what's the thinking behind the change? `dir` ends up being `null` so we signal that downstream it should never attempt to load it from the default `./prompts` folder and error instead?

pavelgj (approved):

js/genkit/src/genkit.ts
pavelgj: reasonable I guess.

7hokerz: @pavelgj, it's not being used, but it's for consistency with the updated behavior. 
i referenced a suggestion from Gemini Code Assist.

PR #5044, opened 2026-04-01
feat(js/core): added middleware to reflection api

pavelgj:
Checklist (if applicable):


js/core/src/reflection-v2.ts
```diff
@@ -346,10 +346,28 @@ export class ReflectionServerV2 {
...
     const { type } = ReflectionListValuesParamsSchema.parse(request.params);
+    if (type !== 'defaultModel' && type !== 'middleware') {
+      this.sendError(
+        request.id,
+        400,
+        `'type' ${type} is not supported. Only 'defaultModel' and 'middleware' are supported`
+      );
+      return;
+    }
     const values = await this.registry.listValues(type);
```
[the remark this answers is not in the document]
pavelgj: That's fine. v1 and v2 values should be consistent. If anything, it was wrong before (too loose)

MichaelDoyle (approved):

js/core/src/reflection-v2.ts
```diff
@@ -346,10 +346,28 @@ export class ReflectionServerV2 {
   private async handleListValues(request: JsonRpcRequest) {
     if (!request.id) return;
     const { type } = ReflectionListValuesParamsSchema.parse(request.params);
+    if (type !== 'defaultModel' && type !== 'middleware') {
+      this.sendError(
+        request.id,
+        400,
```
MichaelDoyle: Are these supposed to be spec compliant with JSON-RPC? i.e. `-32602.` (invalid params).

pavelgj: good catch!

PR #5054, opened 2026-04-02
fix(py): parallelize tool calls in generate loop

huangjeff5:
When processing a single model message, we should process all the tool requests in parallel rather than sequentially.

### DIFF
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -655,10 +655,8 @@ async def resolve_tool_requests(
     revised_model_message = message.model_copy(deep=True)
 
     work: list[tuple[int, Action, ToolRequestPart]] = []
-    i = 0
-    for tool_request_part in message.content:
+    for i, tool_request_part in enumerate(message.content):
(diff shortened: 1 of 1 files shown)


py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -653,34 +654,36 @@ async def resolve_tool_requests(
...
         if tool_request.name not in tool_dict:
             raise RuntimeError(f'failed {tool_request.name} not found')
         tool = tool_dict[tool_request.name]
-        tool_response_part, interrupt_part = await _resolve_tool_request(tool, tool_req_root)
+        work.append((i, tool, tool_req_root))
+
+    if not work:
+        return (None, Message(role=Role.TOOL, content=[]), None)
 
+    outs = await asyncio.gather(*[_resolve_tool_request(tool, trp) for _, tool, trp in work])
```
apascal07, line 673: Not that familiar with this but don't you need to return exceptions from this and handle them? Anything that isn't an interrupt.

huangjeff5: Interrupts are caught inside _resolve_tool_request and turned into normal return values. Gather will see real failures though, and those will bubble back up as an Exception to the ai.generate(...) call.

PR #5059, opened 2026-04-02
fix(py): enforce object root JSON Schema for tool inputs at define_tool

huangjeff5:
Validate tool input JSON Schema when registering a tool with `define_tool`. Immediately after `register_action`, `_check_tool_input_schema_is_object` enforces these rules:

1. Allowed:

- **Empty schema `{}`** — allowed for zero-argument tools (Pydantic `TypeAdapter(object).json_schema()` produces `{}`).
- **Root `type` is the string `object`** (case-insensitive) — required for any non-empty schema that declares a root type.

2. Rejected:
- root `type` expressed as a JSON Schema **array** (unions such as `["object", "null"]` or `["string", "object"]`). 
- any other root `type` (scalar, `array`, etc.). Errors tell developers to use a Pydantic model or TypedDict with an object-shaped schema.

### DIFF
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -704,8 +704,8 @@ def _to_pending_response(request: ToolRequestPart, response: ToolResponsePart) -
 async def _resolve_tool_request(tool: Action, tool_request_part: ToolRequestPart) -> tuple[Part | None, Part | None]:
     """Execute a tool and return (response_part, interrupt_part)."""
     try:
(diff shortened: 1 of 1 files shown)


py/packages/genkit/src/genkit/_ai/_tools.py
apascal07, line 1: Would this be better to enforce at the Genkit tool define level to force users to take an object as the input to any tool so the field can have a description for the model to interpret?

huangjeff5: Good idea. Let's go with that.

Issue #393, opened 2024-06-12
Type error in retrievers [JS] 


cabljac:
**Describe the bug**
The types should complain in the options for the pinecone retriever. I get a runtime error:
```sh
Using DevFlowStateStore. Root: /var/folders/y9/0j765brj643gjz3ggwkpn2740000gn/T/.genkit/47a1bd8a68530c0ff153aeb3c40b4c32/flows
Error[/dev-run-action-wrapper/filmRetrieverFlow/filmRetrieverFlow, TypeError] {
  path: '/dev-run-action-wrapper/filmRetrieverFlow/filmRetrieverFlow',
  name: 'TypeError',
  message: "Cannot read properties of undefined (reading 'namespace')",
  stack: "TypeError: Cannot read properties of undefined (reading 'namespace')\n" +
    '    at /Users/jacob/invertase/genkit/genkit-samples/naive-rag/pinecone/node_modules/genkitx-pinecone/lib/index.js:156:37\n' +
    '    at Generator.next (<anonymous>)\n' +
    '    at fulfilled (/Users/jacob/invertase/genkit/genkit-samples/naive-rag/pinecone/node_modules/genkitx-pinecone/lib/index.js:47:24)\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
  source: 'ts',
  sourceVersion: '0.5.0'
}
```
**To Reproduce**

The following code gives no type errors, but it should. The retriever won't work *unless* you specify `k` which is the number of results you wish to return.

```typescript
import { config } from "dotenv";
config();
import { index, retrieve } from "@genkit-ai/ai/retriever";
import { configureGenkit } from "@genkit-ai/core";
import { defineFlow } from "@genkit-ai/flow";
import { textEmbeddingGecko, vertexAI } from "@genkit-ai/vertexai";
import { z } from "zod";
import { pineconeRetrieverRef } from "genkitx-pinecone";
import { pineconeIndexerRef } from "genkitx-pinecone";
import { pinecone } from "genkitx-pinecone";

if (!process.env.PINECONE_API_KEY) {
  throw new Error("PINECONE_API_KEY is required");
}

const fakeData = [
  {
    content: [
      {
        text: "A film about dodging bullets",
      },
    ],
    metadata: {
      name: "The Matrix",
    },
  },
  {
    content: [
      {
        text: "A film about a space opera",
      },
    ],
    metadata: {
      name: "Star Wars",
    },
  },
  {
    content: [
      {
[...]

PR #5065, opened 2026-04-05
fix(retriever): enforce schema-based validation for options to prevent runtime crashes

pranjal2410719:
## Summary

Fixes a mismatch between TypeScript types and runtime behavior in retrievers.

Previously, `options` was defined as `z.any().optional()`, allowing invalid calls without required fields (e.g., `k` for Pinecone), leading to runtime crashes.

## Changes

- Removed `z.any()` usage for `options`
- Removed global `k` from common schema
- Added runtime validation using plugin-defined `__configSchema` for both retriever and indexer

## Result

- Prevents runtime crashes due to missing required options
- Aligns TypeScript types with actual plugin requirements
- Enables flexible, plugin-specific validation

### DIFF
--- a/genkit-tools/common/src/types/retriever.ts
+++ b/genkit-tools/common/src/types/retriever.ts
@@ -25,7 +25,7 @@ import { DocumentDataSchema } from './document';
  */
 export const RetrieverRequestSchema = z.object({
   query: DocumentDataSchema,
-  options: z.any().optional(),
+  options: z.undefined(),
 });
 export type RetrieverRequest = z.infer<typeof RetrieverRequestSchema>;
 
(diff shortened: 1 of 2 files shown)


pranjal2410719:
Hi maintainers 👋

I’ve signed the CLA and addressed the issue described in #393 by introducing schema-based validation for retriever/indexer options.

Please let me know if any changes or improvements are needed — happy to iterate!

Thanks for your time 🙌

pranjal2410719:
Thanks for the detailed review 🙌
Updated the schema to use `z.unknown().optional()` instead of `z.undefined()` to avoid breaking existing integrations.
Also removed the manual validation since Genkit already handles schema validation internally — good catch on that
Please let me know if this aligns better with the expected approach.

MichaelDoyle:
Hey @pranjal2410719 I don't think we'll go with this approach. The Gemini feedback mostly illustrates why.

For the most part, I think this is WAI - `RetrieverRequestSchema` simply passes the options though allowing it to work with any underlying retriever. Retriever's tend to be written by individual app developers, because they can really talk to any underlying data store (too many to provide plugins for) -- hence the generic nature.

One thing you can look at is how Model plugins do validation to get a better idea of how a base Schema with more concrete subtypes could be implemented for individual retrievers.

Were you mostly having trouble with the Dev UI? Or where is the major pain point for you?

PR #5073, opened 2026-04-08
feat(py): Add DAP helpers to registry + fix business logic in Registry.list_actions()

huangjeff5:
- Wires Dynamic Action Provider (DAP) into Registry.resolve_action: define_dynamic_action_provider now attaches the DynamicActionProvider to the registered DAP Action, and resolution uses get_action(kind, name). Hand-registered providers that still resolve via run({kind, name}) → Action keep working.
- Adds parse_dap_qualified_name and resolve_action_by_key support for /dynamic-action-provider/<provider>:<innerKind>/<innerName>.

Update 4/23: 
I ended up making a few more related improvements to the code, this was the most natural place to put it to prevent merge conflicts.
- Refactor: Removed hand-written ActionMetadata class; re-use the shared `ActionMetadata` type provided through the type generation script.

Update 4/28:
- Refactor: Simplified the logic in unify `list_actions()` and `list_resolvable_actions()`.

### DIFF
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -229,6 +229,11 @@ def extract_action_args_and_types(
 # =============================================================================
 
 
+# Attribute name used to attach a ``DynamicActionProvider`` (cache + helpers)
+# onto the placeholder ``Action`` registered for a DAP. The registry only
+# stores the ``Action``; the provider rides along on it as a Python attribute.
+# Code holding the ``Action`` recovers the provider via
+# ``getattr(action, GENKIT_DYNAMIC_ACTION_PROVIDER_ATTR, None)``.
 GENKIT_DYNAMIC_ACTION_PROVIDER_ATTR = '_genkit_dynamic_action_provider'
 
 
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -546,20 +546,6 @@ class Action(Generic[InputT, OutputT, ChunkT]):
             self._metadata[ActionMetadataKey.OUTPUT_KEY] = self._output_schema
 
 
-class ActionMetadata(BaseModel):
-    """Action metadata for registry and reflection."""
-
-    kind: ActionKind
-    name: str
-    description: str | None = None
-    input_schema: object | None = None
-    input_json_schema: dict[str, object] | None = None
-    output_schema: object | None = None
-    output_json_schema: dict[str, object] | None = None
-    stream_schema: object | None = None
-    metadata: dict[str, object] | None = None
-
-
 def _make_tracing_wrapper(
     name: str,
     kind: ActionKind,
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -274,7 +274,7 @@ def parse_action_key(key: str) -> tuple[ActionKind, str]:
     return kind, name
 
 
-def create_action_key(kind: ActionKind, name: str) -> str:
+def create_action_key(kind: ActionKind | str, name: str) -> str:
     """Create '/<kind>/<name>' key."""
     return f'/{kind}/{name}'
 
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -19,6 +19,7 @@
 import asyncio
 import inspect
 import json
+import re
 import time
 from collections.abc import AsyncIterator, Awaitable, Callable, Generator, Mapping
 from contextlib import contextmanager
@@ -237,21 +238,19 @@ def parse_dap_qualified_name(name: str) -> tuple[str, str, str] | None:
     Used when the action key kind is ``dynamic-action-provider`` and the name
     references a nested action exposed by a provider (e.g. MCP tools).
 
+    Pattern: ``[provider]:[inner_kind]/[inner_name]`` — no slashes in the
+    provider segment (``plugin/foo`` is not a valid provider host).
+
     Returns:
         ``(provider_name, inner_kind, inner_name)`` if the string matches the
         pattern; otherwise ``None`` so callers can treat the name as a plain
         dynamic-action-provider id.
     """
-    if ':' not in name or '/' not in name:
+    # Pattern: [provider]:[inner_kind]/[inner_name]; no '/' or ':' in provider.
+    match = re.match(r'^([^/:]+):([^/:]+)/(.+)$', name)
+    if not match:
         return None
-    colon = name.index(':')
-    provider = name[:colon]
-    rest = name[colon + 1 :]
-    if '/' in provider:
-        return None
-    if '/' not in rest:
-        return None
-    inner_kind, inner_name = rest.split('/', 1)
+    provider, inner_kind, inner_name = match.groups()
     if not provider or not inner_kind or not inner_name:
         return None
     return (provider, inner_kind, inner_name)
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -247,6 +247,8 @@ def parse_dap_qualified_name(name: str) -> tuple[str, str, str] | None:
     colon = name.index(':')
     provider = name[:colon]
     rest = name[colon + 1 :]
+    if '/' in provider:
+        return None
     if '/' not in rest:
         return None
     inner_kind, inner_name = rest.split('/', 1)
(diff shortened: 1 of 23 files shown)


py/packages/genkit/src/genkit/_core/_registry.py
```diff
@@ -128,6 +143,44 @@ async def async_schema_resolver(name: str) -> dict[str, object] | str:
         self._plugins: dict[str, Plugin] = {}
         self._plugin_init_tasks: dict[str, asyncio.Task[None]] = {}
 
+    # -------------------------------------------------------------------------
+    # Child registry support
+    # -------------------------------------------------------------------------
+
+    def new_child(self) -> 'Registry':
```
huangjeff5: type shouldn't be quoted in strings

py/packages/genkit/src/genkit/_core/_registry.py
```diff
@@ -448,38 +504,85 @@ async def resolve_action(self, kind: ActionKind, name: str) -> Action | None:
...
+                        if resolved is not None:
+                            return resolved
+                    except Exception as e:
+                        logger.debug(
+                            f'Dynamic action provider {provider_action.name} failed for {kind}/{name}',
+                            exc_info=e,
+                        )
+                        continue
+                    continue
                 try:
```
huangjeff5, line 612: Not sure what this path (521-531) is doing exactly. What legacy?

py/packages/genkit/tests/genkit/core/action_test.py
```diff
@@ -72,6 +73,15 @@ def test_parse_action_key_invalid_format() -> None:
             parse_action_key(key)
 
 
+def test_parse_dap_qualified_name() -> None:
+    """Parse provider:innerKind/innerName segments."""
+    assert parse_dap_qualified_name('my-dap:tool/echo') == ('my-dap', 'tool', 'echo')
+    assert parse_dap_qualified_name('plugin/foo:model/bar') == ('plugin/foo', 'model', 'bar')
```
huangjeff5: plugin/foo should not be considered a valid host

py/packages/genkit/src/genkit/_core/_action.py
```diff
@@ -228,6 +228,35 @@ def extract_action_args_and_types(
...
+    Returns:
+        ``(provider_name, inner_kind, inner_name)`` if the string matches the
+        pattern; otherwise ``None`` so callers can treat the name as a plain
+        dynamic-action-provider id.
+    """
+    if ':' not in name or '/' not in name:
+        return None
+    colon = name.index(':')
+    provider = name[:colon]
+    rest = name[colon + 1 :]
```
ifielker: Can just use a regex for this I think:
   ``` 
    # Pattern: [provider]:[inner_kind]/[inner_name]
    # Expects no slashes in the provider
    pattern = r'^([^/:]+):([^/:]+)/(.+)$'
    
    match = re.match(pattern, name)
    if match:
        return match.groups()  # Returns (provider, inner_kind, inner_name)
    return None
```

File: py/packages/genkit/src/genkit/_core/_dap.py
Comment (ifielker) [inline py/packages/genkit/src/genkit/_core/_dap.py:88]: I know it's not part of the PR, but this is based on the initial implementation of DAP. JS implementation has changed now and does not pass anything in to the DAP action. Python should probably match (for a future PR)

File: py/packages/genkit/src/genkit/_core/_registry.py [hunk 5]
```diff
@@ -513,7 +612,15 @@ async def list_actions(self, allowed_kinds: list[ActionKind] | None = None) -> l
 
                 if allowed_kinds and meta.kind not in allowed_kinds:
                     continue
+                seen.add((meta.kind, meta.name))
                 metas.append(meta)
+
+        # Include parent actions not shadowed by local plugins.
```
ifielker: I spotted a subtle but dangerous shadowing ambiguity here with child registries, specifically regarding how tools are resolved vs. how they are listed.

Currently, `list_actions()` populates the `seen` set __only__ using actions advertised by the child registry's *plugins*. It completely ignores actions provided by the child's Dynamic Action Providers (e.g., an MCP server) or local `self._entries`.

This creates a serious discrepancy between `list_actions()` and `resolve_action()`.

__The Scenario:__

1. A __Parent Registry__ has a standard plugin installed that provides a tool named `fetch_weather`.
2. A __Child Registry__ is created ephemerally for a specific generate call and registers a DAP (like an MCP server) that *also* exposes a tool named `fetch_weather`.

__What happens?__

- __During Resolution:__ `resolve_action(ActionKind.TOOL, 'fetch_weather')` correctly queries the child's DAPs *before* falling back to the parent. The MCP server intercepts the call and the tool is executed successfully. The parent is properly shadowed.
- __During Listing:__ `list_actions()` completely ignores the MCP server when building the `seen` set. It falls back to the parent, sees the parent plugin's `fetch_weather` metadata, and blindly appends it to the list since it wasn't marked as `seen`.

__The Result:__ Any system relying on `list_actions()` (like the DevUI or reflection APIs) will receive the parent plugin's description and input schema for `fetch_weather`. However, when the framework actually executes the tool based on that schema, the call is routed to the child MCP server's tool—which might expect completely different parameters, causing the execution to crash or behave unpredictably.

__Suggested Fix:__ To mirror the JavaScript implementation (`listResolvableActions()`), Python's `list_actions()` needs to build a comprehensive `seen` set that accurately reflects *everything* the child shadows. We should iterate over the child's DAPs (using the newly added `get_action_metadata_record()`) and local `self._entries`, adding their `(kind, name)` tuples to the `seen` set before we fall back to appending the parent's metadata.

(In fact I might advise against allowing searches for tool/fetch_weather to go into all the DAPs searching for tools because that can create ambiguity if two mcps both register tools with the same name. Which ever happened to be registered first will be the one that is found, and that makes for hard to find bugs.)

py/packages/genkit/src/genkit/_core/_dap.py
huangjeff5: Noted. Will work on this on a followup PR.

py/packages/genkit/src/genkit/_core/_registry.py
huangjeff5: OK I spent some time refactoring this part of the Registry. I noticed some other stuff that didn't match JS implementation. 

Now I have Dev UI wired up to call `list_resolved_actions()`, which implements the behavior above (looking at plugin's list_actions() first, then registered entries, then DAP actions. Then fallback to parent.

I agree with your point about not allowing searches for tool/fetch_weather to go into all the DAPs, although I'll leave it to a followup PR.

py/packages/genkit/src/genkit/_core/_action.py
huangjeff5: Fixed this.

py/packages/genkit/src/genkit/_core/_registry.py
```diff
@@ -76,6 +81,33 @@
 )
 
 
+def _reflection_payload_for_registered_action(action: Action) -> dict[str, Any]:
```
pavelgj: there should be ActionMetadata pydantic class generated which you can probably use here for the type...

py/packages/genkit/src/genkit/_core/_registry.py
```diff
@@ -76,6 +81,33 @@
...
+        'name': action.name,
+        'type': action.kind,
+        'description': action.description,
+        'inputSchema': action.input_schema,
+        'outputSchema': action.output_schema,
+        'metadata': action.metadata,
+    }
+
+
+def _reflection_payload_for_action_metadata(meta: ActionMetadata) -> dict[str, Any]:
```
pavelgj: same here

py/packages/genkit/src/genkit/_core/_registry.py
```diff
@@ -204,6 +290,80 @@ async def resolve_actions_by_kind(self, kind: ActionKind) -> dict[str, Action]:
             await self._trigger_lazy_loading(action)
         return actions
 
+    async def list_actions(self) -> dict[str, Action]:
```
pavelgj: you might be able to get rid of this and only have list_resolvable_actions, which can be renamed to list_actions

py/packages/genkit/src/genkit/_core/_registry.py
```diff
@@ -84,17 +116,29 @@ class Registry:
...
-    def __init__(self) -> None:
-        """Initialize an empty Registry instance."""
+        Args:
+            parent: Optional parent registry.  When provided this is a *child*
+                registry that falls back to the parent for any lookup that
+                returns ``None`` locally.  Use ``new_child()`` as the
+                preferred factory rather than passing ``parent`` directly.
+        """
+        self._parent: Registry | None = parent
+        self._default_model: str | None = None
```
huangjeff5: don't need default_model on Registry field I believe (should be stored as a value)

pavelgj (approved):

PR #5075, opened 2026-04-08
feat(py): integrate DAP into generate call, expand wildcards

huangjeff5:
## Summary

- **Child registry per generate:** `_generate_action` creates an `effective_registry` (child of the root registry) at the start of each generate call. Tool resolution and resource lookup use this child so DAP-resolved tools are scoped to the call and do not pollute the root registry.
- **Wildcard tool expansion:** `expand_wildcard_tools()` expands patterns such as `provider:tool/*` and `provider:tool/prefix*` into concrete tool names via `DynamicActionProvider.list_action_metadata`. Expansion runs before parameter resolution so the model receives a full `ToolDefinition` list.
- **Tool execution:** `resolve_tool_requests` receives `effective_registry` so execution uses the child’s resolved action cache.

## Additional Changes 
- Modified resolve_action() such that when a DAP qualified name is passed in, it will resolve the DAP and retrieve the action from the DAP directly.
- Otherwise, it will proceed normally (look at the entries cache, then search through plugins). No search through DAP.
- Added `assert_valid_tool_names()` to block tool name collision scenarios when resolving DAP tools in `ai.generate`. Prevent registered tool + DAP tool collision, as well as collision with the same name across multiple DAP tools.
- While we're refactoring `resolve_action`, I noticed the JS equivalent also doesn't have "search plugins for the action name as fallback" logic. I removed that for consistency and also to simplify the code.


py/packages/genkit/tests/genkit/ai/dynamic_tools_generate_test.py
```diff
@@ -0,0 +1,236 @@
...
+    call_log: list[str] = []
+
+    class EchoInput(BaseModel):
+        text: str
+
+    async def echo_fn(inp: EchoInput) -> str:
+        call_log.append(inp.text)
+        return f'echoed:{inp.text}'
+
+    echo_action = ai.registry.register_action(
```
huangjeff5: Doesn't this put it in the Registry as a normal tool? I thought we're testing a tool that's only provided through a DAP.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -61,6 +62,46 @@
 logger = get_logger(__name__)
 
 
+async def expand_wildcard_tools(registry: Registry, tool_names: list[str]) -> list[str]:
```
ifielker, line 69: This works a bit differently from the javascript implementation and there is a critical bug here regarding how the expanded tool names are returned.

Currently, the function strips the provider prefix and appends only the unqualified tool name:

```python
tool_name = meta.get('name')
if tool_name:
    expanded.append(str(tool_name))
```

This creates a serious conflict resolution issue because it drops the explicit provider binding. When the generation loop later attempts to resolve this unqualified name (e.g., `'echo'`), the registry will fall back to querying all registered DAPs in registration order to find a match.

__The Bug Scenario:__

1. You have two MCP servers registered: `mcp1` (registered first) and `mcp2` (registered second).
2. Both servers expose a tool named `'echo'`.
3. The user explicitly requests tools from only the second server: `tools=['mcp2:tool/*']`.
4. This function expands that request to just `['echo']`.
5. When the framework resolves `'echo'`, it asks the DAPs in order. `mcp1` replies first saying "I have echo!" and its tool is executed instead of `mcp2`'s.

__The Fix:__ To maintain the explicit provider binding (and match how the JS implementation behaves), we should reconstruct the fully-qualified DAP key so the registry knows exactly which provider to query later on:

```python
tool_name = meta.get('name')
if tool_name:
    expanded.append(f'/dynamic-action-provider/{provider_name}:{action_type}/{tool_name}')
```

This guarantees the executed tool will strictly belong to the provider the user requested.

It would also be a good idea to have a test for this scenario.

py/packages/genkit/tests/genkit/ai/dynamic_tools_generate_test.py
```diff
@@ -0,0 +1,244 @@
...
+    assert sorted(result) == ['get_time', 'get_weather']
+
+
+@pytest.mark.asyncio
+async def test_non_wildcard_names_pass_through() -> None:
+    """Non-wildcard names are returned unchanged."""
+    registry = Registry()
+    result = await expand_wildcard_tools(registry, ['my_tool', 'other_tool'])
+    assert result == ['my_tool', 'other_tool']
+
```
ifielker, line 119: Here's a test that shows the bug I mentioned: 
```@pytest.mark.asyncio
async def test_wildcard_tools_avoids_shadowing_conflict() -> None:
    """Explicit wildcard provider paths should not be shadowed by earlier providers."""
    ai = Genkit()
    pm, _ = define_programmable_model(ai)

    call_log: list[str] = []

    class Inp(BaseModel):
        x: str

    async def echo1_fn(inp: Inp) -> str:
        call_log.append('mcp1')
        return 'echo 1'

    async def echo2_fn(inp: Inp) -> str:
        call_log.append('mcp2')
        return 'echo 2'

    # Detached Actions (not registered in root registry directly)
    echo1_action = Action(name='echo', kind=ActionKind.TOOL, fn=echo1_fn, metadata={'name': 'echo'})
    echo2_action = Action(name='echo', kind=ActionKind.TOOL, fn=echo2_fn, metadata={'name': 'echo'})

    async def dap1_fn() -> DapValue:
        return {'tool': [echo1_action]}

    async def dap2_fn() -> DapValue:
        return {'tool': [echo2_action]}

    # Register mcp1 first. If resolution falls back to an unqualified lookup, mcp1 will "win".
    ai.define_dynamic_action_provider('mcp1', dap1_fn)
    ai.define_dynamic_action_provider('mcp2', dap2_fn)

    # The model calls the 'echo' tool
    pm.responses = [
        _tool_call_response('echo', {'x': 'hello'}),
        _text_response('finished'),
    ]

    response = await ai.generate(
        model='programmableModel',
        prompt='use echo',
        # Crucially, we explicitly request tools from mcp2 ONLY
        tools=['mcp2:tool/*'],
    )

    assert response.text == 'finished'
    
    # If the bug is present, this will fail because it will fall back to the unqualified 
    # global loop and find mcp1's 'echo' tool instead.
    assert call_log == ['mcp2']
```

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: OK good catch, fixed this issue. The scenario above should also be blocked by `assert_valid_tool_names()`, since if there are two tools in scope with the same short name that will raise.

ifielker (approved):

PR #5079, opened 2026-04-08
feat(genkit-tools/cli): support ephemeral runtime commands

ssbushi:
Supports specifying runtime startup with runtime commands.  

Eg: 

` genkit flow:run simpleEcho '"hello world"' -- pnpm run dev`

Checklist (if applicable):
- [x] Tested (manually)


genkit-tools/cli/src/utils/manager-utils.ts
```diff
@@ -236,13 +236,73 @@ export async function runWithManager(
...
+  projectRoot: string,
+  runtimeCommand: string[],
+  fn: (manager: BaseRuntimeManager) => Promise<void>
+) {
+  const devEnv = await getDevEnvVars(projectRoot, {
+    experimentalReflectionV2: true,
+  });
+  const { envVars, telemetryServerUrl, reflectionV2Port } = devEnv;
+
+  process.env.GENKIT_TELEMETRY_SERVER = telemetryServerUrl;
```
pavelgj: this feels more complicated than expected... what's the difference from `runWithManager`

ssbushi: Mostly the reflectionv2 stuff and stopping the manager at the end. 

What seems complicated here?

pavelgj: Gemini and I think runWithEphemeralManager and runWithManager can be combined into a single function reducing duplication:

```ts
export interface RunWithManagerOptions {
  /** Command to start the runtime process. If provided, an ephemeral manager is used. */
  runtimeCommand?: string[];
  experimentalReflectionV2?: boolean;
}
export async function runWithManager(
  projectRoot: string,
  fn: (manager: BaseRuntimeManager) => Promise<void>,
  options?: RunWithManagerOptions
) {
  const useEphemeral = options?.runtimeCommand && options.runtimeCommand.length > 0;
  let manager: BaseRuntimeManager;
  let oldLevel = logger.level;
  try {
    if (useEphemeral) {
      const devEnv = await getDevEnvVars(projectRoot, {
        experimentalReflectionV2: options?.experimentalReflectionV2 ?? true,
      });
      const { envVars, telemetryServerUrl, reflectionV2Port } = devEnv;
      process.env.GENKIT_TELEMETRY_SERVER = telemetryServerUrl;
      logger.level = 'warn';
      const result = await startDevProcessManager(
        projectRoot,
        options!.runtimeCommand![0],
        options!.runtimeCommand!.slice(1),
        {
          experimentalReflectionV2: options?.experimentalReflectionV2 ?? true,
          healthCheck: true,
          envVars,
          telemetryServerUrl,
          reflectionV2Port,
          nonInteractive: true,
        }
      );
      manager = result.manager;
    } else {
      manager = await startManager({ 
        projectRoot, 
        manageHealth: false,
        experimentalReflectionV2: options?.experimentalReflectionV2
      });
    }
  } catch (e) {
    logger.error('Failed to start manager', e);
    process.exit(1);
  } finally {
    if (useEphemeral) {
      logger.level = oldLevel;
    }
  }
  try {
    await fn(manager);
  } catch (err) {
    // ... error handling (same as before) ...
  } finally {
    // Always stopping is safer as it cleans up watchers, but we MUST stop if we started a process.
    if (useEphemeral || manager) {
      await manager.stop();
    }
  }
}
```

pavelgj: https://github.com/genkit-ai/genkit/pull/5151

pavelgj (approved):

PR #5114, opened 2026-04-13
feat(ollama): detect model capabilities dynamically via /api/show

intel352:
## Summary

Replace the hardcoded `toolSupportedModels` and `mediaSupportedModels` allowlists in the Ollama plugin with dynamic capability detection from Ollama's `/api/show` endpoint.

## Problem

The Ollama plugin maintains static string arrays of model names that support tools and media. Newly released models (e.g. `gemma4`, which Ollama reports as tool-capable via its API) don't work with tool calling until someone manually updates these lists and ships a new Genkit release. This creates an unnecessary maintenance burden and blocks users from using new models with tools.

## Solution

Ollama's `POST /api/show` endpoint returns a `capabilities` array (e.g. `["completion", "vision", "tools", "thinking"]`) that accurately reports what each model supports. This PR queries that endpoint when determining model capabilities, falling back to the existing static lists when the server doesn't report capabilities (older Ollama versions).

## Changes

- `go/plugins/ollama/ollama.go`:
  - Add `getModelCapabilities()` — queries `/api/show` for a model's capabilities
  - Add `modelSupportsFromCapabilities()` — derives `ModelSupports` from capability strings, with static-list fallback
  - Update `DefineModel()` to use dynamic detection when no explicit opts provided
  - Update `ListActions()` and `ResolveAction()` to use dynamic detection
- `go/plugins/ollama/ollama_test.go`:
  - Add `TestGetModelCapabilities` — mock server tests for tools/no-tools/unknown models
  - Add `TestModelSupportsFromCapabilities` — unit tests for dynamic and fallback paths

## Backward Compatibility

- When `/api/show` returns capabilities → used directly (new behavior)
- When `/api/show` doesn't return capabilities or fails → falls back to existing static allowlists (existing behavior preserved)
- Explicit `ModelOptions` passed to `DefineModel()` are still respected (no change)

## Testing

All existing tests pass. New tests cover:
- Capability detection from mock `/api/show` endpoint
- Fallback to static lists when no capabilities reported
- Models in the static list still work via fallback
- Unknown models default to no tools/media

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/go/plugins/ollama/ollama.go
+++ b/go/plugins/ollama/ollama.go
@@ -164,8 +164,16 @@ func (o *Ollama) listLocalModels(ctx context.Context) ([]ollamaLocalModel, error
 }
 
 func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.ModelOptions) ai.Model {
-	// Detect capabilities before acquiring the lock to avoid holding it
-	// during HTTP I/O.
+	// Check the init guard first under a brief lock — before any I/O — so
+	// that a forgotten Init() panics immediately rather than after a timeout.
+	o.mu.Lock()
+	if !o.initted {
+		o.mu.Unlock()
+		panic("ollama.Init not called")
+	}
+	o.mu.Unlock()
+
+	// Detect capabilities outside the lock to avoid holding it during HTTP I/O.
 	var modelOpts ai.ModelOptions
 	if opts != nil {
 		modelOpts = *opts
@@ -184,11 +192,9 @@ func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.M
 		}
 	}
 
+	// Re-acquire lock for the registration step.
 	o.mu.Lock()
 	defer o.mu.Unlock()
-	if !o.initted {
-		panic("ollama.Init not called")
-	}
 	meta := &ai.ModelOptions{
 		Label:        "Ollama - " + model.Name,
 		Supports:     modelOpts.Supports,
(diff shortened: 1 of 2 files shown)


MichaelDoyle:
Thanks @intel352 - we'll take a look soon. Sorry for the delay.

apascal07:
@cabljac @adesinah any updates here?

adesinah (changes requested):
Thanks for this contribution, @intel352 

The idea is definitely the right direction. I found a few compatibility cases that need addressing before merge, though.

First, the previous implementation only enabled tools for chat models. That guard has been removed, so a non-chat model can now advertise tool support when `/api/show` reports tools or the model appears in the fallback list. Because non-chat models use /api/generate, where this plugin does not send tool definitions, those tools would be accepted by Genkit but silently omitted from the Ollama request.

Second, fallback does not preserve the previous behavior for dynamically discovered models. `ListActions` and `ResolveAction` previously used `defaultOllamaSupports`, which enabled tools and media. They now use the static allowlists when `/api/show` fails or is unavailable, causing unknown or newer models to lose those capabilities. This is particularly important because the PR describes fallback as backward compatible.

Finally, `len(caps) > 0` cannot distinguish an unavailable capability response from a successful but empty capabilities array. In the latter case, the code can incorrectly re-enable capabilities from the static lists. It would be safer for `getModelCapabilities` to return an explicit success/error signal alongside the array.

Could we add regression tests for these three cases and preserve the previous fallback behavior per call site?

adesinah:
@intel352 I have addressed the required changes mentioned above here #5921 - it also includes your initial commits. So I will be closing this PR.

adesinah:
Closes for #5921

PR #5116, opened 2026-04-14
feat(js): ai.countTokens support

ifielker:
Checklist (if applicable):


js/genkit/src/genkit.ts
```diff
@@ -764,6 +765,13 @@ function registerActionV2(
       resolvedAction,
       { namespace: plugin.name }
     );
+    if (isModelAction(resolvedAction) && resolvedAction.__tokenCounterAction) {
+      registry.registerAction(
+        'model-token-counter',
+        resolvedAction.__tokenCounterAction,
+        { namespace: plugin.name }
+      );
+    }
```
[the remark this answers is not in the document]
ifielker: There are currently no background models with token counting capability. We will add it when those models are available.

js/ai/src/generate.ts
```diff
@@ -796,3 +800,106 @@ export function tagAsPreamble(msgs?: MessageData[]): MessageData[] | undefined {
...
+        // Return a dummy response to safely unwind the middleware chain without
+        // executing the actual model generation, since we only want to intercept the mutated request.
+        return {
+          message: { role: 'model', content: [] },
+          finishReason: 'stop',
+        };
+      }
+      const currentMiddleware = resolvedMiddleware[index];
+      if (currentMiddleware.model) {
+        return currentMiddleware.model(req, {}, async (modifiedReq) =>
```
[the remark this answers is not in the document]
ifielker: Actually, these two locations are invoking two different middleware interface types:

1. __`model.ts`__ is iterating over `ModelMiddlewareArgument[]`, which can be `ModelMiddlewareWithOptions`. The signature for `MiddlewareWithOptions` is `(req, options: ActionRunOptions<S> | undefined, next) => Promise<O>`. Because it explicitly allows `undefined`, `model.ts` can safely pass `undefined`.

2. __`countTokens` (`generate.ts`)__ has already resolved the middleware via `resolveMiddleware`, so it is invoking `GenerateMiddlewareDef.model`. The signature for this internal hook strictly requires an object: `(req, ctx: ActionRunOptions<any>, next) => Promise<GenerateResponseData>`. It does not allow `undefined`.

If we were to change `GenerateMiddlewareDef.model` to accept `ActionRunOptions<any> | undefined` just to match `model.ts`, we would introduce a breaking change for all developers who write custom Genkit middleware. They would suddenly have to perform null checks (e.g., `if (ctx)`) every time they hook into the generation lifecycle. Furthermore, the main `generate` flow actually always passes an object (`actionOpts`) to this hook, guaranteeing `ctx` is never undefined in production.

Because of this strict type requirement in `GenerateMiddlewareDef.model`, we must pass `{}` in `countTokens`. And because `model.ts` is using a different interface that accepts `undefined`, and is out of scope for this PR, we should leave them as they are.

js/ai/src/generate.ts
```diff
@@ -796,3 +800,60 @@ export function tagAsPreamble(msgs?: MessageData[]): MessageData[] | undefined {
...
+    ...(resolvedModel.config || {}),
+    ...(request.config || {}),
+  };
+
+  const counterActionName = `/model-token-counter/${resolvedModel.modelAction.__action.name}`;
+  const counterAction = await registry.lookupAction<
+    typeof GenerateRequestSchema,
+    typeof GenerationUsageSchema,
+    TokenCounterAction<CustomOptions>
+  >(counterActionName);
```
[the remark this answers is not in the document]
ifielker: done

js/ai/src/generate.ts
```diff
@@ -796,3 +800,60 @@ export function tagAsPreamble(msgs?: MessageData[]): MessageData[] | undefined {
...
+  const request = await toGenerateRequest(registry, {
+    ...resolvedOptions,
+    tools,
+    resources,
+  });
+
+  request.config = {
+    ...(resolvedModel.config || {}),
+    ...(request.config || {}),
+  };
```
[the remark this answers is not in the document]
ifielker: done

js/ai/src/model/middleware.ts
```diff
@@ -199,9 +200,9 @@ const CONTEXT_ITEM_TEMPLATE = (
   return out;
 };
 
-export function augmentWithContext(
+export function augmentWithContext<O = GenerateResponseData>(
   options?: AugmentWithContextOptions
-): ModelMiddleware {
+): SimpleMiddleware<GenerateRequestData, O> {
```
pavelgj, line 205: why these changes?

js/ai/src/generate.ts
```diff
@@ -796,3 +800,106 @@ export function tagAsPreamble(msgs?: MessageData[]): MessageData[] | undefined {
...
+    resolvedOptions.use
+  );
+  const resolvedMiddleware = await resolveMiddleware(
+    childRegistry,
+    middlewareRefs
+  );
+  maybeRegisterDynamicMiddlewareTools(childRegistry, resolvedMiddleware);
+
+  let interceptedRequest = request;
+  if (resolvedMiddleware && resolvedMiddleware.length > 0) {
```
pavelgj: why? there's no middleware for counters...

js/ai/src/generate.ts
```diff
@@ -796,3 +800,106 @@ export function tagAsPreamble(msgs?: MessageData[]): MessageData[] | undefined {
...
+      } else {
+        return dispatchModel(index + 1, req);
+      }
+    };
+    await dispatchModel(0, request);
+  }
+
+  const counterActionName = `/model-token-counter/${resolvedModel.modelAction.__action.name}`;
+  const counterAction =
+    resolvedModel.modelAction.__tokenCounterAction ||
```
pavelgj, line 874: in what case is `resolvedModel.modelAction.__tokenCounterAction` not set?

js/genkit/src/genkit.ts
```diff
@@ -764,6 +765,13 @@ function registerActionV2(
       resolvedAction,
       { namespace: plugin.name }
     );
+    if (isModelAction(resolvedAction) && resolvedAction.__tokenCounterAction) {
```
pavelgj: I'm not a fan of exposing too many implementation details here. With background action we use a helper (registerBackgroundAction), thinking `registerModelAction` is in order here....

js/ai/src/generate.ts
```diff
@@ -796,3 +800,106 @@ export function tagAsPreamble(msgs?: MessageData[]): MessageData[] | undefined {
...
+      const currentMiddleware = resolvedMiddleware[index];
+      if (currentMiddleware.model) {
+        return currentMiddleware.model(req, {}, async (modifiedReq) =>
+          dispatchModel(index + 1, modifiedReq || req)
+        );
+      } else {
+        return dispatchModel(index + 1, req);
+      }
+    };
+    await dispatchModel(0, request);
```
pavelgj: why call the model?

js/ai/src/model/middleware.ts
ifielker: When we add middleware that modifies the request, it needs to be run for both generate and countTokens otherwise it might not be valid (think expanding and including large things). countTokens outputs usage and generate outputs GenerateRequestData. ModelMiddleware = SimpleMiddleware<GenerateRequestData, GenerateResponseData> so when calling augmentWithContext(...) nothing changes, but when calling augmentWithContext\<UsageData\> it will work for countTokens. (augmentWithContext is input only, there is no output, so it doesn't matter). Perhaps we should have made an InputMiddleware and and OutputMiddleware or something to differentiate them so the input ones can be used easily with countTokens too.

js/ai/src/generate.ts
ifielker: ai.countTokens takes a fully formed request. Middleware can modify the request. Example: when the model itself doesn't support X so the middleware modifies the request to a format the model does support - Like downloading an image from an unsupported url and including the image in the media instead of the url only. That's a large difference in tokens used. If we countTokens on the unmodified request it will include the length of the url. If we countTokens on the post-middleware request it will include the length of the base64 string...

js/ai/src/generate.ts
ifielker: That's how we get the modified request. We essentially apply all the middleware to the request and then return a dummy response to safely unwind the middleware chain without executing the model and then we have the mutated request that we can give to countTokens.

js/ai/src/generate.ts
ifielker: It's not set if the countTokens was not defined during the model definition (not all models have token counting endpoints in Gemini API for example. Theoretically if someone wanted a tokenCounter that is not tied to a model, they could add one using something like: 
```
const genericTokenCounter = action(
  {
    actionType: 'model-token-counter',
    name: 'generic-counter',
    inputSchema: GenerateRequestSchema,
    outputSchema: GenerationUsageSchema,
  },
  async (request) => {
    return {
      totalTokens: 42,
    };
  }
);
``` 
and then we would still find it using this search. Unlikely, but it keeps the possibility open. Also the error for not finding the countTokens action will be the standard "action not found".

js/ai/src/generate.ts
pavelgj: that is true that middleware can modify the request, but also middleware can do a lot of other things (even with potential side-effects)... I'm worried that it may not be obvious to the user and can lead to undesired side-effects.

I think that countTokens should not take a GenerateRequest but just a list of Messages.

js/genkit/src/genkit.ts
ifielker: Good idea! Done.

js/ai/src/generate.ts
ifielker: That leads to inaccuracies in token counting though... and potentially large ones. The details of the request including middleware and everything else that happens to it is important if we want accurate token counting. And there's a BIG difference between a url string and a downloaded media dataurl...

ifielker: Also I suspect our users will often want something like: 
```
const request = {model, config, ...};
const tokens = await ai.countTokens(request);
if (tokens.totalTokens < MAX_TOKENS_ALLOWED) {
  const response = await ai.generate(request);
  // ...
}
```
and it would probably be bad if tokens.totalTokens was thousands of tokens off the actual answer...

PR #5131, opened 2026-04-15
feat(js): allow middleware generate hook manipulate message/turn index

pavelgj:
## Summary
This PR refactors the Genkit generate middleware API to allow hooks to observe and modify execution state (turn and message indices). This was necessary to fix a bug in the `filesystem` middleware where injected file contents were not being streamed to the user with correct sequence indices.
Additionally, this PR includes a fix for a broken `RunOptions` signature in the `Action` type in `js/core/src/action.ts`.
### Core Framework (`js/ai`)
- **Middleware API Refactor**: Updated the `generate` hook in `GenerateMiddlewareDef` to receive an envelope containing `{ request, currentTurn, messageIndex }` instead of just the request object. This allows middleware to update indices when injecting messages.
- **Generate Action Loop**: Updated `generateActionImpl` to pass this envelope down the chain and respect updates to `currentTurn` and `messageIndex`.
- **Chunk Wrapping**: Added logic to automatically wrap raw data streamed by middleware into `GenerateResponseChunk` objects with safe defaults (`previousChunks: []`, `parser: undefined`) to avoid runtime errors.
### Plugins (`js/plugins/middleware`)
- **Filesystem Middleware**: Updated to use the new envelope API. It now correctly streams file contents injected by the `read_file` tool to the user's output stream using the correct sequential `messageIndex`.
- **Skills Middleware**: Updated to match the new envelope signature (pass-through).
### Core (`js/core`)
- **Action Type Fix**: Fixed broken `RunOptions` signature in the `Action` type in `js/core/src/action.ts`.

Checklist (if applicable):

### DIFF
--- a/js/ai/src/generate/action.ts
+++ b/js/ai/src/generate/action.ts
@@ -277,6 +277,11 @@ async function generateActionImpl(
     context,
   } = args;
 
+  const format = await resolveFormat(registry, rawRequest.output);
+
+  const sharedPreviousChunks: GenerateResponseChunkData[] = [];
+  const parser = format?.handler(rawRequest.output?.jsonSchema).parseChunk;
+
   if (middleware && middleware.length > 0) {
     const dispatchGenerate = async (
       index: number,
@@ -294,6 +299,7 @@ async function generateActionImpl(
           abortSignal: ctx.abortSignal,
           streamingCallback: ctx.onChunk,
           context: ctx.context,
+          sharedPreviousChunks,
         });
       }
       const currentMiddleware = middleware[index];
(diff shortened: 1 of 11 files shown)


pavelgj:
@apascal07 @huangjeff5 this is something to consider for Go/py

js/ai/src/generate/action.ts
```diff
@@ -296,14 +298,50 @@ async function generateActionImpl(
...
+        const wrappedOnChunk = ctx.onChunk
+          ? (c: GenerateResponseChunk | GenerateResponseChunkData) => {
+              if (c instanceof GenerateResponseChunk) {
+                ctx.onChunk!(c);
+              } else {
+                ctx.onChunk!(
+                  new GenerateResponseChunk(c, {
+                    index: c.index !== undefined ? c.index : messageIndex,
+                    role: c.role !== undefined ? c.role : 'model',
+                    previousChunks: [],
```
apascal07: `previousChunks` is always empty? Is that right?

pavelgj: yeah.. I'm not quite sure what's the right thing to do here... this is the case when middlware emits raw/extra chunks. Should they get appended to the accumulated chunks that are fed to the format parser? accumulated chunks are for model output...

apascal07 (approved):

js/ai/src/generate/action.ts
pavelgj: Ok, I think we should be accumulating them... so done.

js/ai/src/generate/action.ts
```diff
@@ -277,39 +277,86 @@ async function generateActionImpl(
...
-          dispatchGenerate(index + 1, modifiedReq || req, opts || ctx)
+        const wrappedOnChunk = ctx.onChunk
+          ? (c: GenerateResponseChunk | GenerateResponseChunkData) => {
+              if (c instanceof GenerateResponseChunk) {
+                ctx.onChunk!(c);
+              } else {
+                const chunk = new GenerateResponseChunk(c, {
+                  index: c.index !== undefined ? c.index : messageIndex,
+                  role: c.role !== undefined ? c.role : 'model',
+                  previousChunks: sharedPreviousChunks,
```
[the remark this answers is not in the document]
pavelgj: done

PR #5157, opened 2026-04-20
refactor(js/core): restrict automatic shutdown signal handlers to development environment only

pavelgj:


MichaelDoyle (approved):
Awesome, thanks!

I *think* there are some handlers like this either in the `google-cloud` plugin or genkit instrumentation code to flush metrics as well, that we might want to take a second look at.

PR #5127, opened 2026-04-15
fix(go): remove gemini-3-pro-preview and claude-3-haiku@20240307

andyma-star:
## Summary
- Remove deprecated model references in Go plugins:
  - `gemini-3-pro-preview` (GoogleAI)
  - `claude-3-haiku@20240307` (Vertex Model Garden)
- Update the affected Go sample and live test to use active model IDs without changing existing workflow logic.
- Keep docs aligned by removing deprecated model mention from the Go Google GenAI README.

## Changes
- `go/plugins/googlegenai/models.go`
  - Filter out deprecated dynamically listed models (`gemini-3-pro-preview`).
- `go/plugins/vertexai/modelgarden/models.go`
  - Remove deprecated Anthropic model entry (`claude-3-haiku@20240307`).
- `go/plugins/googlegenai/googleai_live_test.go`
  - Replace deprecated model usage with `gemini-2.5-pro`.
- `go/samples/multipart-tools/main.go`
  - Replace deprecated model usage with `googleai/gemini-2.5-pro`.
- `go/plugins/googlegenai/README.md`
  - Remove deprecated Gemini model mention.

## Issues
- Follows issue: https://github.com/invertase/genkit-private/issues/101

Checklist (if applicable):
- [X] PR title is following https://www.conventionalcommits.org/en/v1.0.0/
- [X] Tested (manually, unit tested, etc.)
- [X] Docs updated (updated docs or a docs bug required)

PR #5173, opened 2026-04-23
chore(go): deprecate gemini-3-pro-preview and claude-3-haiku@20240307

cabljac:
## Summary

Deprecates two Go P0 models from the parity tracker, treating each according to its shutdown status:

Supersedes https://github.com/genkit-ai/genkit/pull/5127

- **`gemini-3-pro-preview`** (GoogleAI) — shutdown reached 2026-03-09. Removed from samples, the live test, README, and filtered out of dynamic discovery as a defensive guard in case the provider list API still returns it.
- **`claude-3-haiku@20240307`** (Vertex Model Garden) — shutdown 2026-08-23. Rather than hard-removing ~4 months before the provider shutdown (which would silently break existing users), the entry is kept in `AnthropicModels` with `Stage: ai.ModelStageDeprecated`. The existing middleware at `go/ai/model_middleware.go:259` already emits a warn log on generate for deprecated models, so users get a migration window. A follow-up after Aug 23 can hard-remove.

This is the first use of `ai.ModelStageDeprecated` in a Go plugin; the constant and middleware have existed since the `ModelStage` enum was introduced but have never been wired up by a plugin. It follows the same Stage-field pattern already used for `ModelStageStable` / `ModelStageUnstable` elsewhere in `googlegenai/models.go` and other plugins.

### Latent bug fix

`go/plugins/internal/anthropic/anthropic.go` was reconstructing `ai.ModelOptions` in `DefineModel` but dropping `info.Stage`. Any Anthropic model with `Stage` set would have that field silently discarded before registration, preventing the middleware warning from firing. Propagates `Stage` through.

### Relation to existing work

Supersedes #5127 — same scope, but keeps haiku callable-with-warning through its shutdown window instead of immediately returning nil, and adds the missing Stage propagation.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet` clean on changed packages
- [x] `go test ./plugins/googlegenai/` — new `TestIsDeprecatedGenAIModel` passes (4 cases)
- [x] `go test ./plugins/internal/anthropic/ ./plugins/vertexai/modelgarden/` pass
- [x] `go test ./ai/...` pass (middleware validation still happy)
- [ ] Live smoke: run `go/samples/multipart-tools/main.go` with a real API key to confirm the sample still works on `gemini-2.5-pro`
- [ ] Live smoke: call `ai.Generate` with `claude-3-haiku@20240307` and confirm the `"model is deprecated and may be removed in a future release"` warn log fires

### DIFF
--- a/go/plugins/googlegenai/models.go
+++ b/go/plugins/googlegenai/models.go
@@ -107,6 +107,13 @@ const (
 )
 
 var (
+	// deprecatedGenAIModels are filtered out of dynamic discovery so they
+	// are not registered as callable actions. This guards against the provider
+	// still returning a post-shutdown model from the list API.
+	deprecatedGenAIModels = map[string]struct{}{
+		"gemini-3-pro-preview": {},
+	}
+
 	// eventually, Vertex AI and Google AI models will match, in the meantime,
 	// keep them sepparated
 	vertexAIModels = []string{
@@ -373,6 +380,9 @@ func listGenaiModels(ctx context.Context, client *genai.Client) (genaiModels, er
 
 		name := strings.TrimPrefix(item.Name, "publishers/google/")
 		name = strings.TrimPrefix(name, "models/")
+		if isDeprecatedGenAIModel(name) {
+			continue
+		}
 
 		// The Vertex AI backend does not populate SupportedActions,
 		// so we fall back to name-based categorization.
@@ -398,4 +408,9 @@ func listGenaiModels(ctx context.Context, client *genai.Client) (genaiModels, er
 	}
 
 	return models, nil
+}
+
+func isDeprecatedGenAIModel(name string) bool {
+	_, ok := deprecatedGenAIModels[name]
+	return ok
 }
--- a/go/plugins/googlegenai/models.go
+++ b/go/plugins/googlegenai/models.go
@@ -380,32 +380,39 @@ func listGenaiModels(ctx context.Context, client *genai.Client) (genaiModels, er
 
 		name := strings.TrimPrefix(item.Name, "publishers/google/")
 		name = strings.TrimPrefix(name, "models/")
-		if _, ok := deprecatedGenAIModels[name]; ok {
-			continue
-		}
-
-		// The Vertex AI backend does not populate SupportedActions,
-		// so we fall back to name-based categorization.
-		if slices.Contains(item.SupportedActions, "embedContent") || strings.Contains(name, "embed") {
-			models.embedders = append(models.embedders, name)
-			continue
-		}
-
-		if strings.Contains(name, "imagen") {
-			models.imagen = append(models.imagen, name)
-			continue
-		}
-
-		if strings.Contains(name, "veo") {
-			models.veo = append(models.veo, name)
-			continue
-		}
-
-		// Only include models with known generative prefixes.
-		if strings.HasPrefix(name, "gemini") || strings.HasPrefix(name, "gemma") {
-			models.gemini = append(models.gemini, name)
-		}
+		categorizeModel(&models, name, item.SupportedActions)
 	}
 
 	return models, nil
+}
+
+// categorizeModel routes a discovered model name into the appropriate bucket
+// of m. It returns true if the name was placed and false if the model was
+// filtered out or did not match any known prefix.
+//
+// supportedActions is the list of API actions advertised by the provider for
+// this model. The Vertex AI backend does not populate it, so name-based
+// matching is used as a fallback.
+func categorizeModel(m *genaiModels, name string, supportedActions []string) bool {
+	if _, ok := deprecatedGenAIModels[name]; ok {
+		return false
+	}
+
+	if slices.Contains(supportedActions, "embedContent") || strings.Contains(name, "embed") {
+		m.embedders = append(m.embedders, name)
+		return true
+	}
+	if strings.Contains(name, "imagen") {
+		m.imagen = append(m.imagen, name)
+		return true
+	}
+	if strings.Contains(name, "veo") {
+		m.veo = append(m.veo, name)
+		return true
+	}
+	if strings.HasPrefix(name, "gemini") || strings.HasPrefix(name, "gemma") {
+		m.gemini = append(m.gemini, name)
+		return true
+	}
+	return false
 }
--- a/go/plugins/googlegenai/models.go
+++ b/go/plugins/googlegenai/models.go
@@ -380,7 +380,7 @@ func listGenaiModels(ctx context.Context, client *genai.Client) (genaiModels, er
 
 		name := strings.TrimPrefix(item.Name, "publishers/google/")
 		name = strings.TrimPrefix(name, "models/")
-		if isDeprecatedGenAIModel(name) {
+		if _, ok := deprecatedGenAIModels[name]; ok {
 			continue
 		}
 
@@ -408,9 +408,4 @@ func listGenaiModels(ctx context.Context, client *genai.Client) (genaiModels, er
 	}
 
 	return models, nil
-}
-
-func isDeprecatedGenAIModel(name string) bool {
-	_, ok := deprecatedGenAIModels[name]
-	return ok
 }
--- a/go/plugins/internal/anthropic/anthropic.go
+++ b/go/plugins/internal/anthropic/anthropic.go
@@ -55,6 +55,7 @@ func DefineModel(client anthropic.Client, provider, name string, info ai.ModelOp
 		Supports:     info.Supports,
 		Versions:     info.Versions,
 		ConfigSchema: configSchema,
+		Stage:        info.Stage,
 	}
 
 	return ai.NewModel(api.NewName(provider, name), meta, func(
--- a/go/plugins/googlegenai/README.md
+++ b/go/plugins/googlegenai/README.md
@@ -83,7 +83,7 @@ Genkit automatically discovers available models supported by the [Go GenAI SDK](
 
 Commonly used models include:
 
-- **Gemini Series**: `gemini-3-pro-preview`, `gemini-3-flash-preview`, `gemini-2.5-flash`, `gemini-2.5-pro`
+- **Gemini Series**: `gemini-3-flash-preview`, `gemini-2.5-flash`, `gemini-2.5-pro`
 - **Imagen Series**: `imagen-3.0-generate-001`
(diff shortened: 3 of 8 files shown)


go/plugins/googlegenai/models.go
```diff
@@ -373,6 +380,9 @@ func listGenaiModels(ctx context.Context, client *genai.Client) (genaiModels, er
 
 		name := strings.TrimPrefix(item.Name, "publishers/google/")
 		name = strings.TrimPrefix(name, "models/")
+		if isDeprecatedGenAIModel(name) {
```
adesinah: We could get rid of the `isDeprecatedGenAIModel` function and resort to a one-line check:
`if _, ok := deprecatedGenAIModels[name]; ok { continue } `

go/plugins/googlegenai/models.go
adesinah, line 373: We should add a test for `listGenaiModels` to ensure the filter takes effect.

go/plugins/googlegenai/models.go
```diff
@@ -107,6 +107,13 @@ const (
 )
 
 var (
+	// deprecatedGenAIModels are filtered out of dynamic discovery so they
+	// are not registered as callable actions. This guards against the provider
+	// still returning a post-shutdown model from the list API.
+	deprecatedGenAIModels = map[string]struct{}{
+		"gemini-3-pro-preview": {},
```
adesinah, line 114: I have a bit of concern around this: If the API ever returns `gemini-3-pro-preview-002` or `gemini-3-pro-preview-latest` which are common pattern, the filter misses them. If I recall, the old gemini-pro family had `-001`, `-002`. This is worth investigating, in my opinion.

go/plugins/internal/anthropic/anthropic.go
adesinah, line 41: I suggest we have a direct test calling ```
DefineModel(..., ai.ModelOptions{Stage:
  ai.ModelStageDeprecated})
``` and asserting the registered model's meta exposes `Stage: ModelStageDeprecated`.

**Review by adesinah — CHANGES_REQUESTED** [review]
Changes look good. I added a few comments. 
In addition, I think the model deprecation logic needs to be standardized across providers. Using Anthropic as an example, the deprecation of the model acts this way:
This middleware runs around every Generate call, so:
  - The model still resolves and executes (no early return).
  - A WARN log fires every call with the model name attached.
  - The user can keep their code running while they migrate — the warning is the migration window.

Contrast with the Gemini path — gemini-3-pro-preview is filtered at registration time (listGenaiModels), so genkit.LookupModel("googleai", "gemini-3-pro-preview") returns nil and Generate errors out with a "model not registered" before any middleware can warn.

File: go/plugins/googlegenai/models.go (hunk 1 above)
Comment (cabljac) [reply go/plugins/googlegenai/models.go]: Resolved in 8e8bac994.

File: go/plugins/googlegenai/models.go
Comment (cabljac) [reply go/plugins/googlegenai/models.go:373]: want to clarify here before acting - the package has two test patterns (live, unit) - do you think a live test would be fine? not executed in CI , gated via t.Skip on missing `GEMINI_API_KEY`

A unit test we'd have to extract the loop here maybe so we can construct fixtures, but i feel like adding some test-only signature is something i'd usually avoid

File: go/plugins/internal/anthropic/anthropic.go
Comment (cabljac) [reply go/plugins/internal/anthropic/anthropic.go:41]: Resolved in e104143fe — added `TestDefineModel_PropagatesStage` table test covering `Deprecated`, `Unstable`, and unset. Reads the registered action's metadata via the embedded `ActionDef.Desc()` and asserts `metadata["model"]["stage"]` round-trips.

File: go/plugins/googlegenai/models.go (hunk 2 above)
Comment (cabljac) [reply go/plugins/googlegenai/models.go:114]: If these ever appeared it would likely be a re-launch, so im not sure we should mark it as deprecated explicitely? dropping it from discovery in that case would mean more work to add it back in?

Comment (adesinah) [reply go/plugins/googlegenai/models.go:114]: Yes, I agree. It's good that we are aware of that possible situation. 
I am ok with marking this point as resolved.

File: go/plugins/googlegenai/models.go
Comment (adesinah) [reply go/plugins/googlegenai/models.go:373]: Fair point on the test-only signature concern — agreed, I'd avoid that.
 
One catch on the live-test angle: the filter's purpose (per the comment of the function) is to catch the case where the API still returns the dead model. I believe a live test will pass whether the filter exists or not as long as the API stops returning the name, so it can't actually distinguish "filter works" from "API moved on."
 
I suggest this: extract the for-loop body into a pure `categorizeModel(m *genaiModels, name string, supportedActions []string) bool`. That's a real function, not a test seam, and the unit test becomes a one-pager that doesn't touch the SDK.  Example below:

[ I use AI to generate the code below ]
1. Extract the categorizer

  In `models.go`, replace the existing for-loop with a call to a new pure function:

  ```
// categorizeModel routes a discovered model name into the appropriate bucket
  // of m. It returns true if the name was placed and false if the model was
  // filtered out or did not match any known prefix.
  //
  // supportedActions is the list of API actions advertised by the provider for
  // this model. The Vertex AI backend does not populate it, so name-based
  // matching is used as a fallback.
  func categorizeModel(m *genaiModels, name string, supportedActions []string) bool {
        if isDeprecatedGenAIModel(name) {
                return false
        }

        if slices.Contains(supportedActions, "embedContent") || strings.Contains(name, "embed") {
                m.embedders = append(m.embedders, name)
                return true
        }
        if strings.Contains(name, "imagen") {
                m.imagen = append(m.imagen, name)
                return true
        }
        if strings.Contains(name, "veo") {
                m.veo = append(m.veo, name)
                return true
        }
        if strings.HasPrefix(name, "gemini") || strings.HasPrefix(name, "gemma") {
                m.gemini = append(m.gemini, name)
                return true
        }
        return false
  }
```

 ```
 func listGenaiModels(ctx context.Context, client *genai.Client) (genaiModels, error) {
        models := genaiModels{}

        for item, err := range client.Models.All(ctx) {
                if err != nil {
                        return genaiModels{}, fmt.Errorf("failed to list models: %w", err)
                }

                name := strings.TrimPrefix(item.Name, "publishers/google/")
                name = strings.TrimPrefix(name, "models/")
                categorizeModel(&models, name, item.SupportedActions)
        }

        return models, nil
  }
```


  2. Add the unit test

  Extend `models_test.go`:

  ```
package googlegenai

  import (
        "slices"
        "testing"
  )

  func TestCategorizeModel(t *testing.T) {
        tests := []struct {
                name             string
                modelName        string
                supportedActions []string
                wantPlaced       bool
                // pickBucket returns the slice that should have received modelName.
                // nil means "no bucket should have received it."
                pickBucket func(*genaiModels) []string
        }{
                {
                        name:       "deprecated model is filtered out",
                        modelName:  "gemini-3-pro-preview",
                        wantPlaced: false,
                        pickBucket: nil,
                },
                {
                        name:             "embedder via supportedActions",
                        modelName:        "text-embedding-004",
                        supportedActions: []string{"embedContent"},
                        wantPlaced:       true,
                        pickBucket:       func(m *genaiModels) []string { return m.embedders },
                },
                {
                        name:       "embedder via name fallback",
                        modelName:  "text-embedding-005",
                        wantPlaced: true,
                        pickBucket: func(m *genaiModels) []string { return m.embedders },
                },
                {
                        name:       "imagen routed to imagen bucket",
                        modelName:  "imagen-3.0-generate-001",
                        wantPlaced: true,
                        pickBucket: func(m *genaiModels) []string { return m.imagen },
                },
                {
                        name:       "veo routed to veo bucket",
                        modelName:  "veo-3.0-generate-001",
                        wantPlaced: true,
                        pickBucket: func(m *genaiModels) []string { return m.veo },
                },
                {
                        name:       "gemini routed to gemini bucket",
                        modelName:  "gemini-2.5-pro",
                        wantPlaced: true,
                        pickBucket: func(m *genaiModels) []string { return m.gemini },
                },
                {
                        name:       "gemma routed to gemini bucket",
                        modelName:  "gemma-2",
                        wantPlaced: true,
                        pickBucket: func(m *genaiModels) []string { return m.gemini },
                },
                {
                        name:       "unknown prefix is skipped",
                        modelName:  "totally-unrelated-model",
                        wantPlaced: false,
                        pickBucket: nil,
                },
        }

        for _, tc := range tests {
                t.Run(tc.name, func(t *testing.T) {
                        var m genaiModels
                        got := categorizeModel(&m, tc.modelName, tc.supportedActions)
                        if got != tc.wantPlaced {
                                t.Fatalf("categorizeModel placed=%v, want %v", got, tc.wantPlaced)
                        }
                        if tc.pickBucket == nil {
                                return
                        }
                        bucket := tc.pickBucket(&m)
                        if !slices.Contains(bucket, tc.modelName) {
                                t.Errorf("expected %q in bucket, got %v", tc.modelName, bucket)
                        }
                })
        }
  }

  func TestListGenaiModelsFiltersDeprecated(t *testing.T) {
        // Drives the same categorization the live loop would, without touching
        // the SDK. Proves a deprecated name dropped at categorization time never
        // reaches any bucket — which is the contract listGenaiModels relies on.
        var m genaiModels
        categorizeModel(&m, "gemini-3-pro-preview", nil)
        categorizeModel(&m, "gemini-2.5-pro", nil)

        if slices.Contains(m.gemini, "gemini-3-pro-preview") {
                t.Errorf("deprecated model leaked into gemini bucket: %v", m.gemini)
        }
        if !slices.Contains(m.gemini, "gemini-2.5-pro") {
                t.Errorf("expected gemini-2.5-pro in bucket, got %v", m.gemini)
        }
  }
```

cabljac: Good catch on the live test — agreed, it can't distinguish "filter works" from "API moved on." Went with your refactor proposal: extracted `categorizeModel(m *genaiModels, name string, supportedActions []string) bool` in 72a8a47c8. The loop in `listGenaiModels` is now thin; tests cover each bucket + the deprecated filter + an unknown-prefix case, plus a focused `TestListGenaiModelsFiltersDeprecated` that asserts the filter keeps `gemini-3-pro-preview` out.

Inlined the deprecated-map lookup directly in `categorizeModel` rather than re-adding `isDeprecatedGenAIModel` — keeps consistent with the earlier comment about avoiding a one-liner helper.

adesinah (approved):
Looks pretty good to me!

PR #5177, opened 2026-04-23
feat(go/plugins/googlegenai): Support Virtual Try-On 001

cabljac:
Adds `virtual-try-on-001` image-editing model to the googlegenai plugin.

Introduces `ModelTypeVirtualTryOn`, a `VirtualTryOnConfig` struct mirroring the JS `ImagenTryOnConfigSchema`, and a `generateVirtualTryOn` function that calls the Vertex AI `:predict` endpoint directly. Person and product images are identified by a `type` key in `ai.Part.Metadata` (exposed as `PartMetadataTypePersonImage` / `PartMetadataTypeProductImage`), matching the JS plugin convention. Media parts are parsed via the shared `plugins/internal/uri.Data` helper that `veo.go` already uses.

## Testing
image
image
image
image

## Update: migrated to the genai SDK's RecontextImage

The first pass called the Vertex `:predict` endpoint directly, since the Go SDK had no matching method. It does now, so generation goes through `client.Models.RecontextImage` and the hand-rolled request/response structs, URL assembly and raw HTTP post are gone (~120 lines).

The model's config type changed with it, from the bespoke `VirtualTryOnConfig` to `genai.RecontextImageConfig`:

| Before (`VirtualTryOnConfig`) | Now (`genai.RecontextImageConfig`) |
| --- | --- |
| `sampleCount` | `numberOfImages` |
| `seed`, `baseSteps`, `personGeneration` | unchanged |
| `safetySetting` | `safetyFilterLevel` |
| `storageUri` | `outputGcsUri` |
| `outputOptions.{mimeType,compressionQuality}` | `outputMimeType`, `outputCompressionQuality` |
| `location` | dropped — taken from the client |
| — | gains `addWatermark`, `enhancePrompt`, `labels`, `httpOptions` |

**Breaking:** `googlegenai.VirtualTryOnConfig` and `VirtualTryOnOutputOpts` are removed. Nothing outside this PR referenced them.

Two things follow from using the SDK type. The dev UI config editor now renders descriptions for every try-on field (via a `ricOverrides` entry, like Gemini, Imagen and Veo), and errors carry a status — `InvalidArgument` for a missing or unreadable image part, `Unimplemented` for streaming or a non-Vertex backend, and provider errors through `wrapAPIError`.

Response handling was hardened alongside: `translateImagenCandidates` assumed every candidate carries image bytes, but an RAI-filtered candidate has none (an empty `Image` on Vertex, nil on Gemini) and `outputGcsUri` returns a URI. Those cases used to panic or emit an empty media part; filter reasons are now reported, and block the response when nothing survives. This affects Imagen too, not just try-on.

Live coverage: the try-on subtests exercise inline and `gs://` inputs, and assert that `numberOfImages` and `outputMimeType` reach the API — all four pass against Vertex.


cabljac:
addressed both pieces of feedback in the latest push:

1. **Media + data parts (partially adopted).** Accepted `IsData` alongside `IsMedia` — matches `gemini.go:543-553` and `tools.go:57-63`. I skipped the `image/` content-type prefix check though: the helper also passes `gs://` URIs through, and callers often don't set `ContentType` on those. A strict prefix filter would silently drop valid inputs. The JS parity path at `js/plugins/google-genai/src/vertexai/converters.ts` doesn't validate content type either, and Vertex returns a clear error for non-image payloads.

2. **Blocked response (adopted).** Switched from a plain error to `ai.FinishReasonBlocked` with `FinishMessage`, mirroring `veo.go:299` and `compat_oai/generate.go:393`.

go/plugins/googlegenai/virtual_try_on.go
```diff
@@ -0,0 +1,277 @@
...
+// Uses the shared uri.Data helper (the same one veo.go uses) so the parsing
+// logic for data:<mime>[;base64],<data> URIs is consistent across plugins.
+// gs:// URIs are passed through to the API as-is.
+func extractMediaByType(input *ai.ModelRequest, typ string) []virtualTryOnImage {
+	var out []virtualTryOnImage
+	for _, msg := range input.Messages {
+		for _, p := range msg.Content {
+			if !p.IsMedia() {
+				continue
+			}
```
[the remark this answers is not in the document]
cabljac: Already handled — `extractMediaByType` at [virtual_try_on.go:129](https://github.com/firebase/genkit/blob/feat/go-googlegenai-virtual-try-on/go/plugins/googlegenai/virtual_try_on.go#L129) continues on `!p.IsMedia() && !p.IsData()`, so both `PartMedia` and `PartData` are accepted. The `metadata["type"]` filter (person_image / product_image) already scopes the parts to VTO-specific media — a non-image part would never have those metadata keys in practice.

go/plugins/googlegenai/virtual_try_on.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil, fmt.Errorf("virtual try-on: %s: %s", resp.Status, string(data))
+	}
+
+	var vr virtualTryOnPredictResponse
+	if err := json.Unmarshal(data, &vr); err != nil {
+		return nil, fmt.Errorf("virtual try-on: unmarshaling response: %w", err)
+	}
+	if len(vr.Predictions) == 0 {
+		return nil, fmt.Errorf("virtual try-on: no predictions returned (possibly content-filtered)")
+	}
```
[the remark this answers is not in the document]
cabljac: Returning `FinishReasonBlocked` with an empty response is the intended idiom here — the caller can distinguish "model declined to produce output" from a hard error, matching how the Imagen / Gemini paths in this plugin handle the same situation. Returning an error would force every caller to handle it as a failure mode when it often isn't (safety filters, unsupported inputs, etc.). See the current block at [virtual_try_on.go:272–282](https://github.com/firebase/genkit/blob/feat/go-googlegenai-virtual-try-on/go/plugins/googlegenai/virtual_try_on.go#L272-L282).

go/plugins/googlegenai/virtual_try_on.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+	}
+	return out
+}
+
+func toVirtualTryOnRequest(input *ai.ModelRequest, cfg *VirtualTryOnConfig) (*virtualTryOnPredictRequest, error) {
+	persons := extractMediaByType(input, PartMetadataTypePersonImage)
+	products := extractMediaByType(input, PartMetadataTypeProductImage)
+	if len(persons) == 0 {
+		return nil, fmt.Errorf("virtual try-on requires a media part with metadata.type=%q", PartMetadataTypePersonImage)
```
adesinah: Should be wrapped in NewPublicError() to preserve Genkit error classes. We should do this for other places in this function. Example:
```go
return nil, NewPublicError(core.INVALID_ARGUMENT,
      fmt.Sprintf("virtual try-on requires a media part with metadata.type=%q", PartMetadataTypePersonImage), nil)
```

go/plugins/googlegenai/virtual_try_on.go
```diff
@@ -0,0 +1,279 @@
...
+	if cc.Project == "" {
+		return nil, fmt.Errorf("virtual try-on requires a Vertex AI project id")
+	}
+
+	body, err := json.Marshal(payload)
+	if err != nil {
+		return nil, fmt.Errorf("virtual try-on: marshaling request: %w", err)
+	}
+
+	url := fmt.Sprintf(
```
adesinah: If a user configures `&googlegenai.VertexAI{ProjectID: "...", Location: "global"}`, that would break VTO call breaks for them. Let's extract a shared helper since two implementations now need it:

```go
  // vertexPredictURL builds the Vertex AI `:predict` URL for a publisher model,
  // handling the "global" location which uses the bare aiplatform host instead
  // of the regional <location>-aiplatform variant.
  func vertexPredictURL(apiVersion, location, project, model string) string {
      host := "aiplatform.googleapis.com"
      if location != "global" {
          host = location + "-" + host
      }
      return fmt.Sprintf(
          "https://%s/%s/projects/%s/locations/%s/publishers/google/models/%s:predict",
          host, apiVersion, project, location, model,
      )
  }
```

With a unit test:
```
func TestVertexPredictURL(t *testing.T) {
      tests := []struct {
          location string
          want     string
      }{
          {"us-central1",
  "https://us-central1-aiplatform.googleapis.com/v1/projects/p/locations/us-central1/publishers/google/models/m:predict"},
          {"global", "https://aiplatform.googleapis.com/v1/projects/p/locations/global/publishers/google/models/m:predict"},
      }
      for _, tc := range tests {
          if got := vertexPredictURL("v1", tc.location, "p", "m"); got != tc.want {
              t.Errorf("location %q: got %s, want %s", tc.location, got, tc.want)
          }
      }
  }
```


  Then both call sites become one-liners:
  // virtual_try_on.go
`  url := vertexPredictURL("v1", location, cc.Project, model)`

go/plugins/googlegenai/models.go
```diff
@@ -52,6 +52,17 @@ var (
...
 		LongRunning: true,
 	}
+
+	// VirtualTryOnSupports describes model capabilities for the virtual try-on
+	// image editing models. Unlike regular Imagen models these accept a system
+	// role (used to carry the person/product image part metadata).
+	VirtualTryOnSupports = ai.ModelSupports{
+		Multiturn:  false,
+		Tools:      false,
+		SystemRole: true,
```
adesinah: Is there any benefit to setting this "SystemRole: true", and not set to false Like done in Imagen.

adesinah (changes requested):
Looks pretty good. I left a few comments. Also, we should consider the following:
  Everything testable without the API is currently uncovered:
  - ClassifyModel("virtual-try-on-001") == ModelTypeVirtualTryOn
  - virtualTryOnConfigFromRequest across its four type cases + the default
  - extractMediaByType — typed filtering, gs:// vs data:, malformed URIs
  - toVirtualTryOnRequest error paths (no person, no product, both)
  - URL construction (regional vs global)
  - translateVirtualTryOnResponse mapping
  - Non-Vertex backend rejection in generateVirtualTryOn

go/plugins/googlegenai/virtual_try_on.go
cabljac: Done across both commits.

`78bf815d9` shipped the global-host fix first as a focused commit; `68c33fce7` then wrapped all 7 error returns and added unit tests covering the new error classes.

Classes used:
- `INVALID_ARGUMENT` for the 6 caller-fixable paths: missing person/product parts, streaming mode, non-Vertex backend, empty location, empty project.
- `FAILED_PRECONDITION` for the `cc.HTTPClient == nil` invariant — matches the `streaming.go` precedent for "system state not right for operation" rather than caller input.

Test assertions check the class via `errors.As(err, *core.UserFacingError)` so future drift would fail the build.

go/plugins/googlegenai/virtual_try_on.go
cabljac: Done in two parts.

`78bf815d9` is the minimal inline fix: bare `aiplatform.googleapis.com` for `global`, `<region>-aiplatform.googleapis.com` otherwise. Matches JS plugin's `getUrl()` at `client.ts:271-275`.

`68c33fce7` extracts the URL construction into a `vertexPredictURL(location, project, model)` helper colocated in `virtual_try_on.go` (only 1 caller in Go googlegenai today — grepped to confirm). Mostly there to make the global vs regional branch testable; happy to lift it into a shared package the moment a second publisher-model caller lands.

go/plugins/googlegenai/models.go
cabljac: Flipped to `false` in 08e681d95.

You were right — `extractMediaByType` walks `input.Messages` regardless of `msg.Role`, so VTO has no functional need for SystemRole=true. The old comment claiming it was needed for image metadata didn't hold up. Now matches Imagen/Veo.

adesinah (approved):
Looks good to me!

apascal07:
Can you rebase on main and I'll take a look after?

PR #5178, opened 2026-04-23
feat(go/plugins/googlegenai): Support tuned Gemini endpoint models

cabljac:
Accepts Vertex AI tuned Gemini endpoints addressed either by the short form `endpoints/ID` or the full resource path `projects/PROJECT/locations/LOCATION/endpoints/ID`.

`ClassifyModel` now routes these names to `ModelTypeGemini`, and a new `resolveVertexModelName` helper expands the short form to the full path before the `google.golang.org/genai` SDK transforms it — avoiding the SDK's default `publishers/google/models/` prefix, which only applies to first-party models. `VertexAI.DefineModel` also accepts tuned endpoint names with a `nil` options argument and falls back to the default Gemini capability set.

Mirrors `js/plugins/google-genai/src/vertexai/client.ts`.

## Testing
image

### DIFF
--- a/go/plugins/googlegenai/gemini.go
+++ b/go/plugins/googlegenai/gemini.go
@@ -145,6 +145,26 @@ func newModel(client *genai.Client, name string, opts ai.ModelOptions) ai.Model 
 	return ai.NewModel(api.NewName(provider, name), meta, fn)
 }
 
+// resolveVertexModelName prepares a model name for the google.golang.org/genai
+// SDK. The SDK transforms most names into `publishers/google/models/NAME`,
+// which is wrong for tuned endpoints. For a short-form tuned endpoint name
+// (`endpoints/ID`), this expands it to the full resource path
+// `projects/PROJECT/locations/LOCATION/endpoints/ID` using the client's
+// configured project and location. Other names are returned unchanged.
+func resolveVertexModelName(client *genai.Client, name string) string {
+	if !isTunedGeminiName(name) {
+		return name
+	}
+	if strings.HasPrefix(name, "projects/") {
+		return name
+	}
+	cc := client.ClientConfig()
+	if cc.Backend != genai.BackendVertexAI || cc.Project == "" || cc.Location == "" {
+		return name
+	}
+	return fmt.Sprintf("projects/%s/locations/%s/%s", cc.Project, cc.Location, name)
+}
+
 // generate requests generate call to the specified model with the provided
 // configuration.
 func generate(
@@ -157,6 +177,7 @@ func generate(
 	if model == "" {
 		return nil, errors.New("model not provided")
 	}
+	model = resolveVertexModelName(client, model)
 
 	cache, err := handleCache(ctx, client, input, model)
 	if err != nil {
--- a/go/plugins/googlegenai/gemini.go
+++ b/go/plugins/googlegenai/gemini.go
@@ -146,6 +146,26 @@ func newModel(client *genai.Client, name string, opts ai.ModelOptions) ai.Model 
 	return ai.NewModel(api.NewName(provider, name), meta, fn)
 }
 
+// resolveVertexModelName prepares a model name for the google.golang.org/genai
+// SDK. The SDK transforms most names into `publishers/google/models/NAME`,
+// which is wrong for tuned endpoints. For a short-form tuned endpoint name
+// (`endpoints/ID`), this expands it to the full resource path
+// `projects/PROJECT/locations/LOCATION/endpoints/ID` using the client's
+// configured project and location. Other names are returned unchanged.
+func resolveVertexModelName(client *genai.Client, name string) string {
+	if !isTunedGeminiName(name) {
+		return name
+	}
+	if strings.HasPrefix(name, "projects/") {
+		return name
+	}
+	cc := client.ClientConfig()
+	if cc.Backend != genai.BackendVertexAI || cc.Project == "" || cc.Location == "" {
+		return name
+	}
+	return fmt.Sprintf("projects/%s/locations/%s/%s", cc.Project, cc.Location, name)
+}
+
 // generate requests generate call to the specified model with the provided
 // configuration.
 func generate(
@@ -158,6 +178,7 @@ func generate(
 	if model == "" {
 		return nil, errors.New("model not provided")
 	}
+	model = resolveVertexModelName(client, model)
 
 	cache, err := handleCache(ctx, client, input, model)
 	if err != nil {
--- a/go/plugins/googlegenai/googlegenai.go
+++ b/go/plugins/googlegenai/googlegenai.go
@@ -207,24 +207,33 @@ func (ga *GoogleAI) DefineModel(g *genkit.Genkit, name string, opts *ai.ModelOpt
 // The second argument describes the capability of the model.
 // Use [IsDefinedModel] to determine if a model is already defined.
 // After [Init] is called, only the known models are defined.
+//
+// Tuned Gemini endpoints are accepted in either the short form
+// `endpoints/ID` or the full resource path
+// `projects/PROJECT/locations/LOCATION/endpoints/ID`. When opts is nil the
+// caller gets the default Gemini capability set.
 func (v *VertexAI) DefineModel(g *genkit.Genkit, name string, opts *ai.ModelOptions) (ai.Model, error) {
 	v.mu.Lock()
 	defer v.mu.Unlock()
 	if !v.initted {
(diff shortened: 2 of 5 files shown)


go/plugins/googlegenai/googlegenai.go
```diff
@@ -207,24 +207,33 @@ func (ga *GoogleAI) DefineModel(g *genkit.Genkit, name string, opts *ai.ModelOpt
...
-	}
 
 	if opts == nil {
-		var ok bool
-		modelOpts, ok := models[name]
-		if !ok {
-			return nil, fmt.Errorf("VertexAI.DefineModel: called with unknown model %q and nil ModelOptions", name)
+		if isTunedGeminiName(name) {
+			defaults := GetModelOptions(name, vertexAIProvider)
+			opts = &defaults
```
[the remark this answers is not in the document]
cabljac: Traced this and the flow is fine — tuned endpoints get the full Gemini Supports, not nil. Path:

1. `isTunedGeminiName("endpoints/ID")` returns true ([model_type.go:52](https://github.com/firebase/genkit/blob/feat/go-googlegenai-tuned-gemini/go/plugins/googlegenai/model_type.go#L52)).
2. `ClassifyModel` returns `ModelTypeGemini` via the tuned fall-through at [model_type.go:38–42](https://github.com/firebase/genkit/blob/feat/go-googlegenai-tuned-gemini/go/plugins/googlegenai/model_type.go#L38-L42).
3. `GetModelOptions` ([models.go:275–280](https://github.com/firebase/genkit/blob/feat/go-googlegenai-tuned-gemini/go/plugins/googlegenai/models.go#L275-L280)) doesn't find the tuned name in `supportedGeminiModels`, so it falls back to `defaultGeminiOpts`, which has full `Supports` (Multimodal).

No nil Supports at the call site. `go build` / `go vet` clean and CI is green.

go/plugins/googlegenai/gemini.go
```diff
@@ -145,6 +145,26 @@ func newModel(client *genai.Client, name string, opts ai.ModelOptions) ai.Model
 	return ai.NewModel(api.NewName(provider, name), meta, fn)
 }
 
+// resolveVertexModelName prepares a model name for the google.golang.org/genai
+// SDK. The SDK transforms most names into `publishers/google/models/NAME`,
+// which is wrong for tuned endpoints. For a short-form tuned endpoint name
+// (`endpoints/ID`), this expands it to the full resource path
+// `projects/PROJECT/locations/LOCATION/endpoints/ID` using the client's
+// configured project and location. Other names are returned unchanged.
+func resolveVertexModelName(client *genai.Client, name string) string {
```
adesinah, line 155: Could we add tests for this function?

go/plugins/googlegenai/model_type.go
```diff
@@ -35,11 +35,29 @@ func ClassifyModel(name string) ModelType {
...
 }
 
+// isTunedGeminiName reports whether name refers to a Vertex AI tuned Gemini
+// endpoint, either by its short form (`endpoints/ID`) or its fully qualified
+// resource path (`projects/PROJECT/locations/LOCATION/endpoints/ID`).
+func isTunedGeminiName(name string) bool {
+	if strings.HasPrefix(name, "endpoints/") {
+		return true
+	}
+	if strings.HasPrefix(name, "projects/") && strings.Contains(name, "/endpoints/") {
```
adesinah: Should we include `/locations/` check too?

go/plugins/googlegenai/gemini.go
```diff
@@ -145,6 +145,26 @@ func newModel(client *genai.Client, name string, opts ai.ModelOptions) ai.Model
...
+	if !isTunedGeminiName(name) {
+		return name
+	}
+	if strings.HasPrefix(name, "projects/") {
+		return name
+	}
+	cc := client.ClientConfig()
+	if cc.Backend != genai.BackendVertexAI || cc.Project == "" || cc.Location == "" {
```
adesinah, line 163: If a user reaches this branch with an empty Project or Location, the function returns `endpoints/12345` unchanged. This means the SDK then turns it into
   `publishers/google/models/endpoints/12345` and the call will definitely fail with a confusing `404`. We could do either of the following:
  - Change signature to `(string, error)` and return a clear **"tuned endpoint requires Vertex AI backend with ProjectID and Location"** error, or
  - Validate at `DefineModel` time so the failure shows up at registration instead of call time.

go/plugins/googlegenai/googlegenai.go
```diff
@@ -207,24 +207,33 @@ func (ga *GoogleAI) DefineModel(g *genkit.Genkit, name string, opts *ai.ModelOpt
 // The second argument describes the capability of the model.
 // Use [IsDefinedModel] to determine if a model is already defined.
 // After [Init] is called, only the known models are defined.
+//
+// Tuned Gemini endpoints are accepted in either the short form
+// `endpoints/ID` or the full resource path
+// `projects/PROJECT/locations/LOCATION/endpoints/ID`. When opts is nil the
+// caller gets the default Gemini capability set.
 func (v *VertexAI) DefineModel(g *genkit.Genkit, name string, opts *ai.ModelOptions) (ai.Model, error) {
```
adesinah, line 215: `DefineModel` registers `endpoints/12345` even if `v.ProjectID` / `v.Location`
  weren't set, then the failure manifests at `Generate` time. I know this is a bit out of the scope of this PR, but it's a small change that makes a lot of difference.  My point is, `DefineModel` already knows it's running on a VertexAI, so it should refuse a tuned name unless the client has the project/location it'll need to qualify it.

adesinah (changes requested):
Good changes, I left a few comments.

go/plugins/googlegenai/gemini.go
cabljac: Resolved in b437b45bb — added `TestResolveVertexModelName` 

Uses real `*genai.Client` instances with `HTTPClient: &http.Client{}` 

still reviewing this

go/plugins/googlegenai/model_type.go
cabljac: Resolved in 1be0c0922 — added the `/locations/` check and a test case asserting `projects/p/endpoints/999` (missing `/locations/`) returns false.

go/plugins/googlegenai/gemini.go
cabljac: Took another look at this. Both this and the related comment on `DefineModel` come down to the same invariant: when can `v.ProjectID` / `v.Location` (or the corresponding fields on `client.ClientConfig()`) be empty by the time tuned-name logic runs?

`VertexAI.Init` enforces both upfront (`googlegenai.go:120-124`, `:128-135`):

```go
if projectID == "" {
    projectID = os.Getenv("GOOGLE_CLOUD_PROJECT")
    if projectID == "" {
        panic("Vertex AI requires setting GOOGLE_CLOUD_PROJECT ...")
    }
}

// ... same shape for Location with GOOGLE_CLOUD_LOCATION / GOOGLE_CLOUD_REGION fallbacks
```

`DefineModel` is gated behind `if !v.initted { return nil, errors.New("VertexAI plugin not initialized") }`, so by the time it runs, both fields are guaranteed non-empty. `v.gclient` is built from those same fields, so `client.ClientConfig().Project` and `.Location` are also non-empty when `resolveVertexModelName` reads them in `generate()`.

So the "confusing 404" path described here isn't reachable through the supported flow — adding a `(string, error)` signature or a `DefineModel`-time check would be belt-and-braces against a state the public API already forbids, and the new branches would sit untested since there's no way to construct that state without bypassing `Init`.

I'd prefer to leave the local code as-is and let `Init`'s panic remain the single source of truth for that invariant. Happy to add a doc comment on `resolveVertexModelName` making the precondition explicit if it's worth surfacing.

go/plugins/googlegenai/googlegenai.go
cabljac: Replied on the related `resolveVertexModelName` thread: https://github.com/genkit-ai/genkit/pull/5178#discussion_r3279492408 — same invariant. `VertexAI.Init` already panics on empty `ProjectID` / `Location` (`googlegenai.go:120-124`, `:128-135`), and `DefineModel` is gated behind `v.initted`, so the failure mode described here isn't reachable through the supported flow. I'd prefer to leave `DefineModel` as-is and let `Init` remain the single source of truth.

adesinah: Agreed. Marking this as resolved.

adesinah (approved):

PR #5179, opened 2026-04-23
feat(go/plugins/googlegenai): Add Gemini 3.1 Flash preview models

cabljac:
Adds the two Gemini 3.1 preview models on both backends:

- `gemini-3.1-flash-lite-preview`
- `gemini-3.1-flash-image-preview`

Registered as Multimodal / `ModelStageUnstable` on GoogleAI and VertexAI. Also extends the `imagen-gemini` sample with two generate helpers (text-only and image+caption) and four flows exercising both models on each backend, gated by a `hasVertexAIConfig()` check so GoogleAI-only environments still run.

Closes the four `gemini-3.1-flash-*-preview` rows on the Go parity tracker.

## Test plan

- [x] `go test ./plugins/googlegenai/...`
- [x] `go build ./samples/imagen-gemini/...`
- [x] Live: run the sample with `GEMINI_API_KEY` (and optionally `GOOGLE_CLOUD_PROJECT`/`GOOGLE_CLOUD_LOCATION`) and invoke the four new flows from the dev UI.

Evidence of manual testing:
Screenshot 2026-04-23 at 13 04 19

## gemini31LiteGoogleAIFlow
Screenshot 2026-04-23 at 13 02 41

Screenshot 2026-04-23 at 13 06 50

### DIFF
--- a/go/plugins/googlegenai/models.go
+++ b/go/plugins/googlegenai/models.go
@@ -90,6 +90,9 @@ const (
 
 	gemini25Pro = "gemini-2.5-pro"
 
+	gemini31FlashLitePreview  = "gemini-3.1-flash-lite-preview"
+	gemini31FlashImagePreview = "gemini-3.1-flash-image-preview"
+
 	imagen3Generate001     = "imagen-3.0-generate-001"
 	imagen3FastGenerate001 = "imagen-3.0-fast-generate-001"
 
@@ -115,6 +118,8 @@ var (
 		gemini25Flash,
 		gemini25FlashLite,
 		gemini25Pro,
+		gemini31FlashLitePreview,
+		gemini31FlashImagePreview,
 
 		imagen3Generate001,
 		imagen3FastGenerate001,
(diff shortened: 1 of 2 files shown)


Ehesp (approved):

go/plugins/googlegenai/models.go
```diff
@@ -171,6 +178,18 @@ var (
 			Supports: &Multimodal,
 			Stage:    ai.ModelStageStable,
 		},
+		gemini31FlashLitePreview: {
+			Label:    "Gemini 3.1 Flash Lite Preview",
+			Versions: []string{},
+			Supports: &Multimodal,
+			Stage:    ai.ModelStageUnstable,
```
apascal07, line 185: We'll deprecate "stage" in general but for now, any model that is known (even if it's marked "preview"... which seems to be the default state for Gemini models nowadays) should be `ModelStageStable` so it doesn't trigger the warning.

apascal07 (approved):

PR #5231, opened 2026-05-04
feat(go/plugins/googlegenai): improve model config UX in dev UI

### LINKED ISSUE #5064: [Go] Improve model configuration
From a UX standpoint it would be nice to do the following:

1. **REQUEST** If we can order the options so that important ones are up top. See JS for a consistent ordering. In JS we have "common" options that apply to all providers that end up towards the top, then model provider specific following from that.
1. **REQUEST** Have help text bubbles (needs field descriptions on the schema)
1. **REQUEST** Hide system instruction (its provided in messages instead)

There also seems to be an issue with "Tool config" for Gemini. There is some missing metadata somewhere that causes these "Nested object settings" defaults to display.

**REQUEST** Irrespective of that, though if possible, can these be 2 separate config options (function calling and retrieval)?.

Image

For reference, here is Go vs TS

Image Image

apascal07:
Fixes #5064.

The Genkit Developer UI form-renders model configs from the JSON Schema delivered as `customOptions` on each model action. For Go, that schema comes from reflecting `genai.GenerateContentConfig` (and the Imagen/Veo equivalents) — and those upstream SDK structs don't carry JSON Schema descriptions, so the form has no help bubbles. Worse, several Gemini fields are managed by Genkit primitives (system prompts, output schemas, caching) and rejected when supplied directly, so the form invites users to fill in fields that won't work.

This change layers per-property metadata onto the reflected schema before it's exposed:

- **Descriptions** for every supported top-level field on the three configs, mirroring the JS plugin's tone where the field is the same.
- **Hidden fields** for the Gemini config: any field the plugin currently errors on, plus `candidateCount` (which the plugin pins to 1).

No reordering, no flattening of nested objects, no munging of dev-UI submitted requests. Override metadata is curated in the plugin (`go/plugins/googlegenai/config_overrides.go`) so it stays decoupled from upstream SDK changes.

## What's hidden on the Gemini config

Each entry below is rejected at runtime in [gemini.go:289-306](https://github.com/genkit-ai/genkit/blob/ap/go-model-config-ux/go/plugins/googlegenai/gemini.go#L289-L306) when supplied via the dev UI, so removing them from the schema makes the form match reality:

| Field | Genkit primitive |
|---|---|
| `systemInstruction` | `ai.WithSystemPrompt` |
| `cachedContent` | `ai.WithCacheTTL` |
| `responseSchema` | `ai.WithOutputType` / `ai.WithOutputSchema` |
| `responseMimeType` | `ai.WithOutputType` / `ai.WithOutputSchema` |
| `responseJsonSchema` | `ai.WithOutputSchema` |
| `candidateCount` | pinned to 1 by the plugin |

## What we explicitly didn't do

- **No reordering.** Field order is benign; the dev UI's existing primitives-first sort already buckets primitives ahead of nested objects.
- **No flattening of `toolConfig`.** It stays a nested object, matching `genai.ToolConfig`.
- **No request shape munging.** Requests submitted from the dev UI flow through `configFromRequest` unchanged.

## Test plan

- [x] `go test ./plugins/googlegenai/...` — new tests assert hidden fields are absent and described fields carry the expected description; described fields not in the SDK type fail loudly so the override map stays honest under upstream drift.
- [x] `go vet ./plugins/googlegenai/...`
- [x] Manually inspected the produced schema JSON — confirmed `systemInstruction`/`cachedContent`/etc. absent at top-level and that descriptions land on the right fields. Nested `responseJsonSchema` inside function tool schemas is preserved (it's not the same field, just a name collision in a sub-schema).

## Checklist

- [x] Tested (unit tests + manual schema inspection)
- [ ] Docs updated — n/a (no doc-visible API change; descriptions surface only in the dev UI)

### DIFF
--- a/go/plugins/googlegenai/config_overrides.go
+++ b/go/plugins/googlegenai/config_overrides.go
@@ -15,21 +15,27 @@ import (
 // SDK structs do not carry JSON Schema descriptions, and a few of their
 // fields are managed by Genkit primitives and rejected when supplied
 // directly, so we curate that information here.
+//
+// All path application is best-effort: if the upstream SDK renames or
+// removes a field, the corresponding entry silently no-ops rather than
+// panicking. The cost is that stale entries quietly stop applying — caught
+// by the snapshot tests, not at runtime.
 type configOverrides struct {
-	// descriptions maps a JSON property name to the help text shown as the
-	// field's tooltip in the dev UI. Top-level only.
+	// descriptions maps a JSON property path to the help text shown as the
+	// field's tooltip in the dev UI. Keys may be a top-level property name
+	// ("temperature") or a dotted path with "[]" denoting a descent into an
+	// array's item shape ("safetySettings[].category").
 	descriptions map[string]string
-	// hidden lists JSON property paths to remove from the schema. Each entry
-	// is either a top-level property name ("systemInstruction") or a dotted
-	// path with "[]" denoting a descent into an array's item shape
-	// ("tools[].functionDeclarations"). Use this for fields the plugin
-	// rejects at runtime or that the Genkit framework manages directly.
+	// hidden lists JSON property paths to remove from the schema. Same
+	// notation as descriptions. Use this for fields the plugin rejects at
+	// runtime or that the Genkit framework manages directly.
 	hidden []string
 }
 
 // gccOverrides controls dev UI presentation of [genai.GenerateContentConfig].
 var gccOverrides = configOverrides{
 	descriptions: map[string]string{
+		// Top-level fields.
 		"temperature":                "Controls the degree of randomness in token selection. Lower values produce more deterministic responses; higher values produce more diverse or creative ones.",
 		"topP":                       "Considers tokens whose cumulative probability exceeds this value. Lower values constrain the model to high-probability tokens; higher values allow more variety.",
 		"topK":                       "Limits sampling to the K most likely tokens at each step. Lower values constrain output; higher values allow more variety.",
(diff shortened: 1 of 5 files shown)


MichaelDoyle (approved):
Thanks! This is a nice improvement.

MichaelDoyle:
Hey @apascal07 - curious, would you consider (in the future) maintaining the field ordering? Maybe 2.0 (breaking change?) I think we lose the ordering defined in the structs due to serializing as `map[string]any`.

But in theory, we could preserve the order:

 In `go/core/api/action.go`:
 ```go
 type ActionDesc struct {
     // ...
     InputSchema  any `json:"inputSchema"`  // Changed from map[string]any
     OutputSchema any `json:"outputSchema"` // Changed from map[string]any
     // ...
 }
 ```

 In `go/internal/base/json.go`:
 ```go
 func SchemaAsMap(s *jsonschema.Schema) any {
     jsb, err := s.MarshalJSON()
     if err != nil {
         log.Panicf("failed to marshal schema: %v", err)
     }
     // Return as RawMessage to preserve order
     return json.RawMessage(jsb)
 }

MichaelDoyle:
I realized I did not really say "why". :)

It's a bit of a pickle. On the one hand, field ordering is not guaranteed, and probably shouldn't be expected. On the other hand, there are _a lot_ of supported options, and growing. In the Dev UI, "common" attributes are expected to be at the top. Temperature slider comes to mind as most important.

The alternative is for the Dev UI to have its own interpretation of the right field order, and apply it generally.

apascal07:
Can add it but the order feels arbitrary... alphabetical (default for marshaling in Go I believe) at least is predictable when we are not in control of the config, although related fields are not next to each other which is one downside. I intentionally omitted it for now...

PR #5232, opened 2026-05-05
feat(go): trace GenerateActionOptions as input for generate calls

MichaelDoyle:
Currently all of the generate options (such as use/middleware, model, etc) are missing from the trace because `Request` is set as the input. Switching to `Options` is the most idiomatic thing for Go at this time (IMO). 

This PR:

- Updates the trace span for generate to populate `Options` as the input
- Allow custom stepName for generate spans

**Note: This is a bit different from Typescript, because in Go, recursive generate calls do not accumulate message history in the `GenerateActionOptions`, which Typescript does.** For recursive calls, each generate will have the same exact `Options`, even though the underlying request is growing.

We _could_ set the full `GenerateParams` as the input, but there are two problems (a) it "leaks" internal details and (b) it's confusing to have 2 message arrays.

In any case, the full message history will accumulate in the model spans beneath each generate call.

image

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -372,16 +372,20 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
 	var runGenerate func(context.Context, *GenerateParams) (*ModelResponse, error)
 
 	runGenerate = func(ctx context.Context, params *GenerateParams) (*ModelResponse, error) {
-		req := params.Request
-		currentTurn := params.Iteration
-		messageIndex := params.MessageIndex
+		name := params.Options.StepName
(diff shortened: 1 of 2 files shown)


MichaelDoyle:
@pavelgj @apascal07 I debated this one a couple times, back and forth between putting `GenerateActionOptions ` versus `GenerateParams` in the trace. Let me know if you have opinions (see analysis in the desc).

apascal07:
`GenerateActionOptions` LGTM, `GenerateParams` is a "private" type only for the middleware hook.

apascal07 (approved):

apascal07:
Thanks for this!

Issue #4683, opened 2026-02-15
[Go] ModelResponse.History() returns incomplete history when streaming in `compat_oai` plugin


ijteleox:
**Describe the bug**
When using the `compat_oai` plugin with streaming enabled, `ModelResponse.History()` only returns the model's response message instead of the full conversation history. This is because the streaming code path sets `Request` to an empty `&ai.ModelRequest{}` instead of the actual request.

**To Reproduce**
1. Use the `compat_oai/openai` plugin with `ai.WithStreaming(cb)`
2. Pass conversation history via `ai.WithMessages(h...)`
3. Call `resp.History()` on the returned `ModelResponse

**Expected behavior**
Expected: returns all input messages plus the model's response message
Actual: returns only the model's response message (since `Request.Messages` is empty)

**Screenshots**

**Runtime (please complete the following information):**
 - OS: MacOS
 - Version 1.4.0

** Go version
 -  1.25.4

**Additional context**
Root cause

In `plugins/compat_oai/generate.go`, the `Generate` method dispatches to two different code paths:

- **Non-streaming** (`generateComplete`) correctly sets `resp.Request = req` ([line 414](https://github.com/firebase/genkit/blob/main/go/plugins/compat_oai/generate.go#L414))
- **Streaming** (`generateStream`) does not accept the `req` parameter at all, and `convertChatCompletionToModelResponse` initializes `Request` to an empty `&ai.ModelRequest{}` ([line 343](https://github.com/firebase/genkit/blob/main/go/plugins/compat_oai/generate.go#L343))

### Suggested fix

Pass `req *ai.ModelRequest` through to `generateStream` and set `resp.Request = req` on the response before returning, matching what `generateComplete` already does.

hugoaguirre:
@ijteleox I'll take a look at the issue, I'll get back to you

PR #5239, opened 2026-05-06
fix(go): compat_oai plugin returns incomplete history when streaming

MikeRez0:
Changes fix this issue
https://github.com/genkit-ai/genkit/issues/4683

Checklist (if applicable):

### DIFF
--- a/go/plugins/compat_oai/generate.go
+++ b/go/plugins/compat_oai/generate.go
@@ -441,7 +441,7 @@ func (g *ModelGenerator) generateComplete(ctx context.Context, req *ai.ModelRequ
 		return nil, fmt.Errorf("failed to create completion: %w", err)
 	}
 
-	resp, err := convertChatCompletionToModelResponse(completion)
+	resp, err := convertChatCompletionToModelResponse(completion, req)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


MichaelDoyle:
Thanks @MikeRez0! Can you get the CLA taken care of? In the mean time we'll get the review done.

MikeRez0:
Already signed

MikeRez0:
@apascal07 Hi, Alex! Can you review changes, please? I want to use stream mode in my project

apascal07:
A few requested changes:

1. Keep `convertChatCompletionToModelResponse` pure — instead of threading `req` through it, set `resp.Request = req` in `generateStream` after the call, mirroring what `generateComplete` already does. That avoids changing the converter's signature and keeps a single pattern for both paths.
2. Delete the commented-out `// Request: &ai.ModelRequest{},` line rather than leaving it in.
3. With the above, the existing `resp.Request = req` in `generateComplete` stays correct. As written, it's now redundant with the assignment inside the converter.
4. Add a regression test for the streaming path that asserts `resp.Request.Messages` (or `resp.History()`) preserves the input messages — `TestGenerator_Stream` currently discards the response with `_`, so this bug could silently return.

apascal07 (changes requested):
Comments above.

apascal07 (approved):

PR #5251, opened 2026-05-06
feat(js): implemented defineAgent

pavelgj:
Introduces **`defineAgent`** and friends: a first-class, multi-turn agent API for Genkit JS. An agent bundles a model/prompt, conversational state, tool loops, interrupts, persistence, streaming, and abort/resume into a single ergonomic, transport-agnostic surface that runs identically **in-process on the server** and **over HTTP from a browser**.

### Highlights

- **Three ways to define an agent**, from zero-config to full control.
- **Stateful, multi-turn chats** that thread `snapshotId`/`state` for you.
- **Typed custom state** with Zod validation (`stateSchema`).
- **Pluggable persistence** via `SessionStore` (in-memory + file stores included), with server-managed sessions resumable by `sessionId`.
- **Streaming** with live custom-state patches over the wire (JSON Patch).
- **Interrupts & resume** (`restart` / `respond`) for human-in-the-loop tools.
- **Abort, detach (background turns) + heartbeats**, and graceful failure with last-good state.
- **`clientTransform`** to redact/reshape state and stream chunks before they leave the server.
- **Transport-agnostic**: the same `AgentAPI` is returned by `ai.defineAgent(...)` on the server and `remoteAgent(...)` on the client.

---

### 1. Define an agent

The headline API — one call wires up a prompt and a multi-turn agent:

```ts
const agent = ai.defineAgent({
  name: 'assistant',
  model: 'googleai/gemini-flash-latest',
  system: 'You are a helpful assistant.',
  tools: [searchTool],
});
```

Prompt-driven (reuse a registered `.prompt`), with typed `promptInput`:

```ts
const agent = ai.definePromptAgent({
  promptName: 'assistant',
  promptInput: { role: 'pirate', tone: 'cheerful' }, // type-checked vs prompt's input schema
});
```

Full control via a custom turn handler:

```ts
const agent = ai.defineCustomAgent({ name: 'custom' }, async (runner, input) => {
  // own the turn: call models, tools, mutate session state, return a result
});
```

### 2. Chat: send & stream

`chat()` returns a stateful conversation that tracks `state`, `messages`, `snapshotId`, and `sessionId` so you don't thread them by hand:

```ts
const chat = agent.chat();

const res = await chat.send('Hello!');
console.log(res.text, res.finishReason); // 'Hi!  ...', 'stop'

// streaming
const turn = chat.sendStream('Write a poem');
for await (const chunk of turn.stream) {
  process.stdout.write(chunk.text ?? '');
}
const final = await turn.response;
```

### 3. Typed custom state

```ts
const agent = ai.defineAgent({
  name: 'cart',
  model: 'googleai/gemini-flash-latest',
  stateSchema: z.object({ items: z.array(z.string()) }),
});

const chat = agent.chat({ state: { items: [] } });
await chat.send('add milk');
console.log(chat.state); // { items: ['milk'] }  (validated against stateSchema)
```

### 4. Persistence & resuming sessions

Attach a `SessionStore` to make an agent **server-managed**: turns are persisted as snapshots and a chat can be resumed later by `sessionId` (no need to track snapshot IDs) or by an exact `snapshotId`:

```ts
import { FileSessionStore } from '@genkit-ai/ai/session-stores';

const agent = ai.defineAgent({
  name: 'assistant',
  model: 'googleai/gemini-flash-latest',
  store: new FileSessionStore('./sessions'),
});

const chat = agent.chat();
await chat.send('Remember my name is Ada.');
const { sessionId } = chat;

// later, somewhere else:
const resumed = await agent.loadChat({ sessionId });
await resumed.send('What is my name?'); // -> "Ada"
```

Built-in stores: `InMemorySessionStore` and `FileSessionStore` (with filesystem snapshot watching + configurable poll fallback). The `SessionStore` interface uses an atomic mutator-based `saveSnapshot` to make concurrent writes (e.g. an abort racing a completion) safe.

### 5. Interrupts & resume (human-in-the-loop)

When a tool interrupts, the turn finishes with `finishReason: 'interrupted'`. Resume by either re-running the tool (`restart`) or supplying the answer directly (`respond`):

```ts
let res = await chat.send('Transfer $100 to Bob');
if (res.finishReason === 'interrupted') {
  res = await chat.resume({
    respond: [confirmTransfer.respond(res, { approved: true })],
  });
}
```

### 6. Abort, detach & graceful failure

```ts
// abort an in-flight turn
const turn = chat.sendStream('long task...');
await chat.abort();

// background / detached turns with heartbeat-based liveness
const task = await chat.detach('run this in the background');
for await (const status of task.watch()) { /* pending -> completed/expired */ }
```

Failures resolve gracefully (`finishReason: 'failed'`) and preserve the **last-good snapshot/state** instead of corrupting the session; genuine API misuse throws `AgentInitError`.

### 7. Redact / reshape state for clients

```ts
const agent = ai.defineAgent({
  name: 'assistant',
  model: 'googleai/gemini-flash-latest',
  clientTransform: {
    state: (s) => ({ ...s, secret: undefined }), // redact at rest / in responses
    chunk: (c) => c,                              // reshape or drop stream chunks in flight
  },
});
```

### 8. Same API, server or browser

The server returns an `AgentAPI`; the client gets the exact same shape from `remoteAgent(...)`, so UI code is identical whether it runs in-process or over HTTP:

```ts
import { remoteAgent } from 'genkit/beta/client';

const agent = remoteAgent({ url: 'https://my-app/agents/assistant' });
const chat = agent.chat();
const turn = chat.sendStream('Hello from the browser');
for await (const chunk of turn.stream) { render(chunk); }
```

---

### What's inside

- `js/ai/src/agent.ts` — `defineAgent` / `definePromptAgent` / `defineCustomAgent`, schemas, session runner, finish reasons, error handling.
- `js/ai/src/agent-core.ts` — transport-agnostic `AgentAPI` / `AgentChat` / `AgentResponse`, in-process transport, custom-state patching.
- `js/ai/src/session.ts`, `js/ai/src/session-stores.ts` — `Session`, `SessionStore`, `InMemorySessionStore`, `FileSessionStore`, snapshot model + heartbeats.
- `js/ai/src/json-patch.ts` — minimal, pollution-safe JSON Patch used for streaming custom-state updates.
- `js/genkit/src/client/*` — `remoteAgent` HTTP client.
- `js/genkit/src/beta.ts` / `genkit-beta.ts` — `ai.defineAgent` etc. + `ai.currentSession()` exposed on the beta surface.
- Extensive tests: `agent_test.ts`, `agent_client_test.ts`, `session-stores_test.ts`, `json-patch_test.ts`.

**Finish reasons:** `stop`, `interrupted`, `failed`, `aborted`, `blocked`, `detached`, `unknown`.
**Snapshot statuses:** `pending`, `completed`, `failed`, `aborted`, `expired`.


js/ai/src/agent.ts
```diff
@@ -0,0 +1,1104 @@
...
+  runWithSession,
+  type SessionStoreOptions,
+} from './session.js';
+
+/**
+ * Schema for initializing an agent turn.
+ */
+export const AgentInitSchema = z.object({
+  snapshotId: z.string().optional(),
+  newSnapshotId: z.string().optional(),
```
apascal07: ?

js/ai/src/agent.ts
```diff
@@ -0,0 +1,1104 @@
...
+        });
+      }
+      const snapshot = await config.store.getSnapshot(snapshotId, {
+        context: getContext(),
+      });
+      if (snapshot) {
+        const previousStatus = snapshot.status;
+        snapshot.status = 'aborted';
+        await config.store.saveSnapshot(snapshot, { context: getContext() });
+        return previousStatus;
```
apascal07: I'm not sure this logic is helpful. It should return `aborted` *unless* it has reached a different terminal state in which case that status should be returned instead.

js/ai/src/agent.ts
pavelgj: removed

js/ai/src/agent.ts
pavelgj: fixed

js/ai/src/agent.ts
```diff
@@ -0,0 +1,874 @@
...
+  }
+) {
+  let cachedPromptAction: PromptAction | undefined;
+
+  const fn: AgentFn<unknown, State> = async (
+    sess,
+    { sendChunk, abortSignal }
+  ) => {
+    await sess.run(async (input) => {
+      const promptInput = {};
```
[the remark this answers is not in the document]
pavelgj: by design

js/ai/src/agent.ts
```diff
@@ -0,0 +1,874 @@
...
+      // Tag every history message so we can identify them after render.
+      const history = (sess.session.getMessages() || []).map((m) => ({
+        ...m,
+        metadata: { ...m.metadata, [historyTag]: true },
+      }));
+
+      // Let the prompt control where history is placed (e.g. dotprompt
+      // {{history}}).  When the prompt has no explicit `messages` config
+      // the render helper simply appends history after system/user.
+      const genOpts = await cachedPromptAction.__executablePrompt.render(
```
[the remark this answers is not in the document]
pavelgj: it's ok

js/ai/src/session-stores.ts
```diff
@@ -0,0 +1,522 @@
...
+  /**
+   * Resolves the file path for a given snapshotId: `<prefix>/<snapshotId>.json`.
+   */
+  private async getFilePath(
+    snapshotId: string,
+    options?: SessionStoreOptions
+  ): Promise<string> {
+    const dir = this.prefixDir(options);
+    await this.ensureDir(dir);
+    return path.join(dir, `${snapshotId}.json`);
```
apascal07: `snapshotId` comes straight off the wire (the abort/getSnapshot actions are bare `z.string()`) and lands here in `path.join` with no checks, so something like `../../foo` reads/writes outside the prefix and busts the per-tenant prefix isolation. Worth rejecting ids that aren't a plain basename (no `/`, `\`, `..`) and asserting the resolved path stays under the prefix dir. Only bites FileSessionStore, but the remote client exposes exactly these actions over HTTP.

js/ai/src/json-patch.ts
```diff
@@ -0,0 +1,379 @@
...
+/**
+ * Walks to the parent container of `tokens`, lazily creating intermediate
+ * objects for missing segments so leniently-applied patches still land.
+ */
+function ensureParent(doc: any, tokens: string[]): any {
+  let cur = doc;
+  for (let i = 0; i < tokens.length - 1; i++) {
+    const token = tokens[i];
+    if (cur == null) return undefined;
+    const next = Array.isArray(cur) ? cur[Number(token)] : cur[token];
```
apascal07, line 383: `__proto__` slips through here: the existing prototype is a non-null object, so we skip creating a fresh `{}` and walk into it, then `setValue` writes onto `Object.prototype`. Since this module is a public export and gets applied to server-sent patches in the browser, a path like `/__proto__/x` pollutes globally. Easy fix: reject `__proto__`/`prototype`/`constructor` tokens when parsing the pointer.

js/ai/src/session-stores.ts
```diff
@@ -0,0 +1,522 @@
...
+    );
+  }
+
+  async saveSnapshot(
+    snapshotId: string | undefined,
+    mutator: SnapshotMutator<S>,
+    options?: SessionStoreOptions
+  ): Promise<string | null> {
+    // Read the current snapshot when an ID is provided.
+    const current = snapshotId
```
apascal07, line 483: This read and the `writeFile` just below it aren't atomic, but the `SessionStore` contract (and the abort-aware mutator that checks `current.status`) assume they are, so two concurrent saves can have a `completed` write clobber a concurrent `aborted`. InMemory is fine since it's a sync Map; this one needs a per-id lock. While here: `writeFile` truncates in place, so a concurrent reader in `getSnapshot` can hit torn JSON and throw. Writing to a temp file + rename fixes both.

js/ai/src/agent.ts
```diff
@@ -0,0 +1,1863 @@
...
+            setCustomMetadataAttribute('agent:snapshotId', snapshotId);
+          }
+
+          this.notifyEndTurn(snapshotId, finishReason);
+
+          // The turn span's output is the session state this turn produced -
+          // applies to both client- and server-managed agents.
+          return { state: this.session.getState() };
+        });
+        this.turnIndex++;
```
apascal07, line 616: An aborted turn rejects out of `generate` and lands in this catch, so it gets recorded and reported as `failed` with an error rather than `aborted`, unless the concurrent abort write happened to land first (the race from the file store). Probably want to check `abortController.signal.aborted` here and set `finishReason: 'aborted'` with no error, and skip the failed snapshot write.

js/ai/src/agent-core.ts
```diff
@@ -0,0 +1,959 @@
...
+      return { snapshotId: this.snapshotId };
+    }
+    if (this.clientState) {
+      return { state: this.clientState };
+    }
+    return this.connectInit ?? {};
+  }
+
+  /** Applies a completed turn's output to the running aggregates. */
+  private applyOutput(raw: AgentOutput<State>): void {
```
apascal07, line 669: No in-flight guard on the chat: `buildInit` reads `this.snapshotId`/`clientState` and `applyOutput` mutates them, so two overlapping `send`/`sendStream` calls branch from the same parent and race here on messages/snapshotId/state (and a client-managed turn ships the live `clientState` the other turn is still mutating). The interface kind of implies turns are serialized, so either queue/reject while one is in flight, or document it as single-flight.

js/ai/src/session.ts
```diff
@@ -14,170 +14,407 @@
...
+  private version: number = 0;
+
+  /** Stable identifier that correlates traces across agent turns. */
+  readonly sessionId: string;
+
+  constructor(initialState: SessionState<S>) {
+    super();
+    this.sessionId = initialState.sessionId || globalThis.crypto.randomUUID();
+    initialState.sessionId = this.sessionId;
+    this.state = initialState;
```
apascal07: This stores `initialState` by reference (and mutates `.sessionId` on the caller's object the line above). Combined with `getMessages()` returning `this.state.messages` directly at line 271, a handler that mutates either reaches back into caller/chat state. A `structuredClone` here and returning a copy from `getMessages` would keep it isolated.

js/ai/src/agent.ts
```diff
@@ -0,0 +1,1863 @@
...
+    // if the leaf is a non-resumable turn (e.g. a `failed`/`aborted`/`pending`
+    // turn) walk back over its parent chain to the last-good (`completed`)
+    // snapshot. When the session has no resumable snapshot (e.g. a first-turn
+    // failure) seed a fresh session bound to the requested sessionId so
+    // subsequent turns can find it.
+    let snapshot = await store.getSnapshot({
+      sessionId: init.sessionId,
+      context: getContext(),
+    });
+    while (snapshot && snapshot.status !== 'completed') {
```
apascal07, line 979: This walk has no cycle/visited guard, so a self-referential or cyclic `parentId` is an infinite loop of store reads (and a long non-resumable tail is O(n) round-trips). A max-depth or visited set, with `FAILED_PRECONDITION` on a cycle, would be safer.

js/ai/src/agent-core.ts
```diff
@@ -0,0 +1,959 @@
...
+    return turn.response;
+  }
+
+  sendStream(
+    input: string | AgentInput,
+    opts?: { abortSignal?: AbortSignal }
+  ): AgentTurn<State> {
+    const agentInput = toAgentInput(input);
+    if (agentInput.message) {
+      this.messages.push(agentInput.message);
```
apascal07, line 847: Two small things around this eager push: (1) if `opts.abortSignal` is already aborted we still push the message and dispatch the turn before short-circuiting, worth bailing at the top of `sendStream`; (2) on a failed/aborted turn that returns no `state.messages`, this pushed message sticks around in `this.messages` with no reply and gets re-sent next turn, so probably only commit it on success.

js/ai/src/agent-core.ts
```diff
@@ -0,0 +1,959 @@
...
+        for await (const raw of rawStream) {
+          const chunk = new AgentChunkImpl<State>(raw, previousText);
+          previousText = chunk.accumulatedText;
+          // Keep the locally tracked custom state live mid-stream by applying
+          // each streamed JSON Patch to it, then surface the resulting
+          // post-patch custom state on the chunk as `chunk.custom` so consumers
+          // get an explicit change notification (with a fresh object reference
+          // for `===`-based change detection). The first patch of a turn is a
+          // whole-document replace that re-bases us onto the server baseline.
+          if (raw.customPatch) {
```
apascal07, line 880: `customPatch` chunks are applied positionally with no sequence/version, and `applyPatch` is lenient (missing members are silent no-ops), so a dropped or reordered patch gives silently-wrong state instead of an error. Fine over the in-order SSE channel today, but a per-turn sequence number + contiguity check (or at least documenting the in-order/no-loss transport requirement) would make it robust.

js/ai/src/genkit-ai.ts
```diff
@@ -343,6 +345,22 @@ export class GenkitAI {
     return run(name, funcOrInput, this.registry);
   }
 
+  /**
+   * Gets the current session if running within one, otherwise throws a {@link GenkitError}.
+   *
+   * @beta
+   */
+  currentSession<S = any>(): Session<S> {
```
apascal07, line 353: This new base `currentSession` throws `GenkitError(FAILED_PRECONDITION)`, but the `GenkitBeta` override (genkit-beta.ts ~169) still throws `SessionError` for the same condition. Two error types for one method down the inheritance chain is a footgun, worth collapsing to one.

js/ai/src/session.ts
```diff
@@ -14,170 +14,407 @@
...
+
+/**
+ * Artifact generated during a session turn.
+ */
+export type Artifact = z.infer<typeof ArtifactSchema>;
+
+/**
+ * Events signifying a session snapshot persistence point.
+ */
+export const SnapshotEventSchema = z.enum(['turnEnd', 'invocationEnd']);
```
apascal07: On the Go side we dropped `event` entirely: no field, and no `invocationEnd` snapshot. We treat the turn as the unit of persistence, so it's only ever turn-end writes. Are we still doing invocation end writes? If not then definitely don't need the event field.

js/genkit/src/client/agent.ts
```diff
@@ -0,0 +1,160 @@
...
+ * const chat = agent.chat();
+ * const res = await chat.send('Weather in Tokyo?').response;
+ * console.log(res.text);
+ * ```
+ */
+export function remoteAgent<State = unknown>(
+  options: RemoteAgentOptions
+): AgentAPI<State> {
+  const { url } = options;
+  const stateUrl = options.stateUrl ?? `${url}/state`;
```
apascal07: Go serves these as `/getSnapshot` and `/abortSnapshot`, not `/state` + `/abort`, so the paths need to be adjusted in the client SDK. Are you attached to those path names? I prefer these for being 1:1 mappings to the actions since we're not following REST paths.

js/ai/src/agent.ts
```diff
@@ -0,0 +1,1863 @@
...
+  };
+
+  const getSnapshotDataAction = defineAction(
+    registry,
+    {
+      name: config.name,
+      description: `Gets snapshot data for ${config.name} by snapshotId or sessionId`,
+      actionType: 'agent-snapshot',
+      inputSchema: GetSnapshotRequestSchema,
+      outputSchema: z.any(), // SessionSnapshot Schema
```
apascal07: How come this isn't `SessionSnapshotSchema`?

js/genkit/src/client/agent.ts
pavelgj: I'm fine with `/getSnapshot` but `/abortSnapshot` doesn't make sense... we don't abort a snapshot, we abort a detached turn... let's do `/getSnapshot` but `/abort` instead of `/abortSnapshot`?

js/ai/src/session.ts
pavelgj: unused vestiges... will cleanup.

js/genkit/src/client/agent.ts
apascal07: Good point... SGTM.

apascal07: `AbortRequest`/`AbortResponse` or `AgentAbortRequest`/`AgentAbortResponse`? The latter to avoid collisions in shared schemas, prefer the former though.

pavelgj: in schemas.... probably more verbose is better.

js/ai/src/session-stores.ts
pavelgj: Fixed. FileSessionStore now validates snapshotId is a plain basename (rejecting /, \, .., ., NUL) and asserts the resolved path stays under the prefix dir before any read/write. Added path-traversal tests.

js/ai/src/json-patch.ts
pavelgj: Fixed. parsePointer now rejects __proto__/prototype/constructor tokens for every op (path and from), so /__proto__/x throws instead of walking into the prototype. Added prototype-pollution tests.

js/ai/src/session-stores.ts
pavelgj: Fixed. Saves are now serialized per resolved file path with an in-process lock (a completed write can no longer clobber a concurrent aborted), and writes go through a temp file + atomic rename so a concurrent getSnapshot reader never sees torn JSON.

js/ai/src/agent.ts
pavelgj: Fixed. The catch now checks abortController.signal.aborted and, when aborted, records finishReason: aborted with no error and skips the failed snapshot write.

js/ai/src/agent-core.ts
pavelgj: Documented as single-flight: added a note on sendStream that the client does not guard against overlapping send/sendStream calls on the same chat (they would race on messages/snapshotId/clientState); turns are assumed serialized.

js/ai/src/session.ts
pavelgj: Fixed. Session now structuredClones the initial state in the constructor and getMessages() returns a copy, so a handler mutating either can no longer reach back into caller/chat state.

js/ai/src/agent.ts
pavelgj: Fixed. The sessionId parent-chain walk now carries a visited set and throws FAILED_PRECONDITION on a cycle instead of looping forever.

js/ai/src/agent-core.ts
pavelgj: Fixed both: sendStream now bails at the top if opts.abortSignal is already aborted (no message push, no turn dispatch), and the eagerly-pushed user message is rolled back when a failed/aborted turn returns no authoritative state.messages, so it is not re-sent next turn.

js/ai/src/agent-core.ts
pavelgj: Documented the in-order/no-loss transport requirement on applyCustomPatch: customPatch chunks carry no sequence number and are applied positionally, so the transport MUST deliver them in order with no loss (the in-order SSE channel satisfies this). Each turn still begins with a whole-document replace, so any corruption cannot persist past a turn boundary. Left the sequence-number machinery out for now.

js/ai/src/genkit-ai.ts
pavelgj: Fixed by removing the redundant GenkitBeta.currentSession override entirely - it was identical to the base except for the error type, so GenkitBeta now inherits the base GenkitError(FAILED_PRECONDITION). Single error type for the method now.

js/ai/src/session.ts
pavelgj: Good catch - removed it. We were still doing an invocation-end write, but nothing read the event field for control flow (resume/leaf logic keys off status/parentId). Dropped invocationEnd, the event field, and SnapshotEventSchema to match Go; the post-handler write now just defaults to a completed snapshot that the version guard skips when nothing changed.

js/genkit/src/client/agent.ts
pavelgj: Renamed the client path /state -> /getSnapshot to match Go. Kept /abort rather than /abortSnapshot per the discussion (we abort a detached turn, not a snapshot). Updated the client SDK option (getSnapshotUrl) and tests.

js/ai/src/agent.ts
pavelgj: Switched it to SessionSnapshotSchema (added a Zod schema mirroring SessionSnapshot in session.ts). It was z.any() as a placeholder; now properly typed/discoverable.

js/ai/src/agent.ts
```diff
@@ -0,0 +1,874 @@
...
+  };
+
+  const getSnapshotDataAction = defineAction(
+    registry,
+    {
+      name: config.name,
+      description: `Gets snapshot data for ${config.name} by snapshotId`,
+      actionType: 'agent-snapshot',
+      inputSchema: z.string(),
+      outputSchema: z.any(), // SessionSnapshot Schema
```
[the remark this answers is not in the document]
pavelgj: Done - getSnapshotDataAction now uses SessionSnapshotSchema (a Zod schema mirroring SessionSnapshot) instead of z.any().

js/ai/src/agent.ts
```diff
@@ -0,0 +1,874 @@
...
+  /**
+   * Executes the flow handler against incoming input messages sequentially.
+   */
+  async run(fn: (input: AgentInput) => Promise<void>): Promise<void> {
+    for await (const input of this.inputCh) {
+      if (input.messages) {
+        this.session.addMessages(input.messages);
+      }
+
+      const turnSnapshotId = this.newSnapshotId || crypto.randomUUID();
```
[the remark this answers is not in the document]
pavelgj: Not an issue here: the code uses globalThis.crypto.randomUUID() (the Web Crypto API), not the bare crypto global. globalThis.crypto is standard across Node 18+, Cloudflare Workers, Deno, and browsers; switching to node:crypto would actually break the non-Node runtimes Genkit targets, so leaving as-is.

apascal07 (approved):

PR #5255, opened 2026-05-07
feat(go/ai): per-tool strict schema control via WithStrict option

### LINKED ISSUE #5230: Allow optional strict tool use for anthropic sdk

Allow the user to specify non-strict tool schemas. When native structured output was introduced in[ !4701](https://github.com/genkit-ai/genkit/pull/4701) it also forced all tools to strict. This has several disadvantages:
1. You can only do 20 total tools
2. You can only have 24 optional fields. 
3. Other seemingly arbitrary compile time restrictions
4. You have a slow compilation time on first message since it has to compile

Either 2 options:
1. Allow the user to be able to specify which tools are strict.
   a. From what I can see it's all or nothing
2. Allow the user to turn on strict tool use. 

**Additional context**

When the user sends tools that don't match the restraints you get the below error:
```
Schema is too complex for compilation. Try reducing the number of tools or simplifying tool schemas."
```

Anthony-Bible:
Closes #5230 

This is backwards compatible so users who don't define strict will be automatically marked as strict (previous behavior). Right now we define a strict parameter in `go/ai/gen.go`, but it's only used for the anthropic plugin however I believe it sets the foundation to use `strict` in other providers (I noticed there was a todo for the openai plugin). I've tested locally and it properly sets tools as strict that are marked as strict, tools that are marked as not strict are not strict and tools with no strict field set are still strict

### DIFF
--- a/go/ai/gen.go
+++ b/go/ai/gen.go
@@ -529,6 +529,9 @@ type ToolDefinition struct {
 	Name string `json:"name,omitempty"`
 	// OutputSchema is a valid JSON Schema describing the output of the tool.
 	OutputSchema map[string]any `json:"outputSchema,omitempty"`
+	// Strict controls whether strict schema validation is enforced for this tool.
+	// When nil, the provider's default is used.
+	Strict *bool `json:"strict,omitempty"`
 }
 
 // A ToolRequest is a message from the model to the client that it should run a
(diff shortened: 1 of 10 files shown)


apascal07:
Hi @Anthony-Bible, I refactored some bits to be more generally applicable to OAI as well, please verify on your end that this still accomplishes what you wanted.

apascal07 (approved):

Anthony-Bible:
@apascal07 thanks, this is perfect and still works for me. I say it's good to merge if you're also okay with that.

PR #5261, opened 2026-05-07
feat(js/plugins/google-genai): Support for Gemini 3.1 Flash Lite

ifielker:
Checklist (if applicable):

### DIFF
--- a/js/testapps/basic-gemini/src/index.ts
+++ b/js/testapps/basic-gemini/src/index.ts
@@ -126,7 +126,7 @@ ai.defineFlow('combine tools and builtins', async () => {
 
 ai.defineFlow('basic-hi', async () => {
   const { text } = await ai.generate({
-    model: googleAI.model('gemini-flash-lite-latest'),
+    model: googleAI.model('gemini-3.1-flash-lite'),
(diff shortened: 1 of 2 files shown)


js/testapps/basic-gemini/src/index.ts
```diff
@@ -126,7 +126,7 @@ ai.defineFlow('combine tools and builtins', async () => {
 
 ai.defineFlow('basic-hi', async () => {
   const { text } = await ai.generate({
-    model: googleAI.model('gemini-flash-lite-latest'),
+    model: googleAI.model('gemini-3.1-flash-lite'),
```
pavelgj, line 129: This is more maintainable...

```suggestion
    model: googleAI.model('gemini-flash-lite-latest'),
```

pavelgj (approved):

js/testapps/basic-gemini/src/index.ts
ifielker: yes, but I specifically wanted to try out 3.1-flash-lite rather than whatever -latest is pointed to. Don't worry, I do this for every new model (to make sure it actually works).

PR #5297, opened 2026-05-11
feat(go/plugins/vertexai/modelgarden): add Mistral via per-model rawPredict

cabljac:
> ⚠️ **Draft, stacked on #5296.** The compat_oai-based half (Claude Opus 4.5, Llama, shared scaffolding) lives there; this PR adds the Mistral / Codestral plugin on top. Base is set to the compat_oai branch — once #5296 merges I will rebase onto `main` and mark this Ready for Review.

Adds a Mistral / Codestral plugin to the Vertex AI Model Garden Go plugin. Vertex does not serve Mistral via the OpenAI-compatible `/endpoints/openapi/chat/completions` endpoint that `compat_oai` uses for Llama; that endpoint returns `400 FAILED_PRECONDITION` even with a subscribed project. Mistral is served only at per-model `…/publishers/mistralai/models/<id>:rawPredict` (or `:streamRawPredict` for SSE), and the rawPredict response is already OpenAI-shaped JSON.

## Approach

`mistral_transport.go` introduces an `http.RoundTripper` that wraps the oauth2 transport, intercepts outbound `/chat/completions` requests, reads `model` + `stream` from the body, rewrites the URL to the per-model rawPredict (or streamRawPredict) path, and delegates back. The body bytes are preserved (incl. `tools`, `response_format`, `stream_options`), `GetBody` is set for openai-go retries, and the model id is `url.PathEscape`-encoded to prevent path / query injection. The inner oauth2 transport still adds the Bearer token, so authentication is unchanged.

`MistralModels` keys are bare ids (e.g. `mistral-small-2503`), so the openai-go SDK already serialises the form rawPredict expects. `MistralModel(g, id)` and `Mistral.DefineModel(name, opts)` strip an optional `mistralai/` prefix as a convenience for callers used to publisher-qualified ids.

Llama is unchanged from #5296 — it still uses the OpenAI-compat endpoint. Only Mistral takes the rawPredict detour.

## Models

- `mistral-small-2503`, `mistral-medium-3`, `mistral-ocr-2505`, `codestral-2` (all under the `mistralai` Vertex publisher).

## Tests

- `mistral_transport_test.go` — pure in-process tests covering rawPredict / streamRawPredict URL rewrites, publisher-prefix strip, non-chat passthrough, missing-model error, body field preservation (`tools`, `response_format`), `GetBody` population, URL escaping of malformed model ids, and the `io.ReadAll` error path closing the body.
- `mistral_live_test.go` — basic generation, codestral generation, bare-id lookup, and streaming through rawPredict (credential-gated).
- Mistral subtests added to `define_model_test.go` (pre-Init) and `internal_test.go` (nil opts).
- Mistral + Codestral flows added to `go/samples/modelgarden`.

## Testing

image
image

Supersedes the Mistral half of #5175.

### DIFF
--- a/go/plugins/vertexai/modelgarden/mistral_transport.go
+++ b/go/plugins/vertexai/modelgarden/mistral_transport.go
@@ -1,0 +1,125 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package modelgarden
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// mistralVertexTransport rewrites outbound OpenAI-shaped chat completion
+// requests to Vertex's per-model Mistral rawPredict (or streamRawPredict)
+// URLs. Vertex's OpenAI-compatible /endpoints/openapi/chat/completions does
+// not serve Mistral; only the publisher-qualified rawPredict path does. The
+// request and response bodies are already in OpenAI shape, so only the URL
+// needs to change. The inner RoundTripper (typically an oauth2.Transport)
+// continues to add the Bearer token.
+type mistralVertexTransport struct {
+	inner    http.RoundTripper
+	project  string
+	location string
+}
+
+// RoundTrip rewrites any …/chat/completions request to the Mistral
+// rawPredict URL for the model named in the request body. The body itself
+// is preserved byte-for-byte so the openai-go SDK's framing (stream flag,
+// stream_options, tools, response_format) reaches Vertex unchanged.
+func (t *mistralVertexTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// The openai-go SDK emits POST .../chat/completions for chat requests.
+	// Any other path (probes, future SDK changes) is passed through.
+	if !strings.HasSuffix(req.URL.Path, "/chat/completions") {
+		return t.inner.RoundTrip(req)
+	}
+
+	var bodyBytes []byte
+	if req.Body != nil {
+		var err error
+		bodyBytes, err = io.ReadAll(req.Body)
+		// Close unconditionally so a read error does not leak the underlying
+		// reader (openai-go has no way to recover the original body once we
+		// return — GetBody would not be set on the un-rewritten request).
+		closeErr := req.Body.Close()
+		if err != nil {
+			return nil, fmt.Errorf("modelgarden mistral transport: read body: %w", err)
+		}
+		if closeErr != nil {
+			return nil, fmt.Errorf("modelgarden mistral transport: close body: %w", closeErr)
+		}
+	}
+
+	model, stream, err := peekModelAndStream(bodyBytes)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


go/plugins/vertexai/modelgarden/mistral_transport.go
```diff
@@ -0,0 +1,125 @@
...
+
+	suffix := "rawPredict"
+	if stream {
+		suffix = "streamRawPredict"
+	}
+
+	// PathEscape the model id so a hostile/malformed id (e.g. one containing
+	// "/", "?", or "#") cannot inject extra path segments, query strings, or
+	// fragments into the outbound URL.
+	newURL, err := url.Parse(fmt.Sprintf(
```
adesinah, line 89: Same issue as pointed out for VTO here: https://github.com/genkit-ai/genkit/pull/5177/changes#r3277351735
Although, from research, Mistral on Vertex today appears to only be available in regional locations
  (us-central1, europe-west4), so this may never bite in practice — but the inconsistency is real.

go/plugins/vertexai/modelgarden/models.go
```diff
@@ -137,3 +137,31 @@ var LlamaModels = map[string]ai.ModelOptions{
...
+// rawPredict URLs where the publisher lives in the URL path, not the model
+// field; the Mistral plugin's HTTP transport assembles those URLs. The
+// "mistralai/" prefix is still accepted by MistralModel as a convenience for
+// callers used to the publisher-qualified form.
+var MistralModels = map[string]ai.ModelOptions{
+	"mistral-medium-3": {
+		Label:    "Mistral Medium 3",
+		Supports: &internal.BasicText,
+	},
+	"mistral-ocr-2505": {
```
adesinah: This does not look right to me - basically, we are saying "OCR with media disabled". I believe the genkit middleware will reject any media part in a request to this model. But the entire purpose of `mistral-ocr-2505 `is OCR — ingesting images and returning text.

That said, should we be using `Multimodal` here?

adesinah: In addition to the above comment, while testing the feature, here are my findings: 
- The OpenAI chat-completions request body ({messages: [{content: [{type:image_url,...}]}]}) is not the schema Mistral OCR accepts.
- Mistral OCR wants something like {document: {type:document_url, document_url:"..."}}.
- Vertex correctly returns 400 when sent the wrong shape — no silent failure, just a hard reject.
- No amount of capability-flag tuning fixes this. The transport itself can't bridge two different request schemas.

I  recommend dropping OCR from this PR and tracking it as a follow-up: `&internal.Multimodal `(the only configuration that lets a media part reach the transport at all), calling the model with a properly attached image part returns `400` Bad Request from Vertex. Mistral OCR on Vertex uses a document-ingestion request schema (`{document: {type, document_url|document_base64}}`) that does not match the OpenAI chat-completions shape (`{messages: [{content: [{type:image_url, image_url}]}]}`) emitted by compat_oai and forwarded by mistralVertexTransport. The transport plumbing is sound for chat models; it cannot serve OCR. OCR support would require a separate code path with its own request serializer and response parser, similar to how `googlegenai.Lyria` and `googlegenai.VirtualTryOn` each have dedicated generate* functions.

adesinah (changes requested):
The PR looks good. I left some comments.

cabljac:
@adesinah dropped mistral-ocr-2505 in 9f252b2ee.

Double-checked: Vertex Mistral OCR accepts `{document:{type,document_url|document_base64}}` on `:rawPredict`, distinct from chat-completions. The `@mistralai/mistralai-gcp` SDK exposes a separate `client.ocr.process()` namespace for it. The transport in this PR only rewrites the URL; the body stays OpenAI chat-completions, so OCR would 400 with no useful signal.

The JS plugin has the same issue (it also lists `mistral-ocr-2505` but only calls `client.chat.complete()`), so "JS parity" actually means "both broken". Will file a follow-up for proper OCR support via a dedicated generator and link both Go + JS bugs.

Also: branch is rebased onto the latest `feat/go-modelgarden-compat-oai` (which now carries 4d1c13050, the date-versioned alias fix from #5296).

adesinah (approved):
Looks good!

adesinah (approved):
Looks good to me!

PR #5306, opened 2026-05-12
fix(client): make space after "data:" optional in SSE parser

### LINKED ISSUE #3728: [JS] streamFlow fails to parse events without a space after “data:” (non-compliant with SSE spec)
**Describe the bug**
The Genkit client `streamFlow` fails to parse Server-Sent Events (SSE) where the data line is emitted as `data:foo` (no space after the colon).
According to the [SSE specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), whitespace after the colon is optional, but the Genkit parser currently can't parse events without the space.

For example, a compliant Spring WebFlux controller using the built-in ServerSentEventHttpMessageWriter serializes as:

`data:foo`

However, the Genkit client expects:

`data: foo`

**To Reproduce**
1.	Run a simple Spring Boot / WebFlux controller:
```java
@RestController
public class SseController {
    @PostMapping("/stream")
    public Flux<String> stream() {
        return Flux.just("{\"message\":\"foo\"}");
    }
}
```
2.	Connect to /stream using the Genkit client that listens to an SSE endpoint:
```typescript
import { streamFlow } from 'genkit/beta/client'

const flow = streamFlow({
    url: '/stream'
})

for await (const chunk of flow.stream) {
    console.log(chunk)
}
```
3.	Observe that no events are processed, error in console that JSON cannot be parsed:
```
SyntaxError: Unexpected non-whitespace character after JSON at position 9 (line 1 column 10)
    at JSON.parse (<anonymous>)
    at client.js:87:26
```

**Expected behavior**
The Genkit client should accept both `data:foo` and `data: foo` forms, per the SSE spec.
The parser should trim at most one optional leading space after the colon, not require it.

**Screenshots**
N/A – issue reproducible via console logs.

 - OS: macOS
 - Version 26 (Tahoe)

**Node version**
 -  v22.13.1

**Additional context**
The issue is in https://github.com/firebase/genkit/blob/main/js/genkit/src/client/client.ts#L125, where event is hardcoded.

serhiizghama:
## Problem

`streamFlow` (`js/genkit/src/client/client.ts`) strips a hardcoded `"data: "` (6 characters) from each SSE event before passing the rest to `JSON.parse`. Per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), the single space after `data:` is optional — sources are allowed to emit `data:foo`. When they do (e.g. Spring WebFlux's built-in `ServerSentEventHttpMessageWriter`), the current parser consumes the first character of the payload and `JSON.parse` throws:

```
SyntaxError: Unexpected non-whitespace character after JSON at position 9
```

Fixes #3728.

## Solution

Strip the `data:` prefix first, then strip at most one optional leading space — matching the SSE spec's "if value starts with a U+0020 SPACE character, remove it from value" rule.

## Testing

Added `tests/client_test.ts` covering both `data:{...}` (no space) and `data: {...}` (conventional) event forms. Local run:

```
$ pnpm test
# tests 127
# pass 127
# fail 0
```

### DIFF
--- a/js/genkit/src/client/client.ts
+++ b/js/genkit/src/client/client.ts
@@ -138,11 +138,17 @@ async function __flowRunEnvelope({
     }
     // If buffer includes the delimiter that means we are still receiving chunks.
     while (buffer.includes(__flowStreamDelimiter)) {
-      const chunk = JSON.parse(
-        buffer
-          .substring(0, buffer.indexOf(__flowStreamDelimiter))
-          .substring('data: '.length)
-      );
+      const event = buffer.substring(0, buffer.indexOf(__flowStreamDelimiter));
+      // Per the SSE spec, the space after `data:` is optional, so strip the
+      // prefix and then at most one leading space.
+      // https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation
+      let payload = event.startsWith('data:')
+        ? event.substring('data:'.length)
+        : event;
+      if (payload.startsWith(' ')) {
+        payload = payload.substring(1);
+      }
+      const chunk = JSON.parse(payload);
       if (chunk.hasOwnProperty('message')) {
         sendChunk(chunk.message);
       } else if (chunk.hasOwnProperty('result')) {
--- a/js/genkit/src/client/client.ts
+++ b/js/genkit/src/client/client.ts
@@ -149,6 +149,9 @@ async function __flowRunEnvelope({
         payload = payload.substring(1);
       }
       const chunk = JSON.parse(payload);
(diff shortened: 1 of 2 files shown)


serhiizghama:
Friendly ping — this is still failing for SSE sources that omit the space after `data:`. The fix and tests are ready; would appreciate a review when you get a chance.

serhiizghama:
Gentle nudge — the missing space after `data:` in the SSE parser still trips real-world servers. PR's small and green. Happy to adjust if you'd like.

PR #5335, opened 2026-05-15
feat(js/plugins/google-genai): Better debug traces

ifielker:
Also removed deprecated Gemma3 models, fixed a bug with Veo video urls and fixed embedder default dimensions.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/common/utils.ts
+++ b/js/plugins/google-genai/src/common/utils.ts
@@ -35,6 +35,45 @@ import {
   isObject,
 } from './types.js';
 
+export function buildTraceMetadataInput(
+  url: string,
+  fetchOptions: RequestInit,
+  traceOptions: {
+    request?: unknown;
+    model?: string;
+    clientOptions?: { timeout?: number };
+  }
+): Record<string, unknown> {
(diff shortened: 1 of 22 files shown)


js/plugins/google-genai/src/common/utils.ts
```diff
@@ -35,6 +35,45 @@ import {
...
+  };
+
+  if (safeHeaders['x-goog-api-key']) {
+    safeHeaders['x-goog-api-key'] = redactString(
+      safeHeaders['x-goog-api-key']
+    )!;
+  }
+  if (safeHeaders['Authorization']) {
+    safeHeaders['Authorization'] = redactString(safeHeaders['Authorization'])!;
+  }
```
ssbushi, line 61: Not sure if this will expand in the future, but should these headers be part of a const collection?

ssbushi (approved):

js/plugins/google-genai/src/common/utils.ts
ifielker: If it expands, I'll add them to a const collection.

PR #5342, opened 2026-05-18
fix(ai): retry middleware honors abortSignal and skips AbortError

### LINKED ISSUE #5331: retry middleware silently retries AbortError, ignoring abortSignal passed to generateStream
## Summary

The `retry` model middleware (from `genkit/model/middleware`) catches and retries **any** non-`GenkitError`, which includes `AbortError`. This means that when a caller passes `abortSignal` to `generateStream` and then aborts it mid-stream, the retry middleware intercepts the resulting `AbortError` and sleeps through exponential backoff before retrying the model call — completely defeating the abort.

## Steps to Reproduce

1. Call `ai.generateStream({ ..., abortSignal: mySignal, use: [retry({ maxRetries: 5 })] })`
2. Abort `mySignal` while a model call is in-flight
3. The fetch throws `AbortError`, but `retry` catches it and retries up to 5 times with exponential backoff (up to 60s total sleep) before the error surfaces

## Root Cause

In `packages/ai/src/model/middleware.ts`, the retry logic is:

```ts
} else {
  shouldRetry = true; // retries ANY non-GenkitError, including AbortError
}
if (shouldRetry) {
  onError?.(error, i + 1);
  await new Promise((resolve) => setTimeout(resolve, delay)); // not signal-aware
  continue;
}
```

`AbortError` is not a `GenkitError`, so `shouldRetry` is always `true` for it. The backoff sleep is also not signal-aware, so even if the caller detects the retry it still waits out the full delay.

## Suggested Fix

Check `abortSignal?.aborted` before deciding to retry, and short-circuit immediately if aborted. The `abortSignal` is available on the request object passed to the middleware:

```ts
// Option A: check signal on the request before retrying
if (req.abortSignal?.aborted) {
  throw error; // don't retry aborts
}

// Option B: special-case AbortError by name
if ((error as { name?: string }).name === 'AbortError') {
  throw error;
}
```

Option B is simpler and doesn't require threading the signal reference through. The backoff sleep should also be made signal-aware:

```ts
await new Promise((resolve, reject) => {
  const timer = setTimeout(resolve, delay);
  req.abortSignal?.addEventListener('abort', () => {
    clearTimeout(timer);
    reject(new DOMException('Aborted', 'AbortError'));
  }, { once: true });
});
```

## Impact

Any application that uses `abortSignal` with `generateStream` and has the `retry` middleware in the stack (which is the recommended pattern in the docs) will find that user cancellation does not stop model generation for up to `maxRetries × maxDelayMs` of additional time.

## Workaround

Replace the built-in `retry` middleware with a custom abort-aware version that checks `signal.aborted` before each attempt and during the backoff sleep.

## Environment

- genkit version: 1.33.0
- `@genkit-ai/ai` version: 1.33.0

serhiizghama:
## Problem

The \`retry\` model middleware retries **any** non-\`GenkitError\`, including \`AbortError\`. When a caller passes \`abortSignal\` to \`generateStream\`/\`generate\` and aborts mid-stream, the middleware intercepts the resulting \`AbortError\` and sleeps through exponential backoff before retrying the model call — completely defeating the abort. The backoff sleep itself is also not signal-aware, so even if a later attempt detects the abort, the user still waits out the full delay.

Fixes #5331.

## Solution

- Convert \`retry\` to a \`ModelMiddlewareWithOptions\` so it can read \`opts.abortSignal\`.
- Short-circuit at three points:
  - Before every attempt if the signal is already aborted.
  - On any caught \`AbortError\` (matched by \`error.name === 'AbortError'\` or \`error.code === 'ABORT_ERR'\`).
  - During the backoff sleep — \`abortableDelay\` clears the timer and rejects with \`signal.reason\` when the signal fires.
- Existing \`__setTimeout\` hook is preserved so the test helper (\`TEST_ONLY.setRetryTimeout\`) still works.

The retry function's public return type changes from \`ModelMiddleware\` to \`ModelMiddlewareWithOptions\`. Both are valid \`ModelMiddlewareArgument\` entries, so usage via \`use: [retry()]\` is unaffected.

## Testing

Three new cases in \`tests/model/middleware_test.ts\`:

- \`should not retry on AbortError\` — model throws an \`AbortError\`; verifies the model is called exactly once and no backoff sleep runs.
- \`should not retry once abortSignal is already aborted\` — pre-aborted signal; verifies retry bails before sleeping.
- \`should abort the backoff sleep when signal fires mid-retry\` — uses the real \`setTimeout\` so the abort listener fires; verifies the 10s sleep is interrupted and the second attempt is skipped.

All 38 retry/fallback/system-prompt/context tests in the suite pass:

\`\`\`
pnpm test:single tests/model/middleware_test.ts
# tests 38 / pass 38 / fail 0
\`\`\`

\`pnpm check\` (tsc) is clean.

### DIFF
--- a/js/ai/src/model/middleware.ts
+++ b/js/ai/src/model/middleware.ts
@@ -24,6 +24,7 @@ import type {
   MessageData,
   ModelInfo,
   ModelMiddleware,
+  ModelMiddlewareWithOptions,
   Part,
 } from '../model.js';
 import { resolveModel } from '../model.js';
@@ -334,7 +335,9 @@ const DEFAULT_FALLBACK_STATUSES: StatusName[] = [
  * });
  * ```
  */
-export function retry(options: RetryOptions = {}): ModelMiddleware {
+export function retry(
+  options: RetryOptions = {}
+): ModelMiddlewareWithOptions {
   const {
     maxRetries = 3,
     statuses = DEFAULT_RETRY_STATUSES,
@@ -345,18 +348,24 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
     onError,
   } = options;
 
-  return async (req, next) => {
+  return async (req, opts, next) => {
+    const abortSignal = opts?.abortSignal;
     let lastError: any;
     let currentDelay = initialDelayMs;
     for (let i = 0; i <= maxRetries; i++) {
+      if (abortSignal?.aborted) {
+        throw abortSignal.reason ?? new Error('Aborted');
+      }
       try {
-        return await next(req);
+        return await next(req, opts);
       } catch (e) {
         lastError = e;
         const error = e as Error;
         if (i < maxRetries) {
           let shouldRetry = false;
-          if (error instanceof GenkitError) {
+          if (isAbortError(error) || abortSignal?.aborted) {
+            shouldRetry = false;
+          } else if (error instanceof GenkitError) {
             if (statuses.includes(error.status)) {
               shouldRetry = true;
             }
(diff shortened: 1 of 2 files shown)


pavelgj:
I'm fixing this as part of a bigger change to the retry middleware (in the new `@genkit-ai/middleware` package) -- https://github.com/genkit-ai/genkit/pull/5343

PR #5343, opened 2026-05-19
feat(js): honor Retry-After header & never retry AbortError/ToolInterruptError

### LINKED ISSUE #5270: retry middleware — RESOURCE_EXHAUSTED retries use exponential backoff without consulting Retry-After header
## Finding

The `retry` middleware in `js/plugins/middleware/src/retry.ts` includes `RESOURCE_EXHAUSTED` in its default retry statuses but uses pure exponential backoff without consulting the provider’s `Retry-After` header.

When a provider returns HTTP 429 with `Retry-After: 60`, the retry middleware fires at `initialDelayMs` (default 1000ms) intervals — inside the provider cooldown window. All retries fail until the window expires naturally.

## Evidence

`DEFAULT_RETRY_STATUSES` includes `RESOURCE_EXHAUSTED`.

Retry delay computation:

```typescript
let delay = currentDelay;
if (!noJitter) {
  delay = delay + 1000 * Math.pow(2, i) * Math.random();
}
await new Promise((resolve) => __setTimeout(resolve, delay));
currentDelay = Math.min(currentDelay * backoffFactor, maxDelayMs);
```

No `Retry-After` header consulted at any point.

## The correct behavior

When `RESOURCE_EXHAUSTED` is received, check the response headers for `Retry-After`. If present, use that value as the minimum delay before retrying. Fall back to exponential backoff only when the header is absent.

## Related pattern

`retry-after-ignored-exponential-backoff`

Corpus reference: https://github.com/SirBrenton/pitstop-truth

### LINKED ISSUE #5331: retry middleware silently retries AbortError, ignoring abortSignal passed to generateStream

pavelgj:
Fixes #5270 
Fixes #5331

### What

Make the retry middleware to respect the `Retry-After` HTTP header from providers, and prevents retrying errors that should never be retried.

### Changes

**Core (`@genkit-ai/core`)**
- Added `ErrorResponseMetadata` interface with `retryAfterMs` and `headers` fields
- Added optional `responseMetadata` property to `GenkitError` (not serialized in `toJSON()` — in-process only)

**Retry middleware (`@genkit-ai/middleware`)**
- Uses `responseMetadata.retryAfterMs` as a floor for the retry delay (`Math.max(computedDelay, retryAfterMs)`)
- Never retries `AbortError` (from `AbortController`) or `ToolInterruptError` (control-flow signal), regardless of error type
- Added 4 new unit tests covering both behaviors

**Provider plugins** — all now capture `Retry-After` from HTTP responses and pass it as `responseMetadata` on `GenkitError`:
- **google-genai**: `googleai/client.ts` and `vertexai/client.ts` read from `Response.headers`
- **compat-oai**: reads from OpenAI SDK `APIError.headers`
- **anthropic**: added try/catch around `runner.run()` to convert Anthropic SDK `APIError` → `GenkitError` with status mapping (429→RESOURCE_EXHAUSTED, 401→UNAUTHENTICATED, etc.) and `Retry-After` propagation

### Notes
- `parseRetryAfterMs` supports both delay-seconds and HTTP-date formats (RFC 7231 §7.1.3), with guards against empty/whitespace input
- `retryAfterMs` uses `!== undefined` checks to preserve `Retry-After: 0` as a valid "retry immediately" signal
- The anthropic plugin previously let raw SDK errors propagate; it now normalizes them to `GenkitError` with proper status codes

### DIFF
--- a/js/plugins/anthropic/src/models.ts
+++ b/js/plugins/anthropic/src/models.ts
@@ -15,13 +15,19 @@
  * limitations under the License.
  */
 
+import { APIError } from '@anthropic-ai/sdk';
 import type {
   GenerateRequest,
   GenerateResponseData,
   ModelReference,
   StreamingCallback,
 } from 'genkit';
-import { z } from 'genkit';
+import {
+  GenkitError,
+  z,
+  type ErrorResponseMetadata,
+  type StatusName,
+} from 'genkit';
 import type { GenerateResponseChunkData, ModelAction } from 'genkit/model';
 import { modelRef } from 'genkit/model';
 import { model } from 'genkit/plugin';
@@ -36,6 +42,18 @@ import {
   type ClaudeRunnerParams,
 } from './types.js';
 import { checkModelName, isKnownKey } from './utils.js';
+
+/**
+ * Parses a `Retry-After` header value into milliseconds.
+ * Supports delay-seconds and HTTP-date formats (RFC 7231 §7.1.3).
+ */
+function parseRetryAfterMs(value: string): number | undefined {
+  const seconds = Number(value);
+  if (!isNaN(seconds) && seconds >= 0) return seconds * 1000;
+  const date = new Date(value);
+  if (!isNaN(date.getTime())) return Math.max(0, date.getTime() - Date.now());
+  return undefined;
+}
 
 // This contains all the Anthropic config schema types
 type ConfigSchemaType = AnthropicConfigSchemaType;
@@ -148,11 +166,53 @@ export function claudeRunner<TConfigSchema extends z.ZodTypeAny>(
     const runner = isBeta
       ? (betaRunner ??= new BetaRunner(runnerParams))
       : (stableRunner ??= new Runner(runnerParams));
-    return runner.run(normalizedRequest, {
-      streamingRequested,
-      sendChunk,
-      abortSignal,
-    });
+    try {
+      return await runner.run(normalizedRequest, {
+        streamingRequested,
+        sendChunk,
+        abortSignal,
+      });
+    } catch (e) {
+      if (e instanceof APIError) {
+        let status: StatusName = 'UNKNOWN';
+        switch (e.status) {
+          case 429:
+            status = 'RESOURCE_EXHAUSTED';
+            break;
+          case 401:
+            status = 'UNAUTHENTICATED';
+            break;
+          case 403:
+            status = 'PERMISSION_DENIED';
+            break;
+          case 400:
+            status = 'INVALID_ARGUMENT';
+            break;
+          case 500:
+            status = 'INTERNAL';
+            break;
+          case 503:
+            status = 'UNAVAILABLE';
+            break;
+          case 529:
+            status = 'UNAVAILABLE';
+            break;
+        }
+        const retryAfterHeader = e.headers?.get?.('retry-after');
+        const retryAfterMs = retryAfterHeader
+          ? parseRetryAfterMs(retryAfterHeader)
+          : undefined;
+        const responseMetadata: ErrorResponseMetadata | undefined = retryAfterMs
+          ? { retryAfterMs }
+          : undefined;
+        throw new GenkitError({
+          status,
+          message: e.message,
+          responseMetadata,
+        });
+      }
+      throw e;
+    }
   };
 }
 
--- a/js/plugins/anthropic/src/models.ts
+++ b/js/plugins/anthropic/src/models.ts
@@ -48,6 +48,7 @@ import { checkModelName, isKnownKey } from './utils.js';
  * Supports delay-seconds and HTTP-date formats (RFC 7231 §7.1.3).
  */
 function parseRetryAfterMs(value: string): number | undefined {
+  if (!value || !value.trim()) return undefined;
   const seconds = Number(value);
   if (!isNaN(seconds) && seconds >= 0) return seconds * 1000;
   const date = new Date(value);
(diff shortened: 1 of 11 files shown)


js/plugins/anthropic/src/models.ts
```diff
@@ -37,6 +43,19 @@ import {
 } from './types.js';
 import { checkModelName, isKnownKey } from './utils.js';
 
+/**
+ * Parses a `Retry-After` header value into milliseconds.
+ * Supports delay-seconds and HTTP-date formats (RFC 7231 §7.1.3).
+ */
+function parseRetryAfterMs(value: string): number | undefined {
```
ifielker, line 50: Do we have utils that plugins can use? Because I feel like otherwise we are just copy/pasting this thing around everywhere... and if it is subtly wrong somehow... then we will need to change it everywhere... Can we share it in a genkit/utils kind of thing? Or better, can we make something that will take the  error and give back a responseMetadata object (based on headers, status... other?) ? So we can put whatever we need to in there and not have to change it everywhere when there's a new error handling whatever?

js/plugins/compat-oai/src/model.ts
```diff
@@ -632,9 +646,18 @@ export function openAIModelRunner(
...
+          (e.headers as any)?.['retry-after'];
+        const retryAfterMs = retryAfterHeader
+          ? parseRetryAfterMs(retryAfterHeader)
+          : undefined;
+        const responseMetadata: ErrorResponseMetadata | undefined =
+          retryAfterMs !== undefined ? { retryAfterMs } : undefined;
         throw new GenkitError({
           status,
           message: e.message,
+          responseMetadata,
```
ifielker, line 660: Then this just becomes something like `import { metadataFromError } from genkit/utils` and then `responseMetadata: metadataFromError(e),` and we don't need all the other stuff everywhere?

apascal07 (approved):

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -665,4 +665,29 @@ export function isKnownKey<T extends object>(
...
+  }
+  // Try as HTTP-date (e.g., "Mon, 19 May 2026 12:00:00 GMT")
+  const date = new Date(value);
+  if (!isNaN(date.getTime())) {
+    return Math.max(0, date.getTime() - Date.now());
+  }
+  return undefined;
+}
+
+export const TEST_ONLY = { aggregateResponses, parseRetryAfterMs };
```
ifielker: Did you forget the tests? Why add it to TEST_ONLY?

js/plugins/anthropic/src/models.ts
pavelgj: I know... I thought about this. I decided against putting things like this into `genkit` package. Really don't want to get into maintaining util like this in genkit. Opted for "acceptable level of duplication" for now.

js/plugins/google-genai/src/common/utils.ts
pavelgj: looks like leftovers some something... the fn is actually exported, so removed from TEST_ONLY but also added tests.

js/plugins/anthropic/src/models.ts
```diff
@@ -148,11 +167,52 @@ export function claudeRunner<TConfigSchema extends z.ZodTypeAny>(
...
+            break;
+          case 500:
+            status = 'INTERNAL';
+            break;
+          case 503:
+            status = 'UNAVAILABLE';
+            break;
+          case 529:
+            status = 'UNAVAILABLE';
+            break;
```
MichaelDoyle, line 198: nit: up to you

```suggestion
          case 503:
          case 529:
            status = 'UNAVAILABLE';
            break;
```

MichaelDoyle (approved):

PR #5376, opened 2026-05-21
fix(js/ai): correctly tag registered interrupt as a non-dynamic tool

pavelgj:


MichaelDoyle:
Can you add interrupts to `/tool.v2`? Or does that not make sense.

I think DevUI is going to skip these if it detects any other v2 tools (we assume all are of a single type).

ifielker (approved):

pavelgj:
> Can you add interrupts to `/tool.v2`? Or does that not make sense.
> 
> I think DevUI is going to skip these if it detects any other v2 tools (we assume all are of a single type).

Hmm... I'm not sure if it's a breaking change to convert to interrupt to a tool.v2... will have to defer that.
But, I don't think the Dev UI behaviour is correct. That determination must be done per-tool. Today we only define `tool.v2` for multipart tools, regular tools are still defined as `tool`

MichaelDoyle:
I’ll double check, could be mistaken on the logic. I thought even tool (v1)
got wrapped as a tool.v2 (i.e duplicated) though. In fact, I think gemini
pointed that out to me when I reviewed this PR.

PR #5377, opened 2026-05-21
fix(js/middleware/filesystem): return tool response on filesystem tool errors

pavelgj:
Return a proper tool response instead of injecting a user-role message when a filesystem tool call fails. Anthropic requires a tool response after every tool call — injecting a user message breaks its protocol. Updates tests to assert on tool-role messages with error output.

### DIFF
--- a/js/plugins/middleware/src/filesystem.ts
+++ b/js/plugins/middleware/src/filesystem.ts
@@ -115,25 +115,19 @@ export const filesystem: GenerateMiddleware<typeof FilesystemOptionsSchema> =
             if (e.name === 'ToolInterruptError') throw e;
 
             if (filesystemToolNames.includes(req.toolRequest.name)) {
-              const errorPart = {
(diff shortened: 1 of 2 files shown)


MichaelDoyle (approved):

MichaelDoyle:
re: `when a filesystem tool call fails`, should we return a proper tool response when it succeeds too? meaning, not inject a user message? e.g. json-stringified?

My rational is consistency in (a) model behavior and (b) observability; right now, the tool response appears "missing" even though it's not.

pavelgj:
> re: `when a filesystem tool call fails`, should we return a proper tool response when it succeeds too? meaning, not inject a user message? e.g. json-stringified?
> 
> My rational is consistency in (a) model behavior and (b) observability; right now, the tool response appears "missing" even though it's not.

are you talking about read_file? That's a slightly different problem... and we do return the tool response, it just doesn't contain the file contents... 
I'm not against it... but separate thing.

PR #5380, opened 2026-05-21
feat(cli): Add trace:get CLI command

### LINKED ISSUE #5379: [Dev UI] Add trace lookup command in Genkit CLI

ssbushi:
Fixes: https://github.com/genkit-ai/genkit/issues/5379

Usage: 

```
$ genkit trace:get 5aa367e7d3895db78339f6148aab29b4
Telemetry API running on http://localhost:4033
{
  "traceId": "5aa367e7d3895db78339f6148aab29b4",
  "displayName": "generate",
  "startTime": 1778700071212,
  "endTime": 1778700072191.0662,
....
```

```
$ genkit trace:list -l 2 --name generate
Telemetry API running on http://localhost:4033
Found 2 traces:

ID:       34aeabf91dc21a1c8d43a9482d45657b
Type:     util
Name:     generate
Status:   error
Time:     5/26/2026, 4:17:15 PM
Duration: 328.314453125ms
Input:    {"model":"googleai/gemini-2.0-flash-lite","messages":[{"role":"user","content":[{"text":"asdasd"}...
---
ID:       558dbdc4477acbe9c91db5b73dde2b67
Type:     util
Name:     generate
Status:   success
Time:     5/26/2026, 4:16:57 PM
Duration: 2094.606689453125ms
Input:    {"model":"googleai/gemini-2.0-flash","messages":[{"role":"user","content":[{"text":"asdasd"}]}],"...
Output:   {"message":{"role":"model","content":[{"text":"I am an AI language model, and I don't have access...
---
```

Checklist (if applicable):

### DIFF
--- a/genkit-tools/cli/src/commands/trace-list.ts
+++ b/genkit-tools/cli/src/commands/trace-list.ts
@@ -47,9 +47,9 @@ export interface TraceListOptions {
 export const traceList = new Command('trace:list')
   .description('list traces')
   .option('-l, --limit <number>', 'limit the number of returned traces', '15')
-  .option('--status <status>', 'filter by trace status')
-  .option('--type <type>', 'filter by trace type')
-  .option('--name <name>', 'filter by trace name')
+  .option('--status <status>', 'filter by root span status')
+  .option('--type <type>', 'filter by root span type')
+  .option('--name <name>', 'filter by root span name')
   .option('--continuation-token <token>', 'continuation token for pagination')
   .action(async (options: TraceListOptions) => {
     const projectRoot = await findProjectRoot();
--- a/genkit-tools/cli/src/commands/trace-list.ts
+++ b/genkit-tools/cli/src/commands/trace-list.ts
@@ -1,0 +1,149 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { TraceQueryFilter } from '@genkit-ai/tools-common';
+import type { BaseRuntimeManager } from '@genkit-ai/tools-common/manager';
+import {
+  findProjectRoot,
+  logger,
+  stackTraceSpans,
+} from '@genkit-ai/tools-common/utils';
+import { Command } from 'commander';
+import { runWithManager } from '../utils/manager-utils';
+
+function formatTruncated(val: any, maxLen = 100): string {
+  if (val === undefined) return '';
+  const parsed = typeof val === 'string' ? val : JSON.stringify(val);
+  return parsed.length > maxLen
+    ? parsed.substring(0, maxLen - 3) + '...'
+    : parsed;
+}
+
+export interface TraceListOptions {
+  limit: string;
+  status?: string;
+  type?: string;
+  name?: string;
+  continuationToken?: string;
+}
+
+/** Command to list traces. */
+export const traceList = new Command('trace:list')
+  .description('list traces')
+  .option('-l, --limit <number>', 'limit the number of returned traces', '15')
+  .option('--status <status>', 'filter by trace status')
+  .option('--type <type>', 'filter by trace type')
+  .option('--name <name>', 'filter by trace name')
+  .option('--continuation-token <token>', 'continuation token for pagination')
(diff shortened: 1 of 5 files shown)


pavelgj (approved):
LGTM

A future feature request: add `--json` flag for all/most CLI commands to output all data in JSON format. It is super useful because one can pipe output to `jq` and filter/transform the results. Handy when writing scripts to automate stuff.

genkit-tools/cli/src/commands/trace-list.ts
```diff
@@ -0,0 +1,149 @@
...
+  continuationToken?: string;
+}
+
+/** Command to list traces. */
+export const traceList = new Command('trace:list')
+  .description('list traces')
+  .option('-l, --limit <number>', 'limit the number of returned traces', '15')
+  .option('--status <status>', 'filter by trace status')
+  .option('--type <type>', 'filter by trace type')
+  .option('--name <name>', 'filter by trace name')
```
MichaelDoyle: Are these really properties of the *trace*?

Are these span name/type/status? Of the root span?

genkit-tools/cli/src/commands/trace-list.ts
```diff
@@ -0,0 +1,149 @@
...
+  limit: string;
+  status?: string;
+  type?: string;
+  name?: string;
+  continuationToken?: string;
+}
+
+/** Command to list traces. */
+export const traceList = new Command('trace:list')
+  .description('list traces')
```
MichaelDoyle, line 48: What is the sort order? By time? desc? asc? Let's make sure the command/help text describes that properly.

MichaelDoyle:
> LGTM
> 
> A future feature request: add `--json` flag for all/most CLI commands to output all data in JSON format. It is super useful because one can pipe output to `jq` and filter/transform the results. Handy when writing scripts to automate stuff.

I think this will be very important potentially for agents as well.

genkit-tools/cli/src/commands/trace-list.ts
ssbushi: Technically, OTEL has no trace construct, we created that for Genkit. Trace is a span-tree right? Using these attributes make sense to me. This is also what we do in the TraceViewer page in the Dev UI, which is what this is based on.

genkit-tools/cli/src/commands/trace-list.ts
ssbushi: The sort order is not controlled in this command, but I will make a note of it.

genkit-tools/cli/src/commands/trace-list.ts
MichaelDoyle: Say what now? :) 

No, I would not say trace is a genkit concept, I would say it's an Otel concept, which you've correctly defined as a tree of spans.

https://opentelemetry.io/docs/concepts/signals/traces/

We should be precise, because we want agents to be able to understand it _exactly_.

genkit-tools/cli/src/commands/trace-list.ts
MichaelDoyle: Thanks - it is sorted, though, right? By time? Just that there's no sort criteria exposed?

genkit-tools/cli/src/commands/trace-list.ts
pavelgj: Technically the status/name/type are on the root span, not on the trace... but... we don't make that distinction clear to the user, they are high-level Genkit concepts, I think it's easier to attribute these to the trace, for simplicity.

The correct wording would be:

```
  .option('--status <status>', 'filter by status attribute value on the root span')
  .option('--type <type>', 'filter by type attribute value on the root span')
  .option('--name <name>', 'filter by name attribute value on the root span')
```

but I'm not a fan....

pavelgj: another option:

```
  .option('--status <status>', 'filter by root action status')
  .option('--type <type>', 'filter by root action type')
  .option('--name <name>', 'filter by root action name')
```

pavelgj:
```suggestion
  .option('--status <status>', 'filter by root action status')
  .option('--type <type>', 'filter by root action type')
  .option('--name <name>', 'filter by root action name')
```

ssbushi: I think root action is even more confusing. The agent should not care about the details of how Genkit behaves and builds traces (via actions etc). I share the same sentiment as pavel here https://github.com/genkit-ai/genkit/pull/5380#discussion_r3311388371

@MichaelDoyle Well, Pavel has drilled into me several times that Traces are not an OTEL thing. Not saying that they are completely new to Genkit, but that we do not use traces as designed by the OTEL spec. Maybe that is a moot point to the current conversation, so I will concede that I misspoke earlier. 

Regardless, I think the simpler terminology is the better approach -- if trace is a span-tree and has no inherent trace attributes, what is the difference between saying trace vs root span? Also, while these CLI changes are primarily for agentic use, these can also be used by devs and should not too pedantic in terminology?

ssbushi: Changing to `root span` 🏳️

PR #5387, opened 2026-05-22
fix(js): rewrite relative .js specifiers to .mjs in ESM output

### LINKED ISSUE #5386: [JS] genkit ESM entrypoint re-exports from CJS, breaking z under Vite SSR (Angular SSR)
**Describe the bug**

`genkit/lib/index.mjs` wildcard-re-exports from `./common.js` (CJS) instead of the sibling `./common.mjs` (ESM) the same build already emits. Under Vite's SSR module runner — used by Angular's SSR dev server — named bindings from `genkit` (`z`, etc.) resolve to `undefined`, and the first call crashes with:

```
TypeError: Cannot read properties of undefined (reading 'object')
    at z.object(...)
    at SSRCompatModuleRunner.directRequest (vite/dist/node/module-runner.js)
```

Native Node ESM resolves the same import correctly, so the bug only surfaces in Vite-SSR-style runners.

**To Reproduce**

1. Scaffold an Angular SSR app and install Genkit:
   ```bash
   ng new --ssr repro && cd repro
   npm install genkit @genkit-ai/google-genai @genkit-ai/express
   ```
2. Create `src/genkit/flow.ts`:
   ```ts
   import { genkit, z } from 'genkit';

   const ai = genkit({ plugins: [] });
   const Schema = z.object({ foo: z.string() }); // crashes here
   ```
3. Register a route in `src/server.ts`:
   ```ts
   import { expressHandler } from '@genkit-ai/express';
   import express from 'express';
   // ...
   app.use(express.json());
   app.post('/api/flow', expressHandler(/* a flow using the schema above */));
   ```
4. `npm start`, then `curl -X POST http://localhost:4200/api/flow -H 'Content-Type: application/json' -d '{"data":{}}'`.

Confirming it's Vite-SSR-specific, the same import works under plain Node ESM:
```bash
node --input-type=module -e "import { z } from 'genkit'; console.log(typeof z.object)"
# function
```

Root cause is visible in `node_modules/genkit/lib/index.mjs`:
```js
export * from "./common.js";   // CJS file — should be "./common.mjs"
```
`common.mjs` already exists in the same directory; the build (tsup) appears to emit the ESM entry without rewriting the wildcard re-export's extension.

**Expected behavior**

`import { z } from 'genkit'` should resolve `z` to the zod namespace under Vite's SSR module runner, the same way it does under native Node ESM.

The fix is a narrow build-config change: rewrite `export * from './foo.js'` → `export * from './foo.mjs'` in ESM-format output only. The `.mjs` files already exist; the rewrite just points the wildcard re-exports at them. Regular named imports (`import { … } from './foo.js'`) are deliberately left alone so the CJS-interop path still resolves for plugin source files that use extensionless relative imports.

**Screenshots**

N/A.

- OS: macOS (Darwin 24.6.0)
- Version: tested with `genkit@1.35.0`, `@angular/cli@21.2.12`, Vite (bundled with `@angular/build@21.2.x`)

**Node version**
- `node --version` → `v25.0.0` (also reproduced behavior consistent with Node 20+ ESM/CJS interop; bug is in Vite's runner, not Node)

**Additional context**

User-side workarounds while waiting for a fix:

1. **Recommended — import `z` from `zod` directly:**
   ```ts
   import { z } from 'zod'; // pin ^3.23.8 to match Genkit's transitive version
   ```
2. **Alternative — externalize Genkit in `angular.json`** so Vite delegates loading to Node:
   ```json
   "externalDependencies": ["genkit", "@genkit-ai/core", "@genkit-ai/ai", "@genkit-ai/express", "@genkit-ai/google-genai"]
   ```

Why Vite SSR breaks where Node doesn't: ECMAScript `export *` deliberately excludes the `default` export, and Vite's SSR module runner surfaces CJS module shapes differently than Node's native CJS-ESM interop, so named bindings re-exported through a wildcard from a CJS file aren't visible to the importer.

Likely also affected (not verified here): any framework using Vite's SSR module runner (e.g., Nuxt, SvelteKit). Confirmation from users on those stacks welcome.

chrisraygill:
## Summary

Fixes #5386. Tsup's `bundle: false` setting means esbuild transpiles each source file individually and preserves its import paths as-is, so source-level `from './foo.js'` came through unchanged in both CJS and ESM output. The emitted `lib/*.mjs` files were wildcard-re-exporting from sibling `./*.js` (CJS) files, which Vite's SSR module runner cannot statically expand into named bindings — `import { z } from 'genkit'` resolved to `undefined` and crashed at `z.object(...)` under Angular SSR.

This adds an esbuild plugin to `tsup.common.ts` that, **for ESM-format builds only**, rewrites quoted relative specifiers ending in `.js` to `.mjs` before compilation. The matching `.mjs` files are already emitted by tsup; this just points the ESM bundle at them.

## What changes / what doesn't

- **CJS output:** untouched (the plugin early-returns when `format !== 'esm'`).
- **ESM output:** internal cross-file specifiers go from `./foo.js` → `./foo.mjs`. The `.mjs` targets already exist in `lib/`.
- **Public package surface:** unchanged — `package.json` `exports` already routes ESM consumers to `.mjs` and CJS consumers to `.js`. Both target files still exist.
- **Type declarations (`.d.ts` / `.d.mts`):** generated through a separate DTS pipeline; not affected.
- Not a breaking change for any consumer.

## Implementation note

The regex matches any quoted relative path ending in `.js`. A scan over the affected source trees (`core/src`, `ai/src`, `genkit/src`, `plugins/google-genai/src`, `plugins/express/src`) found exactly one match outside an `import ... from` clause — a dynamic `await import('./reflection-v2.js')` — which the rewrite correctly retargets. No false-positive risk.

## Test plan

- [x] Reproduced the original crash under Angular SSR (Angular CLI 21.2.x, `genkit@1.35.0` from npm): `TypeError: Cannot read properties of undefined (reading 'object')` at `z.object(...)`.
- [x] Built `core`, `ai`, `genkit`, `plugins/google-genai`, `plugins/express` with the patch; verified `lib/index.mjs` now reads `export * from "./common.mjs"` across all five packages.
- [x] Linked the rebuilt packages into the Angular SSR reproduction and ran the original `import { genkit, z } from 'genkit'` flow end-to-end: `POST /api/recipeGeneratorFlow` returns HTTP 200 with the expected structured output.
- [ ] CI build across the full monorepo (will run via PR checks).

### DIFF
--- a/js/tsup.common.ts
+++ b/js/tsup.common.ts
@@ -18,15 +18,22 @@ import type { Plugin } from 'esbuild';
 import { readFile } from 'node:fs/promises';
 
 /**
- * Rewrites relative `.js` import/export specifiers to `.mjs` for ESM-format
- * builds. With `bundle: false`, esbuild transpiles each source file
- * individually and preserves its import paths as-is, so the ESM output of
- * `import './foo.js'` keeps pointing at `./foo.js` (CJS) instead of
- * `./foo.mjs`. That breaks Vite's SSR module runner (e.g. Angular SSR),
- * which can't surface named exports re-exported through a CJS file.
+ * Rewrites `export * from './foo.js'` to `export * from './foo.mjs'` in
+ * ESM-format builds. With `bundle: false`, esbuild transpiles each source
+ * file individually and preserves its import paths as-is, so the ESM
+ * output of a wildcard re-export keeps pointing at the sibling CJS `.js`
+ * file. Vite's SSR module runner (e.g. Angular SSR) cannot statically
+ * expand `export *` through a CJS file, leaving named bindings (`z`,
+ * `genkit`, …) undefined for consumers.
+ *
+ * Scope is intentionally narrow: only `export *` wildcard re-exports are
+ * affected. Regular `import`/`import-from` statements keep their `.js`
+ * targets so existing CJS-interop paths (which tolerate extensionless
+ * `require`) continue to work for plugins whose source uses extensionless
+ * relative imports.
  */
-const rewriteJsExtensionsForEsm: Plugin = {
-  name: 'rewrite-js-extensions-for-esm',
+const rewriteWildcardReexportsForEsm: Plugin = {
+  name: 'rewrite-wildcard-reexports-for-esm',
   setup(build) {
     if (build.initialOptions.format !== 'esm') return;
     const loaders: Record<string, 'ts' | 'tsx' | 'js' | 'jsx'> = {
@@ -47,8 +54,8 @@ const rewriteJsExtensionsForEsm: Plugin = {
         if (!loader) return null;
         const source = await readFile(args.path, 'utf8');
         const rewritten = source.replace(
-          /(['"`])(\.\.?\/[^'"`]+?)\.js\1/g,
-          '$1$2.mjs$1'
+          /(export\s*\*\s*from\s*)(['"`])(\.\.?\/[^'"`]+?)\.js\2/g,
+          '$1$2$3.mjs$2'
         );
         return { contents: rewritten, loader };
       }
@@ -66,7 +73,7 @@ export const defaultOptions = {
   entry: ['src/**/*.ts'],
   bundle: false,
   treeshake: false,
-  esbuildPlugins: [rewriteJsExtensionsForEsm],
+  esbuildPlugins: [rewriteWildcardReexportsForEsm],
 };
 
 /**
--- a/js/tsup.common.ts
+++ b/js/tsup.common.ts
@@ -39,17 +39,20 @@ const rewriteJsExtensionsForEsm: Plugin = {
       mjs: 'js',
       cjs: 'js',
     };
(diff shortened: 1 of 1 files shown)


js/tsup.common.ts
```diff
@@ -14,6 +14,55 @@
...
+      async (args) => {
+        const ext = args.path.match(/\.([^.]+)$/)?.[1] ?? '';
+        const loader = loaders[ext];
+        if (!loader) return null;
+        const source = await readFile(args.path, 'utf8');
+        const rewritten = source.replace(
+          /(export\s*\*\s*from\s*)(['"`])(\.\.?\/[^'"`]+?)\.js\2/g,
+          '$1$2$3.mjs$2'
+        );
+        return { contents: rewritten, loader };
```
pavelgj, line 61: Minor optimization... 

```suggestion
        if (rewritten === source) return null;
        return { contents: rewritten, loader };
```

pavelgj (approved):

PR #5401, opened 2026-05-23
fix(go/plugins/googlegenai): use native constrained generation with tools

### LINKED ISSUE #5400: [Go] Gemini models flagged `ConstrainedSupportNoTools`, blocking native structured output when tools are present
**Describe the bug**

The Gemini model metadata in `go/plugins/googlegenai` declares `Constrained: ai.ConstrainedSupportNoTools` for every Gemini model (via the shared `Multimodal` `ModelSupports`). The runtime in `go/ai/generate.go` interprets this as "native constrained generation is supported, but **not** when tools are present" and disables the native path for any combined `ai.WithTools(...)` + `ai.WithOutputType(...)` call.

There is also a second, redundant gate inside the plugin's request translation in [`go/plugins/googlegenai/gemini.go:318`](https://github.com/genkit-ai/genkit/blob/main/go/plugins/googlegenai/gemini.go#L318) which skips setting `ResponseMIMEType` whenever `len(input.Tools) > 0`. Because `gcc.ResponseSchema` is gated on `gcc.ResponseMIMEType != ""` in [`gemini.go:327`](https://github.com/genkit-ai/genkit/blob/main/go/plugins/googlegenai/gemini.go#L327), the schema is also never set on tool-using calls.

Net effect on a tool-using call with a schema:

1. `supportsConstrained(true)` in [`go/ai/generate.go:870-902`](https://github.com/genkit-ai/genkit/blob/main/go/ai/generate.go#L870-L902) returns `false` because of `ConstrainedSupportNoTools`.
2. [`go/ai/generate.go:299-317`](https://github.com/genkit-ai/genkit/blob/main/go/ai/generate.go#L299-L317) sets `outputCfg.Constrained = false`, injects the JSON schema as text instructions into the user prompt, and sets `outputCfg.Schema = nil`.
3. The plugin's `gemini.go` translation never sets `ResponseMIMEType` or `ResponseSchema` on the Gemini API request.

Result:

- Higher prompt-token cost (schema serialized into every request).
- Lower reliability than native structured output (the model can produce malformed JSON or echo schema metadata back in the response).
- No incremental constrained streaming from the API.

The Gemini API has supported combined `tools` + `responseSchema` / `responseJsonSchema` since Gemini 2.0. The `NoTools` flag predates that support and was never relaxed when newer models were registered.

**To reproduce**

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/firebase/genkit/go/plugins/googlegenai"
)

type Result struct {
	Title string   `json:"title"`
	Items []string `json:"items"`
}

func main() {
	ctx := context.Background()
	g := genkit.Init(ctx, genkit.WithPlugins(&googlegenai.GoogleAI{}))

	getThings := genkit.DefineTool(g, "getThings", "returns things",
		func(ctx *ai.ToolContext, _ struct{}) ([]string, error) {
			return []string{"a", "b", "c"}, nil
		},
	)

		ai.WithModelName("googleai/gemini-2.5-flash"),
		ai.WithPrompt("Call getThings, then return a titled list."),
		ai.WithTools(getThings),
		ai.WithOutputType(Result{}),
	)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(resp.Text())
}
```

Observed: the request sent to the Gemini API has no `responseMimeType` / `responseSchema`. The user message contains an injected `"Output should be in JSON format and conform to the following schema..."` block with the full JSON schema serialized as text.

**Expected behavior**

For Gemini 2.0+ models, `ai.WithTools(...)` + `ai.WithOutputType(...)` (or `ai.WithOutputSchema(...)`) should route through the native path: `ResponseMIMEType = "application/json"` and `ResponseSchema` set directly on the Gemini API call, with no schema injection into the prompt.

**Root cause**

1. `Multimodal` defaults every Gemini model to `Constrained: ai.ConstrainedSupportNoTools`:

   - [go/plugins/googlegenai/models.go:34](https://github.com/genkit-ai/genkit/blob/main/go/plugins/googlegenai/models.go#L34)

2. The plugin's request translation has its own tools-vs-output gate:

   - [go/plugins/googlegenai/gemini.go:318](https://github.com/genkit-ai/genkit/blob/main/go/plugins/googlegenai/gemini.go#L318) — `if hasOutput && len(input.Tools) == 0 { ... ResponseMIMEType = ... }`
   - [go/plugins/googlegenai/gemini.go:327](https://github.com/genkit-ai/genkit/blob/main/go/plugins/googlegenai/gemini.go#L327) — `ResponseSchema` is gated on `ResponseMIMEType != ""`.

The runtime decision in [go/ai/generate.go:299-300](https://github.com/genkit-ai/genkit/blob/main/go/ai/generate.go#L299-L300) and the `supportsConstrained` helper in [go/ai/generate.go:870-902](https://github.com/genkit-ai/genkit/blob/main/go/ai/generate.go#L870-L902) then force the prompt-injection fallback whenever tools are present.

**Proposed fix**

1. Change `Multimodal.Constrained` from `ai.ConstrainedSupportNoTools` to `ai.ConstrainedSupportAll` in `go/plugins/googlegenai/models.go`. The `ai.ConstrainedSupport` enum already includes `ConstrainedSupportAll` ([gen.go:307](https://github.com/genkit-ai/genkit/blob/main/go/ai/gen.go#L307)), which means "constrained generation in all cases, including with tools" — exactly the behavior modern Gemini models actually have.
2. Remove the `&& len(input.Tools) == 0` guard in `gemini.go:318` so `ResponseMIMEType` (and therefore `ResponseSchema`) is set on tool-using calls.

This is safe to do unconditionally now: every model in `supportedGeminiModels` is Gemini 2.0 or newer (`gemini-2.0-flash`, `gemini-2.0-flash-lite`, `gemini-2.5-flash`, `gemini-2.5-flash-lite`, `gemini-2.5-pro`, `gemini-3.1-flash-lite-preview`, `gemini-3.1-flash-image-preview`), and all Gemini 1.x models are deprecated and absent from `supportedGeminiModels`. There are no live model entries that would regress.

Not a breaking change for any current model.

**Runtime**

- OS: macOS 15.6
- Genkit: latest on `main`
- Go: 1.25.3

**Additional context**

This is the Go counterpart to genkit-ai/genkit#5392 (JS) and is being filed alongside an equivalent Python issue. If a future Gemini model lacks native tools+constrained support, it can opt out by declaring `Constrained: ai.ConstrainedSupportNoTools` explicitly in its model-specific `ModelOptions.Supports`. The default in `Multimodal` should reflect what the current model family actually supports.

chrisraygill:
## Summary

Fixes #5400. Every Gemini model in `go/plugins/googlegenai` declared `Constrained: ai.ConstrainedSupportNoTools` (via the shared `Multimodal` `ModelSupports`), which made `model.supportsConstrained(hasTools=true)` return false for any combined `ai.WithTools(...)` + `ai.WithOutputType(...)` call. The runtime in `go/ai/generate.go` then set `outputCfg.Constrained = false`, injected the JSON schema as text instructions into the user message, and cleared `outputCfg.Schema`.

A second layer of the same staleness lived in the plugin's request translation: `ResponseMIMEType` was gated on `len(input.Tools) == 0`, and `ResponseSchema` is downstream-gated on `ResponseMIMEType != ""`, so even after flipping the metadata the schema would not have reached the Gemini API for tool-using calls.

The Gemini API has supported tools + native constrained generation since Gemini 2.0. Every model in `supportedGeminiModels` is Gemini 2.0 or newer. After this PR, tool-using calls go through the native path: `ResponseMIMEType = "application/json"` and `ResponseSchema` are set on the Gemini request directly.

## Changes

- `go/plugins/googlegenai/models.go` — flip `Multimodal.Constrained` from `ai.ConstrainedSupportNoTools` to `ai.ConstrainedSupportAll`.
- `go/plugins/googlegenai/gemini.go` — drop the `&& len(input.Tools) == 0` guard and the stale `// JSON mode is not compatible with tools` comment.

## Not a breaking change

All Gemini 1.x models are deprecated and absent from `supportedGeminiModels`. The current entries (`gemini-2.0-flash`, `gemini-2.0-flash-lite`, `gemini-2.5-flash`, `gemini-2.5-flash-lite`, `gemini-2.5-pro`, `gemini-3.1-flash-lite-preview`, `gemini-3.1-flash-image-preview`) all support combined tools + constrained generation at the API level. A future model that lacks this support can still opt out by declaring `Constrained: ai.ConstrainedSupportNoTools` explicitly in its model-specific `ModelOptions.Supports`.

Note: this PR only touches the `googlegenai` plugin. The shared `go/plugins/internal/models.go` still defines a `Multimodal` with `ConstrainedSupportNoTools` because it is consumed by `vertexai/modelgarden` (Anthropic/Llama models on Vertex), whose constrained-generation behavior is unrelated to Gemini and out of scope here.

## Related

This is the Go counterpart to the JS fix in #5393. A parallel Python issue and PR follow.

## Checklist

- [x] Tested: `go test ./plugins/googlegenai/... -count=1` passes locally.
- [x] No docs change required (the public API and behavior contract are unchanged; only the metadata default flips for current Gemini models).


adesinah (changes requested):
The change looks good to me. One thing I would suggest is adding or updating a unit test for the actual translated request. The existing TestConvertRequest still has old assertions saying `MIME/schema` should be empty when tools exist, but that request does not set `Output.Format` or `ContentType`, so it passes without covering the new behavior. For this change, we could add a case with Tools,
  `Output.Constrained: true, Output.Format: "json"` or `ContentType: "application/json",` and assert:
```
  ResponseMIMEType == "application/json"
  ResponseSchema != nil
  FunctionDeclarations still present
```

chrisraygill:
@adesinah good catch, thanks. You're right that the previous \`TestConvertRequest\` covered the new behavior only vacuously: the request set \`Output.Constrained: true\` but never \`Output.Format\`/\`ContentType\`, so \`ResponseMIMEType\` stayed empty and the old "should be empty when tools are present" assertions passed without exercising the change.

Fixed in 5c4af6343:
- Added \`Output.Format: "json"\` to the request (it already had \`Tools\` + \`Output.Constrained: true\` + a schema), so it now matches the exact case you described.
- Flipped the stale assertions to verify the new behavior: \`ResponseMIMEType == "application/json"\` and \`ResponseSchema != nil\`.
- The existing assertions confirming \`FunctionDeclarations\` are still merged in (alongside the preserved \`GoogleSearch\` tool) remain, so we now assert constrained JSON output and tools coexist in a single translated request.

Verified: \`go test ./...\` passes for the package, \`go vet\` and \`gofmt\` clean. PTAL when you have a moment.

adesinah (approved):
Thanks, @chrisraygill . All the changes look good to me. Approved.

PR #5404, opened 2026-05-23
fix(js/ai): avoid any inference for MiddlewareRef

7hokerz:
 - Removed the explicit `z.ZodTypeAny` annotation from `MiddlewareRefSchema` so `MiddlewareRef` is inferred as a concrete object type instead of `any`.

## Explanation
In the case of the previous `genkit/model/middleware` retry middleware, there were no particular issues because the `ModelMiddleware` type was explicitly used.

However, the newly introduced `middlewareRef` was inferred as `any` due to being specified as `z.zodTypeAny`, which triggers a warning. [no-unsafe-assignment](https://typescript-eslint.io/rules/no-unsafe-assignment/)

Although not required, `no-unsafe-assignment` is used in most type-safe projects, so this change is highly important.

This change keeps the serialized middleware boundary flexible while avoiding unnecessary any leakage in public types.

Checklist (if applicable):

### DIFF
--- a/js/ai/src/model-types.ts
+++ b/js/ai/src/model-types.ts
@@ -25,9 +25,9 @@ import {
 export { Part, PartSchema };
 
 /** Zod schema for a serializable reference to a registered generate middleware. */
-export const MiddlewareRefSchema: z.ZodTypeAny = z.object({
+export const MiddlewareRefSchema = z.object({
   name: z.string(),
-  config: z.any().optional(),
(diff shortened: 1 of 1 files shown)


7hokerz:
@pavelgj , Could you please take a look at this when you have a moment?

7hokerz:
@pavelgj , Could you please take a look at this when you have a moment?

PR #5440, opened 2026-06-03
feat(go/plugins/bedrock): add Amazon Bedrock plugin

adesinah:
Adds the first-party Genkit Go plugin for Amazon Bedrock — go/plugins/bedrock — covering generation, streaming, tool use, embeddings, reranking, image generation, and prompt caching, plus a runnable sample, README, and a gated live-test matrix.

## Credentials & region

Uses the standard AWS credential chain (config.LoadDefaultConfig). Region resolves Bedrock.Region → AWS_REGION → ~/.aws/config with no hardcoded default — Init panics if no region is resolvable, so a config-less environment can't silently talk to us-east-1. A pre-built aws.Config can be passed for custom credential providers, fakes, or shared config.

## Testing

  - Unit tests (bedrock_test.go, multimodal_test.go) need no AWS access and run by default — including offline coverage of reasoning-signature round-tripping, streaming tool-block assembly, and malformed-stream error handling.
  - Live tests are opt-in per-model via flags and skip unless the corresponding flag is provided. Matrix covers Claude
  sync/stream/tool, Nova sync, Titan/Cohere/Nova embedders, and Cohere rerank.

## Known limitations (v1)

One region per plugin instance; embedders embed a single modality per document (image takes precedence); no streaming for embedders/image gen; image gen returns one 1024×1024 PNG; only Cohere Rerank's text path is wired. See the README for the full list.

## Checklist (if applicable):


cabljac:
Closing, we will open a PR against the existing plugin

PR #5441, opened 2026-06-04
core(reflection): bind reflection server to loopback and validate Host header

adilburaksen:
## Summary

The JS reflection server (`js/core/src/reflection.ts`), which auto-starts in dev (`GENKIT_ENV=dev`), calls `server.listen(this.port, ...)` with no host argument, so Node binds it to **all interfaces (0.0.0.0)**, and the only middleware is `express.json()` — no Host/Origin validation. As a result the reflection API (`/api/runAction`, `/api/actions`, `/api/values`, `/api/__quitquitquit`) is reachable unauthenticated:
- directly from other hosts on the local network (0.0.0.0 bind), and
- from a web page open in the developer's browser via DNS-rebinding (no Host check → rebinding makes the request same-origin, so CORS does not apply).

`/api/runAction` executes any registered action with caller-supplied input/context, and `/api/__quitquitquit` is a simple `GET` that terminates the process.

## Changes

1. **Bind to loopback by default** — `listen()` now binds `localhost` (loopback). Added an optional `host` to `ReflectionServerOptions` (default `"localhost"`) so non-default setups remain configurable.
2. **Host-header validation** — a middleware (registered before the routes) rejects requests whose `Host` is not a loopback address, defeating DNS-rebinding even on the loopback bind.
3. **`__quitquitquit` Origin guard** — defense-in-depth: reject the shutdown `GET` if it carries an `Origin` header (a legitimate CLI/in-process shutdown does not).

A small correctness change accompanies (1): passing a host to `listen()` makes binding asynchronous, so `start()` now awaits the `listening` event before reading `server.address()` (the existing post-listen init — registry warm-up and runtime-file write — is preserved).

## Testing

`tsc --noEmit` clean; prettier clean. Added 4 tests (non-loopback Host → 403; `localhost`/`127.0.0.1` Host → 200; cross-origin `__quitquitquit` → 403); existing reflection tests still pass (8/8).

### DIFF
--- a/js/core/src/reflection.ts
+++ b/js/core/src/reflection.ts
@@ -43,12 +43,36 @@ export type RunActionResponse = z.infer<typeof RunActionResponseSchema>;
 export interface ReflectionServerOptions {
   /** Port to run the server on. Actual port may be different if chosen port is occupied. Defaults to 3100. */
   port?: number;
+  /** Host/interface to bind the server to. Defaults to `localhost` (loopback only). */
+  host?: string;
   /** Body size limit for the server. Defaults to `30mb`. */
   bodyLimit?: string;
   /** Configured environments. Defaults to `dev`. */
   configuredEnvs?: string[];
   /** Display name that will be shown in developer tooling. */
   name?: string;
+}
+
+/**
+ * Checks whether a `Host` header refers to a loopback interface.
+ *
+ * Used to reject cross-origin/DNS-rebinding requests against the reflection
+ * server, which is bound to loopback and intended only for local dev tooling.
+ */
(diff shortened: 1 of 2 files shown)


adilburaksen:
@pavelgj mind giving this a look? The reflection server currently binds without restricting to loopback and doesn't validate the Host header, so it's reachable/rebindable during dev. The change binds it to loopback and adds the Host check. Dev-tooling only — let me know if the binding default should be configurable.

pavelgj:
Sorry, missed this PR, but I believe this was fixed in https://github.com/genkit-ai/genkit/pull/5751

PR #5449, opened 2026-06-04
fix(js): Add logs in js middleware

shrutip90:
Verified logs display in dev UI.

Checklist (if applicable):
- [x] Tested (manually)

### DIFF
--- a/js/ai/src/model/middleware.ts
+++ b/js/ai/src/model/middleware.ts
@@ -371,7 +371,7 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
           let reason = '';
           if (NEVER_RETRY_ERROR_NAMES.includes(error?.name)) {
             shouldRetry = false;
-            reason = `error name "${error?.name}" is explicitly non-retryable`;
+            reason = `error name "${error?.name}" is non-retryable`;
           } else if (error instanceof GenkitError) {
             if (statuses.includes(error.status)) {
               shouldRetry = true;
@@ -389,7 +389,7 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
               delay = delay + 1000 * Math.pow(2, i) * Math.random();
             }
             logger.logStructuredWarn(
-              `Retry attempt ${i + 1} of ${maxRetries} triggered in ${Math.round(delay)}ms due to error: ${error?.message || String(error)}`,
+              `Retry ${i + 1} of ${maxRetries} in ${Math.round(delay)}ms due to error: ${error?.message || String(error)}`,
               {
                 'genkit.middleware.name': 'retry',
                 'genkit.middleware.retry.attempt': i + 1,
@@ -403,7 +403,7 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             continue;
           } else {
             logger.logStructuredWarn(
-              `Request failed with error: ${error?.message || String(error)}. Skipping retry because ${reason}.`,
+              `Retry skipped for ${error?.message || String(error)} because of ${reason}`,
               {
                 'genkit.middleware.name': 'retry',
                 'genkit.middleware.retry.skipped_reason': reason,
@@ -414,7 +414,7 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
           }
         } else {
           logger.logStructuredWarn(
-            `All retry attempts exhausted (${maxRetries}). Last error: ${error?.message || String(error)}`,
+            `Retry attempts exhausted (${maxRetries}). Last error: ${error?.message || String(error)}`,
             {
               'genkit.middleware.name': 'retry',
               'genkit.middleware.retry.max_attempts': maxRetries,
@@ -483,12 +483,13 @@ export function fallback(
         onError?.(e);
         let lastError: any = e;
         for (const model of models) {
-          const resolved = await resolveModel(ai.registry, model);
-          const targetModelName =
-            resolved.modelAction?.__action?.name || String(model);
+          let targetModelName = String(model);
           try {
+            const resolved = await resolveModel(ai.registry, model);
+            targetModelName =
+              resolved.modelAction?.__action?.name || targetModelName;
             logger.logStructuredWarn(
-              `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${targetModelName}...`,
+              `Falling back to model ${targetModelName} due to error ${lastError.status} ${lastError.message}`,
               {
                 'genkit.middleware.name': 'fallback',
                 'genkit.middleware.fallback.target_model': targetModelName,
@@ -502,12 +503,8 @@ export function fallback(
               onError?.(e2);
               continue;
             }
-            const statusStr =
-              e2 instanceof GenkitError
-                ? `status ${e2.status}`
-                : 'non-Genkit error';
             logger.logStructuredWarn(
-              `Fallback model ${targetModelName} failed with ${statusStr}: ${e2.message || String(e2)}. Aborting fallback sequence.`,
+              `Aborting fallback sequence for unrecoverable error: ${e2.message || String(e2)}`,
               {
                 'genkit.middleware.name': 'fallback',
                 'genkit.middleware.fallback.target_model': targetModelName,
@@ -518,7 +515,7 @@ export function fallback(
           }
         }
         logger.logStructuredWarn(
-          `All fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
+          `Fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
           {
             'genkit.middleware.name': 'fallback',
           },
@@ -526,10 +523,8 @@ export function fallback(
         );
         throw lastError;
       } else {
-        const statusStr =
-          e instanceof GenkitError ? `status ${e.status}` : 'non-Genkit error';
         logger.logStructuredWarn(
-          `Request failed with ${statusStr}: ${e.message || String(e)}. Skipping fallback.`,
+          `Skipping fallback for unhandled error: ${e.message || String(e)}`,
           {
             'genkit.middleware.name': 'fallback',
           },
--- a/js/ai/src/model/middleware.ts
+++ b/js/ai/src/model/middleware.ts
@@ -344,6 +344,8 @@ const DEFAULT_FALLBACK_STATUSES: StatusName[] = [
  * });
  * ```
  */
+const NEVER_RETRY_ERROR_NAMES = ['AbortError', 'ToolInterruptError'];
+
 export function retry(options: RetryOptions = {}): ModelMiddleware {
   const {
     maxRetries = 3,
@@ -366,9 +368,15 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
         const error = e as Error;
         if (i < maxRetries) {
           let shouldRetry = false;
-          if (error instanceof GenkitError) {
+          let reason = '';
+          if (NEVER_RETRY_ERROR_NAMES.includes(error?.name)) {
+            shouldRetry = false;
+            reason = `error name "${error?.name}" is explicitly non-retryable`;
+          } else if (error instanceof GenkitError) {
             if (statuses.includes(error.status)) {
               shouldRetry = true;
+            } else {
+              reason = `status "${error.status}" is not retryable`;
             }
           } else {
             shouldRetry = true;
@@ -380,13 +388,39 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             if (!noJitter) {
               delay = delay + 1000 * Math.pow(2, i) * Math.random();
             }
-            logger.warn(
-              `Request failed: ${error?.message || String(error)}. Retrying in ${Math.round(delay)}ms... (Attempt ${i + 1} of ${maxRetries})`
+            logger.logStructuredWarn(
+              `Retry attempt ${i + 1} of ${maxRetries} triggered in ${Math.round(delay)}ms due to error: ${error?.message || String(error)}`,
+              {
+                'genkit.middleware.name': 'retry',
+                'genkit.middleware.retry.attempt': i + 1,
+                'genkit.middleware.retry.max_attempts': maxRetries,
+                'genkit.middleware.retry.delay_ms': Math.round(delay),
+              },
+              error
             );
             await new Promise((resolve) => __setTimeout(resolve, delay));
             currentDelay = Math.min(currentDelay * backoffFactor, maxDelayMs);
             continue;
+          } else {
+            logger.logStructuredWarn(
+              `Request failed with error: ${error?.message || String(error)}. Skipping retry because ${reason}.`,
+              {
+                'genkit.middleware.name': 'retry',
+                'genkit.middleware.retry.skipped_reason': reason,
+                'genkit.middleware.retry.max_attempts': maxRetries,
+              },
+              error
+            );
           }
+        } else {
+          logger.logStructuredWarn(
+            `All retry attempts exhausted (${maxRetries}). Last error: ${error?.message || String(error)}`,
+            {
+              'genkit.middleware.name': 'retry',
+              'genkit.middleware.retry.max_attempts': maxRetries,
+            },
+            error
+          );
         }
         throw error;
       }
@@ -442,29 +476,67 @@ export function fallback(
   return async (req, next) => {
     try {
       return await next(req);
-    } catch (e) {
-      if (e instanceof GenkitError && statuses.includes(e.status)) {
+    } catch (e: any) {
+      const isFallbackable =
+        e instanceof GenkitError && statuses.includes(e.status);
+      if (isFallbackable) {
         onError?.(e);
         let lastError: any = e;
         for (const model of models) {
+          const resolved = await resolveModel(ai.registry, model);
+          const targetModelName =
+            resolved.modelAction?.__action?.name || String(model);
           try {
-            const resolved = await resolveModel(ai.registry, model);
-            logger.warn(
-              `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${resolved.modelAction.__action.name}...`
+            logger.logStructuredWarn(
+              `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${targetModelName}...`,
+              {
+                'genkit.middleware.name': 'fallback',
+                'genkit.middleware.fallback.target_model': targetModelName,
+              },
+              lastError
             );
             return await resolved.modelAction(req);
-          } catch (e2) {
+          } catch (e2: any) {
             lastError = e2;
             if (e2 instanceof GenkitError && statuses.includes(e2.status)) {
               onError?.(e2);
               continue;
             }
+            const statusStr =
+              e2 instanceof GenkitError
+                ? `status ${e2.status}`
+                : 'non-Genkit error';
+            logger.logStructuredWarn(
+              `Fallback model ${targetModelName} failed with ${statusStr}: ${e2.message || String(e2)}. Aborting fallback sequence.`,
+              {
+                'genkit.middleware.name': 'fallback',
+                'genkit.middleware.fallback.target_model': targetModelName,
+              },
+              e2
+            );
             throw e2;
           }
         }
+        logger.logStructuredWarn(
+          `All fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
+          {
+            'genkit.middleware.name': 'fallback',
+          },
+          lastError
+        );
         throw lastError;
+      } else {
+        const statusStr =
+          e instanceof GenkitError ? `status ${e.status}` : 'non-Genkit error';
+        logger.logStructuredWarn(
+          `Request failed with ${statusStr}: ${e.message || String(e)}. Skipping fallback.`,
+          {
+            'genkit.middleware.name': 'fallback',
+          },
+          e
+        );
+        throw e;
       }
-      throw e;
     }
   };
 }
--- a/js/ai/src/model/middleware.ts
+++ b/js/ai/src/model/middleware.ts
@@ -388,7 +388,7 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             if (!noJitter) {
               delay = delay + 1000 * Math.pow(2, i) * Math.random();
             }
-            logger.logStructuredWarn(
+            logger.warn(
               `Retry ${i + 1} of ${maxRetries} in ${Math.round(delay)}ms due to error: ${error?.message || String(error)}`,
               {
                 'genkit.middleware.name': 'retry',
@@ -402,7 +402,7 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             currentDelay = Math.min(currentDelay * backoffFactor, maxDelayMs);
             continue;
           } else {
-            logger.logStructuredWarn(
+            logger.warn(
               `Retry skipped for ${error?.message || String(error)} because of ${reason}`,
               {
                 'genkit.middleware.name': 'retry',
@@ -413,7 +413,7 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             );
           }
         } else {
-          logger.logStructuredWarn(
+          logger.warn(
             `Retry attempts exhausted (${maxRetries}). Last error: ${error?.message || String(error)}`,
             {
               'genkit.middleware.name': 'retry',
@@ -488,7 +488,7 @@ export function fallback(
             const resolved = await resolveModel(ai.registry, model);
             targetModelName =
               resolved.modelAction?.__action?.name || targetModelName;
-            logger.logStructuredWarn(
+            logger.warn(
               `Falling back to model ${targetModelName} due to error ${lastError.status} ${lastError.message}`,
               {
                 'genkit.middleware.name': 'fallback',
@@ -503,7 +503,7 @@ export function fallback(
               onError?.(e2);
               continue;
             }
-            logger.logStructuredWarn(
+            logger.warn(
               `Aborting fallback sequence for unrecoverable error: ${e2.message || String(e2)}`,
               {
                 'genkit.middleware.name': 'fallback',
@@ -514,7 +514,7 @@ export function fallback(
             throw e2;
           }
         }
-        logger.logStructuredWarn(
+        logger.warn(
           `Fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
           {
             'genkit.middleware.name': 'fallback',
@@ -523,7 +523,7 @@ export function fallback(
         );
         throw lastError;
       } else {
-        logger.logStructuredWarn(
+        logger.warn(
           `Skipping fallback for unhandled error: ${e.message || String(e)}`,
           {
             'genkit.middleware.name': 'fallback',
--- a/js/ai/src/model/middleware.ts
+++ b/js/ai/src/model/middleware.ts
@@ -15,6 +15,7 @@
  */
 
 import { GenkitError, StatusName } from '@genkit-ai/core';
+import { logger } from '@genkit-ai/core/logging';
 import { HasRegistry } from '@genkit-ai/core/registry';
 import { Document } from '../document.js';
 import { injectInstructions } from '../formats/index.js';
@@ -379,6 +380,9 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             if (!noJitter) {
               delay = delay + 1000 * Math.pow(2, i) * Math.random();
             }
+            logger.warn(
+              `Request failed: ${error?.message || String(error)}. Retrying in ${Math.round(delay)}ms... (Attempt ${i + 1} of ${maxRetries})`
+            );
             await new Promise((resolve) => __setTimeout(resolve, delay));
             currentDelay = Math.min(currentDelay * backoffFactor, maxDelayMs);
             continue;
@@ -445,6 +449,9 @@ export function fallback(
         for (const model of models) {
           try {
             const resolved = await resolveModel(ai.registry, model);
+            logger.warn(
+              `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${resolved.modelAction.__action.name}...`
+            );
             return await resolved.modelAction(req);
           } catch (e2) {
             lastError = e2;
--- a/js/core/src/logging.ts
+++ b/js/core/src/logging.ts
@@ -145,14 +145,44 @@ class Logger {
     getLogger().level = level;
   }
 
-  logStructured(msg: string, metadata: any) {
-    getLogger().info(msg, metadata);
-    this._emitOtel('info', [], msg, metadata);
+  private _mergeErrorMetadata(metadata: any, err?: any): any {
+    const mergedMetadata = { ...metadata };
+    if (err) {
+      mergedMetadata['exception.type'] = err.name || 'Error';
+      mergedMetadata['exception.message'] = err.message || String(err);
+      if (err.stack) {
+        mergedMetadata['exception.stacktrace'] = err.stack;
+      }
+    }
+    return mergedMetadata;
   }
 
-  logStructuredError(msg: string, metadata: any) {
-    getLogger().error(msg, metadata);
-    this._emitOtel('error', [], msg, metadata);
+  logStructuredInfo(msg: string, metadata: any, err?: any) {
+    const mergedMetadata = this._mergeErrorMetadata(metadata, err);
+    getLogger().info(msg, mergedMetadata);
+    this._emitOtel('info', [], msg, mergedMetadata);
+  }
+
+  logStructured(msg: string, metadata: any, err?: any) {
+    this.logStructuredInfo(msg, metadata, err);
+  }
+
+  logStructuredError(msg: string, metadata: any, err?: any) {
+    const mergedMetadata = this._mergeErrorMetadata(metadata, err);
+    getLogger().error(msg, mergedMetadata);
+    this._emitOtel('error', [], msg, mergedMetadata);
+  }
+
+  logStructuredWarn(msg: string, metadata: any, err?: any) {
+    const mergedMetadata = this._mergeErrorMetadata(metadata, err);
+    getLogger().warn(msg, mergedMetadata);
+    this._emitOtel('warn', [], msg, mergedMetadata);
+  }
+
+  logStructuredDebug(msg: string, metadata: any, err?: any) {
+    const mergedMetadata = this._mergeErrorMetadata(metadata, err);
+    getLogger().debug(msg, mergedMetadata);
+    this._emitOtel('debug', [], msg, mergedMetadata);
   }
 }
 
--- a/js/core/src/logging.ts
+++ b/js/core/src/logging.ts
@@ -180,7 +180,7 @@ class Logger {
     metadata?: Record<string, any>;
     error?: any;
   }): void;
-  info(message: any, metadata?: Record<string, any>, error?: any): void;
+  info(message: any, metadata?: any, error?: any): void;
   info(...args: any[]) {
     this._log('info', ...args);
   }
@@ -190,7 +190,7 @@ class Logger {
     metadata?: Record<string, any>;
     error?: any;
   }): void;
-  debug(message: any, metadata?: Record<string, any>, error?: any): void;
+  debug(message: any, metadata?: any, error?: any): void;
   debug(...args: any[]) {
     this._log('debug', ...args);
   }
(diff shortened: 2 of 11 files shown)


MichaelDoyle:
Let's see if we can add some structured logs here, instead of just strings.

For the catch blocks where we are logging errors, we can follow this convention:
https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/

For middleware more broadly, what are some attributes we always want present? Since these logs will belong to the generate span, it will be important to have some basic structure around it.

e.g. middleware name might be a good attribute to always have.
e.g. whenever we're doing string interpolation, it's a good sign some or all of those details should also be attributes

Looking at retry for example, we could have something like (I haven't fully thought this through):

```js
{
  // we should probably at *least* always include name
  'genkit.middleware.name': 'retry',

  // other genkit attrs we want to standardize on
  'genkit.model': 'gemini-flash-latest',
  'genkit.retry.attempt': 1,
  'genkit.retry.max_attempts': 3,
  'genkit.retry.delay_ms': 500,
}
```

And when we have catch blocks:

 ```typescript
 logger.warn(
   `Failed to read skill file foo`,
   {
     // By Genkit convention (whatever we decide)
     'genkit.middleware.name': 'skills',
     'genkit.skill.path': `/path/to/skill`,

     // by Otel convention
     'exception.type':  'Error',
     'exception.message': 'something bad happened',
     'exception.stacktrace': 'big long stacktrace',
   }
 );
 ```

js/ai/src/model/middleware.ts
```diff
@@ -379,6 +380,9 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             if (!noJitter) {
               delay = delay + 1000 * Math.pow(2, i) * Math.random();
             }
+            logger.warn(
+              `Request failed: ${error?.message || String(error)}. Retrying in ${Math.round(delay)}ms... (Attempt ${i + 1} of ${maxRetries})`
```
MichaelDoyle: See V2 comments below. Although, not sure how muich we want to put into V1 logging.

js/ai/src/model/middleware.ts
```diff
@@ -445,6 +449,9 @@ export function fallback(
         for (const model of models) {
           try {
             const resolved = await resolveModel(ai.registry, model);
+            logger.warn(
+              `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${resolved.modelAction.__action.name}...`
```
MichaelDoyle: See V2 comments below. Although, not sure how muich we want to put into V1 logging.

js/plugins/middleware/src/fallback.ts
```diff
@@ -116,6 +117,9 @@ export const fallback: GenerateMiddleware<typeof FallbackOptionsSchema> =
             ) {
               let lastError: any = e;
               for (const model of models) {
+                logger.warn(
+                  `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${model.name}...`
+                );
```
MichaelDoyle: This is better, but we can still improve. e.g.

> Model `gemini-whatever` failed with status RESOURCE_EXHAUSTED. Falling back to 'gemini-flash-much-better'...

Again, we are missing several important cases.

 1. **Un-fallbackable Original Error (Skipping Fallback)**: The initial model failed, but the status was not in the configured list (e.g. it was a `PERMISSION_DENIED` rather than `UNAVAILABLE`). The fallback was skipped completely with no log explaining why.
 2. **Fallback Model Fails with Un-fallbackable Error (Aborting Fallback)**: One of our fallback models failed, but with a non-matching error status. This aborts the entire fallback sequence immediately without trying any remaining fallback models. This happens silently.
 3. **All Fallback Options Exhausted (Failure)**: We tried the main model and every single configured fallback model, and they all failed. The last error is thrown, but we never log that we reached the end of our fallback list.

js/plugins/middleware/src/retry.ts
```diff
@@ -191,6 +192,9 @@ export const retry: GenerateMiddleware<typeof RetryOptionsSchema> =
                   if (!noJitter) {
                     delay = delay + 1000 * Math.pow(2, i) * Math.random();
                   }
+                  logger.warn(
+                    `Request failed: ${error?.message || String(error)}. Retrying in ${Math.round(delay)}ms... (Attempt ${i + 1} of ${maxRetries})`
+                  );
```
MichaelDoyle, line 208: 1. I think _this specific log_ should just focus on the retry aspects, and not the error itself. Really just a wording change.

    e.g. `Retry attempt ${i + 1} of ${maxRetries} triggered due to error: ${error?.message || String(error)}.`,

2. In general, we're missing some cases.
   1. **Un-retryable Errors**: If a request failed with a non-retryable error (e.g., matching `NEVER_RETRY_ERROR_NAMES` or with an unconfigured status), it would fail silently without logging why the retry was skipped.
   2. **Exhausted Retries**: If all retry attempts failed, the error would simply bubble up and be thrown, with no warning log indicating that we reached the retry ceiling.

js/plugins/middleware/src/skills.ts
```diff
@@ -97,15 +98,24 @@ export const skills: GenerateMiddleware<typeof SkillsOptionsSchema> =
                         path: skillMdPath,
                         description,
                       });
-                    } catch (e) {
-                      // ignore file read errors
+                    } catch (e: any) {
+                      logger.warn(
+                        `Failed to read skill file "${skillMdPath}": ${e.message || String(e)}`
+                      );
```
MichaelDoyle: This may actually log too often, because I think we're checking every dir for a `SKILL.md`?

pavelgj:
> Let's see if we can add some structured logs here, instead of just strings.
> 
> For the catch blocks where we are logging errors, we can follow this convention: https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/
> 
> For middleware more broadly, what are some attributes we always want present? Since these logs will belong to the generate span, it will be important to have some basic structure around it.
> 
> e.g. middleware name might be a good attribute to always have. e.g. whenever we're doing string interpolation, it's a good sign some or all of those details should also be attributes
> 
> Looking at retry for example, we could have something like (I haven't fully thought this through):
> 
> ```js
> {
>   // we should probably at *least* always include name
>   'genkit.middleware.name': 'retry',
> 
>   // other genkit attrs we want to standardize on
>   'genkit.model': 'gemini-flash-latest',
>   'genkit.retry.attempt': 1,
>   'genkit.retry.max_attempts': 3,
>   'genkit.retry.delay_ms': 500,
> }
> ```
> 
> And when we have catch blocks:
> 
> ```ts
> logger.warn(
>   `Failed to read skill file foo`,
>   {
>     // By Genkit convention (whatever we decide)
>     'genkit.middleware.name': 'skills',
>     'genkit.skill.path': `/path/to/skill`,
> 
>     // by Otel convention
>     'exception.type':  'Error',
>     'exception.message': 'something bad happened',
>     'exception.stacktrace': 'big long stacktrace',
>   }
> );
> ```

Good point... at the very least we do need the standardized middleware name. `'genkit.middleware.name'` SGTM

I'm not sure about standardizing on other attributes... might be useful, I suppose... but they would have to be prefixed...  `'genkit.middleware.{middlewareName}.{attrName}'`

js/ai/src/model/middleware.ts
```diff
@@ -445,6 +449,9 @@ export function fallback(
...
+            );
```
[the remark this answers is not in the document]
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Don't see why we can't do these right now...

js/plugins/middleware/src/fallback.ts
shrutip90: Added

js/plugins/middleware/src/retry.ts
shrutip90: Added, thanks!

js/plugins/middleware/src/skills.ts
shrutip90: Updated to log only on other errors.

shrutip90:
> Let's see if we can add some structured logs here, instead of just strings.
> 
> For the catch blocks where we are logging errors, we can follow this convention: https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/
> 
> For middleware more broadly, what are some attributes we always want present? Since these logs will belong to the generate span, it will be important to have some basic structure around it.
> 
> e.g. middleware name might be a good attribute to always have. e.g. whenever we're doing string interpolation, it's a good sign some or all of those details should also be attributes
> 
> Looking at retry for example, we could have something like (I haven't fully thought this through):
> 
> ```js
> {
>   // we should probably at *least* always include name
>   'genkit.middleware.name': 'retry',
> 
>   // other genkit attrs we want to standardize on
>   'genkit.model': 'gemini-flash-latest',
>   'genkit.retry.attempt': 1,
>   'genkit.retry.max_attempts': 3,
>   'genkit.retry.delay_ms': 500,
> }
> ```
> 
> And when we have catch blocks:
> 
> ```ts
> logger.warn(
>   `Failed to read skill file foo`,
>   {
>     // By Genkit convention (whatever we decide)
>     'genkit.middleware.name': 'skills',
>     'genkit.skill.path': `/path/to/skill`,
> 
>     // by Otel convention
>     'exception.type':  'Error',
>     'exception.message': 'something bad happened',
>     'exception.stacktrace': 'big long stacktrace',
>   }
> );
> ```

Thanks! Added structured logs and followed the prefix pattern that Pavel suggested, as I was thinking the same.

js/plugins/middleware/src/fallback.ts
```diff
@@ -109,13 +110,21 @@ export const fallback: GenerateMiddleware<typeof FallbackOptionsSchema> =
...
+            const isFallbackable =
               e instanceof GenkitError &&
-              statuses.includes(e.status as StatusName)
-            ) {
+              statuses.includes(e.status as StatusName);
+            if (isFallbackable) {
               let lastError: any = e;
               for (const model of models) {
+                logger.logStructuredWarn(
+                  `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${model.name}...`,
```
MichaelDoyle: It's "ok" to include the error message, but same feedback as throughout, we should focus on what is happening here (fallback) over the reason.

e.g.
Falling back to model ${model.name}; Last error: ${lastError.status} ${lastError.message}.

js/ai/src/model/middleware.ts
```diff
@@ -438,26 +476,67 @@ export function fallback(
...
-    } catch (e) {
-      if (e instanceof GenkitError && statuses.includes(e.status)) {
+    } catch (e: any) {
+      const isFallbackable =
+        e instanceof GenkitError && statuses.includes(e.status);
+      if (isFallbackable) {
         onError?.(e);
         let lastError: any = e;
         for (const model of models) {
+          const resolved = await resolveModel(ai.registry, model);
```
MichaelDoyle: Did you mean to move this out of the try/catch?

js/ai/src/model/middleware.ts
```diff
@@ -365,9 +368,15 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
         const error = e as Error;
         if (i < maxRetries) {
           let shouldRetry = false;
-          if (error instanceof GenkitError) {
+          let reason = '';
+          if (NEVER_RETRY_ERROR_NAMES.includes(error?.name)) {
+            shouldRetry = false;
+            reason = `error name "${error?.name}" is explicitly non-retryable`;
```
MichaelDoyle: It's explicit in all cases. :) What's a better wording here? Maybe drop the explicitly?

```suggestion
            reason = `error name "${error?.name}" is non-retryable`;
```

js/ai/src/model/middleware.ts
```diff
@@ -379,10 +388,39 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
             if (!noJitter) {
               delay = delay + 1000 * Math.pow(2, i) * Math.random();
             }
+            logger.logStructuredWarn(
+              `Retry attempt ${i + 1} of ${maxRetries} triggered in ${Math.round(delay)}ms due to error: ${error?.message || String(error)}`,
```
MichaelDoyle:
```suggestion
              `Retry ${i + 1} of ${maxRetries} in ${Math.round(delay)}ms due to error: ${error?.message || String(error)}`,
```

js/ai/src/model/middleware.ts
```diff
@@ -379,10 +388,39 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
...
+                'genkit.middleware.retry.delay_ms': Math.round(delay),
+              },
+              error
+            );
             await new Promise((resolve) => __setTimeout(resolve, delay));
             currentDelay = Math.min(currentDelay * backoffFactor, maxDelayMs);
             continue;
+          } else {
+            logger.logStructuredWarn(
+              `Request failed with error: ${error?.message || String(error)}. Skipping retry because ${reason}.`,
```
MichaelDoyle: Paraphrasing, but should be something like:

```suggestion
              `Retry skipped for {$error} because of ${reason}.`,
```

js/ai/src/model/middleware.ts
```diff
@@ -379,10 +388,39 @@ export function retry(options: RetryOptions = {}): ModelMiddleware {
...
+                'genkit.middleware.name': 'retry',
+                'genkit.middleware.retry.skipped_reason': reason,
+                'genkit.middleware.retry.max_attempts': maxRetries,
+              },
+              error
+            );
           }
+        } else {
+          logger.logStructuredWarn(
+            `All retry attempts exhausted (${maxRetries}). Last error: ${error?.message || String(error)}`,
```
MichaelDoyle:
```suggestion
            `Retry attempts exhausted (${maxRetries}). Last error: ${error?.message || String(error)}`,
```

js/ai/src/model/middleware.ts
```diff
@@ -438,26 +476,67 @@ export function fallback(
...
         onError?.(e);
         let lastError: any = e;
         for (const model of models) {
+          const resolved = await resolveModel(ai.registry, model);
+          const targetModelName =
+            resolved.modelAction?.__action?.name || String(model);
           try {
-            const resolved = await resolveModel(ai.registry, model);
+            logger.logStructuredWarn(
+              `Request failed with status ${lastError.status}: ${lastError.message}. Falling back to model ${targetModelName}...`,
```
MichaelDoyle: Same feedback as before, we want to flip this to be about the action being taken.

Paraphrasing:

```suggestion
              `Falling back to model ${targetModelName} due to error $error`,
```

js/ai/src/model/middleware.ts
```diff
@@ -438,26 +476,67 @@ export function fallback(
...
             if (e2 instanceof GenkitError && statuses.includes(e2.status)) {
               onError?.(e2);
               continue;
             }
+            const statusStr =
+              e2 instanceof GenkitError
+                ? `status ${e2.status}`
+                : 'non-Genkit error';
+            logger.logStructuredWarn(
+              `Fallback model ${targetModelName} failed with ${statusStr}: ${e2.message || String(e2)}. Aborting fallback sequence.`,
```
MichaelDoyle: Paraphrasing, but flip it like so:

```suggestion
              `Aborting fallback sequence for unrecoverable/unhandled error ${error}.`,
```

js/ai/src/model/middleware.ts
```diff
@@ -438,26 +476,67 @@ export function fallback(
...
+                'genkit.middleware.name': 'fallback',
+                'genkit.middleware.fallback.target_model': targetModelName,
+              },
+              e2
+            );
             throw e2;
           }
         }
+        logger.logStructuredWarn(
+          `All fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
```
MichaelDoyle:
```suggestion
          `Fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
```

js/ai/src/model/middleware.ts
```diff
@@ -438,26 +476,67 @@ export function fallback(
...
+            'genkit.middleware.name': 'fallback',
+          },
+          lastError
+        );
         throw lastError;
+      } else {
+        const statusStr =
+          e instanceof GenkitError ? `status ${e.status}` : 'non-Genkit error';
+        logger.logStructuredWarn(
+          `Request failed with ${statusStr}: ${e.message || String(e)}. Skipping fallback.`,
```
MichaelDoyle: Paraphrasing, but flip like so:

```suggestion
          `Skipping fallback for unhandled/unconfigured/etc error foo`,
```

js/plugins/middleware/src/fallback.ts
```diff
@@ -130,20 +139,51 @@ export const fallback: GenerateMiddleware<typeof FallbackOptionsSchema> =
...
                     statuses.includes(e2.status as StatusName)
                   ) {
                     continue;
                   }
+                  const statusStr =
+                    e2 instanceof GenkitError
+                      ? `status ${e2.status}`
+                      : 'non-Genkit error';
+                  logger.logStructuredWarn(
+                    `Fallback model ${model.name} failed with ${statusStr}: ${e2.message || String(e2)}. Aborting fallback sequence.`,
```
MichaelDoyle: Same comment as above.

js/plugins/middleware/src/fallback.ts
```diff
@@ -130,20 +139,51 @@ export const fallback: GenerateMiddleware<typeof FallbackOptionsSchema> =
...
+                      'genkit.middleware.name': 'fallback',
+                      'genkit.middleware.fallback.target_model': model.name,
+                    },
+                    e2
+                  );
                   throw e2;
                 }
               }
+              logger.logStructuredWarn(
+                `All fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
```
MichaelDoyle: Same comment as above.

js/plugins/middleware/src/fallback.ts
```diff
@@ -130,20 +139,51 @@ export const fallback: GenerateMiddleware<typeof FallbackOptionsSchema> =
...
+                lastError
+              );
               throw lastError;
+            } else {
+              const statusStr =
+                e instanceof GenkitError
+                  ? `status ${e.status}`
+                  : 'non-Genkit error';
+              logger.logStructuredWarn(
+                `Request failed with ${statusStr}: ${e.message || String(e)}. Skipping fallback.`,
```
MichaelDoyle: Same comment as above.

js/plugins/middleware/src/fallback.ts
```diff
@@ -109,13 +110,21 @@ export const fallback: GenerateMiddleware<typeof FallbackOptionsSchema> =
         model: async (req, ctx, next) => {
           try {
             return await next(req, ctx);
-          } catch (e) {
-            if (
+          } catch (e: any) {
+            const isFallbackable =
```
MichaelDoyle: nit: isHandled?

js/plugins/middleware/src/retry.ts
```diff
@@ -166,11 +167,15 @@ export const retry: GenerateMiddleware<typeof RetryOptionsSchema> =
               const error = e as Error;
               if (i < maxRetries) {
                 let shouldRetry = false;
+                let reason = '';
                 if (NEVER_RETRY_ERROR_NAMES.includes(error?.name)) {
                   shouldRetry = false;
+                  reason = `error name "${error?.name}" is explicitly non-retryable`;
```
MichaelDoyle: same comment as above

js/plugins/middleware/src/retry.ts
```diff
@@ -191,13 +196,42 @@ export const retry: GenerateMiddleware<typeof RetryOptionsSchema> =
                   if (!noJitter) {
                     delay = delay + 1000 * Math.pow(2, i) * Math.random();
                   }
+                  logger.logStructuredWarn(
+                    `Retry attempt ${i + 1} of ${maxRetries} triggered in ${Math.round(delay)}ms due to error: ${error?.message || String(error)}`,
```
MichaelDoyle: same comment as above

js/plugins/middleware/src/retry.ts
```diff
@@ -191,13 +196,42 @@ export const retry: GenerateMiddleware<typeof RetryOptionsSchema> =
...
+                  );
                   await new Promise((resolve) => __setTimeout(resolve, delay));
                   currentDelay = Math.min(
                     currentDelay * backoffFactor,
                     maxDelayMs
                   );
                   continue;
+                } else {
+                  logger.logStructuredWarn(
+                    `Request failed with error: ${error?.message || String(error)}. Skipping retry because ${reason}.`,
```
MichaelDoyle: same comment as above

js/plugins/middleware/src/retry.ts
```diff
@@ -191,13 +196,42 @@ export const retry: GenerateMiddleware<typeof RetryOptionsSchema> =
...
+                      'genkit.middleware.name': 'retry',
+                      'genkit.middleware.retry.skipped_reason': reason,
+                      'genkit.middleware.retry.max_attempts': maxRetries,
+                    },
+                    error
+                  );
                 }
+              } else {
+                logger.logStructuredWarn(
+                  `All retry attempts exhausted (${maxRetries}). Last error: ${error?.message || String(error)}`,
```
MichaelDoyle: same comment as above

MichaelDoyle (commented):
I'm really not sure sure about an API that has

logger.info
logger.logStructuredInfo
logger.warn
logger.logStructuredWarn
etc, etc

Can we consolidate all of this, in a non-breaking way?

e.g.

```typescript
type LogPayload = string | {
  message: string;
  error?: any;
};

info(payload: LogPayload, ...args: any[]) {
       this._log('info', payload, ...args);
     }

warn(payload: LogPayload, ...args: any[]) {
  this._log('warn', payload, ...args);
}

// etc
```

That way somebody can do all of the following?

 ```typescript
 logger.info('Database connection established');

 logger.info('User completed checkout', { userId: 'usr_98234', cartTotal: 129.99 });

 logger.info('External service call failed', error);

 logger.info('Failed to update inventory', { sku: 'SKU-492', qty: 10 }, error);

 logger.info({
   message: 'Cache synchronization complete',
   metadata: { keysSynchronized: 450, elapsedMs: 120 },
   error: syncError
 });

 logger.info('Attempting retry %d of %d in %dms', attempt, maxRetries, delay);
 ```

js/plugins/middleware/src/fallback.ts
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Good catch, sorry, moved it by mistake.

js/ai/src/model/middleware.ts
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Done

js/ai/src/model/middleware.ts
shrutip90: Done

js/plugins/middleware/src/fallback.ts
shrutip90: Done

js/plugins/middleware/src/fallback.ts
shrutip90: Done

js/plugins/middleware/src/fallback.ts
shrutip90: Done

js/plugins/middleware/src/retry.ts
shrutip90: Done

shrutip90:
> I'm really not sure sure about an API that has
> 
> logger.info logger.logStructuredInfo logger.warn logger.logStructuredWarn etc, etc
> 
> Can we consolidate all of this, in a non-breaking way?
> 
> e.g.
> 
> ```ts
> type LogPayload = string | {
>   message: string;
>   metadata?: Record<string, any>;
>   error?: any;
> };
> 
> info(payload: LogPayload, ...args: any[]) {
>        this._log('info', payload, ...args);
>      }
> 
> warn(payload: LogPayload, ...args: any[]) {
>   this._log('warn', payload, ...args);
> }
> 
> // etc
> ```
> 
> That way somebody can do all of the following?
> 
> ```ts
> logger.info('Database connection established');
> 
> logger.info('User completed checkout', { userId: 'usr_98234', cartTotal: 129.99 });
> 
> logger.info('External service call failed', error);
> 
> logger.info('Failed to update inventory', { sku: 'SKU-492', qty: 10 }, error);
> 
> logger.info({
>   message: 'Cache synchronization complete',
>   metadata: { keysSynchronized: 450, elapsedMs: 120 },
>   error: syncError
> });
> 
> logger.info('Attempting retry %d of %d in %dms', attempt, maxRetries, delay);
> ```

Took a stab at this, please take a look. If it looks good, I will update the plugins to stop using logStructured APIs and also clean up the logStructured APIs introduced in this PR.

js/core/src/logging.ts
```diff
@@ -116,43 +116,130 @@ class Logger {
...
+
+  logStructuredWarn(msg: string, metadata: any, err?: any) {
+    this.warn(msg, metadata, err);
   }
 
-  logStructuredError(msg: string, metadata: any) {
-    getLogger().error(msg, metadata);
-    this._emitOtel('error', [], msg, metadata);
+  logStructuredDebug(msg: string, metadata: any, err?: any) {
+    this.debug(msg, metadata, err);
```
MichaelDoyle: I think we should only keep `logStructured` and `logStructuredError` only, right? Just for backwards compat?

MichaelDoyle (commented):
Looks like you've created the new log methods, but we're not using them (and need to clean up the others)

js/core/src/logging.ts
shrutip90: Done

shrutip90:
> Looks like you've created the new log methods, but we're not using them (and need to clean up the others)

Done!

js/core/src/logging.ts
```diff
@@ -116,43 +116,118 @@ class Logger {
...
+          third instanceof Error ? third : undefined
+        );
+      } else if (args.length > 1) {
+        getLogger()[level].apply(getLogger(), args);
+        this._emitOtel(level, args);
+        return;
+      }
+    }
+
+    getLogger()[level](msg, metadata);
```
pavelgj: Every message-only log now emits a trailing `{}` to the underlying logger, since we always call `getLogger()[level](msg, metadata)` even when `metadata` is empty. For winston that's harmless, but the default/console logger will print `hello {}` instead of `hello` — and this affects every existing `logger.info('...')` call in the codebase, not just middleware. Might be worth only passing metadata when it's non-empty:

```ts
if (Object.keys(metadata).length) getLogger()[level](msg, metadata);
else getLogger()[level](msg);
```

(The test at line 51 asserts the `{}` form, so if this is intentional feel free to ignore.)

js/core/src/logging.ts
```diff
@@ -116,43 +116,118 @@ class Logger {
...
+    }
+
+    if (args.length === 0) return;
+
+    let msg = args[0];
+    let metadata: any = {};
+
+    if (typeof msg === 'object' && msg !== null && !(msg instanceof Error)) {
+      metadata = this._mergeErrorMetadata(msg.metadata, msg.error);
+      msg = msg.message;
```
pavelgj: This branch treats any plain object as a `{message, metadata, error}` payload, so a bare object like `logger.info({ foo: 'bar' })` now logs `msg = undefined` with empty metadata instead of printing the object as before. Could we gate this on the presence of a `message` field so arbitrary objects still pass through? e.g. `if (typeof msg === 'object' && msg !== null && !(msg instanceof Error) && 'message' in msg)`.

js/plugins/middleware/src/fallback.ts
```diff
@@ -130,20 +139,43 @@ export const fallback: GenerateMiddleware<typeof FallbackOptionsSchema> =
...
+                `Fallback options exhausted. Last error: ${lastError.message || String(lastError)}`,
+                {
+                  'genkit.middleware.name': 'fallback',
+                },
+                lastError
+              );
               throw lastError;
+            } else {
+              logger.warn(
+                `Skipping fallback for unhandled error:: ${e.message || String(e)}`,
```
pavelgj: nit: double colon in the message ("unhandled error::"). The V1 copy in `js/ai/src/model/middleware.ts` uses a single colon.

pavelgj:
> > I'm really not sure sure about an API that has
> > logger.info logger.logStructuredInfo logger.warn logger.logStructuredWarn etc, etc
> > Can we consolidate all of this, in a non-breaking way?
> > e.g.
> > ```ts
> > type LogPayload = string | {
> >   message: string;
> >   metadata?: Record<string, any>;
> >   error?: any;
> > };
> > 
> > info(payload: LogPayload, ...args: any[]) {
> >        this._log('info', payload, ...args);
> >      }
> > 
> > warn(payload: LogPayload, ...args: any[]) {
> >   this._log('warn', payload, ...args);
> > }
> > 
> > // etc
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That way somebody can do all of the following?
> > ```ts
> > logger.info('Database connection established');
> > 
> > logger.info('User completed checkout', { userId: 'usr_98234', cartTotal: 129.99 });
> > 
> > logger.info('External service call failed', error);
> > 
> > logger.info('Failed to update inventory', { sku: 'SKU-492', qty: 10 }, error);
> > 
> > logger.info({
> >   message: 'Cache synchronization complete',
> >   metadata: { keysSynchronized: 450, elapsedMs: 120 },
> >   error: syncError
> > });
> > 
> > logger.info('Attempting retry %d of %d in %dms', attempt, maxRetries, delay);
> > ```
> 
> Took a stab at this, please take a look. If it looks good, I will update the plugins to stop using logStructured APIs and also clean up the logStructured APIs introduced in this PR.

+1 on consolidating to debug/info/warn/error + keeping logStructured/logStructuredError as back-compat shims.

One concern: disambiguating by sniffing positional args is subtly breaking:

* Every string-only log now appends `{}`, so the console logger prints `hello {}` instead of `hello` (affects all existing `logger.info('...')` calls).
* Bare objects get swallowed: `logger.info({ foo: 'bar' })` logs `undefined` since any object is treated as a `{message, metadata, error}` payload.

Both are fixable with small guards (only pass metadata when non-empty; gate the payload branch on `'message' in msg`) -- left inline.

Bigger picture: since this is for structured OTel logs, I'd drop printf-style support. Then the overloads collapse to two unambiguous shapes -- `(message, metadata?, error?)` and `({ message, metadata, error })` — and both regressions go away. If we keep printf, let's at least add the guards + tests for the ambiguous cases.

shrutip90:
> `(message, metadata?, error?)` and `({ message, metadata, error })`

Thanks! Removed the printf style overload, but we still need the guards, since without the 'message' in msg, any object passed to logger.info() would be treated as a { message, metadata, error } payload object.

js/core/src/logging.ts
shrutip90: Done

js/core/src/logging.ts
shrutip90: Done

js/plugins/middleware/src/fallback.ts
shrutip90: Done

js/plugins/google-cloud/src/telemetry/path.ts
```diff
@@ -92,7 +92,7 @@ class PathsTelemetry implements Telemetry {
     this.pathLatencies.record(latency, pathDimensions);
 
     const displayPath = truncatePath(toDisplayPath(path));
-    logger.logStructuredError(`Error[${displayPath}, ${errorName}]`, {
+    logger.error(`Error[${displayPath}, ${errorName}]`, {
```
MichaelDoyle, line 95: Where you able to test this one (and the others in `google-cloud`), to make sure the resulting payload is the same?

js/core/src/logging.ts
MichaelDoyle: I think we still have an issue here.

When a bare structured object containing a message property is passed (e.g., `logger.info({ message: 'User checkout', userId: '123' }))`, `msg.metadata` and `msg.error` evaluate to `undefined`, causing `metadata` to become `{}` and stripping all other top-level properties from the log output.                                              
                                                                                                                                                                    
Suggested change:                                                                                                                                                   

```diff                                                                                                                                                                    
    if (                                                                                                                                                          
      typeof msg === 'object' &&                                                                                                                                  
      msg !== null &&                                                                                                                                             
      !(msg instanceof Error) &&                                                                                                                                  
      'message' in msg                                                                                                                                            
    ) {                                                                                                                                                           
-     metadata = this._mergeErrorMetadata(msg.metadata, msg.error);                                                                                               
-     msg = msg.message;                                                                                                                                          
+     const { message, metadata: msgMetadata, error: msgError, ...rest } = msg;                                                                                   
+     metadata = this._mergeErrorMetadata(                                                                                                                        
+       { ...rest, ...(msgMetadata || {}) },                                                                                                                      
+       msgError                                                                                                                                                  
+     );                                                                                                                                                          
+     msg = message;                                                                                                                                              
    } else if (msg instanceof Error) { 
```

js/plugins/middleware/src/skills.ts
```diff
@@ -97,15 +98,36 @@ export const skills: GenerateMiddleware<typeof SkillsOptionsSchema> =
...
+                          },
+                          e
+                        );
+                      }
                     }
                   }
                 }
-              } catch (e) {
+              } catch (e: any) {
                 // ignore directory read errors
```
MichaelDoyle: Nit: consider logging _some_ errors here (e.g. permission errors) that could be useful:

```diff
-             } catch (e: any) {                                                                                                                               
-               // ignore directory read errors                                                                                                                
+             } catch (e: any) {                                                                                                                               
+               if (e.code !== 'ENOENT') {                                                                                                                     
+                 logger.warn(                                                                                                                                 
+                   `Failed to scan skill directory "${dirPath}": ${e.message || String(e)}`,                                                                  
+                   {                                                                                                                                          
+                     'genkit.middleware.name': 'skills',                                                                                                      
+                     'genkit.middleware.skills.path': dirPath,                                                                                                
+                   },                                                                                                                                         
+                   e                                                                                                                                          
+                 );                                                                                                                                           
+               }
```

MichaelDoyle (commented):
This is looking very good. A couple final wrinkles to iron out, and then I will LGTM.

js/core/src/logging.ts
```diff
@@ -116,43 +116,133 @@ class Logger {
     }
   }
 
+  private _log(level: string, ...args: any[]) {
+    const currentLevel = getLogger().level || 'info';
+    if (LOG_LEVELS.indexOf(currentLevel) > LOG_LEVELS.indexOf(level)) {
+      return;
+    }
```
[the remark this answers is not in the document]
shrutip90: Done

js/plugins/middleware/src/skills.ts
```diff
@@ -97,15 +98,36 @@ export const skills: GenerateMiddleware<typeof SkillsOptionsSchema> =
                         path: skillMdPath,
                         description,
                       });
-                    } catch (e) {
-                      // ignore file read errors
+                    } catch (e: any) {
+                      if (e.code !== 'ENOENT') {
```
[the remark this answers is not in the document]
shrutip90: Done

js/plugins/middleware/src/skills.ts
shrutip90: Done

js/plugins/google-cloud/src/telemetry/path.ts
shrutip90: Didn't test all of the paths, but here is one of the error logs case test:

Without the changes here: https://screenshot.googleplex.com/3DaGesRaP6Nkvkk
With the changes: http://screenshot.googleplex.com/6SCGnT9gSKemZF7

js/core/src/logging.ts
shrutip90: Sorry for missing this yesterday. Done.

MichaelDoyle (approved):

PR #5472, opened 2026-06-05
feat(go/plugins/googlegenai): polish Deep Research model support

adesinah:
This PR polishes the Deep Research model support by adding the missing models available in the JavaScript plugin and adding a live test.

## Checklist (if applicable):

### DIFF
--- a/go/plugins/googlegenai/deep_research.go
+++ b/go/plugins/googlegenai/deep_research.go
@@ -1,0 +1,957 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package googlegenai
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/firebase/genkit/go/ai"
+	"github.com/firebase/genkit/go/core"
+	"github.com/firebase/genkit/go/core/api"
+	"github.com/firebase/genkit/go/internal/base"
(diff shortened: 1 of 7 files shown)


go/plugins/googlegenai/deep_research.go
```diff
@@ -0,0 +1,962 @@
...
+		clientOptions := deepResearchOptionsFromOperation(client, op)
+		interaction, err := doDeepResearchRequest(ctx, client, clientOptions, http.MethodGet, "interactions/"+op.ID, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		updatedOp := fromDeepResearchInteraction(interaction)
+		restoreDeepResearchMetadata(updatedOp, op)
+		if updatedOp.Done && updatedOp.Output != nil {
+			if req, ok := updatedOp.Metadata["inputRequest"].(*ai.ModelRequest); ok {
```
cabljac: hmm if this `ok` comes back as `false` - we silently dont calc usage? if the op is unmarshalled from JSON ? do we test in this case?

cabljac: Maybe we need some warning or something here if usage etc is lost?

adesinah: In my opinion, there's genuinely nothing to compute usage on when failed. Also, a failed op deliberately carries `op.Error` and no `Output` ModelResponse, so `Output.Usage` — with no Output, would panic anyways.

cabljac: that makes sense!

PR #5473, opened 2026-06-05
feat(go/plugins/googlegenai): register dedicated Gemini TTS models

adesinah:
This PR registers the Gemini TTS models. Those were already available in JS plugin, but missing for Go.

## Checklist (if applicable):


go/plugins/googlegenai/models.go
```diff
@@ -52,6 +52,18 @@ var (
...
+	// TTSSupports describes model capabilities for text-to-speech models
+	// (gemini-*-tts). They emit audio and, unlike conversational Gemini models,
+	// do not support tools, multi-turn history, or system roles.
+	TTSSupports = ai.ModelSupports{
+		Multiturn:  false,
+		Media:      false,
+		Tools:      false,
+		ToolChoice: false,
+		SystemRole: false,
+		Output:     []string{"audio"},
```
cabljac: `Output:     []string{"audio"},` is this correct? I see "media" used elsewhere (including lyria)

cabljac:
This looks good! Left one small question.

go/plugins/googlegenai/models.go
adesinah: Good point. since this is TTS, it means you are getting an `audio` back from the text. media could mean image, video, etc

cabljac: https://github.com/genkit-ai/genkit/blob/783fbf23287484b6514732808c2b9a814848a402/go/ai/formatter.go#L24-L38

I don't think audio is actually supported?

cabljac: I think this is more "format" rather than "modality" - the contentType should carry audio right?

cabljac (approved):
lgtm

PR #5488, opened 2026-06-08
refactor(go/plugins): extract modality string constants

IzaakGough:
## Summary

Refactor the Go `googlegenai` plugin to replace repeated modality string literals with shared constants, and update the related TTS assertion to use the same constant.

## Problem/Root Cause

`models.go` repeated raw modality strings such as `"media"`, `"text"`, `"image"`, and `"video"` across model and embedder capability definitions. Keeping these values duplicated makes the registry harder to maintain and increases the risk of inconsistent updates when a modality value is reused in multiple places.

## Solution/Changes

- Add shared modality constants for media, text, image, and video in the Go Google GenAI plugin.
- Replace repeated string literals in model output and embedder input capability declarations with those constants.
- Update the existing TTS capability test to assert against the shared media constant instead of a raw string literal.

## Testing

- Updated the existing TTS model options test to use the shared modality constant.
- No additional test runs were verified from the PR evidence.


adesinah (changes requested):
The changes look good. This PR needs to convert this chore for other Go plugins too.

Issue #5489, opened 2026-06-09
googlegenai TTS models return 400 INVALID_ARGUMENT when no voice is set


adesinah:
The dedicated Gemini TTS models (`gemini-2.5-flash-preview-tts`, `gemini-2.5-pro-preview-tts`, `gemini-3.1-flash-tts-preview`) fail with **HTTP 400 `INVALID_ARGUMENT`** when invoked with only a prompt and no `speechConfig` — e.g. running the model from the Dev UI. Affects both the Go (`go/plugins/googlegenai`) and JS (`js/plugins/google-genai`) plugins.

### Cause
Gemini `generateContent` TTS requires a `speechConfig` with a voice. The plugins set `responseModalities: ["AUDIO"]` for TTS models but never supply a voice, so the request is invalid. `gemini-2.5-*-tts` tolerates this (server default voice); `gemini-3.1-flash-tts-preview` does not and returns 400.

### Reproduce
Run a TTS model with just a prompt:
```ts
await ai.generate({
  model: googleAI.model("gemini-3.1-flash-tts-preview"),
  prompt: "Say: Genkit is great",
});
```
Result:
```
Error 400, Message: Request contains an invalid argument., Status: INVALID_ARGUMENT, Details: []
```

### Expected
A bare prompt to a dedicated TTS model should produce audio. A sensible default voice should be applied when the caller does not specify one (still overridable via config).

### Proposed fix
For TTS models, when no `speechConfig` is provided, default it to a prebuilt voice (e.g. `Algenib`) in both plugins, guarded so caller-provided voices still win and non-TTS models are untouched.

### Workaround
Pass an explicit voice in the config:
```ts
config: { speechConfig: { voiceConfig: { prebuiltVoiceConfig: { voiceName: "Algenib" } } } }
```

PR #5491, opened 2026-06-09
fix(js/plugins/google-genai): default a voice for Gemini TTS requests

adesinah:
This PR defaults a voice for Gemini TTS requests. Issue here: #5489 

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/googleai/gemini.ts
+++ b/js/plugins/google-genai/src/googleai/gemini.ts
@@ -845,6 +845,15 @@ export function defineModel(
         if (!generationConfig.responseModalities) {
           generationConfig.responseModalities = ['AUDIO'];
         }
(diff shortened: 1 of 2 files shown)


cabljac (approved):
lgtm

cabljac (changes requested):
Sorry, was a bit hasty to approve here - is the check for a missing config sufficient?

adesinah:
> Sorry, was a bit hasty to approve here - is the check for a missing config sufficient?

Good point. - addressed now.

cabljac (approved):
lgtm

ifielker:
Why are we switching the default voice to Algenib? There is already a default if you just leave it blank... and it's not Algenib. So this will result in users who previously had no voice config, suddenly changing from default to Algenib instead... (very different voices). Let's not do that.

PR #5497, opened 2026-06-09
docs(go/plugins/googlegenai): document gemini-3.1 tts model behaviour and add sample

IzaakGough:
## Summary

Document the dedicated Gemini TTS models in the Go `googlegenai` plugin and add a sample showing how to use them, including the Gemini 3.1 PCM-to-WAV handling needed to produce playable output.

## Problem/Root Cause

The Go plugin exposed dedicated Gemini TTS model IDs, but the behavior and usage of those models were not documented. In particular, `gemini-3.1-flash-tts-preview` returns PCM audio rather than a directly playable WAV file, so users needed sample code showing how to decode the inline media response and wrap it in a WAV container.

## Solution/Changes

- Register the dedicated Gemini TTS models with TTS-specific capabilities in the Google AI model registry.
- Add tests covering model classification, declared capabilities, and Google AI-only registration.
- Expand the plugin README with TTS model documentation, usage examples, and a note describing the Gemini 3.1 PCM output behavior.
- Update the Go text-to-speech sample to:
  - return inline media for the Gemini 2.5 TTS flow
  - add a Gemini 3.1 flow that decodes inline PCM audio, converts it to WAV, and returns the generated file path

## Testing

- Added `go/plugins/googlegenai/tts_test.go` covering TTS model registration and capabilities.
- Ran `go test ./plugins/googlegenai/...`
- Ran `go build ./samples/text-to-speech`

### DIFF
--- a/go/plugins/googlegenai/README.md
+++ b/go/plugins/googlegenai/README.md
@@ -26,21 +26,21 @@ You can use either the Google AI (Gemini API) or Vertex AI backend.
 
 ```go
 import (
- "context"
- "log"
+	"context"
+	"log"
 
- "github.com/firebase/genkit/go/genkit"
- "github.com/firebase/genkit/go/plugins/googlegenai"
+	"github.com/firebase/genkit/go/genkit"
+	"github.com/firebase/genkit/go/plugins/googlegenai"
 )
 
 func main() {
- ctx := context.Background()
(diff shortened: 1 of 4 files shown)


IzaakGough:
Closing in favour of https://github.com/genkit-ai/genkit/pull/5473

PR #5506, opened 2026-06-10
feat(tools): include structured output in exported .prompt files

chrisraygill:
## Summary

This is the **server-side counterpart** to the editable output config added to the model runner in **genkit-ai/genkit-ui#1919**. That PR lets you set structured output in the Dev UI model runner; this PR makes the **Export to `.prompt`** action actually include it.

> Linked PR: https://github.com/genkit-ai/genkit-ui/pull/1919

Previously `createPrompt` dropped the generation's `output` config entirely, so a flow that used structured output lost its schema when exported. The exported front matter now includes the output, with the JSON Schema written as **Picoschema**.

### Before
```yaml
---
model: googleai/gemini-flash-latest
config:
  temperature: 0.7
---
```

### After
```yaml
---
model: googleai/gemini-flash-latest
config:
  temperature: 0.7
output:
  format: json
  schema:
    title: string
    servings: integer, number of servings
    ingredients(array):
      name: string
      onSale: boolean
    steps(array): string
---
```

## What changed

- **`CreatePromptRequestSchema`** now accepts `output` (it was being stripped by tRPC input validation before reaching the frontmatter).
- **`createPrompt`** maps the output config onto the frontmatter, reading the schema from `jsonSchema` (generate action shape) or `schema` (model request shape), and mapping the JSON-producing formats (`json`, `jsonl`, `array`, `enum`) onto `json`.
- **`jsonSchemaToPicoschema`** converts an object JSON Schema into the compact Picoschema form: required/optional (`?`), descriptions, enums, scalar and object arrays, nested objects, and `additionalProperties` wildcards. Non-object top-level schemas (a bare array or scalar) pass through as raw JSON Schema, which Dotprompt also accepts.

## Testing

- Unit tests in `genkit-tools/common/tests/utils/prompt_test.ts` cover the converter and the frontmatter mapping (15 cases).
- Verified end to end against a running Dev UI: exporting a structured-output prompt from the model runner now produces the Picoschema `output` block above.

## Notes

- The frontmatter `format` field is limited to `json`/`text`/`media`, so `jsonl`/`array`/`enum` export as `format: json` with the schema preserved. Exact format fidelity could be a follow-up.

## Checklist

- [x] Tested (unit tests + end-to-end via the Dev UI export)
- [ ] Docs updated (not required)

### DIFF
--- a/genkit-tools/common/src/utils/prompt.ts
+++ b/genkit-tools/common/src/utils/prompt.ts
@@ -18,7 +18,90 @@ import { stringify } from 'yaml';
 import type { MessageData, Part } from '../types/model';
 import type { PromptFrontmatter } from '../types/prompt';
 
-export function fromMessages(
+function toFrontmatterSchema(config?: {
+  schema?: unknown;
+  jsonSchema?: unknown;
+}): unknown | undefined {
+  const schema = config?.jsonSchema ?? config?.schema;
+  return schema && typeof schema === 'object' ? schema : undefined;
+}
+
+/**
+ * Maps a generate request's output config onto `.prompt` frontmatter. The
+ * frontmatter format is limited to json/text/media, so the JSON-producing
+ * formats (json, jsonl, array, enum) map onto `json`. Returns undefined when
+ * there is nothing to record.
+ */
+export function toFrontmatterOutput(output?: {
+  format?: string;
+  jsonSchema?: unknown;
+  schema?: unknown;
+}): PromptFrontmatter['output'] | undefined {
+  if (!output) return undefined;
+  const result: NonNullable<PromptFrontmatter['output']> = {};
+  if (output.format === 'text') {
+    result.format = 'text';
+  } else if (output.format === 'media') {
+    result.format = 'media';
+  } else if (output.format) {
+    result.format = 'json';
+  }
+  const schema = toFrontmatterSchema(output);
+  if (schema !== undefined) {
+    result.schema = schema;
+  }
+  return result.format || result.schema ? result : undefined;
+}
+
+/**
+ * Maps a request's input config onto `.prompt` frontmatter. Returns undefined
+ * when there is nothing to record.
+ */
+export function toFrontmatterInput(input?: {
+  schema?: unknown;
+  jsonSchema?: unknown;
+  default?: unknown;
+}): PromptFrontmatter['input'] | undefined {
+  if (!input) return undefined;
+  const result: NonNullable<PromptFrontmatter['input']> = {
+    schema: toFrontmatterSchema(input),
+    default: input.default,
+  };
+  return result.schema || result.default !== undefined ? result : undefined;
+}
+
+/**
+ * Converts a prompt creation request into a complete `.prompt` template file string.
+ */
+export function toPromptFile(request: {
+  model: string;
+  messages: MessageData[];
+  config?: Record<string, unknown>;
+  tools?: { name: string }[];
+  use?: PromptFrontmatter['use'];
+  input?: {
+    schema?: unknown;
+    jsonSchema?: unknown;
+    default?: unknown;
+  };
+  output?: {
+    format?: string;
+    jsonSchema?: unknown;
+    schema?: unknown;
+  };
+}): string {
+  const frontmatter: PromptFrontmatter = {
+    model: request.model.replace('/model/', ''),
+    config: request.config,
+    tools: request.tools?.map((toolDefinition) => toolDefinition.name),
+    use: request.use,
+    input: toFrontmatterInput(request.input),
+    output: toFrontmatterOutput(request.output),
+  };
+  return renderPromptFile(frontmatter, request.messages);
+}
+
+export function renderPromptFile(
   frontmatter: PromptFrontmatter,
   messages: MessageData[]
 ): string {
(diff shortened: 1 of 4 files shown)


genkit-tools/common/src/utils/prompt.ts
```diff
@@ -18,6 +18,119 @@ import { stringify } from 'yaml';
...
+    case 'string':
+    case 'integer':
+    case 'number':
+    case 'boolean':
+    case 'null':
+      return schema.type;
+    default:
+      return 'any';
+  }
+}
```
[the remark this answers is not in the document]
chrisraygill: Good catch. Fixed in 3da00a739: `scalarType` now reads `schema?.type` and falls back to `any` for a null or malformed property, with a regression test covering it.

genkit-tools/common/tests/utils/prompt_test.ts
```diff
@@ -182,3 +186,128 @@ describe('fromMessages', () => {
...
+    });
+  });
+
+  it('maps json-producing formats onto json', () => {
+    expect(
+      toFrontmatterOutput({ format: 'jsonl', jsonSchema: SCHEMA })?.format
+    ).toBe('json');
+  });
+
+  it('keeps text and media formats', () => {
```
MichaelDoyle: Make 2 separate tests, please.

genkit-tools/common/tests/utils/prompt_test.ts
```diff
@@ -182,3 +186,128 @@ describe('fromMessages', () => {
...
+      required: ['title', 'servings'],
+    };
+    expect(jsonSchemaToPicoschema(schema)).toEqual({
+      title: 'string',
+      'subtitle?': 'string, optional subtitle',
+      servings: 'integer',
+    });
+  });
+
+  it('encodes enums, arrays of scalars, arrays of objects, and nested objects', () => {
```
MichaelDoyle: Break this into a separate test for each thing we are asserting (have an "expect" for).

genkit-tools/common/tests/utils/prompt_test.ts
```diff
@@ -182,3 +186,128 @@ describe('fromMessages', () => {
...
+      '(*)': 'number',
+    });
+  });
+
+  it('passes non-object top-level schemas through unchanged', () => {
+    const arraySchema = { type: 'array', items: { type: 'string' } };
+    expect(jsonSchemaToPicoschema(arraySchema)).toBe(arraySchema);
+  });
+
+  it('does not crash on a null or malformed property', () => {
```
MichaelDoyle: nit: "doesn't crash" is good :) but it doesn't describe _what_ it does instead.

A basic improvement: `it('handles null or malformed properties')`
And if there is something special it does then you can be even more descriptive: ``it('returns any for null or malformed properties')``

genkit-tools/common/src/utils/prompt.ts
```diff
@@ -18,6 +18,119 @@ import { stringify } from 'yaml';
...
+  if (s.type === 'object' || s.properties) {
+    return picoObject(s);
+  }
+  return schema;
+}
+
+/**
+ * Maps a generate request's output config onto `.prompt` frontmatter, converting
+ * the JSON Schema to Picoschema. The frontmatter format is limited to
+ * json/text/media, so the JSON-producing formats (json, jsonl, array, enum) map
```
MichaelDoyle: This is interesting -- so there is a mismatch between what can be done in code, and what dotprompt supports?

If so, whenever we do such a conversion, let's leave a comment in the produced dotprompt that explains.
e.g.) `# converted from <type>`

MichaelDoyle (commented):
This is really cool.

Is the conversion code useful for other applications, besides dev ui generating dotprompts? I wonder if that code should live somewhere else. e.g. in the dotprompt library, and we just depend on it here.

genkit-tools/common/src/utils/prompt.ts
pavelgj: JSON Schema to Picoschema conversion is lossy... Picoschema has a very limited subset of features.

genkit-tools/common/tests/utils/prompt_test.ts
chrisraygill: Done in 5c44f3f27: split into separate tests for an enum, an array of scalars, an array of objects, and a nested object, each with its own schema and a single assertion.

genkit-tools/common/tests/utils/prompt_test.ts
chrisraygill: Done in 5c44f3f27: split into `keeps the text format` and `keeps the media format`.

genkit-tools/common/tests/utils/prompt_test.ts
chrisraygill: Done in 5c44f3f27: renamed to `returns any for null or malformed properties`.

genkit-tools/common/src/utils/prompt.ts
chrisraygill: So, how do we think output format set to `jsonl` should work when converted to picoschema? What if the JSON schema includes unsupported features in picoschema?

MichaelDoyle: I moved pico schema out into another branch (stacked on top). can discuss there. 

I think we can make it optional via dev ui, so we have a lossless version via json schema, but can default? to picoSchema or enable it as a toggle-able option.

MichaelDoyle (approved):

PR #5541, opened 2026-06-14
feat(python): agents implementation

huangjeff5:
## Summary

Adds core **Agent** and **Session** support to Genkit Python: a stateful, streaming, multi-turn layer on top of `generate`, with pluggable session persistence, human-in-the-loop interrupt/resume, and a remote HTTP client. Also adds the **Artifacts** middleware (read/write artifact tools + session artifact listing) to the existing middleware framework.

## Define an agent, stream a turn

```python
agent = ai.define_agent(
    name='weatherAgent',
    model='googleai/gemini-flash-latest',
    system='Weather assistant. Use getWeather for weather questions.',
    tools=[get_weather],
    store=InMemorySessionStore(),
)

turn = agent.chat().send('Weather in Paris?')
async for chunk in turn.stream:
    for call in chunk.tool_requests:            # tools light up as they're called
        print(f'  → {call.tool_request.name}')
    if chunk.artifact:                          # artifacts as they're written
        print(f'  [artifact] {chunk.artifact.name}')
        
res = await turn.response
print(res.text, res.finish_reason, res.artifacts)
```

## Persist and resume from a snapshot id

```python
res = await chat.send('Weather in Paris?').response
checkpoint = res.snapshot_id       # your app only stores this string

# later, after a reconnect or server restart:
resumed = await agent.load_chat(snapshot_id=checkpoint)
await resumed.send('What city did I ask about?').response   # still knows "Paris"
```

`FileSessionStore(directory=...)` swaps in for cross-process persistence with the same API.

## Interrupt & resume (human-in-the-loop)

`ToolApproval` pauses a turn before a sensitive tool runs; a human approves, one `resume` continues it.

```python
out = await chat.send('Transfer $500 to account 12345.').response
assert out.finish_reason == AgentFinishReason.INTERRUPTED   # tool pending, not executed

restart = [i.restart(resumed_metadata={'tool_approved': True}) for i in out.interrupts]
out = await chat.resume(restart=restart).response
assert out.finish_reason == AgentFinishReason.STOP
```

## Talking to a remote agent over HTTP

`remote_agent(...)` gives you the exact same `AgentChat` API as a local agent — the turns just run in another process. You never assemble history yourself: the chat keeps a running `messages` view for you.

```python
from genkit.agent import remote_agent

# Server-managed: the remote store owns the session; each turn only sends the
# reply back over the wire, and the client ships a resume handle, not state.
agent = remote_agent(url='http://localhost:8080/bankingAgent', state_management='server')
chat = agent.chat()

res = await chat.send('Transfer $500 to account 12345.').response
res = await chat.send('And $20 to account 999.').response

# History is managed for you — just read it off the chat handle.
print(chat.messages)     # running conversation, both turns
```


py/packages/genkit/src/genkit/_core/_middleware.py
```diff
@@ -175,33 +195,19 @@ class RetryConfig(BaseModel):
             max_retries: int = 3
 
     2. Extend BaseMiddleware with your config model, e.g. Retry:
```
huangjeff5, line 180: Revert these docs changes

py/packages/genkit/src/genkit/agent/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    'ActionRunContext',
+    'AgentFn',
+    'SessionRunner',
+    # Session persistence
+    'Session',
+    'SessionStore',
+    'SnapshotAborter',
+    'supports_abort',
+    'run_with_session',
+    'validate_resume_against_history',
```
huangjeff5: is this needed?

py/packages/genkit/src/genkit/agent/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    'AgentConnection',
+    # Agent function protocol
+    'ActionRunContext',
+    'AgentFn',
+    'SessionRunner',
+    # Session persistence
+    'Session',
+    'SessionStore',
+    'SnapshotAborter',
+    'supports_abort',
```
huangjeff5: should we export supports_abort? where is it used?

py/packages/genkit/tests/genkit/ai/_tools_test.py
```diff
@@ -79,7 +93,7 @@ async def t2(inp: dict, ctx: ToolRunContext) -> str:  # noqa: ARG001
         tool_request=ToolRequest(name='t2', ref='x', input={'q': 1}),
         metadata={'resumed': True},
     )
-    await run_tool_after_restart(action, restart_trp)
+    await run_tool_after_restart(action, restart_trp, asyncio.Event())
```
huangjeff5: allow not passing abort signal in run_tool_after_restart (adjust all test cases)

py/pyproject.toml
```diff
@@ -181,7 +182,7 @@ genkit-plugin-ollama                = { workspace = true }
 genkit-plugin-evaluators            = { workspace = true }
 genkit-plugin-vertex-ai             = { workspace = true }
 [tool.uv.workspace]
-exclude = ["*/shared", "testapps/*"]
+exclude = ["*/shared", "testapps/*", "samples/fastapi-serve", "samples/usechat-serve"]
```
huangjeff5: remove these samples

py/pyproject.toml
```diff
@@ -259,6 +260,7 @@ select = [
 "packages/genkit/src/genkit/_core/_typing.py" = ["E501"]
 # Base model uses **kwargs: Any to match pydantic signatures
 "packages/genkit/src/genkit/_core/_base.py" = ["ANN401"]
+"packages/genkit/src/genkit/_ai/_json_patch.py" = ["ANN401"]
```
huangjeff5, line 230: leave a code comment if this is needed

py/samples/agents/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```bash
+cd py/samples/agents
+uv sync
+uv run python src/examples/basic_samples/01_define_agent_with_store.py
+```
+
+See [`src/examples/README.md`](src/examples/README.md) for the full matrix (`basic_samples/`, `branching_samples/`). Each example defines its agent inline.
+
+## FastAPI / HTTP
```
huangjeff5: Remove this part and below

py/packages/genkit-middleware/src/genkit_middleware/_artifacts.py
```diff
@@ -0,0 +1,226 @@
...
+    new_request.messages = inject_artifact_listing_messages(list(request.messages), listing)
+    return new_request
+
+
+class Artifacts(BaseMiddleware[ArtifactsConfig]):
+    """Session artifact tools plus an injected artifact listing in the system prompt."""
+
+    def tools(self, ctx: GenerateMiddlewareContext) -> list[Any]:
+        scratch = Registry()
+        tools: list[Any] = []
```
huangjeff5: We don't need to register these tools in the registry; just return the inline tools and the framework will register it.

py/packages/genkit/src/genkit/_core/_middleware.py
```diff
@@ -96,6 +97,24 @@ class GenerateHookParams(BaseModel):
 
     Covers one full iteration of the tool loop: a model call plus optional tool
     resolution. ``message_index`` tracks streaming position for this turn.
+
```
huangjeff5: Update docstring

py/pyproject.toml
```diff
@@ -435,6 +438,7 @@ search-path = [
   "packages/genkit-ollama/src",
   "packages/genkit-vertexai/src",
 ]
+ignore-missing-imports = ["genkit.plugins.*"]
```
huangjeff5: remove, not needed

py/packages/genkit/src/genkit/_core/_middleware.py
```diff
@@ -174,33 +178,19 @@ class RetryConfig(BaseModel):
             max_retries: int = 3
 
     2. Extend BaseMiddleware with your config model, e.g. Retry:
-        ai = Genkit()
```
huangjeff5, line 177: revert docstring changes

py/packages/genkit/src/genkit/_ai/_agents/_transports/_http.py
```diff
@@ -0,0 +1,264 @@
...
+        self,
+        *,
+        agent_input: AgentInput,
+        init: AgentInit,
+    ) -> tuple[AsyncIterable[AgentStreamChunk], Awaitable[AgentOutput]]:
+        """Runs a single turn over HTTP using a streaming POST request."""
+        client = get_cached_client('agent_transport')
+
+        payload: dict[str, Any] = {
+            'input': agent_input.model_dump(by_alias=True, exclude_none=True),
```
pavelgj: This request body doesn't match the JS wire protocol, and I don't think that's intended. JS `remoteAgent` posts the callable/flow envelope: `{"data": <AgentInput>, "init": <AgentInit>}`, and negotiates streaming via the `Accept: text/event-stream` header only (see `js/genkit/src/client/client.ts`). JS agents are served with `expressHandler(agent)` (`js/testapps/agents/src/index.ts`), which reads `req.body.data`.

Here we send `{"input": ..., "init": ..., "key"?: ...}` to `{url}?stream=true`, which is the reflection `runAction` shape. Two consequences: a Python `remote_agent` can't call a JS `expressHandler`-served agent (`data` vs `input`), and it also won't match our own FastAPI/Flask/Django flow handlers, which 400 anything not wrapped in `{"data": ...}`.

Can we align the request to `{data, init}` + the `Accept` header so it's interoperable with the JS server?

py/packages/genkit/src/genkit/_ai/_agents/_session_stores/_file_store.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        self.reject_ambiguous = reject_ambiguous_session
+        self.max_persisted_chain_length = max_persisted_chain_length
+        self.directory = directory
+        self.subs: Subs = {}
+        os.makedirs(directory, exist_ok=True)
+
+    def path(self, snapshot_id: str) -> str:
+        """Return the file path for a snapshot ID."""
+        return os.path.join(self.directory, f'{snapshot_id}.json')
```
pavelgj, line 74: `path()` joins the raw `snapshot_id` straight into a filename. Snapshot/session ids can arrive off the wire (the abort/getSnapshot actions take a bare string), so a crafted id like `../../foo` would let a caller read or write outside `directory`. JS guards exactly this in `session-stores.ts` (there's a comment there about `../../foo`). Could we port that sanitization, or at least reject ids containing path separators / `..` before building the path?

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
```diff
@@ -0,0 +1,935 @@
...
+                snapshot_id=existing.snapshot_id if existing else '',
+                parent_id=parent_id or '',
+                status=snap_status,
+                state=state,
+                created_at=existing.created_at if existing and existing.created_at else now,
+                finish_reason=finish_reason,
+                error=error,
+            )
+
+        snap = await self.store.save_snapshot(None, make_snap)
```
pavelgj: Looks like the reserved-snapshot-id flow didn't come over. In JS the turn's `snapshotId` is reserved up front and handed to the handler via `TurnContext` (`reserveSnapshotId` + `TurnContext` in `agent.ts`), so a handler can name external, snapshot-correlated resources (e.g. a git worktree) before the turn runs and then commit under that id. Here the id is minted inside the store on save (`save_snapshot(None, ...)`), and there's no `TurnContext` surface. Is this a deliberate scope cut for v1, or should we carry the reserved-id + per-turn context through? Fine either way, just want it to be a conscious call rather than an accidental gap.

pavelgj (commented):
A few notes from a JS-parity pass (wire protocol + concept-level consistency). Overall this is a really solid port; the type/schema layer, JSON-Patch, artifacts middleware, snapshot lifecycle, and detach/heartbeat all line up well with JS. Inline comments below cover a couple of things worth resolving before merge plus one to confirm is a deliberate scope call.

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
```diff
@@ -0,0 +1,935 @@
...
+    client is validated against it before the session is built.
+    """
+    name = agent_name or 'agent'
+
+    if init.snapshot_id and init.session_id:
+        raise GenkitError(
+            status='INVALID_ARGUMENT',
+            message=(f"Cannot send both 'snapshot_id' and 'session_id' to agent '{name}'. Provide exactly one."),
+        )
+    if (init.snapshot_id or init.session_id) and store is None:
```
pavelgj: JS splits two kinds of pre-turn failure: API misuse (wrong state-management init) throws `AgentInitError` and maps to an HTTP status, while recoverable pre-turn problems (missing snapshot, non-resumable snapshot, invalid custom state) are absorbed into a graceful `finishReason: 'failed'` result carrying last-good state. `load_session` raises `GenkitError` for all of these uniformly. Can you confirm the caller still funnels the recoverable cases into a failed `AgentOutput` rather than propagating? If it does, this is just a naming/structure nit; if not, the behavior diverges from JS.

py/packages/genkit/src/genkit/_ai/_agents/_transports/_http.py
huangjeff5: Updated HttpAgentTransport to post {data, init} and stream via Accept: text/event-stream (no ?stream=true). Get/abort wrap the lookup in {data: ...} as well. Covered by agent_http_wire_test + the rewritten agent_transports_test against a flow-shaped Starlette endpoint. (Plugin handler init threading for Flask/Django is on #5686 with the FastAPI serve_agent work.)

py/packages/genkit/src/genkit/_ai/_agents/_session_stores/_file_store.py
huangjeff5: Added assert_safe_snapshot_id and call it from FileSessionStore.path() so separators / .. / . reject with INVALID_ARGUMENT before join. Tests in agent_session_stores_test.

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
huangjeff5: I see, that makes sense. I'll add it now so we don't have to refactor the flow later. 

Turns with a store now reserve the id up front (reserve_snapshot_id), pass it on TurnContext, and save_snapshot requires that id (no mint-on-None). Sample 18_turn_context_workspace.py shows naming an on-disk workspace after the reserved id during the turn.

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
huangjeff5: Updated to match. State-management misuse on init (e.g. sending state to a store-backed agent) raises AgentInitError out of the agent's bidi action before the turn starts. If init is valid for the agent but the snapshot is missing / not resumable, that becomes an AgentOutput with finish_reason=failed on the server; the client (local or remote) surfaces it as AgentError.

py/packages/genkit/src/genkit/_ai/_agents/_transports/_http.py
```diff
@@ -0,0 +1,264 @@
...
+
+    async def abort_snapshot(self, snapshot_id: str) -> SnapshotStatus | None:
+        """Aborts the specified snapshot on the server."""
+        result = await self._post_json(url=self.abort_url, input_val={'snapshotId': snapshot_id})
+        if result is None:
+            return None
+        return AgentAbortResponse.model_validate(result).status
+
+
+def remote_agent(
```
pavelgj, line 279: One gap vs the JS `remoteAgent`: there's no way to pass `headers` here. JS accepts either a static header map or an async function resolved per request (handy for auth tokens that need refreshing) and applies them to the turn, `getSnapshot`, and `abort` calls. Right now `HttpAgentTransport` only ever sends `Accept`/`Content-Type` (line 168), so there's no way to reach an authenticated remote agent. Could we add a `headers` param (static dict or callable) and thread it through both `_post_json` and the streaming POST? (`agent_name` being unused is fine, it's documented — just flagging headers.)

py/packages/genkit/src/genkit/_ai/_agents/_client.py
```diff
@@ -0,0 +1,1369 @@
...
+    ) -> AgentTurn[StateT]:
+        """Continues a conversation from an interrupt.
+
+        ``respond`` answers paused tool requests; ``restart`` re-runs them with
+        new metadata. Build each part with the interrupt's ``respond`` / ``restart``
+        helpers. ``metadata`` rides along on the resumed tool message (the same
+        ``resume_metadata`` a plain ``generate`` call takes). Sugar for ``send``
+        with a resume payload.
+        """
+        inp = AgentInput(resume=Resume(respond=respond, restart=restart, metadata=metadata))
```
pavelgj, line 1019: Not for this PR, but worth capturing: `resume.metadata` isn't in the canonical agent schema — both `js/ai/src/agent-types.ts` and `genkit-tools/common/src/types/agent.ts` define `resume` as just `{respond, restart}`. So this is currently a Python-only extension: a Python client talking to a JS agent would have it silently dropped, and the Dev UI (which builds resume payloads off the canonical schema) won't round-trip it — so interrupt resumes that lean on `resume.metadata` will likely break in the UI. Suggest a follow-up to add `metadata` to the shared schema so the languages + tooling stay in lockstep.

MichaelDoyle:
@huangjeff5 I noticed we're missing observability for agent init, which TS and Go have.

---

## Background & Rationale
In Genkit:
- **TypeScript** records `init` on span metadata and serializes it to `'genkit:init'` in `js/core/src/tracing/instrumentation.ts`.
- **Go** records `sm.Init` as the `'genkit:init'` attribute in `go/core/tracing/tracing.go`.
- **Python** currently validates and passes `init` payloads via `ActionRunContext.init` in `py/packages/genkit/src/genkit/_core/_action.py`, but `SpanMetadata` lacks an `init` field and does not attach it to exported spans.

---

## Required Changes

### 1. Update Core Tracing (`_tracing.py`)
Add `init` to `SpanMetadata` in `py/packages/genkit/src/genkit/_core/_tracing.py` and serialize it as `'genkit:init'` inside `run_in_new_span`.

```diff
--- a/py/packages/genkit/src/genkit/_core/_tracing.py
+++ b/py/packages/genkit/src/genkit/_core/_tracing.py
@@ -62,6 +62,7 @@ class SpanMetadata(GenkitModel):
     state: Literal['success', 'error'] | None = None
     input: Any | None = Field(default=None)
     output: Any | None = Field(default=None)
+    init: Any | None = Field(default=None)
     is_root: bool | None = None
     metadata: dict[str, Any] | None = None
     path: str | None = None
@@ -175,6 +176,8 @@ def run_in_new_span(
                 span.set_attribute('genkit:metadata:subtype', metadata.subtype)
             if metadata.input is not None:
                 span.set_attribute('genkit:input', _to_json_attr(metadata.input))
+            if metadata.init is not None:
+                span.set_attribute('genkit:init', _to_json_attr(metadata.init))
             if metadata.metadata:
                 for meta_key, meta_value in metadata.metadata.items():
                     span.set_attribute(f'genkit:metadata:{meta_key}', str(meta_value))
```

### 2. Pass `ctx.init` in Action Telemetry (`_action.py`)
In `_run_with_telemetry` (`py/packages/genkit/src/genkit/_core/_action.py`), pass `ctx.init` when constructing `SpanMetadata`.

```diff
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -746,6 +746,7 @@ class Action(Generic[InputT, OutputT, ChunkT, InitT]):
             type='action',
             subtype=str(self._kind),
             input=input,
+            init=ctx.init,
             metadata=extra_metadata or None,
             telemetry_labels={k: str(v) for k, v in (telemetry_labels or {}).items()} or None,
         )
```

### 3. Add Unit Test (`run_in_new_span_test.py`)
Add a test in `py/packages/genkit/tests/genkit/core/run_in_new_span_test.py` verifying that `init` metadata is correctly serialized to the `'genkit:init'` span attribute.

```diff
--- a/py/packages/genkit/tests/genkit/core/run_in_new_span_test.py
+++ b/py/packages/genkit/tests/genkit/core/run_in_new_span_test.py
@@ -83,6 +83,15 @@ def test_writes_input_from_metadata(exporter: InMemorySpanExporter) -> None:
     assert attrs['genkit:metadata:subtype'] == 'tool'
 
 
+def test_writes_init_from_metadata(exporter: InMemorySpanExporter) -> None:
+    with run_in_new_span(SpanMetadata(name='agentRun', type='action', init={'sessionId': 'session-123'})):
+        pass
+
+    span = _by_name(exporter.get_finished_spans(), 'agentRun')
+    attrs = dict(span.attributes or {})
+    assert attrs['genkit:init'] == '{"sessionId": "session-123"}'
+
+
 def test_writes_output_from_metadata_on_success(exporter: InMemorySpanExporter) -> None:
     meta = SpanMetadata(name='answer', type='util')
     with run_in_new_span(meta):
```

py/packages/genkit/src/genkit/_ai/_agents/_client.py
huangjeff5: I see. Sounds good, I think it is the right call to include metadata on Resume (at least that's how the generate Resume object behaves) so I threaded it through from the Agent. (More details here: https://github.com/genkit-ai/genkit/issues/5826). Should be a small change so I'll work on the followup PR.

py/packages/genkit/src/genkit/_ai/_agents/_transports/_http.py
huangjeff5: Added headers, and also removed agent_name.

huangjeff5:
> @huangjeff5 I noticed we're missing observability for agent init, which TS and Go have.
> 
> ## Background & Rationale
> In Genkit:
> 
> * **TypeScript** records `init` on span metadata and serializes it to `'genkit:init'` in `js/core/src/tracing/instrumentation.ts`.
> * **Go** records `sm.Init` as the `'genkit:init'` attribute in `go/core/tracing/tracing.go`.
> * **Python** currently validates and passes `init` payloads via `ActionRunContext.init` in `py/packages/genkit/src/genkit/_core/_action.py`, but `SpanMetadata` lacks an `init` field and does not attach it to exported spans.
> 
> ## Required Changes
> ### 1. Update Core Tracing (`_tracing.py`)
> Add `init` to `SpanMetadata` in `py/packages/genkit/src/genkit/_core/_tracing.py` and serialize it as `'genkit:init'` inside `run_in_new_span`.
> 
> ```diff
> --- a/py/packages/genkit/src/genkit/_core/_tracing.py
> +++ b/py/packages/genkit/src/genkit/_core/_tracing.py
> @@ -62,6 +62,7 @@ class SpanMetadata(GenkitModel):
>      state: Literal['success', 'error'] | None = None
>      input: Any | None = Field(default=None)
>      output: Any | None = Field(default=None)
> +    init: Any | None = Field(default=None)
>      is_root: bool | None = None
>      metadata: dict[str, Any] | None = None
>      path: str | None = None
> @@ -175,6 +176,8 @@ def run_in_new_span(
>                  span.set_attribute('genkit:metadata:subtype', metadata.subtype)
>              if metadata.input is not None:
>                  span.set_attribute('genkit:input', _to_json_attr(metadata.input))
> +            if metadata.init is not None:
> +                span.set_attribute('genkit:init', _to_json_attr(metadata.init))
>              if metadata.metadata:
>                  for meta_key, meta_value in metadata.metadata.items():
>                      span.set_attribute(f'genkit:metadata:{meta_key}', str(meta_value))
> ```
> 
> ### 2. Pass `ctx.init` in Action Telemetry (`_action.py`)
> In `_run_with_telemetry` (`py/packages/genkit/src/genkit/_core/_action.py`), pass `ctx.init` when constructing `SpanMetadata`.
> 
> ```diff
> --- a/py/packages/genkit/src/genkit/_core/_action.py
> +++ b/py/packages/genkit/src/genkit/_core/_action.py
> @@ -746,6 +746,7 @@ class Action(Generic[InputT, OutputT, ChunkT, InitT]):
>              type='action',
>              subtype=str(self._kind),
>              input=input,
> +            init=ctx.init,
>              metadata=extra_metadata or None,
>              telemetry_labels={k: str(v) for k, v in (telemetry_labels or {}).items()} or None,
>          )
> ```
> 
> ### 3. Add Unit Test (`run_in_new_span_test.py`)
> Add a test in `py/packages/genkit/tests/genkit/core/run_in_new_span_test.py` verifying that `init` metadata is correctly serialized to the `'genkit:init'` span attribute.
> 
> ```diff
> --- a/py/packages/genkit/tests/genkit/core/run_in_new_span_test.py
> +++ b/py/packages/genkit/tests/genkit/core/run_in_new_span_test.py
> @@ -83,6 +83,15 @@ def test_writes_input_from_metadata(exporter: InMemorySpanExporter) -> None:
>      assert attrs['genkit:metadata:subtype'] == 'tool'
>  
>  
> +def test_writes_init_from_metadata(exporter: InMemorySpanExporter) -> None:
> +    with run_in_new_span(SpanMetadata(name='agentRun', type='action', init={'sessionId': 'session-123'})):
> +        pass
> +
> +    span = _by_name(exporter.get_finished_spans(), 'agentRun')
> +    attrs = dict(span.attributes or {})
> +    assert attrs['genkit:init'] == '{"sessionId": "session-123"}'
> +
> +
>  def test_writes_output_from_metadata_on_success(exporter: InMemorySpanExporter) -> None:
>      meta = SpanMetadata(name='answer', type='util')
>      with run_in_new_span(meta):
> ```

Fixed and test:
Screenshot 2026-07-27 at 4 53 47 PM
Screenshot 2026-07-27 at 4 55 16 PM

pavelgj (approved):

PR #5573, opened 2026-06-17
fix(core): bind reflection server to loopback only

evilgensec:
## Summary

The JS reflection server (`js/core/src/reflection.ts`) starts with `server.listen(this.port, ...)` and no host argument, so Node binds it to all interfaces (`0.0.0.0`/`::`) even though it logs that it is "running on http://localhost" and it is an unauthenticated local development API (it exposes `/api/runAction`, `/api/notify`, and other endpoints).

The Go (`go/genkit/reflection.go` -> `127.0.0.1`) and Python (`py/.../_reflection.py` -> `host='localhost'`) runtimes already bind loopback. This change aligns the JS runtime with them.

## Change

Pass `'localhost'` as the host to `server.listen(...)` so the reflection dev server is reachable only from the local machine. This matches the existing log message and the Go/Python behavior, and is transparent to the local `genkit` CLI tooling which connects via `http://localhost:${port}`.

### DIFF
--- a/js/core/src/reflection.ts
+++ b/js/core/src/reflection.ts
@@ -436,7 +436,12 @@ export class ReflectionServer {
     });
 
     this.port = await this.findPort();
-    this.server = server.listen(this.port, async () => {
+    // Bind to loopback only. The reflection server is a local development API
+    // with no authentication, so it must not be exposed on all interfaces.
+    // Use the explicit IPv4 loopback address (rather than 'localhost', which on
+    // Node 17+ can resolve to the IPv6 '::1' only) so the bind is deterministic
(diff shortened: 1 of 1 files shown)


evilgensec:
Thanks — applied. Switched the bind from `'localhost'` to the explicit IPv4 loopback `'127.0.0.1'` so the listen is deterministic across Node versions (avoiding the Node 17+ case where `localhost` can bind `::1` only) and aligns with the Go runtime, which binds `127.0.0.1`.

evilgensec:
Gentle bump. The bind is switched to the explicit 127.0.0.1 loopback per the review note, which matches the Go runtime. Could a maintainer take a look when there is a chance? Happy to rebase if the branch has drifted.

PR #5582, opened 2026-06-18
rfc(py): ModelRef and enabling dict-based config with autocomplete and type hints

huangjeff5:


py/docs/rfcs/typed-model-config.md
```diff
@@ -0,0 +1,312 @@
...
+When the model is known, `config={...}` should autocomplete its keys and flag typos —
+without forcing the developer to build a Pydantic object, and without changing the
+runtime passthrough that already works:
+
+```python
+flash = GoogleAI.model('gemini-2.0-flash')           # a typed, inert handle
+await ai.generate(
+    model=flash,
+    prompt='...',
+    config={'temperature': 0.7, 'code_execution': True},  # ✅ autocomplete + typo check
```
huangjeff5: Note that when a ModelRef is passed, config must be a TypedDict. Cannot be a Pydantic model. That's intentional. If you want the flexibility, don't pass in a ModelRef, just use the string.

PR #5587, opened 2026-06-19
fix: bind local dev servers to loopback and validate Host header

### LINKED ISSUE #5581: [security] Genkit local dev servers bind all interfaces with no Host-header validation, enabling DNS-rebinding execution of arbitrary actions and disclosure of flow traces (default `genkit start` flow)
https://github.com/genkit-ai/genkit/security/advisories/GHSA-273q-hccx-23r6 and
as advised by https://issuetracker.google.com/issues/525169408 to create issue:
```
Unfortunately, according to the [OSS VRP rules](https://bughunters.google.com/about/rules/open-source/google-open-source-software-vulnerability-reward-program-rules), product vulnerabilities and other security issues in OT2 and OT3 tiered repositories are not eligible for monetary rewards.

We encourage you to open an issue or submit a pull request directly to the Genkit GitHub repository (https://github.com/genkit-ai/genkit) to address these findings.
```

### Summary
When a developer runs Genkit locally (`genkit start`, or an app in `GENKIT_ENV=dev`), the
local HTTP servers it stands up - the Dev UI/tools API server, the telemetry server, and
(in standalone dev mode) the runtime reflection server - all bind to every network
interface (`0.0.0.0`/`::`) rather than loopback, and none of them validate the `Host`
request header. Their CORS policy (`origin: /^http:\/\/localhost:\d+$/`) only controls
which origins may read responses; it does not block requests, and it is bypassed by DNS
rebinding (where the request is same-origin to the attacker's page). As a result:
- Any host on the developer's network, and any website the developer visits (via DNS
  rebinding), can reach `POST /api/runAction` on the Dev UI server (default port 4000) and
  execute any registered Genkit action - flow, tool, model, prompt - with attacker-supplied
  `input` and `context`, then read the result.
- The telemetry server's SSE endpoint `GET /api/traces/:traceId/stream` (default port 4033)
  sets `Access-Control-Allow-Origin: *` unconditionally, so any web origin that knows a
  `traceId` can read the full trace, including each action's input and output.
- These chain: trigger an action to obtain its `traceId` (returned in the
  `X-Genkit-Trace-Id` response header), then read its trace to exfiltrate flow inputs and
  outputs (model prompts, user data, secrets passing through flows).

### Details
All three servers `listen(port)` with no host argument, so Node binds all interfaces, and
none check `Host`:

Dev UI / tools API server (`genkit-tools/common/src/server/server.ts`):
```ts
app.use(cors({ origin: /^http:\/\/localhost:\d+$/, allowedHeaders: ['Content-Type'] })); // :93
app.post('/api/runAction', bodyParser.json(...), async (req, res) => {                    // :113
  ... const result = await manager.runAction(req.body, ...);                              // :137 executes the action
});
server = app.listen(port, async () => { ... });                                           // :323 no host -> 0.0.0.0
```
`cors()` is a header control, not an access gate: for a request whose `Origin` does not
match (or is absent, as in a DNS-rebound same-origin request) it simply omits
`Access-Control-Allow-Origin` and still runs the handler. There is no `Host` check, so a
request with `Host: attacker.evil.com` is processed - the DNS-rebinding precondition.

Telemetry server (`genkit-tools/telemetry-server/src/index.ts`):
```ts
api.use(cors({ origin: params.corsOrigin || /^http:\/\/localhost:\d+$/, ... }));          // :60
api.get('/api/traces/:traceId/stream', async (request, response, next) => {              // :85
  response.setHeader('Access-Control-Allow-Origin', '*');                                 // :93 unconditional wildcard
  const currentTrace = await params.traceStore.load(traceId);
  response.write(`data: ${JSON.stringify(currentTrace)}\n\n`);                            // streams trace input/output
  ...
});
server = api.listen(params.port, ...);                                                    // :315 no host -> 0.0.0.0
```
The SSE handler overrides the per-app localhost CORS with a hardcoded `*`, so any origin
may read traces (given a `traceId`).

Runtime reflection server (`js/core/src/reflection.ts`), used in standalone dev mode when
`GENKIT_REFLECTION_V2_SERVER` is not set (the `genkit start` CLI sets it -> safe v2
outbound WebSocket): `express()` with no cors/auth/Host check, `POST /api/runAction`
(`:240`) executes any action by `key` with attacker `input`/`context`, and
`server.listen(this.port, cb)` (`:439`) binds all interfaces. It runs when `isDevEnv()`
(`js/genkit/src/genkit.ts:191`).

The attacker-controlled `context` is passed verbatim to `action.run`; flows authorize on
`context.auth`, so a forged `context.auth` impersonates an authenticated principal.

### PoC
Validated live on the default flow: `npx genkit-cli@1.36.0 start --non-interactive -- node
index.mjs` (Dev UI on :4000, telemetry on :4033), with a sample flow:
```js
import { genkit } from 'genkit'; import fs from 'fs';
const ai = genkit({});
ai.defineFlow('pwn', async (input) => { fs.writeFileSync('/tmp/GENKIT_DEVUI_'+input,'x'); return 'executed:'+input; });
```

1) Servers bind all interfaces (LAN-reachable), not loopback:
```
ss -tlnp | grep -E '4000|4033'  ->  *:4000 ... *:4033       (0.0.0.0)
curl http://192.168.0.6:4000/api/__health  ->  HTTP 200      (machine's LAN IP)
```

2) Execute an action with a spoofed Host (DNS-rebinding precondition) - no auth:
```
curl -s -X POST http://127.0.0.1:4000/api/runAction -H 'Content-Type: application/json' \
  -H 'Host: attacker.evil.com' -d '{"key":"/flow/pwn","input":"DEVUI789"}'
-> 200, X-Genkit-Trace-Id: 8d172ff2...e5d0
   {"result":"executed:DEVUI789","telemetry":{"traceId":"8d172ff2...e5d0"},...}
```
The spoofed Host was accepted and the flow ran (`/tmp/GENKIT_DEVUI_DEVUI789` created).

3) Read the trace cross-origin via the wildcard-CORS SSE endpoint:
```
curl -s -i -H 'Origin: https://evil.example.com' \
  http://127.0.0.1:4033/api/traces/8d172ff2...e5d0/stream
-> 200; Access-Control-Allow-Origin: *
   data: {"traceId":"8d172ff2...","spans":{...,"genkit:input":"\"DEVUI789\"",
          "genkit:output":"\"executed:DEVUI789\"","genkit:key":"/flow/pwn",...}}
```
Any web origin reads the flow input and output. Combined: a malicious page DNS-rebinds to
the developer's machine, POSTs `/api/runAction` (port 4000) to run a flow, takes the
`X-Genkit-Trace-Id`, then reads `/api/traces/<id>/stream` (port 4033, `ACAO:*`) to
exfiltrate the inputs/outputs of that and other actions.

### Impact
For any developer running Genkit locally (the standard `genkit start` workflow, default
ports 4000/4033), a host on the same network - or a remote website via DNS rebinding -
can, unauthenticated: execute arbitrary registered actions (flows/tools/models) with
attacker input and a forged `context.auth`, and read flow traces containing inputs and
outputs (model prompts, user/PII data, and secrets that pass through flows). Executing
actions runs the developer's flow/tool code (which commonly performs outbound HTTP, file,
and database operations and consumes paid LLM quota). The standalone-dev reflection server
extends the same execution primitive to apps run in `GENKIT_ENV=dev` without the CLI. Root
causes: binding to all interfaces instead of loopback, no `Host` validation, and an
unconditional wildcard CORS on the trace SSE endpoint.

Suggested remediation: bind all dev servers to `127.0.0.1` only; reject requests whose
`Host` header is not `localhost`/`127.0.0.1[:port]` (this defeats DNS rebinding); remove
the hardcoded `Access-Control-Allow-Origin: *` on the trace SSE endpoint (honor the same
localhost CORS as the rest of the API); and require a per-run secret/token - written to the
runtime info file the trusted CLI already reads - on `/api/runAction` and the trace
endpoints.

serhiizghama:
Fixes #5581 (advisory GHSA-273q-hccx-23r6).

## Problem
When Genkit runs locally (`genkit start`, or an app with `GENKIT_ENV=dev`), the local HTTP servers it stands up bind to every network interface (`0.0.0.0`/`::`) and none validate the `Host` header:

- **Dev UI / tools API server** (`genkit-tools/common/src/server/server.ts`) — `POST /api/runAction` executes any registered action.
- **Runtime reflection server** (`js/core/src/reflection.ts`, standalone dev mode) — same `runAction` execution primitive.
- **Telemetry server** (`genkit-tools/telemetry-server/src/index.ts`) — the trace SSE endpoint sets a hardcoded `Access-Control-Allow-Origin: *`.

The CORS origin check is not an access gate — the handler still runs, and it is bypassed by DNS rebinding (the request is same-origin to the attacker's page). As a result, any host on the developer's network, or any website the developer visits, can execute arbitrary actions (with attacker-supplied `input` and a forged `context.auth`) and exfiltrate flow traces (model prompts, user data, secrets). See the issue for a full PoC.

## Solution
Applies the remediation outlined in the issue across all three dev servers:

- **Bind to the loopback interface** (`localhost`) instead of all interfaces, so the servers are not reachable from other hosts on the network.
- **Reject requests whose `Host` header is not loopback** (`localhost` / `127.0.0.1` / `[::1]`, optional port, plus RFC 6761 `*.localhost`). A browser cannot forge the `Host` header, so this defeats DNS rebinding — which the CORS check alone does not.
- **Remove the unconditional `Access-Control-Allow-Origin: *`** on the telemetry SSE endpoint; the shared `cors()` middleware (localhost origins only) now governs it.
- **`GENKIT_DEV_SERVER_HOST`** opts into exposing the servers on another interface (e.g. a remote dev container); when set, the server binds to that value and the loopback `Host` check is skipped.

The legitimate `localhost` workflow is unchanged. The shared helper lives in `@genkit-ai/tools-common/utils` (reused by the tools and telemetry servers); the reflection server in `js/core` carries a small self-contained copy to avoid a cross-package dependency.

The per-request secret token suggested as additional defense-in-depth in the issue is a larger change (requires coordination with the CLI runtime file) and is left as a follow-up.

## Why this approach
Binding to loopback closes direct access from other hosts; `Host`-header validation closes DNS rebinding (the severe, remotely-triggerable vector); removing the wildcard SSE CORS closes cross-origin trace reads. Together they cover the documented attack chain with no change to the normal local developer experience. Binding to `localhost` (rather than `127.0.0.1`) keeps client/server address-family resolution in sync so existing `http://localhost:<port>` clients keep working on IPv6-first hosts.

## Testing
- `genkit-tools/common`: `jest` — new unit tests for the host helper (loopback detection, default bind host, env override) and an integration test asserting `/api/runAction` rejects a non-loopback `Host` with 403.
- `js/core`: `node --test tests/reflection_test.ts` — new tests asserting the reflection server rejects a non-loopback `Host` (403) and accepts loopback requests; existing reflection + reflection-v2 suites pass.
- `tsc` typecheck and `prettier --check` clean on all changed files.


serhiizghama:
Rebased on main. Conflicts were just my Host-validation tests landing next to the new analytics-events suite and the NOT_FOUND reflection test — kept both, plus the logger import move into the utils barrel. server_test 15/15, reflection_test green, dev_server_security 7/7.

serhiizghama:
Closing — loopback binding landed in #5751, so the remaining part of this isn't needed.

Issue #5542, opened 2026-06-15
feat: Python P0 model parity june 2026


cabljac:
## Goal

> **Progress (4 of 5 slices merged):** #5559, #5588, #5579, #5592 landed. Only the embedder slice (`gemini-embedding-2`) remains.
>
> Supersedes #5495 (same Gemini + embedder rows, without Anthropic).

Bring the Python plugins up to P0 model parity for the June 2026 set. We want
proper first-party support (explicit registration with correct capability flags)
for the models below, not just "it resolves if you pass the string".

| Model Family | API | ModelName |
|---|---|---|
| GoogleAI | Embedders | gemini-embedding-2 |
| GoogleAI | Gemini | gemini-3-pro-image |
| GoogleAI | Gemini | gemini-3.1-flash-image |
| GoogleAI | Gemini | gemini-3.5-flash |
| GoogleAI | Gemini | gemini-3.1-flash-lite-preview |
| GoogleAI | Gemini | gemini-3.1-pro-preview-customtools |
| GoogleAI | Gemini | gemini-3.1-pro-preview |
| GoogleAI | Gemini Image | gemini-3.1-flash-image-preview |
| VertexAI | Gemini | gemini-3.5-flash |
| VertexAI | Gemini | gemini-3.1-flash-lite |
| VertexAI | Gemini | gemini-3.1-pro-preview |
| VertexAI | Gemini Image | gemini-3.1-flash-image |
| VertexAI | Gemini Image | gemini-3-pro-image |
| Anthropic | Claude | claude-opus-4-8 |
| Anthropic | Claude | claude-opus-4-7 |

## How support works here (context for the matrix)

GoogleAI/VertexAI discover models dynamically from the live API in
`list_actions`, so an unregistered model still resolves by string. Its
capabilities come from `google_model_info(name)`, which returns the hand-written
`ModelInfo` from `SUPPORTED_MODELS` if registered, otherwise a guessed fallback
(`GENERIC_IMAGE_MODEL` when the name contains `-image`, else a generic text
default). `SUPPORTED_MODELS` is one shared dict across both APIs.

Anthropic is static: `SUPPORTED_ANTHROPIC_MODELS` is the registry, and unknown
names fall back to `output=['text']` only, with no json/constrained.

So "fallback-only" below means it'll likely run, but Genkit advertises
guessed/degraded capabilities, and things like tools, JSON/constrained output, or
image config can be silently wrong.

## Current state (verified against source)

[...]

PR #5588, opened 2026-06-19
feat(py/plugins/google-genai): register VertexAI Gemini 3.1 text models

hilariie:
## What

Registers the VertexAI Gemini 3.1 text models with explicit `ModelInfo` so they
resolve to real capability metadata instead of the generic text fallback, and
adds `gemini-3.5-flash` to the Vertex enum:

- `gemini-3.1-pro-preview`
- `gemini-3.1-flash-lite`
- `gemini-3.5-flash` (added to `VertexAIGeminiVersion` — it already resolved via
  the shared `SUPPORTED_MODELS` dict but was missing from the Vertex enum/docs)

Part of #5542 (Python model-parity work). VertexAI counterpart to #5559.

## Why

Without registration these names fall through to `DEFAULT_SUPPORTS_MODEL`, which
leaves `output` unset and doesn't reflect each model's real capabilities.
Registering them gives the correct `Supports` metadata (tools, tool-choice,
constrained generation, `text`/`json` output) that the framework and Dev UI
rely on. Adding `gemini-3.5-flash` to `VertexAIGeminiVersion` closes a
discoverability/doc gap on the Vertex side.

## How

Touch points in
`py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py`,
following the existing 3.x wiring:

1. `ModelInfo` definition for `gemini-3.1-flash-lite`.
2. Enum members on `VertexAIGeminiVersion` (`gemini-3.5-flash`,
   `gemini-3.1-pro-preview`, `gemini-3.1-flash-lite`) + docstring table rows.
3. `_add_model(...)` registration for `gemini-3.1-flash-lite`.

Capabilities were verified against the JS and Go registries rather than copied
from generic defaults — all use the full multimodal/tool profile (`tools=True`,
`tool_choice=True`, `constrained=ALL`, `output=['text', 'json']`).

## Testing

- **Unit tests:** added a parametrized test asserting each VertexAI Gemini 3.x
  text model resolves to real capabilities (`tools`, `tool_choice`,
  `constrained=all`, `output=['text', 'json']`) rather than the fallback.
- **Dev UI (VertexAI):** screenshots below show a successful generate against `gemini-3.1-pro-preview`, `gemini-3.1-flash-lite`, and `gemini-3.5-flash`, including a structured/JSON run.

### Dev UI screenshots

Screenshot 2026-06-19 at 10 11 59
Screenshot 2026-06-19 at 10 18 59
Screenshot 2026-06-19 at 10 28 35
Screenshot 2026-06-19 at 12 07 43

### DIFF
--- a/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
+++ b/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
@@ -817,6 +817,8 @@ GEMINI_3_1_FLASH_LITE = ModelInfo(
         constrained=Constrained.ALL,
         output=['text', 'json'],
     ),
+)
+
 GEMINI_IMAGE_SUPPORTS = Supports(
     multiturn=True,
     media=True,
--- a/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
+++ b/py/plugins/google-genai/src/genkit/plugins/google_genai/models/gemini.py
@@ -806,6 +806,19 @@ GEMINI_3_1_FLASH_LITE_PREVIEW = ModelInfo(
     ),
 )
 
+GEMINI_3_1_FLASH_LITE = ModelInfo(
+    label='Google AI - Gemini 3.1 Flash Lite',
+    supports=Supports(
+        multiturn=True,
+        media=True,
+        tools=True,
+        tool_choice=True,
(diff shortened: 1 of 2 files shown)


cabljac (approved):
lgtm

PR #5595, opened 2026-06-22
chore(py/samples): drop redundant uvloop dependency

hilariie:
## What

Removes the bare `uvloop>=0.21.0` dependency from 8 Python samples:
`context`, `gemini-code-execution`, `gemini-context-caching`,
`middleware`, `middleware-coding-agent`, `output-formats`, `prompts`,
and `tool-interrupts`.

## Why

These samples declared `uvloop>=0.21.0` with no platform marker. uvloop
ships no Windows wheels, so the bare pin makes the samples fail to
install on Windows.

The dependency is also redundant: the core `genkit` package already
declares `uvloop>=0.21.0; sys_platform != 'win32'`, and `run_loop`
imports it optionally with an asyncio fallback. None of these samples
import uvloop directly.

## How

Deleted the `uvloop` line from each sample's `pyproject.toml`. On
non-Windows platforms uvloop is still installed transitively via the
`genkit` dependency, so the optional speedup in `run_loop` is unaffected.
On Windows, the install is no longer blocked and `run_loop` falls back to
asyncio as intended.

## Testing

- `grep` confirms no remaining `uvloop` references in any sample
  `pyproject.toml` and no direct `uvloop` imports in sample `src/`.
- Regenerated `py/uv.lock`; `uv lock --check` passes.
- Confirmed uvloop still resolves transitively via `genkit` core
  (`uv pip show uvloop` → `Required-by: genkit`) and imports in a
  synced sample env, so the `run_loop` speedup is unaffected on
  non-Windows.

### DIFF
--- a/py/samples/context/pyproject.toml
+++ b/py/samples/context/pyproject.toml
@@ -7,7 +7,6 @@ dependencies = [
   "genkit-plugin-google-genai",
   "pydantic>=2.0.0",
   "structlog>=24.0.0",
-  "uvloop>=0.21.0",
 ]
 
 [build-system]
--- a/py/samples/gemini-code-execution/pyproject.toml
+++ b/py/samples/gemini-code-execution/pyproject.toml
@@ -7,7 +7,6 @@ dependencies = [
   "genkit-plugin-google-genai",
   "pydantic>=2.10.5",
   "structlog>=25.2.0",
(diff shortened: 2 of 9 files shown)


cabljac (approved):
lgtm!

PR #5596, opened 2026-06-22
feat(py/plugins/google-genai): register Gemini embedding 2

hilariie:
## What

Registers the exact gemini-embedding-2 model name for the Python Google GenAI plugin and carries its embedder metadata through action registration.

- **Dimensions:** 3072
- **Supported input:** `text`, `image`, `video`

Also introduces a `get_embedder_options(name, label)` helper in
`models/embedder.py` that centralises building `EmbedderOptions`
(label, supported input, dimensions) for discovered embedders, and
wires it into the GoogleAI plugin, the VertexAI plugin, and
`_create_embedder_action`.

A slice of #5542 

## Why

`gemini-embedding-2` was not registered, so it carried no dimension
or input-type metadata. Separately, embedder metadata was built inline
in three places with `supports=EmbedderSupports(input=['text'])`
hardcoded, which means that multimodal embedders were always advertised as
text-only. Centralising the lookup fixes that and removes the duplication.

## How

- Add `gemini-embedding-2` to GeminiEmbeddingModels.
- Add the `EMBEDDER_DIMENSIONS` (3072) and a new
  `EMBEDDER_INPUT_SUPPORTS` map (`text`/`image`/`video`).
- Add `get_embedder_options()` which reads both maps (defaulting input
  to `['text']` for models not listed) and returns the `EmbedderOptions`.
- Replace the three inline `EmbedderOptions(...)` constructions with
  calls to the helper.
- Add tests for embedder dimensions, supported inputs, and discovery metadata.

## Validation
- Dev UI tested locally.
- `uv run --directory plugins/google-genai python -m pytest test/` — 242 passed

## Screenshots
image
image
image
Screenshot 2026-06-29 at 14 33 39
Screenshot 2026-06-29 at 14 34 12

### DIFF
--- a/py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
+++ b/py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
@@ -29,6 +29,7 @@ from google.genai import types as genai_types
 
 from genkit import DocumentPart, Embedding, EmbedRequest, EmbedResponse
 from genkit._core._typing import DocumentData
+from genkit.embedder import EmbedderOptions, EmbedderSupports
 from genkit.plugins.google_genai.models.utils import PartConverter
 
 
@@ -47,6 +48,8 @@ class VertexEmbeddingModels(StrEnum):
 class GeminiEmbeddingModels(StrEnum):
     """Embedding models supported by Google-Genai gemini."""
 
+    GEMINI_EMBEDDING_2_PREVIEW = 'gemini-embedding-2-preview'
+    GEMINI_EMBEDDING_2 = 'gemini-embedding-2'
     GEMINI_EMBEDDING_EXP_03_07 = 'gemini-embedding-exp-03-07'
     TEXT_EMBEDDING_004 = 'text-embedding-004'
     GEMINI_EMBEDDING_001 = 'gemini-embedding-001'
@@ -68,6 +71,7 @@ class EmbeddingTaskType(StrEnum):
 EMBEDDER_DIMENSIONS: dict[str, int] = {
     # Google AI
     'gemini-embedding-2-preview': 3072,
+    'gemini-embedding-2': 3072,
     'gemini-embedding-001': 3072,
     # Vertex AI
     'text-embedding-005': 768,
@@ -81,6 +85,21 @@ VERTEX_KNOWN_EMBEDDERS: tuple[str, ...] = (
     'text-multilingual-embedding-002',
     'gemini-embedding-001',
 )
+
+EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
+    'gemini-embedding-2-preview': ['text', 'image', 'video'],
+    'gemini-embedding-2': ['text', 'image', 'video'],
+}
+
+
+def get_embedder_options(name: str, label: str) -> EmbedderOptions:
+    """Return EmbedderOptions metadata for a discovered embedder model."""
+    supports = EMBEDDER_INPUT_SUPPORTS.get(name, ['text'])
+    return EmbedderOptions(
+        label=label,
+        supports=EmbedderSupports(input=supports),
+        dimensions=EMBEDDER_DIMENSIONS.get(name),
+    )
 
 
 class Embedder:
--- a/py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
+++ b/py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
@@ -86,15 +86,27 @@ VERTEX_KNOWN_EMBEDDERS: tuple[str, ...] = (
     'gemini-embedding-001',
 )
 
-EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
+# Advertised input modalities per backend
+GOOGLEAI_EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
     'gemini-embedding-2-preview': ['text', 'image', 'video'],
     'gemini-embedding-2': ['text', 'image', 'video'],
 }
 
 
-def get_embedder_options(name: str, label: str) -> EmbedderOptions:
-    """Return EmbedderOptions metadata for a discovered embedder model."""
-    supports = EMBEDDER_INPUT_SUPPORTS.get(name, ['text'])
+def get_embedder_options(name: str, label: str, is_vertex: bool = False) -> EmbedderOptions:
+    """Return EmbedderOptions metadata for a discovered embedder model.
+
+    Args:
+        name: The bare (unprefixed) model name, e.g. 'gemini-embedding-2'.
+        label: Human-readable label for the embedder.
+        is_vertex: True when resolving for the Vertex backend, which advertises
+            text-only because its embed path cannot deliver media.
+
+    Returns:
+        EmbedderOptions describing the model's label, supported inputs and
+        static dimensions.
+    """
+    supports = ['text'] if is_vertex else GOOGLEAI_EMBEDDER_INPUT_SUPPORTS.get(name, ['text'])
     return EmbedderOptions(
         label=label,
         supports=EmbedderSupports(input=supports),
--- a/py/plugins/google-genai/test/google_plugin_test.py
+++ b/py/plugins/google-genai/test/google_plugin_test.py
@@ -40,7 +40,12 @@ from genkit import (
 )
 from genkit.plugin_api import GENKIT_CLIENT_HEADER
 from genkit.plugins.google_genai import GoogleAI, VertexAI
-from genkit.plugins.google_genai.google import _inject_attribution_headers, googleai_name, vertexai_name
+from genkit.plugins.google_genai.google import (
+    _inject_attribution_headers,
+    _list_genai_models,
+    googleai_name,
+    vertexai_name,
+)
 from genkit.plugins.google_genai.models.embedder import VERTEX_KNOWN_EMBEDDERS
 from genkit.plugins.google_genai.models.gemini import (
     DEFAULT_SUPPORTS_MODEL,
@@ -236,16 +241,17 @@ def test_googleai__resolve_model(
 
 
 @pytest.mark.parametrize(
-    'model_name, expected_model_name, clean_name',
+    'expected_model_name, expected_dimensions, expected_support_inputs',
     [
-        ('gemini-pro-deluxe-max', 'googleai/gemini-pro-deluxe-max', 'gemini-pro-deluxe-max'),
-        ('googleai/gemini-pro-deluxe-max', 'googleai/gemini-pro-deluxe-max', 'gemini-pro-deluxe-max'),
+        ('googleai/gemini-embedding-2', 3072, ['text', 'image', 'video']),
+        ('googleai/gemini-embedding-2-preview', 3072, ['text', 'image', 'video']),
+        ('googleai/custom-embedder', None, ['text']),
     ],
 )
 def test_googleai__resolve_embedder(
-    model_name: str,
     expected_model_name: str,
-    clean_name: str,
+    expected_dimensions: int | None,
+    expected_support_inputs: list[str],
     googleai_plugin_instance: GoogleAI,
 ) -> None:
     """Tests for GoogleAI._resolve_embedder method."""
@@ -256,6 +262,8 @@ def test_googleai__resolve_embedder(
     assert action is not None
     assert action.kind == ActionKind.EMBEDDER
     assert action.name == expected_model_name
+    assert action.metadata['embedder']['dimensions'] == expected_dimensions
+    assert action.metadata['embedder']['supports']['input'] == expected_support_inputs
 
 
 @pytest.mark.asyncio
@@ -270,6 +278,8 @@ async def test_googleai_list_actions(googleai_plugin_instance: GoogleAI) -> None
 
     models_return_value = [
         MockModel(supported_actions=['generateContent'], name='models/gemini-pro'),
+        MockModel(supported_actions=['embedContent'], name='models/gemini-embedding-2'),
+        MockModel(supported_actions=['embedContent'], name='models/gemini-embedding-2-preview'),
         MockModel(supported_actions=['embedContent'], name='models/gemini-embedding-001'),
         MockModel(supported_actions=['generateContent'], name='models/gemini-2.0-flash-tts'),  # TTS
     ]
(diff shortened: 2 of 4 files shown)


py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -68,13 +71,29 @@ class EmbeddingTaskType(StrEnum):
...
     # Vertex AI
     'text-embedding-005': 768,
     'text-multilingual-embedding-002': 768,
     'multimodalembedding@001': 768,
 }
 
+EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
+    'gemini-embedding-2-preview': ['text', 'image', 'video'],
+    'gemini-embedding-2': ['text', 'image', 'video'],
+}
```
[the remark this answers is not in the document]
hilariie: Resolved in #5649 . multimodelembedding@001 also accepts video as input, so it was included as well (https://docs.cloud.google.com/python/docs/reference/vertexai/latest, https://docs.cloud.google.com/vertex-ai/generative-ai/docs/samples/generativeaionvertexai-multimodal-embedding-image-video-text)

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -68,13 +71,29 @@ class EmbeddingTaskType(StrEnum):
 EMBEDDER_DIMENSIONS: dict[str, int] = {
     # Google AI
     'gemini-embedding-2-preview': 3072,
+    'gemini-embedding-2': 3072,
     'gemini-embedding-001': 3072,
     # Vertex AI
     'text-embedding-005': 768,
     'text-multilingual-embedding-002': 768,
     'multimodalembedding@001': 768,
 }
```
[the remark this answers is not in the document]
hilariie: Resolved in #5649

cabljac:
Code and tests look good. One thing before I approve: the screenshots show the model registered in the Dev UI, but that's the metadata, not proof the multimodal path works. Can you make sure to run an embed with both image and video input, confirm you get a vector back, and drop in a screenshot? Otherwise the `input=['text', 'image', 'video']` metadata could be advertising modalities the API rejects, which would mislead users.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -82,6 +86,21 @@ class EmbeddingTaskType(StrEnum):
     'gemini-embedding-001',
 )
 
+EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
```
cabljac: One thing I wanted to check: this map isn't backend-scoped, and I think `VertexAI.resolve()` will build an embedder for any name, so `vertexai/gemini-embedding-2` would also advertise `['text','image','video']`. From what I can see in the SDK, the Vertex embed path (`t_contents_for_embed`) strips every non-text part with just a log warning, so an image+text batch would come back with fewer embeddings than documents and no error, which quietly misaligns doc-to-vector in a store. JS seems to deliberately not register this model on Vertex. Would it make sense to scope the supports per backend, or validate request parts against the advertised supports in `generate()` so the metadata and behavior share one gate?

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -82,6 +86,21 @@ class EmbeddingTaskType(StrEnum):
...
+    'gemini-embedding-2-preview': ['text', 'image', 'video'],
+    'gemini-embedding-2': ['text', 'image', 'video'],
+}
```
cabljac, line 93: Should `multimodalembedding@001` be in this map too? It's in `EMBEDDER_DIMENSIONS` and the JS plugin advertises it as text/image/video, but with the supports lookup now centralized it falls back to `['text']`, so the one Vertex embedder actually built for media gets advertised as text-only. I know the follow-up PR adds it, but as this slice stands alone the map reads inverted (gemini-embedding-2 over-advertises on Vertex, multimodalembedding under-advertises). Worth pulling that entry into this PR?

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -82,6 +86,21 @@ class EmbeddingTaskType(StrEnum):
...
+
+
+def get_embedder_options(name: str, label: str) -> EmbedderOptions:
```
cabljac: The client path for media on GoogleAI looks right to me (`_build_contents` -> `PartConverter` -> inline_data, and the non-Vertex transformer forwards it), but I couldn't find a test that drives `Embedder.generate` with an image part, and I haven't seen evidence `embedContent` accepts media for gemini-embedding-2 server-side. The new tests pin the dict values rather than the behavior. Can you make sure to run an image embed against the real API and drop in a screenshot, and maybe add a generate()-level test asserting what contents reach the mocked client for a media doc?

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -82,6 +86,21 @@ class EmbeddingTaskType(StrEnum):
     'gemini-embedding-001',
 )
 
+EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
+    'gemini-embedding-2-preview': ['text', 'image', 'video'],
+    'gemini-embedding-2': ['text', 'image', 'video'],
```
cabljac, line 92: Small structural thought: we now have three hand-maintained name lists (the enum, `EMBEDDER_DIMENSIONS`, `EMBEDDER_INPUT_SUPPORTS`), and the multimodalembedding case above is what drift between them looks like. compat-oai keeps a single `SUPPORTED_EMBEDDING_MODELS` table with label/dimensions/supports per model, and gemini.py does the same with `SUPPORTED_MODELS`. Would one `KNOWN_EMBEDDERS: dict[name -> record]` here be worth it while the helper is fresh? Fine as a follow-up if you'd rather keep this slice small.

py/plugins/google-genai/test/google_plugin_test.py
```diff
@@ -236,16 +241,17 @@ def test_googleai__resolve_model(
...
+    'expected_model_name, expected_dimensions, expected_support_inputs',
     [
-        ('gemini-pro-deluxe-max', 'googleai/gemini-pro-deluxe-max', 'gemini-pro-deluxe-max'),
-        ('googleai/gemini-pro-deluxe-max', 'googleai/gemini-pro-deluxe-max', 'gemini-pro-deluxe-max'),
+        ('googleai/gemini-embedding-2', 3072, ['text', 'image', 'video']),
+        ('googleai/gemini-embedding-2-preview', 3072, ['text', 'image', 'video']),
+        ('googleai/custom-embedder', None, ['text']),
     ],
 )
 def test_googleai__resolve_embedder(
```
cabljac, line 257: I think this test ends up circular: it resolves `expected_model_name` and then asserts `action.name == expected_model_name`, so input and expectation are the same string and a name-normalization regression can't fail it. (Looks like the old version had the same quirk, so nothing regressed, but since the metadata assertions now depend on `clean_name` stripping the prefix before the dict lookups, could we add one case that resolves a bare unprefixed name and asserts the prefixed action name?)

cabljac (commented):
Nice cleanup of the three inline metadata blocks, and the registration itself looks right (dimensions and label semantics check out, gemini-embedding-001 matches JS exactly). Questions inline, mostly about the supports metadata vs what the runtime delivers - some may be resolved by the stacked follow-up, flagging them here since this slice ships first.

Two notes that don't anchor to a line:

- Cross-SDK drift: JS registers `googleai/gemini-embedding-2` with dimensions undefined (the `commonRef` info override replaces the 3072 default in `js/plugins/google-genai/src/googleai/embedder.ts`), while this PR says 3072. Python looks right to me, so I think that's a JS issue worth filing (we can bundle it with the multimodalembedding 768/`outputDimensionality` gaps we spotted reviewing #5649).
- Test nits, take or leave: `test_embedder_dimensions_known_models` asserts the dict literal against itself so it can't catch a behavioral regression, and the private `_list_genai_models` test seems fully covered by the public `list_actions` test updated in this same PR.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
hilariie: Scoped the supports per backend in `ef2cf835b`

Note that `multimodalembedding@001` is a genuine Vertex media embedder, so the blanket text-only default is deliberately conservative here; #5649 refines the Vertex path to a per-model lookup so it gets its `text/image/video` back.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
hilariie: Deferring this to #5649, which adds `multimodalembedding@001` with `text/image/video`. Worth noting that the "inverted map" concern is now moot for this slice: with the backend scoping in `ef2cf835b`, every Vertex embedder is advertised text-only regardless of the map, so `gemini-embedding-2` no longer over-advertises on Vertex and `multimodalembedding@001` isn't under-advertised relative to its siblings.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
hilariie: Added `test_embedding_forwards_media_parts` in `ef2cf835b`

For the real-API check: 
Method: embeddings here are deterministic, so embedding the same text twice gives a same-input baseline cosine (~1.0). I then embed that same text with an image attached. If the server consumes the image, the text+image vector drops measurably below the baseline; if it drops or ignores the image, it stays baseline-close (or bit-identical). Script generates its own test PNG so it's self-contained.

Result:
Screenshot 2026-07-08 at 14 31 03

The text+image cosine sits well below the same-input baseline of ~1.0, so embedContent is genuinely using the image. 

Reproduction script:
```python
"""Manual real-API proof for PR #5596 review comment 3530491617.

Question under test: does googleai/gemini-embedding-2 actually *consume* an
image on the server, or is the ['text','image','video'] supports entry a claim
the API silently ignores?

Method (the reviewer's "same-vector" check, made rigorous):
  1. Embed the SAME text twice -> establishes a baseline cosine for identical
     input (embeddings are ~deterministic, so this should be ~1.0). This tells
     us how close "no real difference" looks on this model.
  2. Embed that same text WITH an image attached.
  3. Compare. If the image is consumed, the text+image vector diverges from the
     text-only one noticeably *below* the baseline. If the server drops the
     image, the text+image vector is identical (or baseline-close) to text-only
     -> the multimodal advertisement is a lie.

Run it:
    cd py
    GEMINI_API_KEY=your-key uv run python /tmp/embed_media_proof.py
"""

import asyncio
import base64
import io
import math

from PIL import Image

from genkit import Document, DocumentPart, Genkit, Media, MediaPart, TextPart
from genkit.plugins.google_genai import GoogleAI

EMBEDDER = 'googleai/gemini-embedding-2'
TEXT = 'a red and blue gradient square'


def make_test_png() -> bytes:
    """Build a small but non-trivial PNG (a two-tone gradient) in memory."""
    w = h = 64
    img = Image.new('RGB', (w, h))
    px = img.load()
    for y in range(h):
        for x in range(w):
            px[x, y] = (int(255 * x / w), 0, int(255 * y / h))
    buf = io.BytesIO()
    img.save(buf, format='PNG')
    return buf.getvalue()


def cosine(a: list[float], b: list[float]) -> float:
    """Cosine similarity of two vectors (1.0 == same direction)."""
    dot = sum(x * y for x, y in zip(a, b))
    na = math.sqrt(sum(x * x for x in a))
    nb = math.sqrt(sum(y * y for y in b))
    return dot / (na * nb) if na and nb else float('nan')


async def main() -> None:
    ai = Genkit(plugins=[GoogleAI()])

    png = make_test_png()
    data_url = f'data:image/png;base64,{base64.b64encode(png).decode()}'

    image_doc = Document(
        content=[
            DocumentPart(root=TextPart(text=TEXT)),
            DocumentPart(root=MediaPart(media=Media(url=data_url, content_type='image/png'))),
        ]
    )

    # Step 1: same text twice -> baseline for "identical input".
    text_a = await ai.embed(embedder=EMBEDDER, content=Document.from_text(TEXT))
    text_b = await ai.embed(embedder=EMBEDDER, content=Document.from_text(TEXT))
    print(f'text-only:  {len(text_a)} vector(s), dim {len(text_a[0].embedding)}')

    # Step 2: same text + image.
    try:
        with_image = await ai.embed(embedder=EMBEDDER, content=image_doc)
    except Exception as e:  # noqa: BLE001
        print(f'\nAPI REJECTED the media part: {type(e).__name__}: {e}')
        print("VERDICT: FAIL -> flip both GoogleAI entries to ['text'] before merging, and say so on the thread.")
        return
    print(f'with-image: {len(with_image)} vector(s), dim {len(with_image[0].embedding)}')

    v_text, v_image = text_a[0].embedding, with_image[0].embedding
    baseline = cosine(text_a[0].embedding, text_b[0].embedding)  # text vs text (same input)
    image_sim = cosine(v_text, v_image)  # text vs text+image

    print(f'\nexact-identical (text vs text+image): {v_text == v_image}')
    print(f'baseline cosine (text vs text, same input): {baseline:.6f}')
    print(f'cosine (text-only vs text+image):           {image_sim:.6f}')

    # If the image moved the vector clearly below the same-input baseline, it was used.
    if v_text == v_image or image_sim >= baseline - 1e-5:
        print("\nVERDICT: FAIL -> image had no measurable effect (silently ignored or rejected).")
        print("         Flip both GoogleAI entries to ['text'] before merging, and note it on the thread.")
    else:
        print('\nVERDICT: PASS -> the image measurably changed the embedding.')
        print('         The multimodal advertisement is justified. Screenshot this output onto the PR thread.')


if __name__ == '__main__':
    asyncio.run(main())
```

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
hilariie: yeah, this is the cleaner shape, and #5649 is the natural place for it since it also has to turn the `is_vertex` boolean into a per-model Vertex lookup. Keeping this slice small and leaving the consolidation for that follow-up.

py/plugins/google-genai/test/google_plugin_test.py
hilariie: resolved in `ef2cf83`

hilariie:
> Cross-SDK drift: JS registers `googleai/gemini-embedding-2` with dimensions undefined...

Will open an issue

> `test_embedder_dimensions_known_models` asserts the dict literal against itself...

Behavioral coverage now lives in `test_googleai__resolve_embedder` / `test_googleai_list_actions` (ef2cf835b)

cabljac (approved):
lgtm

PR #5614, opened 2026-06-25
feat(go/plugins/firebase/exp): add Firestore session store

apascal07:
Adds `FirestoreSessionStore`, a Cloud Firestore-backed `exp.SessionStore` for the experimental Agent API (`ai/exp`), the Go counterpart to the JS `FirestoreSessionStore`. It persists each session as a chain of per-turn JSON Patch diffs anchored to periodic, sharded full-state checkpoints, with a per-session pointer document for latest-by-session resolution. No single document approaches Firestore's 1 MiB limit (state is sharded), and the number of diff documents read or written per turn is bounded by the checkpoint interval rather than total session length, so the store scales to arbitrarily long sessions (coding agents, long-lived chatbots). Reads reconstruct via document-ID lookups inside a read-only transaction (strongly consistent, no secondary indexes), writes use `RunTransaction` for atomic read-modify-write, and status changes are surfaced through Firestore's native document listener, so a detached turn aborted by one instance is observed by the instance running it.

It reuses the `exp.Diff` / `exp.ApplyPatch` helpers from `ai/exp` (the same self-built RFC 6902 implementation the runtime uses for `customPatch` streaming), so the Go store and the JS store share one diff/apply definition per language.

The PR also consolidates the experimental Firebase package (moves `plugins/firebase/x` → `plugins/firebase/exp`) and tightens the session-store contract (all stores now require a `sessionId`).

### Basic usage

The store resolves its Firestore client from the Firebase plugin, mirroring `NewFirestoreStreamManager`:

```go
import (
    aix "github.com/firebase/genkit/go/ai/exp"
    "github.com/firebase/genkit/go/plugins/firebase"
    firebasex "github.com/firebase/genkit/go/plugins/firebase/exp"
)

g := genkit.Init(ctx, genkit.WithPlugins(&firebase.Firebase{ProjectId: "my-project"}))

store, err := firebasex.NewFirestoreSessionStore[MyState](ctx, g)
if err != nil { ... }

agent := genkit.DefineAgent(g, "assistant",
    aix.InlinePrompt{ai.WithModelName("googleai/gemini-flash-latest")},
    aix.WithSessionStore(store),
)
```

`State` is the user-defined custom-state type carried in `aix.SessionState.Custom`; it must be JSON-serializable.

### Options

```go
store, err := firebasex.NewFirestoreSessionStore[MyState](ctx, g,
    firebasex.WithCollection("chat-sessions"), // default "genkit-sessions"
    firebasex.WithCheckpointInterval(50),      // turns between full-state checkpoints (default 25)
    firebasex.WithShardSize(256*1024),         // max bytes per shard / diff doc (default 512 KiB)
)
```

Options are scoped to the service they configure, enforced at compile time: passing a stream-manager option (`WithTimeout` / `WithTTL`) to `NewFirestoreSessionStore` is a compile error, as is passing a session-store option (`WithShardSize` / `WithCheckpointInterval` / `WithSnapshotPathPrefix`) to `NewFirestoreStreamManager`. Every option rejects an invalid value (empty / zero / negative / nil); the way to request a default is to omit the option.

### Multi-tenant isolation

`WithSnapshotPathPrefix` derives a per-call prefix from context, nesting every document under a tenant-scoped subcollection so one tenant can never address another's snapshots, even holding a snapshot ID:

```go
store, err := firebasex.NewFirestoreSessionStore[MyState](ctx, g,
    firebasex.WithSnapshotPathPrefix(func(ctx context.Context) string {
        return tenantIDFromContext(ctx) // empty result falls back to the "global" prefix
    }),
)
```

### Constructor

```go
// NewFirestoreSessionStore creates a Firestore-backed snapshot store, resolving
// the Firestore client from the Firebase plugin registered with g (the plugin
// must be passed to genkit.Init first). State is the user-defined custom-state
// type; it must be JSON-serializable.
func NewFirestoreSessionStore[State any](
    ctx context.Context, g *genkit.Genkit, opts ...SessionStoreOption,
) (*FirestoreSessionStore[State], error)
```

`*FirestoreSessionStore[State]` implements `aix.SessionStore[State]` and `aix.SnapshotSubscriber`, so it supports detached (background) invocations and their abort.

### Options

```go
// WithCollection sets the root collection (default "genkit-sessions"); two
// companion collections, "<collection>-shards" and "<collection>-pointers", are
// derived from it. Valid for both the session store and the stream manager.
func WithCollection(collection string) CollectionOption

// WithCheckpointInterval sets the number of turns between full-state checkpoints
// (must be >= 1; default 25). Bounds the diff documents touched per turn.
func WithCheckpointInterval(turns int) SessionStoreOption

// WithShardSize sets the max bytes per shard / diff document (must be > 0;
// default 512 KiB). A diff exceeding it is promoted to a sharded checkpoint.
func WithShardSize(bytes int) SessionStoreOption

// WithSnapshotPathPrefix derives a per-call tenant prefix from context for
// per-tenant isolation (an empty result falls back to "global").
func WithSnapshotPathPrefix(fn func(ctx context.Context) string) SessionStoreOption
```

### Storage layout

For collection `<c>` and tenant prefix `<p>` (default `global`):

```
<c>/<p>/snapshots/<snapshotID>            one document per snapshot (diff or checkpoint)
<c>-shards/<p>/shards/<checkpointID>_<i>  the sharded full state for a checkpoint
<c>-pointers/<p>/pointers/<sessionID>     the session's latest snapshot + reconstruction metadata
```

A `diff` document holds only the RFC 6902 patch from its parent's state; a `checkpoint` document holds a full-state materialization, JSON-serialized and split into `shardSize`-byte chunks stored out of band. Every snapshot carries the metadata to reconstruct it with a single batched `GetAll` (its nearest checkpoint, that checkpoint's shard count, and the ordered diff segment from the checkpoint down to it), so the diff documents touched per reconstruction are bounded by the checkpoint interval, not by total session length. `statePatch` and the error are stored as opaque JSON bytes rather than native nested data, which sidesteps Firestore's nested-array restriction and matches Genkit's own JSON serialization.

`GetLatestSnapshot` reads the per-session pointer (greatest `CreatedAt`, ties broken by snapshot ID) and reconstructs from its cached checkpoint metadata, so it needs no composite index.

### `sessionId` now required on every saved snapshot

`SaveSnapshot` across all stores (`InMemorySessionStore`, `FileSessionStore`, and `FirestoreSessionStore`) now rejects a snapshot whose resolved session ID is empty, with `INVALID_ARGUMENT`. The ID is resolved by preserving the existing row's session ID on update, else the row's own `SessionID`; stores never mint or infer one. The agent runtime stamps a session ID on every row it writes, so an empty one indicates misuse (and the Firestore store needs it to key its per-session pointer). The shared conformance subtest `NoSessionWithoutID` becomes `RejectsRowWithoutSessionID`.

### `plugins/firebase/x` → `plugins/firebase/exp`

The experimental Firebase package is consolidated under `exp` (alongside the new session store), sharing one option set. The `FirestoreStreamManager` moves unchanged but for two pre-existing bug fixes surfaced by running its emulator-gated tests:

- `streamEntry.Chunk` / `Output` were `json.RawMessage`, which the Firestore Go SDK cannot decode back from a Bytes value (now `[]byte`; the wire format is unchanged).
- A `UserFacingError` was persisted with its status-prefixed `Error()` string instead of the bare message, which both duplicated the separately-stored status and would have made the subscribe path rebuild a double-prefixed error.

## Testing

Covered by an emulator-backed suite exercising the shared session-store conformance scenarios (session-ID rules, `GetLatestSnapshot` recency by `CreatedAt`, heartbeat semantics) plus Firestore-specific behavior: multi-checkpoint diff-chain reconstruction, large-state sharding, oversized-diff promotion, branching, multi-tenant prefix isolation, live abort propagation through the native listener, and concurrent read-modify-write atomicity. All pass under `-race`. The compile-time option scoping is verified with a negative-compile check.


pavelgj (commented):
Solid port — builds/vets clean and faithfully mirrors the JS `FirestoreSessionStore`. A few **nonblocking nice-to-haves/nits**:

1. **Stale-shard-pruning test** — `writeShards`' prune loop (`for i := count; i < oldShardCount`) is unexercised. Worth a test: re-checkpoint a large leaf, upsert with much smaller state, assert shard count drops and it still reads back. (JS covers this.)

2. **Error/heartbeat round-trip test** — the `Error []byte` (custom `GenkitError.MarshalJSON`) and `HeartbeatAt` encode/decode paths are untested. A save-then-read assertion that both survive would be nice.

3. **`GetAll` ordering comment** — `reconstructFrom` relies on `tx.GetAll` preserving request order for shard stitching (the JS impl indexes by path instead). A brief note citing the SDK ordering guarantee would make the invariant explicit.

pavelgj (approved):

PR #5621, opened 2026-06-25
feat(js/ai/session-stores): add per-session pointer files for faster snapshot lookup

pavelgj:
Introduce a `.pointers` sub-directory holding per-session pointer files that record each session's current leaf snapshot id. This lets `getSnapshot({ sessionId })` resolve via a single pointer read plus one snapshot read instead of scanning and parsing every snapshot file in the prefix directory.

Falls back to a full directory scan when the pointer is missing (legacy store) or stale, rewriting the pointer afterwards so subsequent lookups are fast again. Also generalize `assertSafeSnapshotId` into a reusable `assertSafeId` helper to validate `sessionId` against path traversal.

### DIFF
--- a/js/ai/src/session-stores.ts
+++ b/js/ai/src/session-stores.ts
@@ -219,29 +219,57 @@ export class InMemorySessionStore<S = unknown> implements SessionStore<S> {
 const DEFAULT_SNAPSHOT_WATCH_POLL_INTERVAL_MS = 2000;
 
 /**
- * Validates that a snapshotId is a plain file basename and not a path that
- * could escape the (per-tenant) prefix directory.
+ * Validates that an id is a plain file basename and not a path that could
+ * escape the (per-tenant) prefix directory.
  *
- * `snapshotId` can arrive straight off the wire (the abort/getSnapshot actions
- * accept a bare string), so without this an id like `../../foo` would let a
- * caller read or write outside the prefix and break per-tenant isolation.
+ * Ids (`snapshotId`, `sessionId`) can arrive straight off the wire (the
+ * abort/getSnapshot actions accept a bare string), so without this an id like
+ * `../../foo` would let a caller read or write outside the prefix and break
+ * per-tenant isolation.
  */
-function assertSafeSnapshotId(snapshotId: string): void {
+function assertSafeId(id: string, label: string): void {
   if (
-    !snapshotId ||
-    snapshotId.includes('/') ||
-    snapshotId.includes('\\') ||
-    snapshotId.includes('\0') ||
-    snapshotId === '.' ||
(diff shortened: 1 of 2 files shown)


js/ai/src/session-stores.ts
```diff
@@ -499,6 +650,24 @@ export class FileSessionStore<S = unknown> implements SessionStore<S> {
...
+    // Maintain the per-session pointer so `sessionId` lookups stay fast (one
+    // pointer read plus one snapshot read). Advance the pointer for a new leaf,
+    // refresh it when rewriting the current leaf, and leave it untouched when
+    // upserting an older (non-leaf) snapshot. Snapshots without a `sessionId`
+    // are not addressable by session, so skip the pointer.
+    const sessionId = full.sessionId ?? full.state?.sessionId;
+    if (sessionId) {
+      const isNew = !current;
+      const pointer = await this.readPointer(sessionId, options);
+      if (isNew || !pointer || pointer.currentSnapshotId === id) {
```
apascal07: `!pointer` here can leave the pointer on a non-leaf: if no pointer exists yet and this save rewrites an older snapshot (`!isNew`), it writes the pointer to that non-leaf, and the fast path won't self-heal until the next new leaf. Since a new snapshot is always the leaf and a rewrite never changes leaf-ness, this can collapse to:

```ts
if (!current) {
  await this.writePointer(sessionId, id, options);
}
```

which also avoids rewriting the pointer on every heartbeat/status update.

js/ai/src/session-stores.ts
```diff
@@ -349,6 +381,81 @@ export class FileSessionStore<S = unknown> implements SessionStore<S> {
...
+    let contents: string;
+    try {
+      contents = await fsp.readFile(filePath, 'utf-8');
+    } catch {
+      // Missing / unreadable pointer: fall back to the scan.
+      return undefined;
+    }
+    try {
+      const parsed = JSON.parse(contents) as PointerDoc;
+      return parsed?.currentSnapshotId ? parsed : undefined;
```
apascal07: A non-string `currentSnapshotId` slips through here and later throws a `TypeError` in `assertSafeId` instead of falling back to the scan. A `typeof` check keeps it best-effort:

```ts
return typeof parsed?.currentSnapshotId === 'string' ? parsed : undefined;
```

apascal07 (approved):
One is nitty but might as well.

js/ai/src/session-stores.ts
pavelgj: Done

js/ai/src/session-stores.ts
pavelgj: Done

PR #5637, opened 2026-06-26
feat(go): realtime telemetry for live agents and the basic-agents sample

apascal07:


go/core/tracing/realtime_span_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	if !async {
+		// Use a fresh context, not the action's: a span often ends because its
+		// action context was canceled (the dev UI canceled the run, or the agent
+		// connection closed), and the final trace write must still land.
+		if err := p.client.Save(context.Background(), td); err != nil {
+			slog.Debug("realtime telemetry: failed to save trace", "error", err)
+		}
+		return
+	}
+	p.wg.Add(1)
```
pavelgj, line 110: Slight concern with the `wg.Add(1)` here vs `wg.Wait()` in `wait()`: `OnStart` runs on the span-creating goroutine, so if a span starts concurrently with a `ForceFlush`/`Shutdown` that has already driven the counter to zero, this trips the `sync.WaitGroup` rule that a positive `Add` from zero must happen-before `Wait` — which can panic ("WaitGroup is reused before previous Wait has returned") or drop the save.

In practice flush/shutdown happen at teardown when nothing new is starting, and this is dev-only telemetry, so the real-world risk is low. Mostly flagging it — could you either add a short comment noting that assumption, or guard against new saves once shutdown has begun?

pavelgj (approved):

PR #5648, opened 2026-06-29
fix(go/plugins/ollama): register embedders by model name, not server address

### LINKED ISSUE #5470: fix(go/plugins/ollama): register embedders by model name, not server address
Sub-issue of #2380.

The Go `ollama` plugin registers embedders keyed by **server address** and takes a separate `serverAddress` argument, diverging from JS, which keys by model name (`ollama/<name>`). Two embedders on one server collide, and lookup-by-name doesn't work. The Go embedder info also omits `dimensions` and `input` modality.

## JS behavior

- Registered as `ollama/<name>` with `dimensions` and `supports.input: ['text']`: [embeddings.ts#L62-L77](https://github.com/genkit-ai/genkit/blob/783fbf23287484b6514732808c2b9a814848a402/js/plugins/ollama/src/embeddings.ts#L62-L77).

## Go divergence

- Registered as `ollama/<serverAddress>`, separate `serverAddress` param, no dimensions/supports: [embed.go#L132-L138](https://github.com/genkit-ai/genkit/blob/783fbf23287484b6514732808c2b9a814848a402/go/plugins/ollama/embed.go#L132-L138), with lookups using the same key at [#L152](https://github.com/genkit-ai/genkit/blob/783fbf23287484b6514732808c2b9a814848a402/go/plugins/ollama/embed.go#L152) and [#L158](https://github.com/genkit-ai/genkit/blob/783fbf23287484b6514732808c2b9a814848a402/go/plugins/ollama/embed.go#L158).

## Scope

- Key embedders by model name to match JS.
- Populate embedder info (`dimensions`, `input: ['text']`).
- **Breaking API change** to `DefineEmbedder`/`Embedder`/`IsDefinedEmbedder` signatures — flag for maintainer sign-off.

_Permalinks pinned to upstream `main` @ `783fbf2` (2026-06-04)._

IzaakGough:
Resolves #5470 

## Summary

- Align the Go Ollama embedder API with the JS plugin by registering and looking up embedders as `ollama/<model>` instead of `ollama/<serverAddress>`.
- Populate embedder metadata consistently and harden request handling so model-bound embedders set dimensions and text-input defaults, reject conflicting model overrides, and handle nil options safely.

## Problem

The Go `ollama` plugin diverged from the JS implementation by keying embedders with server address rather than model name. That meant multiple embedders on the same Ollama server could collide, and callers could not look them up by `ollama/<model>` the way other Genkit surfaces expect. The Go embedder registration path also omitted metadata such as `dimensions` and `supports.input`, and the request path allowed invalid or ambiguous options to flow through without validating that the request model matched the model the embedder was defined for.

## Changes

`DefineEmbedder`, `Embedder`, and `IsDefinedEmbedder` now use the model name as the embedder identity, while requests continue to execute against the plugin's configured server address. `DefineEmbedder` now requires `dimensions`, applies default metadata for the label and text input support, and preserves any caller-provided metadata overrides. The embed request path now normalizes nil and typed-nil options, fills in the bound model when it is omitted, and returns an error if a request tries to override the embedder with a different model. The README examples were updated to reflect the new model-based API and the required `dimensions` argument.

*Note: This PR adds validation for the dimensions param (i.e. don't accept non-positive dimensions. I don't think the JS SDK does this currently so not sure if we want this or not)

## Testing

- Added and updated `go/plugins/ollama/embed_test.go` coverage for model-based registration and lookup, default and custom metadata, request option normalization, conflicting model rejection, non-positive dimensions validation, duplicate model registration, and typed-nil options.

### DIFF
--- a/go/plugins/ollama/README.md
+++ b/go/plugins/ollama/README.md
@@ -201,6 +201,8 @@ You can define and use text embedding models hosted on Ollama (e.g., `nomic-embe
 `dimensions` is required and should match the model's embedding size. The embedder is
 retrieved by model name.
 
+Note: Existing callers should replace server-address lookups with model-name lookups and pass dimensions to `DefineEmbedder`
+
 ### Defining an Embedder
 
 ```go
--- a/go/plugins/ollama/README.md
+++ b/go/plugins/ollama/README.md
@@ -198,6 +198,8 @@ resp, err := genkit.Generate(ctx, g,
 ## Embedding Models
 
 You can define and use text embedding models hosted on Ollama (e.g., `nomic-embed-text`).
+`dimensions` is required and should match the model's embedding size. The embedder is
+retrieved by model name.
 
 ### Defining an Embedder
 
--- a/go/plugins/ollama/README.md
+++ b/go/plugins/ollama/README.md
@@ -203,9 +203,9 @@ You can define and use text embedding models hosted on Ollama (e.g., `nomic-embe
 
 ```go
 // Define an embedder model
-o.DefineEmbedder(g, "http://localhost:11434", "nomic-embed-text", nil)
+o.DefineEmbedder(g, "nomic-embed-text", 768, nil)
 
-embedder := ollama.Embedder(g, "http://localhost:11434")
+embedder := ollama.Embedder(g, "nomic-embed-text")
 ```
 
 ### Usage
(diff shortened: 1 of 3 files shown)


adesinah (changes requested):
@IzaakGough thanks for picking this up! The changes look pretty good. Two things I would love for us to add are:
1. Add a test for the original collision bug: Basically, define two embedders on the same Ollama instance/server with different model names, then assert both are registered and distinct. This directly proves issue is fixed.

 2. Since this is a breaking change, let's update README note: I think something like "**Existing callers should replace server-address lookups with model-name lookups and pass dimensions to DefineEmbedder**" should be just fine.

adesinah (approved):
Looks good to me.

PR #5649, opened 2026-06-29
feat(py/plugins/google-genai): Vertex multimodal embeddings via :predict

hilariie:
## What

Adds working image and video embedding support for the Vertex AI `multimodalembedding@001` model in the `google-genai` plugin, and registers correct output dimensions for `multimodalembedding@001` and `text-embedding-004`.

## Why

The plugin advertised `multimodalembedding@001` as accepting image and video input but routed every embed request through `embed_content`. On Vertex AI that call drops non-text parts, so image and video requests failed at runtime with `400 INVALID_ARGUMENT "Empty instances."`. The advertised output dimension of 768 is also not valid for this model; the default is 1408.

## How

- Route multimodal embedders through the Vertex `:predict` endpoint instead of `embed_content`. `Embedder.generate()` checks `_is_multimodal()` and dispatches to `_generate_multimodal()`.
- Build `{text, image, video}` prediction instances per document. Image and video references map `gs://` URLs to `gcsUri` and `data:` URLs to `bytesBase64Encoded`; `http(s)` URLs are rejected, since Vertex only accepts `gs://` there. An optional video segment config is read from document metadata.
- Parse `predictions[]` into one `Embedding` per image, text, and video chunk, tagging each with `metadata.embedType` and preserving video chunk offsets.
- Advertise `multimodalembedding@001` from the curated `VERTEX_KNOWN_EMBEDDERS` list rather than model discovery: the Vertex catalog over-lists embedders and returns `supported_actions=None`, so Vertex embedders are curated, not discovered.
- Resolve dimensions and supported inputs by exact name, with a fallback to the version-stripped base name.
- Register output dimensions: `multimodalembedding` 1408 (model default; valid values 128/256/512/1408) and `text-embedding-004` 768.

## Testing

- `uv run --directory py pytest plugins/google-genai/test/google_plugin_test.py plugins/google-genai/test/models/googlegenai_embedder_test.py` (71 passed)
- `uv run --directory py ruff check` and `ruff format --check` on the changed files (clean)

New tests cover the `:predict` request path and image/video result mapping, curated-list registration of `multimodalembedding@001`, and dimension and input metadata for bare and versioned model names.

## Limitations

- `multimodalembedding@001` accepts a single instance per `:predict` call, so requests with more than one document are rejected with a clear error until batching is implemented. Single document embeds work.
- The advertised dimension (1408) differs from the JS plugin, which still advertises 768. 

## Screenshots
Screenshot 2026-07-06 at 16 47 43
Screenshot 2026-06-29 at 16 46 56
Screenshot 2026-06-29 at 16 50 55
Screenshot 2026-06-29 at 16 53 24

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
@@ -96,13 +96,13 @@ VERTEX_KNOWN_EMBEDDERS: tuple[str, ...] = (
     'multimodalembedding@001',
 )
 
-# Advertised input modalities for multimodal embedders. Each model belongs to a
-# single backend (gemini-embedding-2* is Google AI only; multimodalembedding@001
-# is Vertex only), so a model name is never ambiguous across backends and no
-# backend scoping is needed. Lookups strip the '@version' suffix first.
-EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
+# Advertised input modalities, per backend. Unknown names default to text-only.
+GOOGLEAI_EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
     'gemini-embedding-2-preview': ['text', 'image', 'video'],
     'gemini-embedding-2': ['text', 'image', 'video'],
+}
+
+VERTEX_EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
     'multimodalembedding': ['text', 'image', 'video'],
 }
 
@@ -112,19 +112,21 @@ def _base_name(name: str) -> str:
     return name.split('@', 1)[0]
 
 
-def get_embedder_options(name: str, label: str) -> EmbedderOptions:
+def get_embedder_options(name: str, label: str, is_vertex: bool = False) -> EmbedderOptions:
     """Return EmbedderOptions metadata for a discovered embedder model.
 
     Args:
         name: The bare (unprefixed) model name, e.g. 'gemini-embedding-2'.
         label: Human-readable label for the embedder.
+        is_vertex: True when resolving for the Vertex backend.
 
     Returns:
         EmbedderOptions describing the model's label, supported inputs and
         static dimensions.
     """
     base = _base_name(name)
-    supports = EMBEDDER_INPUT_SUPPORTS.get(name) or EMBEDDER_INPUT_SUPPORTS.get(base) or ['text']
+    supports_map = VERTEX_EMBEDDER_INPUT_SUPPORTS if is_vertex else GOOGLEAI_EMBEDDER_INPUT_SUPPORTS
+    supports = supports_map.get(name) or supports_map.get(base) or ['text']
     dimensions = EMBEDDER_DIMENSIONS.get(name) or EMBEDDER_DIMENSIONS.get(base)
     return EmbedderOptions(
         label=label,
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
@@ -16,8 +16,9 @@
 
 """Google-Genai embedder model."""
 
+import json
 import sys
-from typing import cast
+from typing import Any, cast
 
 if sys.version_info < (3, 11):
     from strenum import StrEnum
@@ -28,7 +29,7 @@ from google import genai
 from google.genai import types as genai_types
 
 from genkit import DocumentPart, Embedding, EmbedRequest, EmbedResponse
-from genkit._core._typing import DocumentData
+from genkit._core._typing import DocumentData, MediaPart, TextPart
 from genkit.embedder import EmbedderOptions, EmbedderSupports
 from genkit_google_genai.models.utils import PartConverter
 
@@ -67,7 +68,10 @@ class EmbeddingTaskType(StrEnum):
     FACT_VERIFICATION = 'FACT_VERIFICATION'
 
 
-# Static dimensions for known embedders.
+# Static dimensions for known embedders. Keys are version-suffix free
+# (e.g. 'multimodalembedding', not 'multimodalembedding@001') because model
+# discovery returns the bare name on some accounts/regions; lookups strip the
+# '@version' suffix before matching (see get_embedder_options).
 EMBEDDER_DIMENSIONS: dict[str, int] = {
     # Google AI
     'gemini-embedding-2-preview': 3072,
@@ -76,42 +80,55 @@ EMBEDDER_DIMENSIONS: dict[str, int] = {
     # Vertex AI
     'text-embedding-005': 768,
     'text-multilingual-embedding-002': 768,
-    'multimodalembedding@001': 768,
+    'multimodalembedding': 1408,  # default; valid dims 128/256/512/1408 (not 768)
 }
 
 
 # Curated set of Vertex AI embedders that are verified to be callable.
+# The Vertex catalog over-lists embedders (and returns supported_actions=None),
+# so embedders are advertised from this list rather than discovered. Multimodal
+# embedders route through the :predict endpoint (see Embedder._is_multimodal).
 VERTEX_KNOWN_EMBEDDERS: tuple[str, ...] = (
     'text-embedding-005',
     'text-multilingual-embedding-002',
     'gemini-embedding-001',
+    'multimodalembedding@001',
 )
 
-# Advertised input modalities per backend
-GOOGLEAI_EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
+# Advertised input modalities for multimodal embedders. Each model belongs to a
+# single backend (gemini-embedding-2* is Google AI only; multimodalembedding@001
+# is Vertex only), so a model name is never ambiguous across backends and no
+# backend scoping is needed. Lookups strip the '@version' suffix first.
+EMBEDDER_INPUT_SUPPORTS: dict[str, list[str]] = {
     'gemini-embedding-2-preview': ['text', 'image', 'video'],
     'gemini-embedding-2': ['text', 'image', 'video'],
+    'multimodalembedding': ['text', 'image', 'video'],
 }
 
 
-def get_embedder_options(name: str, label: str, is_vertex: bool = False) -> EmbedderOptions:
+def _base_name(name: str) -> str:
+    """Strip a trailing '@version' suffix from a model name (e.g. '@001')."""
+    return name.split('@', 1)[0]
+
+
+def get_embedder_options(name: str, label: str) -> EmbedderOptions:
     """Return EmbedderOptions metadata for a discovered embedder model.
 
     Args:
         name: The bare (unprefixed) model name, e.g. 'gemini-embedding-2'.
         label: Human-readable label for the embedder.
-        is_vertex: True when resolving for the Vertex backend, which advertises
-            text-only because its embed path cannot deliver media.
 
     Returns:
         EmbedderOptions describing the model's label, supported inputs and
         static dimensions.
     """
-    supports = ['text'] if is_vertex else GOOGLEAI_EMBEDDER_INPUT_SUPPORTS.get(name, ['text'])
+    base = _base_name(name)
+    supports = EMBEDDER_INPUT_SUPPORTS.get(name) or EMBEDDER_INPUT_SUPPORTS.get(base) or ['text']
+    dimensions = EMBEDDER_DIMENSIONS.get(name) or EMBEDDER_DIMENSIONS.get(base)
     return EmbedderOptions(
         label=label,
         supports=EmbedderSupports(input=supports),
-        dimensions=EMBEDDER_DIMENSIONS.get(name),
+        dimensions=dimensions,
     )
 
 
@@ -147,6 +164,8 @@ class Embedder:
                 'Embed request input is empty: provide at least one document with content '
                 '(for example input: [{"content": [{"text": "your text here"}]}]).'
             )
+        if self._is_multimodal():
+            return await self._generate_multimodal(request)
         contents = await self._build_contents(request)
         config = self._genkit_to_googleai_cfg(request)
         response = await self._client.aio.models.embed_content(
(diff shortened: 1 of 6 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -138,6 +153,116 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+                    )
+                    if segment_config:
+                        video['videoSegmentConfig'] = segment_config
+                    instance['video'] = video
+                else:
+                    raise ValueError(f'Unsupported contentType for multimodal embedding: {content_type!r}')
+
+        if not instance:
+            raise ValueError('Multimodal embed document has no text, image, or video content.')
+        return instance
```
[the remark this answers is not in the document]
hilariie: Resolved in `ed650017`. I concatenated text parts and added guards against multiple images/videos, but did not add the image+video guard because multimodalembedding@001 supports text, image, and video in a single instance and returns separate embeddings for each (the JS plugin allows this too), so rejecting it would block a valid case.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -138,6 +153,116 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+        """Map a media URL to a Vertex image/video reference (gcsUri or base64)."""
+        if url.startswith('gs://') or url.startswith('http'):
+            ref: dict[str, str] = {'gcsUri': url}
+        elif url.startswith('data:'):
+            ref = {'bytesBase64Encoded': url.split(',', 1)[1]}
+        else:
+            ref = {'bytesBase64Encoded': url}
+        if include_mime_type and content_type:
+            ref['mimeType'] = content_type
+        return ref
```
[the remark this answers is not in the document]
hilariie: Resolved in `ed65017`. This change diverges from the JS plugin, which forwards http(s) URLs as a gcsUri. Since Vertex's gcsUri only accepts gs://, failing fast with a clear error is better than an opaque API 400.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -138,6 +153,116 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+
+    @staticmethod
+    def _prediction_to_embeddings(prediction: dict[str, Any]) -> list[Embedding]:
+        """Convert one multimodal prediction into Genkit embeddings.
+
+        A prediction can carry image, text and/or video embeddings. A single
+        video is split into chunks, yielding one embedding per chunk; the chunk
+        offsets are preserved in each embedding's metadata.
+        """
+        embeddings: list[Embedding] = []
```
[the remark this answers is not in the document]
hilariie: Leaving as-is. The boundary is already guarded (isinstance(body, dict) before reading predictions), the parameter is typed dict[str, Any], and a silent return [] would hide an API contract change rather than surface it.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -145,6 +164,141 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
 
+    def _is_multimodal(self) -> bool:
+        """Whether this embedder uses the Vertex multimodal ``:predict`` API.
+
+        The google-genai ``embed_content`` API only accepts text on Vertex (it
+        silently drops image/video parts), so multimodal embedders must call the
+        ``predict`` endpoint with structured ``{text, image, video}`` instances
+        instead. This mirrors the JS plugin's vertexai embedder.
+        """
+        return 'multimodalembedding' in str(self._version).lower()
```
cabljac, line 195: Am I right that `GoogleAI.resolve()` will build an `Embedder` for any name here? If so, I think `ai.embed(embedder='googleai/multimodalembedding@001')` would route into this Vertex-only `:predict` path - and since the SDK only adds the project/location prefix when `client.vertexai` is true, the API-key client would POST to `generativelanguage.googleapis.com/.../:predict` and get an opaque 404. Would it make sense to guard the route on the client, e.g.

```python
return self._client.vertexai and 'multimodalembedding' in self._version.lower()
```

maybe with a clear ValueError for the non-Vertex case? Happy to test this against a real API-key client if useful.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -145,6 +164,141 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+        payload: dict[str, Any] = {'instances': instances}
+        if request.options:
+            dimension = request.options.get('output_dimensionality') or request.options.get('outputDimensionality')
+            if dimension is not None:
+                payload['parameters'] = {'dimension': dimension}
+
+        # google-genai exposes no typed multimodal-embedding method, so reuse the
+        # client's authenticated low-level transport to POST to :predict. For
+        # Vertex, the project/location prefix is added by the SDK automatically.
+        http_response = await self._client._api_client.async_request(
```
cabljac: This leans on google-genai private internals (`_api_client.async_request`, `http_response.body`). It checks out against the locked SDK (1.63.0), but `pyproject.toml` still declares `google-genai>=1.7.0` and I couldn't verify the older signatures - old 1.x seems to predate `SdkHttpResponse`, so an environment resolving an older SDK might fail at runtime in a way the mocked tests can't catch. Should we bump the floor to a version we've verified? And possibly worth a small guarded helper at some point - gemini.py already touches `_api_client` in two places (one with a "breaks if renamed" warning), and the tests mock this seam in three - though that could be a follow-up.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -145,6 +164,141 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+            request: Genkit embed request.
+
+        Returns:
+            EmbedResponse
+        """
+        instances = [self._build_multimodal_instance(doc) for doc in request.input]
+
+        payload: dict[str, Any] = {'instances': instances}
+        if request.options:
+            dimension = request.options.get('output_dimensionality') or request.options.get('outputDimensionality')
```
cabljac: I think options now behave a bit differently across the two embed paths: this one accepts `output_dimensionality` or `outputDimensionality` but drops `task_type`/`title`, while `_genkit_to_googleai_cfg` honors `task_type`/`title` but only snake_case dimensionality. Is that deliberate (does `:predict` have no `task_type` equivalent)? If so, would a debug log when it's passed be worth it, rather than a silent drop? And would normalizing keys once at the top of `generate()` keep the two paths consistent, or is there a reason to keep them separate?

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -145,6 +164,141 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+            ref: dict[str, str] = {'gcsUri': url}
+        elif url.startswith('http'):
+            raise ValueError(
+                'Vertex multimodal embedding does not accept http(s) media URLs. '
+                'Upload the file to Cloud Storage and pass a gs:// URI, or inline it as a data: URL.'
+            )
+        elif url.startswith('data:'):
+            ref = {'bytesBase64Encoded': url.split(',', 1)[1]}
+        else:
+            ref = {'bytesBase64Encoded': url}
```
cabljac, line 321: A few edge cases I wondered about: a comma-less `data:` URL would raise a bare `IndexError` from `split(',', 1)[1]`; a non-base64 data URL (`data:image/png,%89...`, legal per RFC 2397) would ship its payload verbatim as `bytesBase64Encoded`; and the final else assumes any other string (e.g. a local file path) is raw base64, which I think surfaces as an opaque Vertex 400 where the http(s) branch fail-fasts nicely. Also noticed `PartConverter` in `models/utils.py` already parses the mime type out of a data-URL header - could a data URL with no explicit `contentType` recover its mime from there instead of hitting the "Unsupported contentType" raise? Might be one shared parser closes most of these at once, but curious what you think.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
```diff
@@ -145,6 +164,141 @@ async def generate(self, request: EmbedRequest) -> EmbedResponse:
...
+        elif url.startswith('data:'):
+            ref = {'bytesBase64Encoded': url.split(',', 1)[1]}
+        else:
+            ref = {'bytesBase64Encoded': url}
+        if include_mime_type and content_type:
+            ref['mimeType'] = content_type
+        return ref
+
+    @staticmethod
+    def _prediction_to_embeddings(prediction: dict[str, Any]) -> list[Embedding]:
```
cabljac, line 327: One thing worth deciding explicitly: this drops the 1:1 doc-to-embedding alignment the old path had - a text+image doc yields 2 embeddings, a video one per chunk, and unrecognized/empty predictions get skipped - so a `zip(docs, response.embeddings)` indexer would misattribute vectors. JS flattens the same way, so the shape itself seems right for parity. Would it be enough to tag each embedding with its source document index in metadata (alongside `embedType`) and note in the docstring that consumers correlate via metadata rather than position? Maybe a debug log for the skip case too? Open to other approaches if you had one in mind.

cabljac (changes requested):
Real fix and the test coverage on the rejection cases is thorough. A handful of questions below before this lands, mostly about the new :predict path's edges.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
hilariie: Resolved in `05872088`. Yeah you're right, `resolve()` does build an `Embedder` for any name, so `googleai/multimodalembedding@001` would have POSTed `:predict` on an API-key client. I added the guard but threaded an explicit `is_vertex` flag into `Embedder` (set from `plugin_name == VERTEXAI_PLUGIN_NAME` in `_create_embedder_action`) and raise a `ValueError` in `_generate_multimodal` for the non-Vertex case, rather than reading `client.vertexai`. That keeps it consistent with the existing plugin pattern of not reaching into client internals (`VertexAI.__init__` stores `_project`/`_location` for the same reason), and the plugin already knows its surface at construction time. Added a test that a non-Vertex embedder raises before any request is sent.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
hilariie: Resolved in `2b167678` (floor) and `d1750566` (guard). Bumped the floor to `google-genai>=1.63.0. I Checked and the old `>=1.7.0` was never a hard API requirement. It was introduced in the attribution-headers PR (`7283fc2426`) as just the then-current release with no numeric rationale, and `HttpOptions` predates it, so raising it is safe. I also added a defensive guard before the private call: if `_api_client`/`async_request` is missing it raises a clear error naming `google-genai>=1.63.0`, so a stale SDK fails actionably instead of with an opaque `AttributeError`.

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
hilariie: Resolved in `05872088`. Yes, dropping `task_type`/`title` here is deliberate: `multimodalembedding@001`'s `:predict` API only accepts `dimension` under `parameters`. There's no `taskType`/`title` equivalent (those belong to the text embedders like `text-embedding-005`/`gemini-embedding-001`), so forwarding them would be an API error. This path now reads snake_case `output_dimensionality` only (dropped the camelCase alias) to match `_genkit_to_googleai_cfg`, and I fixed the `or` so a literal `0` isn't discarded. On the silent drop, I'd keep it silent rather than log. Ignoring an inapplicable option is normal and is exactly what the text path already does (`_genkit_to_googleai_cfg` reads only the keys it supports); a `debug` log is invisible unless someone runs at DEBUG level, so it wouldn't reach the caller anyway, and singling this key out would imply it's special when it's just an option that doesn't apply to this model. If we ever decide it's worth surfacing, a `warning` might be the right level instead.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
hilariie: Resolved in `05872088`. Fixed both base64 cases: `_media_reference` now finds the `;base64,` marker and takes the substring after it, and a `data:` URL without that marker raises a clear `ValueError` (Vertex `bytesBase64Encoded` requires base64) instead of an `IndexError` on the comma-less form or silently shipping a non-base64 payload. Added tests for the success and no-marker cases. I left the final `else` as-is (unrecognized string -> raw base64) since inputs arrive as `gs://` / `http(s)` / `data:` via `Document.from_media` in practice (can fail-fast there too if you'd prefer). On recovering the mime type from the data-URL header via `PartConverter`: not load-bearing here because `Document.from_media` always carries a `content_type`, so the "Unsupported contentType" raise isn't reachable from a well-formed part. Should we keep a shared parser as a separate cleanup or fold it into this PR?

py/plugins/google-genai/src/genkit/plugins/google_genai/models/embedder.py
hilariie: Resolved in `05872088`. You're right that one document fans out to multiple embeddings here (text+image = two, video = one per chunk) and a positional `zip(docs, embeddings)` would misattribute. JS flattens the same way, so the shape is right for parity. I documented the contract in the docstring: embeddings are told apart by their `embedType` metadata, not by position, so consumers must correlate via metadata rather than zipping positionally. I held off on the `docIndex` field itself, though: `_generate_multimodal` currently rejects multi-document requests, so there's only ever one input doc and the index would be a constant `0`. It carries no signal until batching lands. I'd rather add it (tested against real multi-doc responses) as part of that change than ship dead metadata now. Let me know if you'd prefer the field in regardless.

cabljac (approved):
lgtm

hilariie:
One post-approval change from the rebase: resolving conflicts with #5703, I initially dropped the `is_vertex` scoping that #5596 added to `get_embedder_options`. I've restored it, but as two per-backend supports maps instead of a boolean that forces text, as the boolean would have wrongly forced multimodalembedding@001 to advertise text-only. This matches how the JS plugin scopes embedder metadata (separate per-backend model tables): Vertex advertises multimodal only for `multimodalembedding@001`, Google AI only for `gemini-embedding-2*`, and unknown names default to text-only on both. The `test_vertexai__resolve_embedder_scopes_supports_to_text` test from #5596 is reinstated unchanged. Metadata-only; no runtime behaviour changes.

cabljac (approved):

PR #5652, opened 2026-06-30
feat(go/plugins/ollama): add Ollama structured output support

### LINKED ISSUE #4788: [Go] Add structured output support in Ollama plugin
https://docs.ollama.com/capabilities/structured-outputs

adesinah:
This PR adds native structured output support to the Go Ollama plugin by mapping Genkit constrained output requests to.
 
Closes: #4788 

## Usage Example:
```go
type Answer struct {
	Answer int `json:"answer"`
}

func main() {
	ctx := context.Background()
	modelName := "llama3.2:latest"
	o := &ollama.Ollama{ServerAddress: "http://localhost:11434"}
	g := genkit.Init(ctx, genkit.WithPlugins(o))
	o.DefineModel(g, ollama.ModelDefinition{Name: modelName, Type: "chat"}, nil)
	model := ollama.Model(g, modelName)

		ai.WithModel(model),
		ai.WithPrompt("What is 2 + 2?"),
		ai.WithOutputType(Answer{}),
	)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(resp.Text()) // {"answer":4}

	// You can also provide a schema directly:

	schema := map[string]any{
		"type": "object",
		"properties": map[string]any{
			"answer": map[string]any{"type": "integer"},
		},
		"required": []string{"answer"},
	}

	resp, errs := genkit.Generate(ctx, g,
		ai.WithModel(model),
		ai.WithPrompt("What is 2 + 2?"),
		ai.WithOutputSchema(schema),
	)
	if errs != nil {
		log.Fatal(errs)
	}

	fmt.Println(resp.Text()) // {"answer":4}
}
```

## Side effect: googlegenai's `$ref` resolution got stricter

Ollama's new structured-output path shares a `$ref` resolver with `googlegenai` (both now go through `plugins/internal/jsonschema`). That resolver is stricter than the one `googlegenai` had before, which is a real behavior change worth calling out even though this PR isn't otherwise about that plugin.

- **Before:** `resolveRef` just took the last `/`-separated segment of a ref and looked it up in `$defs`/`definitions` — so a pointer like `#/components/schemas/Foo` would resolve `Foo` even though it was never a `$defs`/`definitions` reference.
- **After:** only a ref that actually starts with `#/$defs/` or `#/definitions/` resolves. A ref shaped like `#/components/schemas/Foo` now fails with `unable to resolve schema reference` instead of silently resolving the wrong thing.

This is more correct — it stops resolving pointers that were never valid `$defs`/`definitions` references — but anyone relying on the old, looser matching will start seeing that error.

## Side effect: `*ai.GenerationCommonConfig` is no longer accepted as this plugin's config

Two more changes are carried along here that aren't really about structured output either:

- `DefineModel`/`newModel` now call `genkit.DefineModelAction`/`ai.NewModelAction` instead of the older `genkit.DefineModel`/`ai.NewModel`.
- `ollamaChatRequest.ApplyOptions` used to take `cfg any`; it now takes the typed `GenerateContentConfig`.

Because `generate`'s config parameter is now typed as `GenerateContentConfig`, the framework infers that as this model's config type and only accepts a `GenerateContentConfig` (or an equivalent `map[string]any`) on requests. A caller who was passing `*ai.GenerationCommonConfig` directly will now get an "invalid config type" error instead of it being silently accepted.

## Verification
```
  go test ./plugins/ollama ./ai
```

Checklist (if applicable):


go/plugins/ollama/ollama.go
```diff
@@ -119,16 +120,30 @@ func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.M
 
 	if opts != nil {
 		modelOpts = *opts
+		// When the caller provides opts without Supports, derive capabilities
+		// from the model name and type (same logic as the nil-opts path) rather
+		// than applying blanket defaults, so Media/Tools are not over-granted.
+		if modelOpts.Supports == nil {
```
cabljac: nit: Does this shift behaviour if `opts` is passed with `Supports` nil? Previously e.g. `DefineModel(g, {Name:"llama3.1", Type:"chat"}, &ai.ModelOptions{Label:"x"})` would register with `Supports == nil`, whereas now it'd derive `Tools: true` (llama3.1 matches the static `toolSupportedModels` list) plus `Constrained`, so that model would start advertising tool/constrained support it didn't before. Just flagging as a small behaviour change. Might be worth a small test pinning the path.

adesinah: I have added a test documenting the behavior change: 
when `DefineModel` is called with custom options but `Supports == nil`,
  Ollama now derives supports from the model `name/type`. However, an explicit `Supports` is still preserved (no overriding).

go/plugins/ollama/ollama.go
```diff
@@ -543,6 +561,160 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+	}
+	return result
+}
+
+// inlineSchemaRefs recursively replaces $ref nodes with the definition they
+// reference. Sibling keywords alongside a $ref are merged into the resolved
+// definition (JSON Schema draft 2019-09+ semantics). Non-map $defs entries
+// (e.g. boolean schemas), unknown $refs, and cycles are left in place.
+// visited tracks names on the current DFS stack to terminate circular $refs.
+func inlineSchemaRefs(v any, defs map[string]any, visited map[string]bool) any {
```
apascal07: Should these be helpers in `core` instead?

adesinah: @apascal07  I don't think we should put it there. Core should return "raw" data while plugin does the **parsing** (please correct me if i am wrong). Also, because `ai` package imports `core`, so a `core` helper that returns *ai.ModelResponse or *ai.Message would invert the dependency and likely create a cycle. Moreover, a small unexported helper in ollama.go (the plugin) is reasonable.

go/plugins/ollama/ollama.go
```diff
@@ -543,6 +561,160 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+			if hasLocalRefsOutsideDefs(item) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func refName(ref string) string {
+	name := ref[strings.LastIndex(ref, "/")+1:]
```
apascal07: `refName` keeps only the last path segment, so any local pointer resolves against the merged `$defs`+`definitions` pool regardless of where it actually points. Two failures, both confirmed by running this code:

```go
{"$defs": {"b": {"type":"string"}},
 "properties": {"a": {"$ref":"#/properties/b"}, "b": {"type":"integer"}}}
// returns a as {"type":"string"}
```

`a` should be an integer, or left untouched.

```go
{"$defs":       {"Tag": {"type":"integer"}},
 "definitions": {"Tag": {"type":"string"}},
 "properties":  {"tag": {"$ref":"#/definitions/Tag"}}}
// returns tag as {"type":"integer"}
```

A `#/definitions/` pointer resolves to the `$defs` entry. The `$defs wins over definitions` test only exercises a `#/$defs/` ref, so it passes anyway.

In both cases `hasLocalRefsOutsideDefs` then reports no remaining refs, `$defs` is deleted, and the corrupted schema is what Ollama compiles into a grammar. Silent wrong output constraints, no error.

Matching on the full prefix (`#/$defs/` or `#/definitions/`) and resolving each ref against its own pool rules both out. Anything unmatched falls through to the leave-in-place path that already exists.

`plugins/googlegenai/schema.go:186` has the same `tkns[len(tkns)-1]` flaw. Pre-existing, but worth not propagating.

go/plugins/ollama/ollama.go
```diff
@@ -543,6 +561,160 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+	if len(output.Schema) > 0 {
+		return resolveSchemaRefs(output.Schema)
+	}
+	return ai.OutputFormatJSON
+}
+
+// resolveSchemaRefs returns a copy of schema with $ref pointers inlined from
+// the top-level $defs / definitions map where possible. Ollama's constrained
+// output engine does not resolve JSON Schema references itself, so schemas
+// produced by Go struct reflection (which use $defs + $ref) should be
```
apascal07: Go struct reflection here does not use `$defs` + `$ref`. `InferJSONSchema` sets `DoNotReference: true` and `Anonymous: true` (`go/internal/base/json.go:130,145`) and emits fully inlined, ref-free schemas, with an explicit assertion of that at `go/internal/base/json_test.go:173-180`. I probed `InferSchemaMap` with repeated nested types, a slice of them, and a self-recursive pointer field: zero `$defs`, zero `$ref`.

So `ai.WithOutputType(Answer{})`, the path in the PR description, never reaches this function. The only way in is a hand-written map through `ai.WithOutputSchema` or `genkit.DefineSchema`. That is legitimate, but it is narrow, and it is carrying roughly 430 of the 798 added lines.

Two things follow. The comment needs to state the real motivation, otherwise the next maintainer preserves this on a premise that was never true here. And the format wiring would be better landing on its own, with ref inlining as a follow-up where the collision, cycle and precedence semantics get their own review. If it stays, `go/plugins/internal/` is a better home than this file: `plugins/googlegenai/schema.go:186` already has a second, weaker copy of the same resolver.

go/plugins/ollama/ollama.go
```diff
@@ -543,6 +561,160 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+			if found && isMap && !visited[name] {
+				visited[name] = true
+				resolved, _ := inlineSchemaRefs(defMap, defs, visited).(map[string]any)
+				visited[name] = false
+				if resolved == nil {
+					return node
+				}
+				// Merge sibling keywords from the $ref node into the resolved
+				// definition; caller annotations (description, title, etc.) win.
+				if len(node) > 1 {
```
apascal07: Siblings win for every keyword, not just the annotations the comment names. Confirmed:

```go
{"$defs": {"Addr": {"type":"object","properties":{"s":{"type":"string"}}}},
 "properties": {"addr": {"$ref":"#/$defs/Addr", "type":"string"}}}
// returns {"addr":{"type":"string","properties":{"s":{"type":"string"}}}}
```

`type: string` carrying `properties`, handed to Ollama's grammar compiler. Draft 2019-09 treats `$ref` siblings as an implicit `allOf`, so this case is unsatisfiable rather than an override.

Restricting the win to the annotation keywords (`description`, `title`, `default`, `examples`, `deprecated`, `readOnly`, `writeOnly`) matches what the comment already says the intent is.

go/plugins/ollama/ollama.go
```diff
@@ -119,16 +120,30 @@ func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.M
 
 	if opts != nil {
 		modelOpts = *opts
+		// When the caller provides opts without Supports, derive capabilities
+		// from the model name and type (same logic as the nil-opts path) rather
+		// than applying blanket defaults, so Media/Tools are not over-granted.
```
apascal07: The rationale is inverted. `ai.NewModel` normalizes `opts.Supports == nil` to `&ModelSupports{}`, all false (`go/ai/generate.go:149-151`), so nothing was over-granted. `DefineModel(g, def, &ai.ModelOptions{Label: "x"})` previously produced a model that supported nothing: no multiturn, no system role, no tools, no media.

The change is right, but it is a silent capability change riding in a structured-output PR. Anyone passing `opts` without `Supports` today gains four capabilities on upgrade. Worth naming in the description, or splitting out, since it stands on its own as a fix.

go/plugins/ollama/ollama_test.go
```diff
@@ -300,6 +301,212 @@ func TestDynamicPlugin(t *testing.T) {
...
+	modelMeta, ok := action.Desc().Metadata["model"].(map[string]any)
+	if !ok {
+		t.Fatal("action metadata missing model map")
+	}
+	supports, ok := modelMeta["supports"].(map[string]any)
+	if !ok {
+		t.Fatal("action metadata missing supports map")
+	}
+	return ai.ModelSupports{
+		Tools:       supports["tools"].(bool),
```
apascal07: nit: this and the `constrained` assertion below panic instead of failing the test if the metadata shape changes. Comma-ok with `t.Fatalf`.

go/plugins/ollama/ollama_test.go
```diff
@@ -300,6 +301,212 @@ func TestDynamicPlugin(t *testing.T) {
...
+	}
+}
+
+func TestGenerate_StructuredOutput(t *testing.T) {
+	// chatResponse is a minimal valid Ollama /api/chat response.
+	chatResponse := `{"model":"llama3","created_at":"2024-01-01T00:00:00Z","message":{"role":"assistant","content":"{\"answer\":\"42\"}"}}`
+	// generateResponse is a minimal valid Ollama /api/generate response.
+	generateResponse := `{"model":"llama3","created_at":"2024-01-01T00:00:00Z","response":"{\"answer\":\"42\"}"}`
+
+	schema := map[string]any{
```
apascal07, line 812: This schema is ref-free, and so is every other one in the table, so nothing here asserts that `ollamaFormatValue` actually applies `resolveSchemaRefs`. `resolveSchemaRefs` is well covered in isolation, but its only caller is not. A `$defs` case in this table would close that.

Separately, these cases drive `gen.generate` directly and skip `validateSupport`, so `chat model, tools present + Constrained=false` documents an assumption about the framework rather than verifying it. One case through `genkit.Generate` against the mock server would cover the wiring that decides whether `format` is sent at all.

apascal07 (changes requested):

adesinah:
@apascal07 thanks for the review. I have addressed the comments.

go/plugins/ollama/ollama.go
```diff
@@ -413,13 +415,15 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
 	if !isChatModel {
-		payload = ollamaModelRequest{
+		modelReq := ollamaModelRequest{
 			Model:  g.model.Name,
 			Prompt: concatMessages(input, []ai.Role{ai.RoleUser, ai.RoleModel, ai.RoleTool}),
 			System: concatMessages(input, []ai.Role{ai.RoleSystem}),
 			Images: images,
 			Stream: stream,
 		}
+		modelReq.Format = ollamaFormatValue(input.Output)
```
apascal07: nit: `Format` can be set in the struct literal.

go/plugins/ollama/ollama.go
```diff
@@ -543,6 +548,185 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+	return ai.OutputFormatJSON
+}
+
+// resolveSchemaRefs returns a copy of schema with $ref pointers inlined from
+// the top-level $defs / definitions map where possible. Ollama's constrained
+// output engine does not resolve JSON Schema references itself, so schemas
+// supplied through WithOutputSchema or a registered schema can therefore be
+// flattened before forwarding. Schemas inferred from Go types are already
+// emitted without references.
+func resolveSchemaRefs(schema map[string]any) map[string]any {
```
apascal07: Continuing the earlier thread on this resolver: the doc comment now states the real motivation, but the shared location is still open. `plugins/googlegenai/schema.go` carries a weaker copy of the same resolver, and these helpers are plain `map[string]any` transforms with no `ai` imports, so `go/plugins/internal` works without the cycle concern.

adesinah:
@apascal07 this is good for re-review now.

go/plugins/ollama/ollama.go
```diff
@@ -224,13 +227,12 @@ func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.M
...
-		Supports:     modelOpts.Supports,
-		Versions:     []string{},
-		ConfigSchema: core.InferSchemaMap(GenerateContentConfig{}),
+		Label:    "Ollama - " + model.Name,
+		Supports: modelOpts.Supports,
+		Versions: []string{},
 	}
 	gen := &generator{model: model, serverAddress: o.ServerAddress, timeout: o.Timeout}
-	return genkit.DefineModel(g, api.NewName(provider, model.Name), meta, gen.generate)
+	return genkit.DefineModelAction(g, api.NewName(provider, model.Name), meta, gen.generate)
```
apascal07, line 232: The move to `DefineModelAction` and the deletion of `ApplyOptions(cfg any)` are a separate change from structured output, and neither the title nor the description mentions them. Fine to carry here, but they should be called out, including the consequence that `*ai.GenerationCommonConfig` is no longer an accepted config type for this plugin.

go/plugins/internal/jsonschema/refs.go
```diff
@@ -0,0 +1,186 @@
...
+
+func inlineRefs(v any, defs, definitions map[string]any, visited map[string]bool) any {
+	switch node := v.(type) {
+	case map[string]any:
+		if ref, ok := node["$ref"].(string); ok {
+			def, found := refTarget(ref, defs, definitions)
+			defMap, isMap := def.(map[string]any)
+			if found && isMap && !visited[ref] && hasOnlyAnnotationSiblings(node) {
+				visited[ref] = true
+				resolved, _ := inlineRefs(defMap, defs, definitions, visited).(map[string]any)
```
apascal07: Each `$ref` re-expands its target from scratch, so a shared `$defs` graph inlines exponentially rather than linearly. A 22-level diamond (each `Ln` referencing `L(n-1)` twice) produced a 247 MB schema in 13.6s here, and that result is then marshaled and POSTed.

Genkit's own `InferSchemaMap` emits no `$defs`, so this only reaches caller-supplied schemas via `WithOutputSchema` or dotprompt, but there is no bound on the work. Memoizing acyclic defs, or abandoning inlining past a node budget and returning the schema untouched, would cap it.

go/plugins/internal/jsonschema/refs.go
```diff
@@ -0,0 +1,186 @@
...
+		if key != "$ref" && key != "$defs" && key != "definitions" && !isAnnotation(key) {
+			return false
+		}
+	}
+	return true
+}
+
+func isAnnotation(key string) bool {
+	switch key {
+	case "description", "title", "default", "examples", "deprecated", "readOnly", "writeOnly":
```
apascal07: `$comment` is missing here, so `{"$ref": "#/$defs/X", "$comment": "..."}` counts as a structural sibling and the ref ships to Ollama unresolved, which is the exact failure this package exists to prevent. `$id` and `$anchor` are worth considering for the same reason.

go/plugins/internal/jsonschema/refs.go
```diff
@@ -0,0 +1,186 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package jsonschema
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ResolveRefs returns a copy of schema with direct $ref pointers into the
```
apascal07: Not quite a copy: `inlineRefs` returns the caller's own node whenever a ref is left in place, and the `len(defs) == 0 && len(definitions) == 0` early return hands back `schema` itself.

Nothing mutates the result today so there is no live bug, but the input is `input.Output.Schema`, which the framework owns. A future caller taking this comment at its word would corrupt it. Either copy consistently or say the result may share structure with the input.

go/plugins/googlegenai/schema.go
```diff
@@ -25,7 +25,7 @@ func toGeminiSchema(originalSchema map[string]any, genkitSchema map[string]any)
 		if !ok {
 			return nil, fmt.Errorf("invalid $ref value: not a string")
 		}
-		s, err := resolveRef(originalSchema, ref)
+		s, err := pluginjsonschema.ResolveRef(originalSchema, ref)
```
apascal07: Separate from the pool selection fix, which is a genuine bug fix: this also tightens googlegenai. The old `resolveRef` split on `/` and took the last token, so `#/components/schemas/Foo` resolved `Foo` out of `$defs`. Those now fail with `unable to resolve schema reference`.

More correct, but it is a behavior change in a plugin this PR is otherwise not about. Worth a line in the description, and ideally its own commit.

go/plugins/ollama/model_test.go
```diff
@@ -48,149 +48,481 @@ func TestOllamaChatRequest_MarshalJSON(t *testing.T) {
...
+			t.Fatalf("marshal error: %v", err)
+		}
+		got := string(data)
+		if strings.Contains(got, `"format"`) {
+			t.Errorf("expected \"format\" key to be absent, got: %s", got)
+		}
+	})
+}
+
+func TestResolveSchemaRefs(t *testing.T) {
```
apascal07: This table only exercises `pluginjsonschema.ResolveRefs`, nothing from `model.go`. Meanwhile `refs_test.go` in the owning package has three cases, two of which are duplicated again in `googlegenai/schema_test.go`. Moving the table into `refs_test.go` keeps the coverage with the code and drops the duplication.

go/plugins/ollama/ollama.go
```diff
@@ -33,9 +33,9 @@ import (
 	"time"
 
 	"github.com/firebase/genkit/go/ai"
-	"github.com/firebase/genkit/go/core"
 	"github.com/firebase/genkit/go/core/api"
 	"github.com/firebase/genkit/go/genkit"
+	pluginjsonschema "github.com/firebase/genkit/go/plugins/internal/jsonschema"
```
apascal07: nit: this needs an alias at every use site because it collides with `invopop/jsonschema`, which this file also imports. Naming the package `schemautil` would avoid that in all three files.

go/plugins/ollama/ollama.go
```diff
@@ -734,6 +736,23 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
+func ollamaFormatValue(output *ai.ModelOutputConfig) any {
+	if output == nil || !output.Constrained {
+		return nil
+	}
+	if len(output.Schema) > 0 {
+		// Ollama's constrained-output engine does not resolve JSON Schema
+		// references, so flatten schemas supplied explicitly by callers.
+		return pluginjsonschema.ResolveRefs(output.Schema)
+	}
+	return ai.OutputFormatJSON
```
apascal07: `GenerateWithRequest` only sets `Constrained` when `opts.Output.JsonSchema != nil`, and that also populates `Schema`, so this branch is unreachable from the framework. Fine as defense in depth, but the `Constrained=true without schema` test case documents a state that cannot occur. Worth saying so in the comment, or dropping both.

go/plugins/ollama/ollama.go
```diff
@@ -604,13 +605,15 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
...
-		payload = ollamaModelRequest{
+		modelReq := ollamaModelRequest{
 			Model:  g.model.Name,
 			Prompt: concatMessages(input, []ai.Role{ai.RoleUser, ai.RoleModel, ai.RoleTool}),
 			System: concatMessages(input, []ai.Role{ai.RoleSystem}),
 			Images: images,
 			Stream: stream,
+			Format: ollamaFormatValue(input.Output),
 		}
+		payload = modelReq
```
apascal07: nit: `modelReq` exists only to be assigned to `payload` on the next line.

go/plugins/ollama/ollama.go
```diff
@@ -590,7 +591,7 @@ func Ptr[T any](v T) *T {
 }
 
 // Generate makes a request to the Ollama API and processes the response.
-func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb func(context.Context, *ai.ModelResponseChunk) error) (*ai.ModelResponse, error) {
+func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, config GenerateContentConfig, cb ai.ModelStreamCallback) (*ai.ModelResponse, error) {
```
apascal07, line 611: The new typed `config` is only consumed on the chat branch: `ollamaModelRequest` has no `options`, `think`, or `keep_alive` fields, and `ApplyOptions` is never called for `Type != "chat"`. `WithConfig(&GenerateContentConfig{Temperature: Ptr(0.5)})` on a generate-type model sends `{"model":...,"prompt":"hi","stream":false}`.

The drop is pre-existing, but the typed parameter makes it read as deliberate. A TODO or a follow-up issue would be enough, especially since this struct is already gaining `Format` here.

go/plugins/ollama/ollama.go
adesinah: Fixed — the comment now says what's actually true here (`WithOutputSchema`/`DefineSchema`, not `WithOutputType`), and the resolver already lives in `plugins/internal/jsonschema` so googlegenai can share it too.

go/plugins/internal/jsonschema/refs.go
adesinah: Fixed — `inlineRefs` now memoizes each def by ref name, so a shared def gets expanded once instead of re-expanding at every occurrence. Added a diamond-shaped regression test to cover it.

go/plugins/internal/jsonschema/refs.go
adesinah: Fixed — `$comment`, `$id`, and `$anchor` are now treated the same as description/title, so a ref with one of those siblings still gets inlined instead of shipping unresolved. Added a test for the `$comment` case.

go/plugins/internal/jsonschema/refs.go
adesinah: Fixed — updated the doc to say the result may share structure with the input instead of claiming it's always a copy, since that's what actually happens on the two paths you flagged.

go/plugins/googlegenai/schema.go
adesinah: Called this out in the PR description — added a section explaining the before/after behavior so it's not a silent surprise. Left it folded into this commit rather than splitting it out, to avoid rewriting history on a branch already in review.

go/plugins/ollama/model_test.go
adesinah: Fixed — moved the table into refs_test.go next to `ResolveRefs`, and dropped the now-redundant standalone mutation test that lived there.

go/plugins/ollama/ollama.go
adesinah: Fixed — renamed the package to `schemautil`, so nothing importing it needs an alias anymore, including this file.

go/plugins/ollama/ollama.go
adesinah: Left the fallback in as defense-in-depth, but added a comment (and updated the matching test's comment) explaining it's unreachable through the framework and only matters for a hand-built ModelOutputConfig.

go/plugins/ollama/ollama.go
adesinah: Fixed — inlined it.

go/plugins/ollama/ollama.go
adesinah: Left the behavior as-is but added a TODO explaining the gap (`ollamaModelRequest` has no options/think/keep_alive fields), so it reads as a known, tracked issue rather than an accident.

go/plugins/ollama/ollama.go
adesinah: Called this out in the PR description — added a section explaining both the `DefineModelAction`/`ApplyOptions` change and the consequence that `*ai.GenerationCommonConfig` no longer works as this plugin's config.

go/plugins/ollama/ollama.go
```diff
@@ -751,6 +758,26 @@ func (g *generator) generate(ctx context.Context, input *ai.ModelRequest, cb fun
 	}
 }
 
+// ollamaFormatValue returns the value for the Ollama API's format field when
+// native constrained output is active, or nil when the framework chose prompt
+// injection (e.g. tools are present or the caller opted out). Passing nil
+// leaves the format field absent from the request, which is correct in both
+// the prompt-injection case and plain text output.
+func ollamaFormatValue(output *ai.ModelOutputConfig) any {
```
apascal07, line 765: `Constrained` is the only gate, so the `enum` format also ships its schema in `format` and the reply stops parsing. Confirmed against a mock server:

```go
ai.WithOutputSchema(map[string]any{
	"type": "object",
	"properties": map[string]any{
		"value": map[string]any{"type": "string", "enum": []any{"RED", "BLUE"}},
	},
}),
ai.WithOutputFormat(ai.OutputFormatEnum),

// sent:      "format": {"properties":{"value":{"enum":["RED","BLUE"],"type":"string"}},"type":"object"}
// returned:  {"value":"RED"}
// Generate:  model failed to generate output matching expected schema:
//            message {value:RED} not in list of valid enums: RED, BLUE
```

`enumHandler` sets `Constrained: true` (`go/ai/format.go:550`) and expects the bare value back, and `objectEnums` supports exactly this nested-property shape. On `main` the same call succeeds: Ollama advertised no constrained support, so the framework injected enum instructions instead. The top-level `{"type":"string","enum":[...]}` shape survives only because `quoteStripper` removes the quotes the grammar adds.

The other plugins gate on the format, not on the flag alone: anthropic on `Format == "json" && Schema != nil && Constrained` (`plugins/internal/anthropic/anthropic.go:399`), compat_oai switches on `Format` (`plugins/compat_oai/generate.go:434`), googlegenai maps enum to `text/x.enum` before setting a schema (`plugins/googlegenai/gemini.go:448`).

go/plugins/internal/schemautil/refs.go
```diff
@@ -0,0 +1,210 @@
...
+	switch key {
+	case "description", "title", "default", "examples", "deprecated", "readOnly", "writeOnly",
+		"$comment", "$id", "$anchor":
+		return true
+	default:
+		return false
+	}
+}
+
+func hasLocalRefsOutsideDefs(v any) bool {
```
apascal07, line 182: The `$defs`/`definitions` skip applies at every depth, not just at the root, so a ref under a property with either name is invisible to this walk and the pool is deleted out from under it. Confirmed:

```go
{"type": "object",
 "properties": {"definitions": {"$ref": "#/$defs/Never"},
                "other":       {"$ref": "#/$defs/Ok"}},
 "$defs": {"Never": false, "Ok": {"type": "string"}}}

// returns {"properties":{"definitions":{"$ref":"#/$defs/Never"},"other":{"type":"string"}},"type":"object"}
```

`Never` is a boolean schema, so `inlineRefs` leaves its ref in place, but `hasLocalRefsOutsideDefs` never reaches it, `$defs` is dropped, and Ollama gets the dangling ref this package exists to prevent. Only the two root keys are definition pools; below that, `definitions` is an ordinary property name.

go/plugins/internal/schemautil/strict.go
```diff
@@ -14,8 +14,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-// Package jsonschema contains JSON Schema helpers shared by Genkit provider plugins.
-package jsonschema
+// Package schemautil contains JSON Schema helpers shared by Genkit provider plugins.
+package schemautil
```
apascal07, line 18: nit: the package rename landed, the directory did not, so the import path still reads `internal/jsonschema` while the identifier is `schemautil`, and the three existing importers still alias it `pluginjsonschema` (`internal/anthropic/anthropic.go:34`, `compat_oai/compat_oai.go:34`, `compat_oai/generate.go:28`). Renaming the directory and dropping those aliases finishes it.

apascal07 (approved):

apascal07 (changes requested):
Maybe pull in `main`, there might have been changed since you last rebased, not sure.

PR #5659, opened 2026-07-01
feat(py/ollama): add configurable transport headers and timeouts

hilariie:
## What

Brings the Ollama Python plugin's transport surface to parity with the JS/Go plugins (slice B2 of #5521, stacked on #5658 ).

- **Configurable request headers**:
  - Static dict headers are applied to the cached per-event-loop Ollama client.
  - Sync/async callable headers are resolved **per request** with `RequestHeaderParams` (`server_address`, model/embedder definition, and model/embed request context), matching JS plugin semantics and supporting expiring auth tokens.
  - Callable-header requests use a fresh Ollama client because the Python SDK bakes headers in at construction; that client's wrapped `httpx` pool is closed on exit.
- **`timeout` kwarg** forwarded to the underlying `httpx` client via `_make_client`.
- **Friendly connection errors**: Ollama SDK transport failures on model, embedder, and `list_actions` paths become `OllamaConnectionError` (subclass of `ConnectionError`) naming the address and pointing at `ollama serve`.
- **Scoped error wrapping**: media URL fetch failures remain raw HTTP errors instead of being misreported as Ollama server outages.

### Latent Bugs Fixed

- `request_headers` was stored on the plugin but never passed to the client factory.
- Callable `request_headers` originally required fresh SDK clients but did not close their underlying `httpx` pools.
- Broad connection-error wrapping could label non-Ollama transport failures, such as media fetch failures, as Ollama server connectivity issues.

## Test Plan

- `uv run pytest plugins/ollama/tests/plugin_api_test.py plugins/ollama/tests/models/ollama_models_test.py --no-cov` -> 118 passed
- Live Ollama smoke test:
  - `list_actions()` returned local `qwen3.5:4b` and `nomic-embed-text:latest`
  - chat generate against `qwen3.5:4b` returned `OK`
  - embed against `nomic-embed-text:latest` returned a 768-dimensional embedding
  - callable `request_headers` context was exercised for list/model/embed paths

### DIFF
--- a/py/plugins/ollama/src/genkit/plugins/ollama/__init__.py
+++ b/py/plugins/ollama/src/genkit/plugins/ollama/__init__.py
@@ -155,6 +155,7 @@ See Also:
     - Genkit documentation: https://genkit.dev/
 """
 
+from genkit.plugins.ollama._errors import OllamaConnectionError
 from genkit.plugins.ollama.models import ModelDefinition, OllamaConfig, OllamaSupports
 from genkit.plugins.ollama.plugin_api import Ollama, ollama_name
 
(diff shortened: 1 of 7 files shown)


cabljac (approved):
lgtm

PR #5667, opened 2026-07-01
fix(go): implement recursive schema reference resolution

MichaelDoyle:
This change fixes "Reference must be canonical" validation errors when running actions through the JSON path (e.g., via the Dev UI).

The Go framework now:
1. Recursively resolves 'genkit:' schema references within nested objects and arrays in ResolveSchema.
2. Ensures schema resolution is performed in runJSONWithTelemetry before input validation.

This ensures that strict JSON schema validators see fully-resolved definitions rather than internal Genkit reference strings.

### DIFF
--- a/go/core/core.go
+++ b/go/core/core.go
@@ -66,22 +66,130 @@ func SchemaRef(name string) map[string]any {
 // Returns the original schema if no $ref is present, or the resolved schema if found.
 // Returns an error if the schema reference cannot be resolved.
 func ResolveSchema(r api.Registry, schema map[string]any) (map[string]any, error) {
+	return resolveSchema(r, schema, nil, 0)
+}
+
+const maxSchemaDepth = 50
+
+func resolveSchema(r api.Registry, schema map[string]any, seen map[uintptr]bool, depth int) (map[string]any, error) {
 	if schema == nil {
 		return nil, nil
 	}
-	ref, ok := schema["$ref"].(string)
-	if !ok {
+	if depth > maxSchemaDepth {
+		return nil, fmt.Errorf("schema reference too deep (possible cycle)")
+	}
+
+	if ref, ok := schema["$ref"].(string); ok {
+		if schemaName, found := strings.CutPrefix(ref, "genkit:"); found {
+			resolved := r.LookupSchema(schemaName)
+			if resolved == nil {
+				return nil, fmt.Errorf("schema %q not found", schemaName)
+			}
+
+			if seen[reflect.ValueOf(resolved).Pointer()] {
+				return schema, nil
+			}
+
+			newSeen := make(map[uintptr]bool, len(seen)+1)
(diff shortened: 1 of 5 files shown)


go/core/core.go
```diff
@@ -66,22 +66,110 @@ func SchemaRef(name string) map[string]any {
 // Returns the original schema if no $ref is present, or the resolved schema if found.
 // Returns an error if the schema reference cannot be resolved.
 func ResolveSchema(r api.Registry, schema map[string]any) (map[string]any, error) {
+	return resolveSchema(r, schema, 0)
+}
+
+const maxSchemaDepth = 100
```
apascal07: This seems... like a lot.

apascal07:
@cabljac this is somewhat related to your work supporting the new JSON schema field in the Gemini plugin (and conversely allowing recursion/references in schema generation) I think, can we validate both pieces of work and make sure they work well together?

go/core/core.go
```diff
@@ -66,22 +66,110 @@ func SchemaRef(name string) map[string]any {
...
+	var newSchema map[string]any
+	for k, v := range schema {
+		resolved, err := resolveValue(r, v, depth+1)
+		if err != nil {
+			return nil, err
+		}
+
+		if newSchema == nil && !isSame(v, resolved) {
+			newSchema = make(map[string]any, len(schema))
+			for k2, v2 := range schema {
```
apascal07, line 130: I think this can drop keys. The inner loop assumes it visits keys in the same order as the outer loop, but Go randomizes map iteration order on each `range`, so they won't line up. If a key was already handled by the outer loop but happens to come after `k` in the inner loop, it never gets copied into `newSchema` and just disappears.

Quick repro with `{"type":"object","properties":{...$ref}}`: when `properties` triggers the clone, `type` drops out of the result about 28% of the time. So the resolved schema can come back missing `type`/`required`/etc, which is the same JSON path this PR is trying to fix.

I think you can just copy all the keys when cloning and drop the break, since the outer loop overwrites the resolved key anyway. The array version below looks fine since it copies over a slice.

MichaelDoyle:
> @cabljac this is somewhat related to your work supporting the new JSON schema field in the Gemini plugin (and conversely allowing recursion/references in schema generation) I think, can we validate both pieces of work and make sure they work well together?

lemme know @cabljac @apascal07  -- I am also happy to back off from the slop cannon on this one. Right now, prompts with input schemas don't work on the dev-ui from what I can tell, so that's what I am aiming to fix.

go/core/core.go
MichaelDoyle: Made it 50. Although Gemini thinks we can/should be lenient here.

go/core/core.go
MichaelDoyle: fixed

Issue #5633, opened 2026-06-26
[plugins] Anthropic Python: dynamic model listing via beta.models.list with static fallback


hilariie:
Parent: #5617 (Phase B, P1: item 19)

## Why

Model discovery is static: listing only returns the hard-coded set. New or access-gated models (including ones intentionally kept out of the static set, like `claude-fable-5` and `claude-mythos-5`; see item 18, #5626) never appear in reflection until someone edits the registry by hand. JS lists models from the API to capture both stable and beta; Go does the same with a short cache.

This keeps the plugin current without code changes and lets approved accounts see the models they actually have access to. It pairs with the registry refresh (item 18, #5626).

## What needs to be done

- Discover available models from the Anthropic API (the listing that includes both stable and beta) and surface them, resolving each through the existing model-info / generic fallback so access-gated IDs appear for approved accounts.
- Fall back to the static set when the API is unreachable. A short cache (Go uses ~1h) is reasonable.

## Acceptance

- Listing returns API-discovered models (stable + beta) when reachable, and falls back to the static set on error.
- Tests cover the dynamic path (mocked) and the static fallback.

## Reference

- Current static listing: `py/plugins/anthropic/src/genkit/plugins/anthropic/plugin.py:131-146`.
- JS target: `js/plugins/anthropic/src/list.ts:37` (`beta.models.list()`).

PR #5684, opened 2026-07-03
fix(py/plugins/anthropic): refresh model registry with sonnet-4-6, fable-5; drop retired claude-3-5 models

### LINKED ISSUE #5626: [plugins] Anthropic Python: refresh model registry (add sonnet-4-6, drop retired 3-5-*)
Parent: #5617 (Phase A, P0: item 18)

## Why

The Python static model registry has drifted from both the JS reference set and the current Anthropic lineup. It is missing `claude-sonnet-4-6` (which the JS `KNOWN_MODELS` set advertises) and still advertises the retired `claude-3-5-haiku` and `claude-3-5-sonnet`. Advertising retired models while omitting a current one is misleading in reflection and the dev UI, and breaks parity with JS.

This is a low-risk parity change: it is isolated to the model-info module, has no upstream or downstream dependencies, and nothing else in the epic blocks on it, so it can land early as a standalone change.

## What needs to be done

- Bring the default known-model set in line with the JS static set and the current API: surface `claude-sonnet-4-6`, stop advertising the retired `claude-3-5-*` models, and keep the existing 4.x entries.
- Keep limited-access and testing-blocked models **out of the default static set** while remaining reachable:
  - `claude-fable-5`: present in the JS set but **deferred here due to testing limitations**.
  - `claude-mythos-5`: access-gated and absent from JS `KNOWN_MODELS`.
  Both must stay usable via generic model references and via dynamic listing (item 19, #5633 ), just not in the default set.

## Acceptance

- `claude-sonnet-4-6` is advertised; retired `claude-3-5-*` entries are no longer advertised; existing 4.x entries are retained.
- `claude-fable-5` and `claude-mythos-5` are absent from the static set but still resolve when referenced by full name.
- Model/registry tests updated; CI green.

## Reference

- Current Python set: `py/plugins/anthropic/src/genkit/plugins/anthropic/model_info.py:166-178`; generic fallback `get_model_info()` at `model_info.py:189-204`.
- JS target: `js/plugins/anthropic/src/models.ts:105-117` (`KNOWN_MODELS`).

hilariie:
Fixes #5626

## Description

The Python Anthropic plugin's static model registry had drifted from the JS plugin and from Anthropic's current lineup. It still advertised `claude-3-5-haiku` and `claude-3-5-sonnet`, both of which were retired in late 2025 / early 2026, while missing `claude-sonnet-4-6` and `claude-fable-5`, which the JS plugin already registers in `KNOWN_MODELS`.

This PR brings the registry back in line:

- Adds `claude-sonnet-4-6`, `claude-sonnet-5`, and `claude-fable-5` with the same capabilities as the other 4.x entries (multiturn, media, tools, system role, `text`/`json` output, constrained generation), matching how JS registers them via `commonRef(..., ADVANCED_MODEL_INFO)`.
- Removes the retired `claude-3-5-haiku` and `claude-3-5-sonnet` entries. The registry stays at 11 models.
- Refreshes the module docstring so the supported models table and usage examples no longer reference retired model names.

One deviation from the issue as written: #5626 originally said to hold off on `claude-fable-5` due to testing limitations. Fable is publicly available now, so it's included here with full parity to JS.

`claude-mythos-5` stays out of the static set on purpose. It's access-gated and absent from the JS registry too, but it remains usable by full name because unknown models already resolve through the generic fallback in `get_model_info()`. There's a test covering exactly that. Dynamic model listing is tracked separately in #5633.

The Vertex AI model garden plugin also references the retired 3.5 names, but that's a separate registry and out of scope for this issue.

## Test plan

- Updated `test_supported_models`: removed the dead `claude-3-5-haiku` media branch, added label/version/constrained assertions for both new models.
- Added `test_mythos_excluded_but_resolvable`: asserts `claude-mythos-5` is not advertised but still resolves via the generic fallback with text-only output.
- Updated the structured output fallback test to use an explicitly unknown model name instead of the removed `claude-3-5-haiku`.
- `pytest plugins/anthropic/tests/` passes (66 tests), ruff format and lint clean.

## Screenshots
Screenshot 2026-07-03 at 13 49 32
Screenshot 2026-07-03 at 14 05 47
Screenshot 2026-07-03 at 14 04 54
Screenshot 2026-07-03 at 14 58 11

### DIFF
--- a/py/plugins/anthropic/src/genkit/plugins/anthropic/__init__.py
+++ b/py/plugins/anthropic/src/genkit/plugins/anthropic/__init__.py
@@ -105,7 +105,7 @@ Architecture Overview::
     │  └── Streaming support                                                  │
     ├─────────────────────────────────────────────────────────────────────────┤
     │  model_info.py - Model Registry                                         │
-    │  ├── SUPPORTED_MODELS (claude-3.5-sonnet, opus, haiku, etc.)            │
+    │  ├── SUPPORTED_MODELS (claude-sonnet-4-5, opus, haiku, etc.)            │
     │  └── Model capabilities and metadata                                    │
     └─────────────────────────────────────────────────────────────────────────┘
 
@@ -118,10 +118,10 @@ Supported Models:
     ┌─────────────────────────────────────────────────────────────────────────┐
     │ Model                     │ Description                                 │
     ├───────────────────────────┼─────────────────────────────────────────────┤
-    │ claude-3-5-sonnet-latest  │ Balanced performance and capability         │
-    │ claude-3-5-haiku-latest   │ Fast and cost-effective                     │
-    │ claude-3-opus-latest      │ Most capable, complex tasks                 │
-    │ claude-sonnet-4-20250514  │ Latest Sonnet model                         │
+    │ claude-sonnet-4-5         │ Balanced performance and capability         │
+    │ claude-haiku-4-5          │ Fast and cost-effective                     │
+    │ claude-opus-4-5           │ Most capable, complex tasks                 │
+    │ claude-sonnet-4-6         │ Latest Sonnet model                         │
     └───────────────────────────┴─────────────────────────────────────────────┘
 
 Key Components:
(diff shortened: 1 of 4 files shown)


cabljac (approved):
lgtm

PR #5696, opened 2026-07-06
fix(js/plugins/google-genai): remove dead contextCache config flag and strip it from requests

### LINKED ISSUE #5593: [JS] Dead `contextCache` config flag on google-genai googleAI models
The `googleAI` model config in `@genkit-ai/google-genai` exposes a `contextCache` boolean:

```ts
// js/plugins/google-genai/src/googleai/gemini.ts:179
contextCache: z
  .boolean()
  .describe('Context caching allows you to save and reuse precomputed input tokens that you wish to use repeatedly.')
  .optional(),
```

It's never consumed. The only occurrence of `contextCache` in the whole plugin is this schema definition. In the request builder it isn't destructured out (`gemini.ts:735-839`), and `removeClientOptionOverrides` (`utils.ts:214`) doesn't strip it, so it gets spread verbatim into `generationConfig`.

Confirmed with a mock-fetch repro: `generate({ config: { contextCache: true } })` against `gemini-2.5-flash` sends this body to the API:

```jsonc
// top-level keys: generationConfig, contents
{ "generationConfig": { "contextCache": true } }
```

So it's passed to the Gemini API as an unknown `generationConfig` field (no `cachedContent` reference, no cache-creation call), where it has no effect.

This is a leftover from the deprecated `@genkit-ai/googleai` plugin's explicit context-caching, removed in #4948. The replacement is implicit caching (Gemini 2.5+ caches common prefixes automatically), which needs no config. Came up while looking at #3362.

The flag reads as a supported feature but does nothing, which is misleading. Proposed fix: remove `contextCache` from `GeminiConfigSchema`. It's a public config field so it's technically a breaking type change, though setting it never had any effect.

Notes:
- googleAI only. `vertexai` config doesn't have it.
- Happy to put up the PR removing it.

cabljac:
Fixes #5593

## Problem

The googleAI `GeminiConfigSchema` exposes a `contextCache` boolean that is never consumed. From what I can see it is a leftover from the deprecated `@genkit-ai/googleai` plugin's explicit context caching (removed in #4948), superseded by implicit caching on Gemini 2.5+ which needs no config.

It is worse than just a dead option: the schema is `.passthrough()`, and `contextCache` is neither destructured out in the request builder nor stripped by `removeClientOptionOverrides`, so a config that sets it forwards the flag verbatim into `generationConfig` on the outbound request as an unknown field.

```ts
// gemini.ts request builder: not in the destructure list...
const { apiKey, safetySettings, /* ... */ serviceTier, ...restOfConfigOptions } = requestOptions;
// ...so it rides along here
const generationConfig: GenerationConfig = {
  ...removeClientOptionOverrides(restOfConfigOptions),
```

## Change

- Remove `contextCache` from `GeminiConfigSchema` (googleAI only; the vertexai config does not have this flag).
- Strip it in the request-builder destructure so a stale config (which still validates, given passthrough) no longer leaks it into the outbound request.
- Add a regression test asserting the serialized `generationConfig` does not contain it.

## Notes

- This is technically a breaking type change for a public config field, but setting it never had any effect.
- The Python plugin has a separate `contextCache` alias (`py/plugins/google-genai/.../gemini.py`) which reads as out of scope for this JS-only issue; happy to follow up there if useful.

## Verification

- `tsx --test tests/googleai/gemini_test.ts`: 34/34 pass (including the new regression test).
- `tsc --noEmit` clean for the plugin; `prettier --check` clean on the touched files.
- `grep contextCache` across `js/` returns only the intentional strip and its test.

### DIFF
--- a/js/plugins/google-genai/src/googleai/gemini.ts
+++ b/js/plugins/google-genai/src/googleai/gemini.ts
@@ -176,13 +176,6 @@ export const GeminiConfigSchema = GenerationCommonConfigSchema.extend({
     .union([z.boolean(), z.object({}).strict()])
     .describe('Enables the model to generate and run code.')
     .optional(),
-  contextCache: z
-    .boolean()
-    .describe(
-      'Context caching allows you to save and reuse precomputed input ' +
-        'tokens that you wish to use repeatedly.'
-    )
-    .optional(),
   functionCallingConfig: z
     .object({
       mode: z.enum(['MODE_UNSPECIFIED', 'AUTO', 'ANY', 'NONE']).optional(),
@@ -751,6 +744,10 @@ export function defineModel(
         tools: toolsFromConfig,
         retrievalConfig,
         serviceTier,
+        // The schema is passthrough, so also strip the removed contextCache
+        // flag; otherwise a stale config would still forward it to the API as
+        // an unknown generationConfig field.
+        contextCache: _contextCache,
         ...restOfConfigOptions
       } = requestOptions;
 
--- a/js/plugins/google-genai/tests/googleai/gemini_test.ts
+++ b/js/plugins/google-genai/tests/googleai/gemini_test.ts
@@ -449,6 +449,27 @@ describe('Google AI Gemini', () => {
         );
       });
 
+      it('strips the removed contextCache flag from the outbound request', async () => {
+        const model = defineModel('gemini-2.5-flash', defaultPluginOptions);
+        mockFetchResponse(defaultApiResponse);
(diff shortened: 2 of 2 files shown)


cabljac:
Closing this one. After discussion with the maintainers: `contextCache` is a public config field, and removing it now would be a breaking type change - so we'd rather not do that ahead of a major version.

The field is still a no-op (leftover from the legacy explicit context caching removed in #4948, superseded by implicit caching on 2.5+), and it'll be flagged for removal in V2 via a code comment. Leaving everything as-is for now.

PR #5698, opened 2026-07-07
fix(tools): restore usage stats for runAction, streamAction, etc

MichaelDoyle:
Currently we track analytics for tRPC. However, when we moved `runAction` to the express server for real time tracing, we lost our metrics. `streamAction` possibly never had metrics.

Add a new middleware that does the same metrics collection for any dev-ui routes directly defined in the express server

### DIFF
--- a/genkit-tools/common/src/server/server.ts
+++ b/genkit-tools/common/src/server/server.ts
@@ -24,8 +24,12 @@ import os from 'os';
 import path from 'path';
 import type { GenkitToolsError } from '../manager';
 import type { BaseRuntimeManager } from '../manager/manager';
-import { writeToolsInfoFile } from '../utils';
-import { logger } from '../utils/logger';
+import { logger, writeToolsInfoFile } from '../utils';
+import {
+  createToolsRequestEvent,
+  extractActionType,
+  recordRequestEvent,
+} from '../utils/analytics';
 import { toolsPackage } from '../utils/package';
 import { downloadAndExtractUiAssets } from '../utils/ui-assets';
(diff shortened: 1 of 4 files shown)


genkit-tools/common/src/server/server.ts
```diff
@@ -83,6 +83,63 @@ class PushableAsyncIterable<T> implements AsyncIterable<T> {
 
 const activeInputStreams = new Map<string, PushableAsyncIterable<any>>();
 
+const loggedExpressRoute = (routeName: string) => {
```
huangjeff5, line 90: nit: this looks almost the same as analyticsEventForRoute, may be worth seeing we can share this logic in a helper.

genkit-tools/common/tests/server_test.ts
```diff
@@ -162,4 +163,153 @@ describe('Tools Server', () => {
...
+          name: 'tools_request',
+          parameters: expect.objectContaining({
+            route: 'runAction',
+            action: '/util/generate',
+            status: 'success',
+          }),
+        })
+      );
+
+      await axios.post(`http://localhost:${port}/api/runAction`, {
```
huangjeff5, line 240: Should we add a test case for streamAction as well?

huangjeff5 (approved):

genkit-tools/common/tests/server_test.ts
MichaelDoyle: done

genkit-tools/common/src/server/server.ts
MichaelDoyle: good idea. thanks!

PR #5699, opened 2026-07-07
feat(py/plugins/amazon-bedrock): add model capability registry

cabljac:
Slice 2 of the Amazon Bedrock Python plugin train, into the `py-bedrock-plugin` base branch.

Ports the Go plugin's `models.go` ([aws-bedrock-go-plugin](https://github.com/genkit-ai/aws-bedrock-go-plugin)):

- **47-entry model capability registry** (Claude 3.x/4.x incl. provisioned-throughput variants, Nova, Cohere Command, Mistral/Pixtral, Jamba, Llama 3.x/4, DeepSeek R1, Writer, TwelveLabs) - verified key-for-key and flag-for-flag against the Go source.
- **Inference-profile prefix stripping** (`global. us-gov. us. eu. jp. apac. au.`) for capability lookup only; the full original model ID is always sent to Bedrock untouched.
- **`get_model_info()`**: known models are `stable` with registry capabilities; unknown chat/text models default to multimodal + tools at the `unstable` stage so newer or inference-profile-only models remain callable without a plugin release; image models advertise media output only; embedding models advertise nothing.

Tests: 12 new (prefix stripping incl. `us-gov.` vs `us.` ordering, registry lookups via profile IDs, unknown-model defaults, image/embedding shapes, registry size + base-ID invariants). `ruff` clean, 24/24 plugin tests pass.

Next slice: generate sync path (Converse).

### DIFF
--- a/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/model_info.py
+++ b/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/model_info.py
@@ -1,0 +1,186 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Model capability registry for the Amazon Bedrock plugin.
+
+Ported from the Go plugin's ``models.go``. Capabilities are keyed by base
+Bedrock model ID; cross-region inference-profile prefixes are stripped for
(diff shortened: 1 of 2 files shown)


cabljac:
Both review points addressed in 075d69db7. Went slightly broader than the suggestion: matching on `arn:` rather than `arn:aws:` so GovCloud (`arn:aws-us-gov:`) and China (`arn:aws-cn:`) partition ARNs also resolve, which matters given the registry handles the `us-gov.` profile prefix. Tests cover inference-profile ARNs, foundation-model ARNs, the GovCloud partition, and opaque application-inference-profile IDs falling back to unstable defaults. Worth noting the Go plugin doesn't reduce ARNs before lookup either, so this may be worth backporting.

cabljac:
Two things on the registry, both for a follow-up rather than an amend here. Does that work for you?

The registry stops at 4.6 (`model_info.py:65-66`). I checked against `list-inference-profiles`: `claude-opus-4-7`, `claude-opus-4-8`, `claude-sonnet-5`, `claude-opus-5` and `claude-fable-5` are all ACTIVE and missing. They still resolve through the unknown-model default, so the only effect is a `stage: unstable` badge. The registry was verified against the Go plugin's `models.go`, which has the same cutoff, so that is where the gap comes from.

nit: `model_info.py:151,171` sets `label=model_name`. Peers use `"<Provider> - <Model>"`, so the Dev UI picker shows bare IDs next to `Anthropic - ...` and `Ollama - ...` entries.

Tracking: #5820

PR #5708, opened 2026-07-08
fix(py/ollama): accumulate streaming content in final ModelResponse

### LINKED ISSUE #5700: [Ollama][Python] Streaming responses return empty `response.text` (final ModelResponse content is discarded)
### Describe the bug

In the Python Ollama plugin, streaming a generation sends text deltas via `ctx.send_chunk()` but discards the aggregated content from the final `ModelResponse`. As a result, `(await response).text` (and `response.message`) is empty even on a fully successful stream.

The offending code is in `OllamaModel.generate()` in `py/plugins/ollama/src/genkit/plugins/ollama/models.py`. After the streaming branch has emitted all chunks, the method explicitly resets the accumulated content to an empty list before building the final response:

```python
if self.is_streaming_request(ctx=ctx):
    content = []

response_message = Message(role=Role.MODEL, content=content)
# ...
return ModelResponse(
    message=Message(role=Role.MODEL, content=content),  # content == []
    usage=...,
)
```

The streaming helpers (`_chat_with_ollama` / `_generate_ollama_response`) send each chunk via `ctx.send_chunk(...)` and then return `None`, so nothing is ever accumulated into the final response.

### Impact

Any consumer using the documented streaming pattern gets an empty string for the final text:

```python
stream, response = ai.generate_stream(model='ollama/...', prompt='...')
async for chunk in stream:
    ...  # chunks arrive fine
final = await response
print(final.text)  # '' — even though the stream succeeded
```

Callers are forced to buffer the streamed chunks themselves to reconstruct the full text, which defeats the purpose of awaiting the final response.

### To Reproduce

1. Configure the Ollama plugin with any chat/generate model.
2. Call `ai.generate_stream(...)`.
3. Consume the async stream (chunks contain text as expected).
4. `await` the final response and read `.text`.

**Expected:** `.text` contains the full aggregated model output (matching the concatenation of the streamed chunks), consistent with non-streaming behavior and with the other language plugins.

**Actual:** `.text` is `''` and `message.content` is `[]`.

### Suggested fix

Accumulate the streamed parts (text / reasoning / tool requests) as chunks are produced and use the aggregated content for the final `ModelResponse` instead of resetting it to `[]`. The final message should be equivalent to the non-streaming response for the same request.

### Runtime

- Plugin: `genkit-plugin-ollama` (Python)
- File: `py/plugins/ollama/src/genkit/plugins/ollama/models.py` (`OllamaModel.generate`)

RM-RAMASAMY:
Fixes #5700

The Python Ollama plugin's streaming path discarded the aggregated content from the final `ModelResponse`. After streaming all chunks via `ctx.send_chunk()`, callers who awaited the final response got an empty `response.text` / `message.content == []`.

**Root cause:**

The streaming helpers (`_chat_with_ollama` / `_generate_ollama_response`) send each chunk via `ctx.send_chunk()` but return `None`, so no content is ever accumulated into the final response. Additionally, `generate()` explicitly resets `content = []` for streaming requests, ensuring the final `ModelResponse` is always empty.

**Fix:**

- `_chat_with_ollama`: accumulate `chunk.message.content` across streamed chunks and return the last `ChatResponse` with the full aggregated text.
- `_generate_ollama_response`: same pattern — accumulate `chunk.response` and return the last `GenerateResponse` with full text.
- `generate()`: remove the `content = []` reset for streaming requests so the aggregated content flows through to the final `ModelResponse`.

This aligns the Python Ollama plugin with the JS and Go implementations, which both return a fully populated final response after streaming.

Checklist:

### DIFF
--- a/py/packages/genkit-ollama/src/genkit_ollama/models.py
+++ b/py/packages/genkit-ollama/src/genkit_ollama/models.py
@@ -277,9 +277,6 @@ class OllamaModel:
         else:
             raise ValueError(f'Unresolved API type: {self.model_definition.api_type}')
 
-        if self.is_streaming_request(ctx=ctx):
-            content = []
-
         response_message = Message(
             role=Role.MODEL,
             content=content,
@@ -367,9 +364,13 @@ class OllamaModel:
                     **extra_kwargs,
                 )
                 idx = 0
+                accumulated_text = ''
+                last_chunk: ollama_api.ChatResponse | None = None
                 async for chunk in chat_response:
                     idx += 1
+                    last_chunk = chunk
                     role = self._from_ollama_role(chunk.message.role)
+                    accumulated_text += chunk.message.content or ''
                     if ctx:
                         ctx.send_chunk(
                             chunk=ModelResponseChunk(
@@ -378,9 +379,9 @@ class OllamaModel:
                                 content=self._build_multimodal_chat_response(chat_response=chunk),
                             )
                         )
-            # For streaming requests, we return None because the response chunks
-            # have already been sent via ctx.send_chunk() above. The async generator
-            # is now exhausted, and the caller should not expect a return value.
+            if last_chunk is not None:
+                last_chunk.message.content = accumulated_text
+                return last_chunk
             return None
         else:
             async with wrap_connection_errors(self._server_address):
@@ -433,8 +434,12 @@ class OllamaModel:
                     **extra_kwargs,
                 )
                 idx = 0
+                accumulated_text = ''
+                last_chunk: ollama_api.GenerateResponse | None = None
                 async for chunk in generate_response:
                     idx += 1
+                    last_chunk = chunk
+                    accumulated_text += chunk.response or ''
                     if ctx:
                         ctx.send_chunk(
                             chunk=ModelResponseChunk(
@@ -443,9 +448,9 @@ class OllamaModel:
                                 content=self._build_generate_response(generate_response=chunk),
                             )
                         )
-            # For streaming requests, we return None because the response chunks
-            # have already been sent via ctx.send_chunk() above. The async generator
-            # is now exhausted, and the caller should not expect a return value.
+            if last_chunk is not None:
+                last_chunk.response = accumulated_text
+                return last_chunk
             return None
         else:
             async with wrap_connection_errors(self._server_address):
--- a/py/packages/genkit-ollama/src/genkit_ollama/models.py
+++ b/py/packages/genkit-ollama/src/genkit_ollama/models.py
@@ -277,6 +277,9 @@ class OllamaModel:
         else:
             raise ValueError(f'Unresolved API type: {self.model_definition.api_type}')
 
+        if not api_response and self.is_streaming_request(ctx=ctx):
+            content = []
+
         response_message = Message(
             role=Role.MODEL,
             content=content,
@@ -313,7 +316,11 @@ class OllamaModel:
                 stored client factory when omitted.
 
         Returns:
-            The chat response from Ollama.
+            The chat response from Ollama. For streaming requests, returns the
+            last streamed chunk with ``message.content``, ``message.thinking``,
+            and ``message.tool_calls`` replaced by the values accumulated across
+            all chunks; other fields reflect the final chunk. Returns ``None``
+            if the stream yielded no chunks.
         """
         # Resolve media URLs first. build_chat_messages may perform HTTP fetches
         # for image parts, and those must stay *outside* wrap_connection_errors so
@@ -365,12 +372,17 @@ class OllamaModel:
                 )
                 idx = 0
                 accumulated_text = ''
+                accumulated_thinking = ''
+                accumulated_tool_calls: list[ollama_api.Message.ToolCall] = []
                 last_chunk: ollama_api.ChatResponse | None = None
                 async for chunk in chat_response:
                     idx += 1
                     last_chunk = chunk
                     role = self._from_ollama_role(chunk.message.role)
                     accumulated_text += chunk.message.content or ''
+                    accumulated_thinking += chunk.message.thinking or ''
+                    if chunk.message.tool_calls:
+                        accumulated_tool_calls.extend(chunk.message.tool_calls)
                     if ctx:
                         ctx.send_chunk(
                             chunk=ModelResponseChunk(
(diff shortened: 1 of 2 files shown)


py/packages/genkit-ollama/src/genkit_ollama/models.py
```diff
@@ -277,9 +277,6 @@ async def generate(
         else:
             raise ValueError(f'Unresolved API type: {self.model_definition.api_type}')
 
-        if self.is_streaming_request(ctx=ctx):
-            content = []
```
hilariie, line 281: Removing this means a stream that yields zero chunks now surfaces 'Failed to get response from Ollama API' as the final message text, where before, streaming callers got empty content. Probably acceptable, but it is a behavior change and that string reads like model output. Could we pin the intended behavior with a small test?

py/packages/genkit-ollama/src/genkit_ollama/models.py
```diff
@@ -367,9 +364,13 @@ async def _chat_with_ollama(
                     **extra_kwargs,
                 )
                 idx = 0
+                accumulated_text = ''
+                last_chunk: ollama_api.ChatResponse | None = None
                 async for chunk in chat_response:
                     idx += 1
+                    last_chunk = chunk
                     role = self._from_ollama_role(chunk.message.role)
+                    accumulated_text += chunk.message.content or ''
```
hilariie, line 382: We only accumulate message.content here, so for think-enabled models, the reasoning that streamed out as chunks never lands in the final response. last_chunk.message.thinking just holds the last delta, which is usually empty on the done chunk. The Go plugin merges all chunk parts including reasoning, so this isn't quite full Go parity yet. Can fix now or can be a follow up, but maybe leave a TODO here if you decide a follow-up?

py/packages/genkit-ollama/src/genkit_ollama/models.py
```diff
@@ -378,9 +379,9 @@ async def _chat_with_ollama(
                                 content=self._build_multimodal_chat_response(chat_response=chunk),
                             )
                         )
-            # For streaming requests, we return None because the response chunks
-            # have already been sent via ctx.send_chunk() above. The async generator
-            # is now exhausted, and the caller should not expect a return value.
+            if last_chunk is not None:
+                last_chunk.message.content = accumulated_text
+                return last_chunk
```
hilariie, line 398: Because we return the last chunk, tool_calls only survive if the model happens to emit them on the final chunk. A tool call arriving mid-stream is dropped from the final response, and the tool loop reads tool requests from the final response. JS has the same gap but you can fix it, then we open a follow-up JS issue.

py/packages/genkit-ollama/src/genkit_ollama/models.py
```diff
@@ -378,9 +379,9 @@ async def _chat_with_ollama(
...
             return None
```
hilariie, line 399: nit: with the old comments gone, the docstring is the only word on the return contract and it only describes the non streaming case. For streaming this is really the last chunk with its content swapped for the accumulated text, every other field is just whatever the final chunk had, and it can still be None if the stream yielded nothing. Worth a line in the Returns section here and in _generate_ollama_response.

py/packages/genkit-ollama/tests/models/ollama_models_test.py
```diff
@@ -427,10 +433,9 @@ async def mock_streaming_chunks() -> AsyncIterator[ollama_api.ChatResponse]:
 
         response = await self.ollama_model._chat_with_ollama(self.request, self.ctx)
 
-        # For streaming requests, the method returns None because response chunks
-        # are sent incrementally via ctx.send_chunk() rather than returned at the end.
-        # This is the expected behavior for streaming APIs.
-        self.assertIsNone(response)
+        self.assertIsNotNone(response)
+        assert response is not None
```
hilariie, line 474: nit: the bare `assert response is not None` already fails the test if it's None and also narrows the type for pyright, so the assertIsNotNone above it is redundant. Same pattern in test_streaming_generate_success below.

RM-RAMASAMY:
Addressed all review feedback in commit 2f9bf3f :

Accumulate thinking across chunks (Comment 2) Both _chat_with_ollama and _generate_ollama_response now accumulate thinking across all streamed chunks, matching Go plugin behavior. Set to None when no thinking content was received.

Accumulate tool_calls across chunks (Comment 3) _chat_with_ollama now accumulates tool_calls across all streamed chunks, so tool calls arriving mid-stream are not lost. This is chat-only — /api/generate does not support tools. Set to None when no tool calls were received.

Handle zero-chunks edge case (Comment 1) When the stream yields no chunks, generate() now returns an empty content list instead of falling through to the error-default message. Added two new tests (test_generate_chat_streaming_zero_chunks and test_generate_generate_streaming_zero_chunks) to cover this path.

Remove redundant assertion (Comment 5) Removed duplicate self.assertIsNotNone(response) — the assert response is not None on the next line already handles both runtime checking and type narrowing.

Update docstrings (Comment 4) Updated the Returns section in both _chat_with_ollama and _generate_ollama_response to document the streaming return contract: last chunk with accumulated fields replaced, other fields from the final chunk, None if no chunks arrived.

hilariie:
Thanks, all review comments are addressed, and the accumulation logic checks out end-to-end. One last thing before I approve: the thinking and tool_calls accumulation added in `2f9bf3f` has no test coverage. I mutation-tested this by deleting the accumulation assignments (or the swap-in on the last chunk), and the entire test suite remains green, so does replacing the helpers' empty-stream return None with a raise. So the exact regressions from my earlier comments could be reintroduced without CI noticing.

Could you add three small tests following the test_streaming_chat_success pattern:

- a multi-chunk stream with thinking set on non-final chunks, asserting the returned response carries the concatenated thinking (and that generate()'s final message gets the ReasoningPart);

- a chunk carrying tool_calls followed by an empty done chunk, asserting the tool call survives into the returned response;

- an empty stream through the real helper, asserting it returns None (pins the documented Returns contract).

With those in, LGTM.

RM-RAMASAMY:
Added the three tests in the Commit 6203d6a: 
Thinking accumulation across chunks, tool_calls surviving from a mid-stream chunk, and empty stream returning None through the real helper.

hilariie (approved):

RM-RAMASAMY:
Hi @hilariie,
Thanks for reviewing my contributions. I appreciate it!

PR #5718, opened 2026-07-09
feat(js/ai/agent): assemble agent messages from streamed chunks

pavelgj:
Add stream-to-message assembly helpers to reconstruct a turn's intermediate messages from streamed `modelChunk` frames. Server-managed agents only return the final response on the wire, so intermediate messages (tool requests/responses, reasoning) must be rebuilt from chunks bucketed by index, coalescing consecutive text-only and reasoning-only deltas.

Also switch `AgentResponseImpl` to a live messages getter so responses reflect the stream that may still be draining when first constructed.

### DIFF
--- a/js/ai/src/agent-core.ts
+++ b/js/ai/src/agent-core.ts
@@ -433,9 +433,11 @@ class MessageAssembler {
 
   /** Folds one streamed model chunk into the accumulating messages. */
   add(chunk: ModelChunk): void {
+    const hasIndex = chunk.index !== undefined;
     const index = chunk.index ?? this.lastIndex;
     this.lastIndex = index;
     let entry = this.byIndex.get(index);
+    const isNewEntry = !entry;
     if (!entry) {
       entry = {
         role: (chunk.role ?? 'model') as MessageData['role'],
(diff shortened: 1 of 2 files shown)


huangjeff5:
I think the assembler corrupts `chat.messages` for server-managed prompt agents on interrupt.

On interrupt, `agentFromPrompt` emits an extra frame after the model message already streamed — and it has no `index`:

```ts
// js/ai/src/agent.ts (~1560)
if (res.finishReason === 'interrupted') {
  const parts = res.message?.content?.filter((p) => !!p.toolRequest) || [];
  if (parts.length > 0) {
    sendChunk({ modelChunk: { role: 'tool', content: parts } });
  }
}
```

In `MessageAssembler.add`, a missing `index` falls back to `lastIndex` — the model message's bucket:

```ts
const index = chunk.index ?? this.lastIndex;
```

So that `role: 'tool'` frame lands on the model message and:

```ts
if (chunk.role) {
  entry.role = chunk.role;          // flips the model message to role 'tool'
}
// toolRequest isn't text/reasoning-only, so mergeParts appends it again:
entry.content = mergeParts(entry.content, content);  // duplicates the tool request
```

Net result — the last entry in `chat.messages` / `res.messages` becomes:

```jsonc
{ "role": "tool", "content": [ /* toolRequest */, /* toolRequest again */ ] }
```

instead of the stored:

```jsonc
{ "role": "model", "content": [ /* toolRequest w/ interrupt metadata */ ] }
```

I think the right fix is: Stream chunks to build intermediate messages, but append the final message directly.

(Note you also need to make sure that the streamed chunk missing an index doesn't swap the role to "tool". I think it does currently).

apascal07 (approved):

huangjeff5 (approved):

Issue #5635, opened 2026-06-26
[plugins] Anthropic Python: stable/beta API split skeleton (api_version + betas)


hilariie:
Parent: #5617 (Phase B, P1 - item 22)

## Why

The Python plugin has no notion of API surface (stable vs. beta). JS exposes a per-request `api_version` plus a plugin-wide default, and attaches the `betas` request parameter only on beta calls. Without this skeleton there is nowhere for the later beta-only features to hang — effort `max`, `task_budget`, Files API media sources, and citations all require a beta surface.

This issue is the enabling skeleton, not the beta features themselves; landing it unblocks the Phase C plumbing.

## What needs to be done

- Introduce stable/beta selection: a plugin-level default and a per-request `api_version`, resolved per request in the order per-request override → plugin default → stable (mirroring the JS resolution).
- Route beta requests through the beta surface and attach the `betas` request parameter only on beta calls; leave stable behavior unchanged.

## Acceptance

- `api_version` works both per-request and as a plugin default, with resolution order matching JS.
- Beta calls go through the beta surface with `betas`; stable calls are unchanged.
- Tests cover stable/beta resolution.

## Reference

- JS target: `js/plugins/anthropic/src/types.ts:35-39, 134-140, 209-218` (`PluginOptions.apiVersion`, per-request `apiVersion`, `resolveBetaEnabled`).

hilariie:
Partially addressed by #5720, which wires the per-request `api_version` + `betas` skeleton (typed `AnthropicConfig`, beta-surface routing, `betas` attached only on beta calls). Two intentional divergences from this issue's "mirror the JS resolution" spec, flagging so they don't get "corrected" back to JS later:

1. `betas` set without `api_version` selects the beta surface. JS silently drops `betas` on stable here, which reads as a bug (features quietly discarded) rather than behavior worth copying, see #5737, so Python honors the `betas` and routes to beta.
2. `api_version: 'stable'` + `betas` raises a `ValidationError` instead of resolving by priority. Since the Python plugin validates eagerly, it rejects the contradictory combination up front rather than silently overriding the explicit `stable` choice.

Both keep the core guarantee (`betas` only sent on beta calls, stable calls unchanged); they just differ on those two edge cases. Can the spec be updated to note them?

Still outstanding for this issue, assumed to land with the PR that closes it:
- Plugin-level `api_version` default (#5720 only adds the per-request override).
- Full resolution order: per-request,  plugin default, stable. Currently only per-request and stable.

PR #5720, opened 2026-07-10
feat(py/plugins/anthropic): add typed AnthropicConfig schema as customOptions

### LINKED ISSUE #5627: [plugins] Anthropic Python: add typed AnthropicConfig schema as customOptions
Parent: #5617 (Phase A, P0: item 33)

## Why

Python advertises a generic `ModelConfig` as `customOptions`, so every Anthropic-specific option (thinking modes, `output_config.effort`, `tool_choice`, `api_version`, `betas`, and `metadata`) is invisible to reflection, the dev UI, and type hints. Today a caller can only discover these by reading the source, and the values are not validated before they reach the API.

Beyond discoverability, a typed config is the scaffold that the rest of the config-driven parity work hangs off (thinking modes, effort, the stable/beta split). Defining it once, up front, means those fields are typed a single time instead of being re-typed by each later change, which is why the epic sequences it before the thinking request-side work (item 11, #5629 ).

## What needs to be done

- Provide a typed Anthropic config that surfaces and validates the Anthropic-specific options: `thinking`, `output_config` (including `effort` and `task_budget`), `tool_choice`, `metadata`, `api_version`, and `betas`, modelled on the JS config and thinking schemas.
- Advertise it as the `customOptions` for every Anthropic model so the options are discoverable and typed in reflection and the dev UI.
- Preserve backward compatibility for callers that currently pass plain-dict configs.

## Acceptance

- Anthropic-specific options are discoverable via reflection / dev UI and are typed and validated.
- The typed config is advertised as `customOptions` for all Anthropic models.
- Existing dict-based configs continue to work; tests cover the schema and its validation rules (e.g. `enabled` requires a budget; `enabled` and `adaptive` are mutually exclusive).

## Reference

- Current `customOptions` wiring: `py/plugins/anthropic/src/genkit/plugins/anthropic/plugin.py:126` and `plugin.py:144`.
- JS target: `js/plugins/anthropic/src/types.ts:68-159` (`ThinkingConfigSchema`, `AnthropicConfigSchema`).

hilariie:
Closes #5627

The Python plugin advertised the generic `ModelConfig` as `customOptions`, so none of the Anthropic-specific options showed up in reflection, or the Dev UI, and nothing validated them before they hit the API. This adds a typed `AnthropicConfig` modelled on the JS schemas (`AnthropicConfigSchema` and `ThinkingConfigSchema` in `js/plugins/anthropic/src/types.ts`) and advertises it for every Anthropic model, including the Vertex Model Garden ones.

What the config covers:

- `thinking`, with the same rules as JS: `enabled` and `adaptive` are mutually exclusive, and `budgetTokens` is required (an integer >= 1024) when thinking is enabled
- `output_config` with `effort` and `task_budget`
- `tool_choice`: `auto`, `any`, `tool`, plus `none`, which JS doesn't have yet but the API supports
- `metadata`, `apiVersion`, `betas`, and the per-request `apiKey` / `version` overrides inherited from the common config

Request building now actually consumes these instead of dropping them. Thinking configs are translated to the SDK shape, `version` overrides the model id, `apiVersion: 'beta'` routes the request through `client.beta.messages`, `apiKey` swaps in a request-scoped client via `client.copy()`, and a user-supplied `output_config` merges with the structured-output format instead of being clobbered by it. Unknown keys still pass through (`extra='allow'`) and are routed via the SDK's `extra_body` escape hatch, so new API options keep working without a plugin release.

```python
from genkit_anthropic import AnthropicConfig, ThinkingConfig

response = await ai.generate(
    model='anthropic/claude-sonnet-4-6',
    prompt='Why is the sky blue?',
    config=AnthropicConfig(
        max_output_tokens=2048,
        thinking=ThinkingConfig(enabled=True, budget_tokens=1024),
    ),
)
```

Plain-dict configs keep working, they just get validated now. That is a small behavior change worth calling out: a dict like `{'thinking': {'enabled': True}}` with no budget used to be forwarded and fail at the API, and now raises a `ValidationError` before the request is sent.

One deliberate divergence from JS: setting `betas` without `apiVersion` selects the beta surface here. JS silently drops the betas on the stable API in that case, which looks like a bug rather than behavior worth copying, so the Python plugin honors them instead.

## Screenshots
Screenshot 2026-07-10 at 15 32 52

## Tests
Tested with unit tests for the validation rules, alias handling, SDK param translation, beta routing, and the advertised JSON schema (104 tests in the plugin plus the Model Garden suite), and verified manually in the Dev UI as shown above.

---
**Thinking validation note:** Python intentionally validates bare `budgetTokens` / `budget_tokens` according to the payload it will send to Anthropic. When a budget is provided without `enabled`, the plugin normalizes it to `{ type: "enabled", budget_tokens: ... }` before calling the SDK, so Python also applies the integer budget rule and rejects fractional values such as `2048.5`. Same for a bare `budgetTokens` with no `enabled` set: it now turns thinking on to match JS, but if the budget is bigger than `max_tokens` (default 4096) the API 400s, nothing guards that yet. Applies through Vertex Model Garden too.

JS currently applies its integer check only when `enabled` is explicit. That may be preserving a looser historical shorthand, or limiting cross-field validation to explicit thinking modes, so this PR does not assume the JS behavior is wrong. The Python behavior is stricter for this edge case because it matches the normalized SDK request shape.

### DIFF
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/config.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/config.py
@@ -191,7 +191,7 @@ class ThinkingConfig(BaseModel):
         if (
             (enabled or budget_implies_enabled)
             and self.budget_tokens is not None
-            and not self.budget_tokens.is_integer()
+            and not float(self.budget_tokens).is_integer()
         ):
             raise ValueError('budgetTokens must be an integer when thinking is enabled')
         return self
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/config.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/config.py
@@ -21,16 +21,28 @@ Extends the shared :class:`ModelConfig` (``version``, ``temperature``,
 
 Unknown keys pass through (``extra='allow'``). Top-level Genkit fields
 accept their usual camelCase aliases, while Anthropic-specific nested keys
-match the Anthropic/JavaScript plugin shape field-by-field.
+match the Anthropic plugin shape field-by-field.
 """
 
 from typing import Annotated, ClassVar, Literal, cast
 
+from anthropic.types.beta.message_create_params import MessageCreateParamsBase as BetaMessageCreateParamsBase
+from anthropic.types.message_create_params import MessageCreateParamsBase
 from pydantic import BaseModel, ConfigDict, Field, WithJsonSchema, model_validator
 from pydantic.alias_generators import to_camel
 from pydantic.config import JsonDict
 
 from genkit import ModelConfig
+
+_STABLE_BODY_KEYS = frozenset(MessageCreateParamsBase.__annotations__)
+_BETA_BODY_KEYS = frozenset(BetaMessageCreateParamsBase.__annotations__)
+
+# Accepted by create() alongside the body fields; `stream` is excluded because Genkit owns streaming.
+_REQUEST_KWARG_KEYS = frozenset({'extra_body', 'extra_headers', 'extra_query', 'timeout'})
+
+STABLE_KWARG_KEYS = _STABLE_BODY_KEYS | _REQUEST_KWARG_KEYS
+BETA_KWARG_KEYS = _BETA_BODY_KEYS | _REQUEST_KWARG_KEYS
+BETA_ONLY_KEYS = _BETA_BODY_KEYS - _STABLE_BODY_KEYS
 
 _NESTED_CONFIG = ConfigDict(extra='allow', populate_by_name=True)
 
@@ -163,14 +175,14 @@ def _anthropic_config_schema_extra(schema: JsonDict) -> None:
 class ThinkingConfig(BaseModel):
     """Extended-thinking configuration.
 
-    ``enabled`` and ``adaptive`` are mutually exclusive, and ``budgetTokens``
-    is required when ``enabled`` is true.
+    ``enabled``, ``adaptive`` and ``disabled`` are mutually exclusive, and
+    ``budgetTokens`` is required when ``enabled`` is true.
     """
 
     model_config = _NESTED_CONFIG
 
     enabled: bool | None = None
-    # float, not int: adaptive mode allows a fractional budget it ignores; integers enforced only when enabled.
+    # Adaptive mode allows a fractional budget it ignores; integers enforced only when enabled.
     budget_tokens: float | None = Field(default=None, alias='budgetTokens', ge=1024)
     adaptive: bool | None = None
     display: Literal['summarized', 'omitted'] | None = None
@@ -187,6 +199,8 @@ class ThinkingConfig(BaseModel):
 
         if enabled and adaptive:
             raise ValueError('Cannot use both enabled and adaptive thinking modes simultaneously')
+        if disabled and (enabled or adaptive):
+            raise ValueError('Cannot disable thinking and request an enabled or adaptive thinking mode simultaneously')
         if enabled and self.budget_tokens is None:
             raise ValueError('budgetTokens is required when thinking is enabled')
         if (
@@ -302,9 +316,26 @@ class AnthropicConfig(ModelConfig):
         description='Anthropic beta feature headers to enable for this request.',
     )
 
+    def beta_only_fields(self) -> set[str]:
+        """Return the names of beta-only request fields set on this config."""
+        present = {
+            name
+            for name, value in (self.__pydantic_extra__ or {}).items()
+            if name in BETA_ONLY_KEYS and value is not None
+        }
+        if self.betas:
+            present.add('betas')
+        if self.output_config is not None and self.output_config.task_budget is not None:
+            present.add('output_config.task_budget')
+        return present
+
     @model_validator(mode='after')
     def _check_api_surface(self) -> 'AnthropicConfig':
-        """Reject betas on the stable surface so an explicit apiVersion is never silently overridden."""
-        if self.api_version == 'stable' and self.betas:
-            raise ValueError("betas require the beta API surface; remove betas or set apiVersion to 'beta'")
+        """Reject beta-only fields on the stable surface so an explicit apiVersion is never silently overridden."""
+        if self.api_version != 'stable':
+            return self
+        beta_only = self.beta_only_fields()
+        if beta_only:
+            names = ', '.join(sorted(beta_only))
+            raise ValueError(f"{names} require the beta API surface; remove them or set apiVersion to 'beta'")
         return self
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/config.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/config.py
@@ -1,0 +1,302 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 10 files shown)


py/packages/genkit-anthropic/src/genkit_anthropic/config.py
```diff
@@ -0,0 +1,302 @@
...
+            raise ValueError('Cannot use both enabled and adaptive thinking modes simultaneously')
+        if enabled and self.budget_tokens is None:
+            raise ValueError('budgetTokens is required when thinking is enabled')
+        if (
+            (enabled or budget_implies_enabled)
+            and self.budget_tokens is not None
+            and not self.budget_tokens.is_integer()
+        ):
+            raise ValueError('budgetTokens must be an integer when thinking is enabled')
+        return self
```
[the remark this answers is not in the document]
hilariie: sticking with `float | None` since it's mirroring the JS plugin.

In `ThinkingConfigSchema` (`js/plugins/anthropic/src/types.ts`, L68-L101), `budgetTokens` is `z.number().min(1024)`, not `z.number().int()`, and the integer check only runs inside `superRefine` when `enabled` is true. So JS accepts a fractional `budgetTokens` when it's ignored (like adaptive mode), which is what `test_thinking_adaptive_allows_fractional_ignored_budget` is covering.

Switching to `int | None` would make Pydantic reject fractional values in every mode, which diverges from JS. It'd also put the model out of sync with the advertised schema: `_THINKING_SCHEMA` declares `budgetTokens` as `type: number` to match `z.number()`, so I'd have to flip that to `integer` as well and drift from the JS-advertised shape. Keeping the model and the advertised schema aligned is the main thing I'm after here, so I'd rather leave it.

py/packages/genkit-anthropic/src/genkit_anthropic/config.py
```diff
@@ -0,0 +1,302 @@
...
+
+    Extends the shared :class:`ModelConfig` with Anthropic-specific options.
+    JSON keys stay snake_case for ``tool_choice`` and ``output_config`` and
+    camelCase elsewhere (``apiVersion``, inherited ``maxOutputTokens``).
+    """
+
+    model_config = ConfigDict(
+        alias_generator=to_camel,
+        extra='allow',
+        json_schema_extra=_anthropic_config_schema_extra,
```
cabljac, line 290: The effort enum is missing `max`, which the API supports (GA on Opus 4.6+). Before this PR `{'output_config': {'effort': 'max'}}` passed through the untyped config and worked, now it fails validation before the request is sent, so this is a small regression for anyone already using it. JS has the same gap (`js/plugins/anthropic/src/types.ts:146`) so it's a faithful port, but since we validate eagerly here the cost is higher on the Python side. Can you add `max` to the Literal and the advertised schema (plus a test)? I'll flag the JS enum separately.

Related nit while you're in there: `_THINKING_SCHEMA` advertises `budgetTokens` as `type: number` but the validator rejects fractional values when thinking is enabled. Worth advertising `integer` so the Dev UI matches runtime behaviour.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
```diff
@@ -183,52 +297,65 @@ def _build_usage(self, response: AnthropicMessage, basic_usage: ModelUsage) -> M
...
+        # copy() keeps every other client setting and shares the pooled HTTP transport.
+        return self.client.copy(api_key=config.api_key)
+
+    def _uses_beta_api(self, config: AnthropicConfig) -> bool:
+        """Whether this request should use the Anthropic beta API surface.
+
+        Unlike the JS plugin, setting ``betas`` alone selects the beta surface
+        so the requested features are honored instead of silently dropped.
+        """
+        return config.api_version == 'beta' or bool(config.betas)
```
cabljac: I agree with the divergence here, silently dropping betas on stable reads like a JS bug rather than behaviour worth copying (filed #5737 for the JS side). One gap though: with an explicit `api_version: 'stable'` plus `betas` set, betas win and the request goes to the beta surface, which silently overrides an explicit user choice, the same failure mode we're diverging to avoid. JS documents per-request `apiVersion` as the highest-priority signal, so this also inverts that ordering.

Since the PR already rejects invalid thinking configs before the request is sent, I think a validator raising on the `stable` + `betas` conflict is the consistent move:

```python
@model_validator(mode='after')
def _check_api_surface(self) -> 'AnthropicConfig':
    if self.api_version == 'stable' and self.betas:
        raise ValueError("betas require the beta API surface; remove betas or set apiVersion to 'beta'")
    return self
```

Can you also add a test for the conflict case? I couldn't see one covering `stable` + `betas`.

cabljac (changes requested):
Overall this is in good shape, the config surface matches the JS shapes closely and the test coverage on validation rules is solid. Two things I'd like resolved before merge (inline comments), plus one coordination point:

This PR partially implements the stable/beta split that #5635 scopes as its own issue, and diverges from that issue's written spec ("mirror the JS resolution", "attach `betas` only on beta calls"). I think the divergence is the right call (see #5737), but can you drop a comment on #5635 so the spec gets updated rather than someone later "fixing" this back to the JS behaviour? #5635's plugin-level `api_version` default is still missing here, which I'm assuming lands with that issue's PR, fine to defer if so.

On the gemini-code-assist suggestion (`budget_tokens` as `int`): I'm OK with `float` + the explicit integer checks given the documented adaptive/fractional edge case, but a short code comment explaining why it isn't just `int` would save the next reader the same question.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
hilariie: resolved in [b5fe38c](https://github.com/genkit-ai/genkit/pull/5720/commits/b5fe38c5e8ba7f095cd1cece91f5ef6116a8bc28)

py/packages/genkit-anthropic/src/genkit_anthropic/config.py
hilariie: Resolved in [b5fe38c](https://github.com/genkit-ai/genkit/pull/5720/commits/b5fe38c5e8ba7f095cd1cece91f5ef6116a8bc28)

hilariie:
Addressed all review comments in [b5fe38c](https://github.com/genkit-ai/genkit/pull/5720/commits/b5fe38c5e8ba7f095cd1cece91f5ef6116a8bc28) and left a comment on #5635

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
```diff
@@ -95,6 +105,77 @@ def _to_tool_input_schema(schema: dict[str, Any] | None) -> dict[str, Any]:
...
+        result['budget_tokens'] = int(budget_tokens)
+        return result
+
+    if disabled:
+        result['type'] = 'disabled'
+        return result
+
+    if thinking_type is not None:
+        result['type'] = thinking_type
+    return result or None
```
cabljac: One edge case I think is worth checking: `{'thinking': {'display': 'summarized'}}` passes validation (no mode set isn't an error), so this returns `{'display': 'summarized'}` and we'd send a `thinking` object with no `type`. From what I can see the API rejects thinking blocks without a `type`, and JS's `toAnthropicThinkingConfig` returns `undefined` for this input (drops thinking entirely), but worth verifying both. If that's right, I think we should match JS, something like:

```python
if 'type' not in result:
    return None
```

Can you check what the API actually does with a display-only thinking config and what JS does here, and align before merge?

cabljac (approved):
Approving, all my earlier asks look addressed (max effort in the Literal and advertised schema, budgetTokens as integer in the schema, stable + betas rejection broadened to all beta-only fields, and the float comment). Nice work on the extra test coverage.

One small thing flagged inline on `_to_anthropic_thinking_config`, I believe it's a one-liner. Can you sort that before merging.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
hilariie: I've resolved in `6fe5a19`. Also made edits to test files to reflect this change.

Yeah you are right. In JS, `toAnthropicThinkingConfig` in runner/base.ts falls through every
branch and returns undefined for a display-only config. It also only ever
attaches display on the adaptive branch, so it can't emit a typeless thinking
object at all.

As for the API, I did a live request for this and it errors as expected.

Issue #5617, opened 2026-06-25
[plugins] port Anthropic to Python


cabljac:
Port of the **Anthropic** plugin to Python.

This is a feature lift to get the python plugin at parity with the JS one.

We will use this issue as a parent issue to track overall progress and feature parity.
---

## Anthropic plugin: Python parity with JS and Go

**Scope: Python only.** This issue lifts the Python plugin up to the JS standard. No JS or Go changes are in scope — the JS and Go columns in the matrix below are the *reference target* for what Python should match, not a work list. We are not adding new features to JS/Go; any JS/Go observations noted here (e.g. the beta `cacheControl()` no-op) are FYI only and must be tracked in their own separate issues.

Reference policy: match JS where possible, prefer Go only where its implementation is genuinely better. Limited to the direct Anthropic API (Vertex model-garden and Bedrock excluded).

Versions surveyed: JS `@genkit-ai/anthropic` 0.3.0 (`@anthropic-ai/sdk ^0.97.1`), Go `go/plugins/anthropic` (SDK v1.23.0), Python `genkit-plugin-anthropic` 0.7.0 (`anthropic>=0.40.0`), including PR #5592 which registers `claude-opus-4-7` and `claude-opus-4-8`. API docs rechecked 2026-06-25.

### Summary

Python already covers the core surface: text, streaming (text and tool deltas), tool calling, system prompts, images, PDF/text documents, prompt caching (including cache token usage), a basic `thinking` request config, structured output via the current stable `output_config.format` shape, a system-prompt JSON fallback, lazy model resolution, and loop-local async clients.

The gaps fall into five areas, in rough order of impact:

1. **Extended thinking is not surfaced (P0).** Python can request basic thinking, but reasoning is never returned to the caller, streamed, or round-tripped. Both JS and Go do this.
[...]

PR #5724, opened 2026-07-13
feat(py/plugins/anthropic): surface extended thinking (extract, stream, signature round-trip)

### LINKED ISSUE #5628: [plugins] Anthropic Python: surface extended thinking (extract + stream + signature)
Parent: #5617 (Phase A, P0: items 12, 13, 14)

## Why

Extended thinking can be requested, but the reasoning is never given back to the caller. Reasoning blocks are dropped when converting the response (only `text` and `tool_use` are mapped) and are never streamed (only text and tool-input deltas are handled). Both JS and Go return reasoning, stream it, and round-trip its signature.

This is the single highest-impact gap in the epic: without it, callers cannot read a model's reasoning, cannot show it as it streams, and cannot continue a reasoning turn. Preserving the signature matters specifically because Anthropic requires it when reasoning is sent back on a later turn; drop it and multi-turn thinking breaks.

This issue covers the response side only; the request-side config modes are tracked separately (item 11, #5629 ). It is independent of the typed config schema, making it the lowest-risk thinking step and immediately testable.

## What needs to be done

- Return thinking to callers as reasoning content, both non-streaming and streaming, and stream thinking deltas as they arrive.
- Preserve the Anthropic reasoning signature so it can be sent back on subsequent turns. Use JS naming (`metadata.thoughtSignature`) and accept the Go-style `metadata.signature` as an input alias.
- Capture redacted-thinking blocks so they can be round-tripped (full redacted-thinking support is a separate P2 item, 15).

## Acceptance

- Reasoning is returned as reasoning parts (non-stream and stream), and thinking deltas are streamed.
- Signatures are preserved and round-trip back to the API on a follow-up turn.
- Tests cover extraction, streaming, and signature round-trip.

## Reference

- Current response conversion (extract reasoning here): `py/plugins/anthropic/src/genkit/plugins/anthropic/models.py:421-439`; streaming `models.py:278-351`.
- Outbound encoding (round-trip the signature here): `_to_anthropic_block` / `_to_anthropic_messages`, `models.py:365-419` — currently has no `ReasoningPart`→`thinking` branch, so a preserved signature has nowhere to go on a follow-up turn. Signature round-trip is only meaningfully testable live (opaque server value), not in unit tests.
- JS/Go return, stream, and round-trip reasoning today.

hilariie:
Closes #5628 (part of the #5617 epic)

## What

Returns extended thinking to callers in the Python Anthropic plugin. Thinking blocks now come back as reasoning parts in both streaming and non-streaming responses, thinking deltas stream as they arrive, and the reasoning signature is preserved so a reasoning turn can be continued later. Redacted thinking blocks are captured and round-tripped too.

## Why

Extended thinking could already be requested via config, but the reasoning never made it back out. The response converter only mapped `text` and `tool_use` blocks, and the stream handler only surfaced text and tool input deltas. So callers could not read the model's reasoning, could not render it while it streamed, and could not continue a reasoning conversation, because the signature (which Anthropic requires when thinking is sent back on a later turn) was dropped. JS and Go already do all of this; this brings Python in line with them.

## How

- `_to_genkit_content` maps `thinking` blocks to `ReasoningPart`, storing the signature under `metadata.thoughtSignature` (JS naming), and maps `redacted_thinking` blocks to `CustomPart` with `custom.redactedThinking`, the same shape JS uses.
- `_generate_streaming` emits reasoning chunks for `thinking_delta` events, and a custom chunk for `redacted_thinking` blocks (those arrive complete in `content_block_start`, no deltas follow). `signature_delta` is intentionally not streamed, matching JS; the signature is recovered from the final message instead.
- `_to_anthropic_block` gets the missing outbound branch: reasoning parts encode as `thinking` blocks with their signature, and raise a clear error if the signature is missing (same error message as JS). `get_thinking_signature` reads `metadata.thoughtSignature` first and falls back to the Go-style `metadata.signature` alias, decoding bytes if needed.
- The outbound check is attribute-based (`part.reasoning`) rather than isinstance. Reasoning parts parsed from JSON resolve to `DataPart` in the core `Part` union, so an isinstance check would silently drop them and break round-trips for Dev UI traffic, dict-built messages, and history produced by Go or JS.
- `cache_control` metadata is no longer attached to `thinking` or `redacted_thinking` blocks. The API rejects it there, and JS never attaches it to those block types.

## Tests

- New unit tests cover extraction (with and without a signature), redacted thinking capture, streaming (thinking deltas, `signature_delta` ignored but present on the final message, redacted chunk emission), outbound encoding, the Go alias (string and bytes), the missing-signature error, empty reasoning parts, JSON-deserialized round-trips for both reasoning and redacted parts, and the `cache_control` exclusion.
- All 90 package tests pass: `uv run pytest packages/genkit-anthropic/tests/`
- The signature is an opaque server value, so the true round-trip can only be verified live. The anthropic sample now has a `thinking_tool_round_trip` flow that enables thinking, streams the reasoning, and completes a tool call in a multi-turn conversation, which exercises the signature round-trip end to end from the Dev UI.

## Screenshots
Screenshot 2026-07-10 at 20 05 38

### DIFF
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
@@ -33,6 +33,7 @@ from anthropic.types import Message as AnthropicMessage
 
 from genkit import (
     Constrained,
+    CustomPart,
     FinishReason,
     MediaPart,
     Message,
@@ -41,6 +42,7 @@ from genkit import (
     ModelResponseChunk,
     ModelUsage,
     Part,
+    ReasoningPart,
     Role,
     TextPart,
     ToolRequest,
@@ -54,6 +56,8 @@ from genkit_anthropic.model_info import get_model_info
 from genkit_anthropic.utils import (
     build_cache_usage,
     get_cache_control,
+    get_redacted_thinking_data,
+    get_thinking_signature,
     maybe_strip_fences,
     to_anthropic_media,
 )
@@ -436,8 +440,9 @@ class AnthropicModel:
     ) -> AnthropicMessage:
         """Handle streaming generation.
 
-        Processes Anthropic streaming events including text deltas and
-        tool-use blocks.  Tool-use blocks arrive as:
+        Processes Anthropic streaming events including text deltas,
+        thinking deltas, redacted thinking blocks, and tool-use blocks.
+        Tool-use blocks arrive as:
 
         1. ``content_block_start`` with ``content_block.type == 'tool_use'``
         2. Zero or more ``content_block_delta`` with ``delta.type == 'input_json_delta'``
@@ -465,6 +470,15 @@ class AnthropicModel:
                                 'name': getattr(block, 'name', ''),
                                 'input_json': '',
                             }
+                    elif getattr(block, 'type', None) == 'redacted_thinking' and hasattr(block, 'data'):
+                        # Redacted thinking arrives complete in the start event; no deltas follow.
+                        ctx.send_chunk(
+                            ModelResponseChunk(
+                                role=Role.MODEL,
+                                index=0,
+                                content=[Part(root=CustomPart(custom={'redactedThinking': block.data}))],  # pyright: ignore[reportAttributeAccessIssue]
+                            )
+                        )
 
                 elif chunk.type == 'content_block_delta' and hasattr(chunk, 'delta'):
                     delta = chunk.delta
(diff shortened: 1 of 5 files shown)


cabljac (approved):
LGTM. Verified the two subtle bits against JS: the attribute-based `part.reasoning` check (JSON-parsed reasoning parts do resolve to `DataPart`, confirmed empirically via `Part.model_validate`, and `stable.ts:100` uses the same truthy check), and chunk `index=0` matching JS's candidate-index semantics in `runner/base.ts`. 90 package tests pass locally.

One test I'd like before merge: a streaming unit test covering thinking -> tool_use interleave (thinking block at index 0, tool_use at index 1 in the same turn). That's the exact path the signature round-trip exists for, and right now it's only exercised by the live sample.

Heads up, not blocking: this and #5720 both rewrite `_generate_streaming` (#5720 threads `client`/`use_beta` through it, this PR rewrites the chunk handling), so whichever lands second will need a careful rebase rather than a mechanical one.

hilariie:
added `test_streaming_thinking_then_tool_use_interleave` to resolve test review comment. Rebased, and changed sample code to use opus 4.8 since #5720's merged changes now allows adaptive thinking.

## Dev UI Screenshot of Opus 4.8

Screenshot 2026-07-17 at 16 48 14

PR #5726, opened 2026-07-13
fix(py/plugins/anthropic): map API errors and Retry-After

### LINKED ISSUE #5630: [plugins] Anthropic Python: map APIError to Genkit status + Retry-After
Parent: #5617 (Phase B, P1: item 21)

## Why

Anthropic API errors surface raw: there is no translation into a Genkit status and no `Retry-After` handling. As a result, callers cannot distinguish rate limiting from authentication failures from bad requests, and clients cannot back off correctly on 429/503/529. JS already maps these.

This is a small, self-contained reliability fix with no ordering constraints relative to the tool-schema default (item 28, #5631 ).

## What needs to be done

- Translate Anthropic API errors into the appropriate Genkit status, matching the JS mapping:
  - rate limited → resource-exhausted
  - unauthorized → unauthenticated
  - forbidden → permission-denied
  - bad request → invalid-argument
  - server error → internal
  - service unavailable / overloaded → unavailable
  - anything else → unknown
- Parse the `Retry-After` header (both delay-seconds and HTTP-date forms) and surface it on the error so clients can back off appropriately.

## Acceptance

- Anthropic failures raise Genkit errors with the correct mapped status.
- `Retry-After` is parsed and surfaced on rate-limit / unavailable responses.
- Tests cover the status mapping and Retry-After parsing.

## Reference

- Current calls have no error mapping: `py/plugins/anthropic/src/genkit/plugins/anthropic/models.py:111-163` and `models.py:278-351`.
- JS target: `js/plugins/anthropic/src/models.ts:50-216` (status switch + `parseRetryAfterMs`).

hilariie:
## Summary

Fixes #5630.

This brings Python Anthropic error handling in line with the existing JS implementation.

The change:

- Maps Anthropic API errors to the equivalent Genkit status codes
- Handles errors from both streaming and non-streaming requests
- Parses numeric and HTTP-date `Retry-After` values
- Exposes retry metadata on `GenkitError` for in-process use
- Keeps response metadata out of callable and reflection serialization
- Updates retry middleware to treat the provider delay as a minimum

## Testing

- Added Anthropic error handling and `Retry-After` coverage
- Added core error serialisation coverage
- Added retry middleware delay precedence coverage
- 113 targeted tests passing

### DIFF
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
@@ -25,6 +25,7 @@ See:
 """
 
 import json
+import math
 import time
 from email.utils import parsedate_to_datetime
 from typing import Any, Protocol, cast
@@ -106,15 +107,17 @@ def _parse_retry_after_ms(value: str) -> float | None:
     except ValueError:
         pass
     else:
-        if seconds >= 0:
-            return seconds * 1000
+        # Check the scaled value: a large finite input can overflow to inf.
+        retry_after_ms = seconds * 1000
+        if seconds >= 0 and math.isfinite(retry_after_ms):
+            return retry_after_ms
 
     try:
         retry_at = parsedate_to_datetime(value)
         if retry_at is None:
             return None
         retry_at_ms = retry_at.timestamp() * 1000
-    except (OverflowError, TypeError, ValueError):
+    except (OSError, OverflowError, TypeError, ValueError):
         return None
     return max(0.0, retry_at_ms - time.time() * 1000)
 
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
@@ -25,16 +25,20 @@ See:
 """
 
 import json
+import time
+from email.utils import parsedate_to_datetime
 from typing import Any, Protocol, cast
 
 import structlog
-from anthropic import AsyncAnthropic
+from anthropic import APIError, AsyncAnthropic
 from anthropic.types import Message as AnthropicMessage
 
 from genkit import (
     Constrained,
     CustomPart,
+    ErrorResponseMetadata,
     FinishReason,
+    GenkitError,
     MediaPart,
     Message,
     ModelRequest,
@@ -50,7 +54,7 @@ from genkit import (
     ToolResponsePart,
 )
 from genkit.model import get_basic_usage_stats
-from genkit.plugin_api import ActionRunContext
+from genkit.plugin_api import ActionRunContext, StatusName
 from genkit_anthropic.config import BETA_KWARG_KEYS, STABLE_KWARG_KEYS, AnthropicConfig
 from genkit_anthropic.model_info import get_model_info
 from genkit_anthropic.utils import (
@@ -74,6 +78,64 @@ class _ModelDumpable(Protocol):
     def model_dump(self, *, exclude_none: bool = False, by_alias: bool = False) -> dict[str, object]:
         """Dump model fields."""
         ...
+
+
+_ANTHROPIC_STATUS_MAP: dict[int, StatusName] = {
+    400: 'INVALID_ARGUMENT',
+    401: 'UNAUTHENTICATED',
+    403: 'PERMISSION_DENIED',
+    429: 'RESOURCE_EXHAUSTED',
+    500: 'INTERNAL',
+    503: 'UNAVAILABLE',
+    529: 'UNAVAILABLE',
+}
+
+
+def _parse_retry_after_ms(value: str) -> float | None:
+    """Parse an HTTP Retry-After value into milliseconds.
+
+    Supports both delay-seconds and HTTP-date values, matching the
+    JavaScript Anthropic adapter.
+    """
+    value = value.strip()
+    if not value:
+        return None
+
+    try:
+        seconds = float(value)
+    except ValueError:
+        pass
+    else:
+        if seconds >= 0:
+            return seconds * 1000
+
+    try:
+        retry_at = parsedate_to_datetime(value)
+        if retry_at is None:
+            return None
+        retry_at_ms = retry_at.timestamp() * 1000
+    except (OverflowError, TypeError, ValueError):
+        return None
+    return max(0.0, retry_at_ms - time.time() * 1000)
+
+
+def _from_anthropic_error(error: APIError) -> GenkitError:
+    """Convert an Anthropic SDK error to its Genkit equivalent."""
+    status_code = getattr(error, 'status_code', None)
+    status = _ANTHROPIC_STATUS_MAP.get(status_code, 'UNKNOWN') if isinstance(status_code, int) else 'UNKNOWN'
+
+    response = getattr(error, 'response', None)
+    retry_after_header = response.headers.get('retry-after') if response is not None else None
+    retry_after_ms = _parse_retry_after_ms(retry_after_header) if retry_after_header else None
+    response_metadata: ErrorResponseMetadata | None = None
+    if retry_after_ms is not None:
+        response_metadata = {'retry_after_ms': retry_after_ms}
+
+    return GenkitError(
+        status=status,
+        message=error.message,
+        response_metadata=response_metadata,
+    )
 
 
 def _to_anthropic_schema(schema: dict[str, Any]) -> dict[str, Any]:
@@ -228,13 +290,16 @@ class AnthropicModel:
 
         logger.debug('Anthropic generate request', model=self.model_name, streaming=bool(streaming))
 
-        if streaming:
-            assert ctx is not None  # streaming requires ctx
-            response = await self._generate_streaming(params, ctx, client=client, use_beta=use_beta)
-        else:
-            active_client = cast(Any, client)
-            messages_client = active_client.beta.messages if use_beta else active_client.messages
-            response = await messages_client.create(**params)
+        try:
+            if streaming:
+                assert ctx is not None  # streaming requires ctx
+                response = await self._generate_streaming(params, ctx, client=client, use_beta=use_beta)
+            else:
+                active_client = cast(Any, client)
+                messages_client = active_client.beta.messages if use_beta else active_client.messages
+                response = await messages_client.create(**params)
+        except APIError as error:
+            raise _from_anthropic_error(error) from error
 
         logger.debug(
             'Anthropic raw API response',
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
@@ -113,10 +113,7 @@ def _parse_retry_after_ms(value: str) -> float | None:
             return retry_after_ms
 
     try:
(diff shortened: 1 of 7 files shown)


py/packages/genkit-anthropic/src/genkit_anthropic/models.py
```diff
@@ -60,6 +64,63 @@
...
+    return max(0.0, retry_at_ms - time.time() * 1000)
+
+
+def _from_anthropic_error(error: APIError) -> GenkitError:
+    """Convert an Anthropic SDK error to its Genkit equivalent."""
+    status_code = getattr(error, 'status_code', None)
+    status = _ANTHROPIC_STATUS_MAP.get(status_code, 'UNKNOWN') if isinstance(status_code, int) else 'UNKNOWN'
+
+    response = getattr(error, 'response', None)
+    retry_after_header = response.headers.get('retry-after') if response is not None else None
```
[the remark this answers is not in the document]
hilariie: This isn't needed for the SDK contract. Response-bearing Anthropic errors use `httpx.Response`, which always exposes headers. Base and connection errors have no response and are already handled by the None path. Supporting custom objects that violate that contract would be outside the scope of this fix.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
```diff
@@ -60,6 +64,63 @@
...
+        if seconds >= 0:
+            return seconds * 1000
+
+    try:
+        retry_at = parsedate_to_datetime(value)
+        if retry_at is None:
+            return None
+        retry_at_ms = retry_at.timestamp() * 1000
+    except (OverflowError, TypeError, ValueError):
+        return None
```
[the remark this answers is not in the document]
hilariie: `Retry-After` comes from `httpx.Headers` as a string, and malformed string dates are reported as `ValueError`. timestamp() can raise OSError for naive dates on some platforms, including Windows so i've added `OSError` to the targeted exception list in `a9d5b284a`. Did not catch every Exception since that would hide unrelated programming errors.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
```diff
@@ -60,6 +64,63 @@
...
+    if not value:
+        return None
+
+    try:
+        seconds = float(value)
+    except ValueError:
+        pass
+    else:
+        if seconds >= 0:
+            return seconds * 1000
```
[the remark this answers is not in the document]
hilariie: resolved in `a9d5b284a`. Follow-up issue on JS required.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
```diff
@@ -76,6 +81,66 @@ def model_dump(self, *, exclude_none: bool = False, by_alias: bool = False) -> d
...
+        pass
+    else:
+        # Check the scaled value: a large finite input can overflow to inf.
+        retry_after_ms = seconds * 1000
+        if seconds >= 0 and math.isfinite(retry_after_ms):
+            return retry_after_ms
+
+    try:
+        retry_at = parsedate_to_datetime(value)
+        if retry_at is None:
```
cabljac: I think this branch is unreachable on 3.10+, `parsedate_to_datetime` raises `ValueError` rather than returning `None`. Harmless either way.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
```diff
@@ -76,6 +81,66 @@ def model_dump(self, *, exclude_none: bool = False, by_alias: bool = False) -> d
...
+    status = _ANTHROPIC_STATUS_MAP.get(status_code, 'UNKNOWN') if isinstance(status_code, int) else 'UNKNOWN'
+
+    response = getattr(error, 'response', None)
+    retry_after_header = response.headers.get('retry-after') if response is not None else None
+    retry_after_ms = _parse_retry_after_ms(retry_after_header) if retry_after_header else None
+    response_metadata: ErrorResponseMetadata | None = None
+    if retry_after_ms is not None:
+        response_metadata = {'retry_after_ms': retry_after_ms}
+
+    return GenkitError(
```
cabljac, line 134: Any reason not to pass `cause=error` here as well as `from error`? `__cause__` is set, but `GenkitError.cause` stays `None`, and `_error.py:245`/`257` branch on `self.cause`, so from what I can see the Anthropic body never makes it into the serialized error. Worth confirming on your side. JS doesn't set it either so I don't feel strongly.

cabljac (commented):
The Anthropic-side mapping looks right to me and matches `js/plugins/anthropic/src/models.ts:180-215`. One open question on the middleware change.

In `_retry.py:83-84`, when `retry_after_ms` exceeds `max_delay_ms` we skip the cap entirely, so the configured ceiling stops applying. From what I can see locally this means a `Retry-After: 86400` sleeps the caller for 24h even with `max_delay_ms=60000`:

```python
retry = Retry(max_retries=1, initial_delay_ms=100, max_delay_ms=60000, no_jitter=True)
# raises RESOURCE_EXHAUSTED with response_metadata={'retry_after_ms': 86_400_000}
# -> asyncio.sleep(86400.0)
```

Worth verifying on your side, but if that's right it's a bit of a footgun for anyone who set `max_delay_ms` specifically to bound how long a call can block, and the value is upstream-controlled. Simplest fix is to keep the unconditional `min(delay_ms, self.config.max_delay_ms)` so retry-after is a floor within the user's ceiling rather than an override of it. That does lose fidelity when the provider asks for longer than we allow, though I'd argue that's exactly the case the knob exists for. Happy to be talked out of it if you think matching JS exactly matters more here.

Related: JS has the same uncapped behaviour in `js/plugins/middleware/src/retry.ts:184-194`, so whichever way we go it'd be good to file a follow-up so the two don't drift.

Also, could you call the `_retry.py` change out in the PR body? The title scopes this to the anthropic plugin, but the change is in shared middleware, so it applies to any future producer or to a user-constructed `GenkitError` carrying `response_metadata`. Anthropic is the only producer today, so nothing else regresses right now, but I'd want the core team to see it.

cabljac (changes requested):

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
hilariie: The body already reaches the serialized error, just via `message` not `cause`. The SDK builds
`err_msg` as `f"Error code: {status} - {body}"`, and `_from_anthropic_error` passes that through.

Serializing a real 429 both ways:

```
WITHOUT cause (current)
  callable   : Error code: 429 - {'type': 'error', 'error': {'type': 'rate_limit_error', ...}}
  reflection : Error code: 429 - {'type': 'error', 'error': {'type': 'rate_limit_error', ...}}

WITH cause=error
  callable   : RateLimitError("Error code: 429 - {'type': 'error', ...}")
  reflection : Error code: 429 - {...}: RateLimitError("Error code: 429 - {...}")
```

So `cause` doesn't add information here; it duplicates the body in the reflection message and
downgrades the callable message from clean text to an exception `repr()`. I've left it as
`raise ... from error`, which still sets `__cause__` so the chain shows up in tracebacks.

hilariie:
> The Anthropic-side mapping looks right to me and matches `js/plugins/anthropic/src/models.ts:180-215`. One open question on the middleware change.
> 
> In `_retry.py:83-84`, when `retry_after_ms` exceeds `max_delay_ms` we skip the cap entirely, so the configured ceiling stops applying. From what I can see locally this means a `Retry-After: 86400` sleeps the caller for 24h even with `max_delay_ms=60000`:
> 
> ```python
> retry = Retry(max_retries=1, initial_delay_ms=100, max_delay_ms=60000, no_jitter=True)
> # raises RESOURCE_EXHAUSTED with response_metadata={'retry_after_ms': 86_400_000}
> # -> asyncio.sleep(86400.0)
> ```
> 
> Worth verifying on your side, but if that's right it's a bit of a footgun for anyone who set `max_delay_ms` specifically to bound how long a call can block, and the value is upstream-controlled. Simplest fix is to keep the unconditional `min(delay_ms, self.config.max_delay_ms)` so retry-after is a floor within the user's ceiling rather than an override of it. That does lose fidelity when the provider asks for longer than we allow, though I'd argue that's exactly the case the knob exists for. Happy to be talked out of it if you think matching JS exactly matters more here.
> 
> Related: JS has the same uncapped behaviour in `js/plugins/middleware/src/retry.ts:184-194`, so whichever way we go it'd be good to file a follow-up so the two don't drift.
> 
> Also, could you call the `_retry.py` change out in the PR body? The title scopes this to the anthropic plugin, but the change is in shared middleware, so it applies to any future producer or to a user-constructed `GenkitError` carrying `response_metadata`. Anthropic is the only producer today, so nothing else regresses right now, but I'd want the core team to see it.

Fixed in `410955ea0`

Two existing tests encoded the uncapped behaviour. Rather than flip their assertions, I raised their `max_delay_ms` so they still exercise the floor semantics they were written for, and added `test_retry_after_is_capped_by_max_delay` for your 24h case.

py/packages/genkit-anthropic/src/genkit_anthropic/models.py
hilariie: removed in `2fc80aaf2`

huangjeff5:
> The Anthropic-side mapping looks right to me and matches `js/plugins/anthropic/src/models.ts:180-215`. One open question on the middleware change.
> 
> In `_retry.py:83-84`, when `retry_after_ms` exceeds `max_delay_ms` we skip the cap entirely, so the configured ceiling stops applying. From what I can see locally this means a `Retry-After: 86400` sleeps the caller for 24h even with `max_delay_ms=60000`:
> 
> ```python
> retry = Retry(max_retries=1, initial_delay_ms=100, max_delay_ms=60000, no_jitter=True)
> # raises RESOURCE_EXHAUSTED with response_metadata={'retry_after_ms': 86_400_000}
> # -> asyncio.sleep(86400.0)
> ```
> 
> Worth verifying on your side, but if that's right it's a bit of a footgun for anyone who set `max_delay_ms` specifically to bound how long a call can block, and the value is upstream-controlled. Simplest fix is to keep the unconditional `min(delay_ms, self.config.max_delay_ms)` so retry-after is a floor within the user's ceiling rather than an override of it. That does lose fidelity when the provider asks for longer than we allow, though I'd argue that's exactly the case the knob exists for. Happy to be talked out of it if you think matching JS exactly matters more here.
> 
> Related: JS has the same uncapped behaviour in `js/plugins/middleware/src/retry.ts:184-194`, so whichever way we go it'd be good to file a follow-up so the two don't drift.
> 
> Also, could you call the `_retry.py` change out in the PR body? The title scopes this to the anthropic plugin, but the change is in shared middleware, so it applies to any future producer or to a user-constructed `GenkitError` carrying `response_metadata`. Anthropic is the only producer today, so nothing else regresses right now, but I'd want the core team to see it.

Agree with this point. I think:

min(max(retry_after_ms, delay_ms), max_delay_ms) is what we want.

huangjeff5 (approved):
Core/middleware changes look good to me.

cabljac (approved):

PR #5728, opened 2026-07-13
fix(py): gate Anthropic structured output on constraints

### LINKED ISSUE #5632: [plugins] Anthropic Python: harden structured output (gate on output_constrained)
Parent: #5617 (Phase B, P1: item 16)

## Why

Native constrained output is triggered by the wrong condition. Today it keys off `output_format == 'json'` plus a schema plus model support, rather than off the caller's actual intent to constrain output. This conflates "I want JSON-ish text" with "I want constrained decoding," and can force native structured output when the caller only wanted JSON text.

The request shape itself is already correct (stable `output_config.format`); the gating is the bug. Note this depends on the SDK floor from #5622 (`anthropic>=0.96.0`); #5626 is the registry refresh, not the SDK bump.

## What needs to be done

- Drive native constrained output off the caller's constrained-output intent (`request.output_constrained`) together with model support, instead of off `output_format == 'json'` plus a schema.
- Keep the existing system-prompt JSON fallback for unconstrained JSON and for models that do not support constrained output, including the existing fence-stripping behaviour.

## Acceptance

- Native `output_config` is used only when the caller asks for constrained output and the model supports it.
- Unconstrained JSON still works via the system-prompt fallback.
- Tests cover constrained vs. unconstrained requests and the unsupported-model fallback.

## Reference

- Current gating: `py/plugins/anthropic/src/genkit/plugins/anthropic/models.py:236-253`; fence-stripping fallback in `utils.py:80-105`.

hilariie:
## What

The Anthropic Python plugin was triggering native constrained output (`output_config`
with a JSON schema) off the wrong condition. It keyed off `output_format == 'json'`
plus a schema plus model support, which conflates "I want JSON-ish text" with "I want
constrained decoding." A caller who only asked for unconstrained JSON could get forced
into native structured output.

This changes the gate to key off the caller's actual intent (`request.output_constrained`)
together with model support, matching how JS and Go already work.

## Why

Parity with JS/Go. Both plugins gate native structured output on the same triple of
format + schema + the caller's constrained intent:

- Go: `i.Output.Format == "json" && i.Output.Schema != nil && i.Output.Constrained`
  (`go/plugins/internal/anthropic/anthropic.go`)
- JS: `output.schema && output.constrained && output.format === 'json'`
  (`js/plugins/anthropic/src/runner/beta.ts`)

In JS/Go the model-capability check lives in core (the `simulateConstrainedGeneration`
middleware in JS, `supportsConstrained` in Go), which downgrades unsupported models to a
prompt-instruction fallback. Python core has no equivalent step yet, so the plugin carries
the model-support check and system-prompt fallback itself. This is the Pythonic stand-in
for that core middleware.

Closes #5632.

## Changes

- Gate `output_config` on `output_schema` present + `output_constrained` truthy +
  `_supports_constrained(has_tools)`, replacing the old `'json' in supports.output` check.
- Add `_supports_constrained`, mirroring the JS/Go capability gate including the
  `no-tools` case.
- Keep the existing system-prompt JSON fallback and fence-stripping behaviour unchanged.

## Testing

- Native `output_config` only when `constrained=True` and the model supports it.
- Fallback to system prompt for unconstrained JSON (`None` and `False`), missing schema,
  unknown models, explicit `constrained=none`, and `no-tools` with tools present.
- `uv run --directory py pytest packages/genkit-anthropic/tests/ -v` passes (71 tests)

### DIFF
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/models.py
@@ -32,6 +32,7 @@ from anthropic import AsyncAnthropic
 from anthropic.types import Message as AnthropicMessage
 
 from genkit import (
+    Constrained,
     FinishReason,
     MediaPart,
     Message,
@@ -235,8 +236,13 @@ class AnthropicModel:
 
         # Handle JSON output constraint
         if request.output_format == 'json':
-            supports_json = 'json' in (self._model_info.supports.output or []) if self._model_info.supports else False
-            if request.output_schema and supports_json:
+            use_native = (
+                request.output_schema is not None
+                and bool(request.output_constrained)
+                and self._supports_constrained(bool(request.tools))
+            )
+            if use_native:
+                assert request.output_schema is not None
                 # Use native structured outputs via output_config.
                 params['output_config'] = {
                     'format': {
@@ -247,7 +253,7 @@ class AnthropicModel:
             else:
                 # Fall back to system prompt instruction.
                 instruction = '\n\nOutput valid JSON. Do not wrap the JSON in markdown code fences.'
-                if request.output_schema:
+                if request.output_schema is not None:
                     schema_str = json.dumps(request.output_schema, indent=2)
                     instruction += f'\n\nFollow this JSON schema:\n{schema_str}'
                 system = (system or '') + instruction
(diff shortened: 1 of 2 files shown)


cabljac (approved):
LGTM. Gating matrix looks right (constrained + capability + `NO_TOOLS` vs tools present), and the prompt fallback still embeds the schema for the unconstrained case. Heads up that this overlaps #5720's `output_config` merge in the same block of `_build_params`, whichever lands second needs a semantic rebase there.

PR #5733, opened 2026-07-14
fix(js/ai): persist serializable message data from prompt agents

### LINKED ISSUE #5712: genkit 1.39.0 `definePromptAgent` + `output` schema returns `DataCloneError: ... could not be cloned`
**Describe the bug**

Genkit 1.39.0 agent API (`definePromptAgent()`) returns the following error when output schema is provided.

A prompt agent whose prompt declares a structured `output` schema throws on the
very first `send()`:

```
AgentError: (message) => {
        return extractJson(message.text);
      } could not be cloned.
  ... DOMException [DataCloneError] ... at structuredClone (node:internal/worker/js_transferable)
  at AgentChatImpl.toAgentError (@genkit-ai/ai/src/agent-core.ts:985)
  at @genkit-ai/ai/src/agent-core.ts:754
```

**To Reproduce**

Clone, npm install, and run the [example repo here](https://github.com/jmcooper/genkit-agent-api-error-repro) which does the following:

```ts
ai.definePrompt(
  {
    name: 'echoPrompt',
    model,
    output: { schema: z.object({ answer: z.string() }) }, // <-- remove this line and the error goes away
  },
  'You are a test assistant.',
);

// A prompt agent backed by a session store.
const agent = ai.definePromptAgent({
  promptName: 'echoPrompt',
  store: new InMemorySessionStore(),
});
```

**Expected behavior**

It should support specifying an output schema when using the agent API and return a response matching the schema.

**Desktop (please complete the following information):**
- OS: Windows
- Bash Terminal
- `genkit` **1.39.0** (JS), `genkit/beta`
- Node.js 24.x
- Models tested:
  - Fake in-process model: No plugin, no API key, no network — reproduced with a fake in-process model
  - gemini-3.1-flash-lite
  - gemini-3-flash-preview

### Using a fake inline model:
```bash
npm install
npm run repro
```

### Using a real model:
To use a real model set the following values in the `.env` (see `.env.sample`) before running the project as described above: 
- Set `USE_REAL_MODEL=true`
- Set `GEMINI_API_KEY`
- Set `GEMINI_MODEL_NAME` (i.e. `GEMINI_MODEL_NAME='gemini-3.5-flash'`)

## Expected
`chat.send()` resolves and `res.output` is `{ answer: 'hello' }`.

## Actual
Throws `AgentError` wrapping a `DataCloneError` from `structuredClone`.

## Isolating variable
Deleting the single `output: { schema }` line makes it run to completion. So the
crash is triggered specifically by combining a **prompt agent** with a
**structured output schema**.

Also isolated (all still crash / all irrelevant):
- **Model** — reproduced with the fake model above, and with `gemini-2.5-flash`,
  `gemini-2.5-flash-lite`, `gemini-2.5-pro`, `gemini-3.5-flash`.
- **Session store** — crashes with `InMemorySessionStore`, `FileSessionStore`,
  and even with **no store** (ephemeral `agent.chat()` / client-managed state).
- **The plain API is fine** — `ai.generate({ output: { schema } })` and
  `ai.prompt(...).stream(input, { ... })` with the *same* output schema work
  perfectly. Only the Agent path fails.

ATKasem:
Fixes #5712

### Bug

A prompt agent whose prompt declares an `output` schema throws on the very first `send()`:

```
DataCloneError: (message) => { return extractJson(message.text); } could not be cloned.
```

Removing the `output` schema makes it go away.

### Root cause

When a prompt declares an `output` schema, the generate response's `.message` is a `Message` instance that carries a non-serializable `parser` function (`(message) => extractJson(message.text)`). `definePromptAgent` pushes that instance straight into the session messages, so when the turn snapshots state to the store, `session.getState()`'s `structuredClone` chokes on the function.

### Fix

Persist `res.message.toJSON()` (plain `MessageData`) instead of the `Message` instance — the same serializable form `GenerateResponse.messages` already uses for history persistence. The live response object is untouched, so callers still get `res.text` and normal output parsing.

### Test

Added a regression test in `js/ai/tests/agent_test.ts` (prompt agent + `output` schema + `InMemorySessionStore`). It throws the `could not be cloned` error before the fix and passes after. Full `agent_test` (94 tests), `agents_spec`, and `session` suites pass; `tsc` is clean.

### Checklist
- [x] PR title follows Conventional Commits
- [x] Tested (unit test added + existing suites)
- [ ] Docs updated (n/a — bug fix)

### DIFF
--- a/js/ai/src/agent.ts
+++ b/js/ai/src/agent.ts
@@ -1560,11 +1560,15 @@ export function definePromptAgent<
           (m) => !m.metadata?.[promptTag]
         );
         if (res.message) {
-          msgs.push(res.message);
+          // Persist the serializable message data, not the `Message` instance:
+          // when the prompt declares an `output` schema the response message
+          // carries a non-cloneable `parser` function, which would break
+          // `structuredClone` when the session state is snapshotted (#5712).
+          msgs.push(res.message.toJSON());
         }
         sess.setMessages(msgs);
       } else if (res.message) {
-        sess.addMessages([res.message]);
+        sess.addMessages([res.message.toJSON()]);
       }
 
       if (res.finishReason === 'interrupted') {
--- a/js/ai/tests/agent_test.ts
+++ b/js/ai/tests/agent_test.ts
@@ -31,6 +31,7 @@ import {
   defineCustomAgent,
   definePromptAgent,
 } from '../src/agent.js';
+import { configureFormats } from '../src/formats/index.js';
 import { definePrompt } from '../src/prompt.js';
 import { InMemorySessionStore } from '../src/session-stores.js';
 import {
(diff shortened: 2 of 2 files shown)


jmcooper:
A couple of versions have been released since this bug was released and this PR submitted. Any chance of this getting merged in soon?

PR #5735, opened 2026-07-14
fix(js/plugins/google-genai): validate per-request baseUrl override host

ibondarenko1:
### What

`calculateRequestOptions` (js/plugins/google-genai/src/googleai/utils.ts) applied a per-request `config.baseUrl` override unconditionally. Genkit flow handlers commonly forward user-controlled config into `ai.generate({ config })`, so a value that reaches the request could redirect the Gemini call to an arbitrary host together with the `x-goog-api-key` header and the full prompt body (SSRF + API-key / prompt exfiltration + response forgery). The SDK did not validate that the override host belongs to a Google endpoint, and did not strip the API key when the host changed.

### Fix

Restrict a per-request `baseUrl` override to the default Google endpoint (`generativelanguage.googleapis.com`) or the host of the plugin-configured `baseUrl`. An override to any other host now throws `PERMISSION_DENIED` (and an invalid URL throws `INVALID_ARGUMENT`).

Developers who legitimately need arbitrary per-request hosts can opt back in explicitly at plugin construction:

```ts
googleAI({ allowCustomBaseUrl: true })
```

The trust boundary is moved to plugin construction (developer-controlled) instead of per-request config (which may flow from end users). The plugin-configured `baseUrl` is still honored, so existing custom-proxy setups keep working.

Vertex AI is unaffected: its `calculateRequestOptions` has no `baseUrl` override (it routes by `location`).

### Tests

Adds `calculateRequestOptions` guard tests to `tests/googleai/utils_test.ts`: default Google host allowed, plugin-configured host allowed, attacker host rejected, invalid URL rejected, and `allowCustomBaseUrl` opt-in bypass.

- `pnpm --filter @genkit-ai/google-genai test` → 30/30 pass
- `pnpm --filter @genkit-ai/google-genai run check` (tsc) → clean

Reported via the Google OSS VRP. Signed-off per the DCO.

### DIFF
--- a/js/plugins/google-genai/src/googleai/antigravity.ts
+++ b/js/plugins/google-genai/src/googleai/antigravity.ts
@@ -170,6 +170,7 @@ export function defineModel(
   const clientOptions: ClientOptions = {
     apiVersion: pluginOptions?.apiVersion,
     baseUrl: pluginOptions?.baseUrl,
+    allowCustomBaseUrl: pluginOptions?.allowCustomBaseUrl,
     customHeaders: pluginOptions?.customHeaders,
     experimental_debugTraces: pluginOptions?.experimental_debugTraces,
   };
--- a/js/plugins/google-genai/src/googleai/deep-research.ts
+++ b/js/plugins/google-genai/src/googleai/deep-research.ts
@@ -43,6 +43,7 @@ import {
 } from './types.js';
 import {
   calculateApiKey,
+  resolveBaseUrlOverride,
   checkApiKey,
   checkModelName,
   extractVersion,
@@ -282,7 +283,9 @@ export function defineModel(
       const newClientOptions: ClientOptions = {
         ...clientOptions,
         apiKey,
-        baseUrl: baseUrl || clientOptions.baseUrl,
+        baseUrl: baseUrl
+          ? resolveBaseUrlOverride(baseUrl, clientOptions)
+          : clientOptions.baseUrl,
         apiVersion: apiVersion || clientOptions.apiVersion,
       };
 
--- a/js/plugins/google-genai/src/googleai/deep-research.ts
+++ b/js/plugins/google-genai/src/googleai/deep-research.ts
@@ -251,6 +251,7 @@ export function defineModel(
   const clientOptions: ClientOptions = {
     apiVersion: pluginOptions?.apiVersion,
     baseUrl: pluginOptions?.baseUrl,
+    allowCustomBaseUrl: pluginOptions?.allowCustomBaseUrl,
     customHeaders: pluginOptions?.customHeaders,
     experimental_debugTraces: pluginOptions?.experimental_debugTraces,
   };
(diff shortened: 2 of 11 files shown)


ibondarenko1:
Good catch, thanks. Fixed in 1176c44: `deep-research.ts` and `lyria.ts` resolved request options manually and took `config.baseUrl` directly, bypassing the guard. Exported `resolveBaseUrlOverride` from `utils.ts` and applied it at both sites (`baseUrl ? resolveBaseUrlOverride(baseUrl, clientOptions) : clientOptions.baseUrl`), so the restriction is now enforced consistently.

I also checked the other model paths: `imagen.ts` and `veo.ts` don't take a per-request `baseUrl` (they use the plugin-level `clientOptions`), and `antigravity.ts` and `gemini.ts` already route through `calculateRequestOptions`, so those are covered. Added direct `resolveBaseUrlOverride` tests: `pnpm --filter @genkit-ai/google-genai test` -> 35/35, `tsc` clean.

ifielker:
It seems like it would be much easier to remove unwanted options from the config before you pass it in... rather than having extra config to allow other config... 

Can you not do something like: 
```
  ai.defineFlow({
  name: 'cleanConfigFlow',
  inputSchema: z.object({input: z.string(), userConfig: z.object({...})}),
 }, async(input) => {
  const cleanConfig = sanitzeUsersConfig(userConfig); // Clean the config before you pass it to ai.generate
  const response = ai.generate({
    ...,
    config: cleanConfig,
  });
});
```

ibondarenko1:
Fair point on the extra config, I'd rather not grow the surface either.

The reason this came through the VRP is the default: once `config.baseUrl` reaches the request, the SDK sends `x-goog-api-key` and the full prompt to that host. Sanitizing in the flow is the right primary control, but it only protects the developer who already knows `baseUrl` is a credential sink, which is the exact gap that creates the bug. I'm happy to add the `sanitizeUserConfig` pattern to the plugin README as the recommended practice.

To drop the extra config: I can remove `allowCustomBaseUrl` and just restrict a per-request `baseUrl` override to the plugin-configured host (set at construction, developer-controlled per your point). Plugin-level proxies keep working; a per-request override to a different host isn't honored. No opt-in flag, default stays safe. Want me to rework it that way?

ifielker:
I appreciate that you want to increase the safety of using Genkit. However, while I think this proposed change would be helpful in a very narrow, specific pattern of Genkit usage. I believe it would make it harder to use it in other different, but equally useful ways. We don't want to decide for you how you use Genkit. If you choose to allow the end-user to give you config, then you should validate/sanitize that config before using it. If on the other hand, you choose not to allow the user to give you any config, but choose to use different (potentially hardcoded) endpoints for different requests based on functionality, for example, then that should not be extra difficult because someone else has a different pattern of Genkit usage. We don't want to favour one usage pattern over the other.

So, while I appreciate the thought and the effort to write this PR, I don't think the plugin is the right place for this. 
For your use case, you could just write a wrapper for yourself that sanitizes and then calls the generate. (And others could write wrappers that decide which endpoint to target per request etc.)

ibondarenko1:
Thanks for the clear explanation. Understood  I won't pursue the
SDK-level runtime restriction further.

The intent was to prevent an application-configured credential from being
implicitly forwarded when a per-request endpoint changes, but I understand
that the project considers request config developer-controlled and expects
applications to validate it at their own trust boundary.

I'll close this PR. If documentation clarifying that per-request baseUrl is
trusted configuration and must not be forwarded from untrusted input would
be useful, I can propose that separately.

Thanks for taking the time to review it.

PR #5753, opened 2026-07-16
feat(js/plugins/google-genai): Multi-region support

ifielker:
Also added apiVersion choice

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/vertexai/client.ts
+++ b/js/plugins/google-genai/src/vertexai/client.ts
@@ -339,6 +339,8 @@ export function getVertexAIUrl(params: {
 
   if (params.clientOptions.kind == 'regional') {
     basePath = `${params.clientOptions.location}-${API_BASE_PATH}`;
+  } else if (params.clientOptions.kind == 'multi-regional') {
+    basePath = `aiplatform.${params.clientOptions.location}.rep.googleapis.com`;
   } else {
     basePath = API_BASE_PATH;
   }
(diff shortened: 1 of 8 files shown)


js/plugins/google-genai/src/vertexai/client.ts
```diff
@@ -339,6 +339,8 @@ export function getVertexAIUrl(params: {
 
   if (params.clientOptions.kind == 'regional') {
     basePath = `${params.clientOptions.location}-${API_BASE_PATH}`;
+  } else if (params.clientOptions.kind == 'multi-regional') {
```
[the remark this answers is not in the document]
ifielker: Irrelevant, kind is a const string in each. I can change them all in a different PR.

js/plugins/google-genai/src/vertexai/utils.ts
```diff
@@ -16,12 +16,14 @@
...
 import { GenkitError, z } from 'genkit';
 import { GoogleAuth } from 'google-auth-library';
-import type {
-  ClientOptions,
-  ExpressClientOptions,
-  GlobalClientOptions,
-  RegionalClientOptions,
-  VertexPluginOptions,
+import {
+  MultiRegionalClientOptions,
```
pavelgj: Minor: `MultiRegionalClientOptions` is only used as a type here, so it can join the `type` imports below for consistency (`isMultiRegionalLocation` is correctly a value import).

```suggestion
  isMultiRegionalLocation,
  type MultiRegionalClientOptions,
```

pavelgj (approved):

js/plugins/google-genai/src/vertexai/utils.ts
ifielker: done

PR #5757, opened 2026-07-17
feat(py): FirestoreSessionStore for google-cloud plugin

huangjeff5:
## Summary

Adds a durable FirestoreSessionStore in `genkit-google-cloud` built on top of the refactored core SessionStore protocol (#6028).

### DIFF
--- a/py/packages/genkit-google-cloud/pyproject.toml
+++ b/py/packages/genkit-google-cloud/pyproject.toml
@@ -44,6 +44,7 @@ classifiers = [
 ]
 dependencies = [
   "genkit",
+  "google-cloud-firestore>=2.14.0",
   "google-cloud-logging>=3.10.0",
   "opentelemetry-exporter-gcp-trace>=1.9.0",
   "opentelemetry-exporter-gcp-monitoring>=1.9.0",
--- a/py/packages/genkit-google-cloud/src/genkit_google_cloud/__init__.py
+++ b/py/packages/genkit-google-cloud/src/genkit_google_cloud/__init__.py
@@ -15,16 +15,23 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-"""Google Cloud Plugin for Genkit.
+"""Google Cloud plugin for Genkit.
 
-This plugin provides Google Cloud observability integration for Genkit,
-enabling telemetry export to Cloud Trace, Cloud Monitoring, and Cloud Logging.
+Exports Genkit telemetry to Cloud Trace, Cloud Monitoring, and Cloud Logging,
+and provides :class:`FirestoreSessionStore` — a durable agent session store
+backed by Firestore. Turns are saved as JSON Patch diffs between periodic
+full-state checkpoints (sharded so no document approaches Firestore's size
+limit), so apps can resume long conversations across processes without
+secondary indexes.
 
 Example:
     ```python
     from genkit import Genkit
     from genkit_google_genai import GoogleAI
-    from genkit_google_cloud import enable_google_cloud_telemetry
+    from genkit_google_cloud import (
+        FirestoreSessionStore,
+        enable_google_cloud_telemetry,
+    )
 
 
     # 1. Enable Google Cloud Trace and Monitoring export
@@ -34,17 +41,22 @@ Example:
     ai = Genkit(plugins=[GoogleAI()], model='googleai/gemini-flash-latest')
     await ai.generate(prompt='Hello, world!')
     # => Traces exported asynchronously to Cloud Trace (latency, tokens, status)
+
+    # 3. Persist agent sessions in Firestore (ADC / FIRESTORE_EMULATOR_HOST)
+    store = FirestoreSessionStore()
+    agent = ai.define_agent(name='assistant', store=store)
     ```
 
 Requirements:
-    - Requires Google Cloud Application Default Credentials (ADC) or explicit credentials.
-    - Telemetry export is disabled by default in local dev environments unless explicitly configured.
+    - Google Cloud Application Default Credentials (ADC), or explicit credentials.
+    - Telemetry export is disabled by default in local dev unless explicitly configured.
 
 See Also:
     - Cloud Trace: https://cloud.google.com/trace
     - Cloud Monitoring: https://cloud.google.com/monitoring
 """
 
+from .session_store.firestore import FirestoreSessionStore
 from .telemetry import add_gcp_telemetry, enable_google_cloud_telemetry
 
 
@@ -58,6 +70,7 @@ def package_name() -> str:
 
 
 __all__ = [
+    'FirestoreSessionStore',
     'add_gcp_telemetry',
     'enable_google_cloud_telemetry',
     'package_name',
--- a/py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
+++ b/py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
@@ -61,6 +61,7 @@ from genkit._ai._agents._session import (
     StateT,
 )
 from genkit._ai._agents._session_stores._util import (
+    TERMINAL_STATUSES,
     SaveFn,
     apply_save,
     iterate_statuses,
@@ -97,13 +98,6 @@ MAX_SHARD_SIZE = 1_000_000
 # Firestore transactions retry internally on contention; after this many
 # attempts the failure surfaces as GenkitError(status='ABORTED').
 DEFAULT_TRANSACTION_MAX_ATTEMPTS = 5
-
-_TERMINAL_STATUSES = frozenset({
-    SnapshotStatus.COMPLETED,
-    SnapshotStatus.FAILED,
-    SnapshotStatus.ABORTED,
-    SnapshotStatus.EXPIRED,
-})
 
 
 @dataclass(eq=False)
@@ -1609,7 +1603,7 @@ class FirestoreSessionStore(SessionStore[StateT], SnapshotSubscriber, Generic[St
                 sub.last_status = status
                 with contextlib.suppress(Exception):
                     sub.queue.put_nowait(status)
-                if status not in _TERMINAL_STATUSES:
+                if status not in TERMINAL_STATUSES:
                     return
                 with contextlib.suppress(Exception):
                     sub.queue.put_nowait(None)
--- a/py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
+++ b/py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
@@ -1,0 +1,1745 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Firestore-backed session store for agent snapshots.
+
+Each turn is persisted as a JSON Patch diff from its parent, with periodic
+full-state checkpoints. Checkpoint state is split across shard documents so no
+single write approaches Firestore's ~1 MiB document limit.
+
+Reads and writes use only document-ID lookups (pointer + snapshot + shards), so
+deployments need no secondary indexes and stay strongly consistent.
+
+Default layout (collection ``genkit-sessions``, prefix ``global``)::
+
+    genkit-sessions/{prefix}/snapshots/{snapshotId}
+    genkit-sessions-shards/{prefix}/shards/{checkpointId}_{index}
+    genkit-sessions-pointers/{prefix}/pointers/{sessionId}
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import json
+import logging
+import threading
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from typing import Any, Generic, Literal
+
+from google.api_core import exceptions as google_exceptions
+from google.cloud import firestore
+from google.cloud.firestore import (
+    AsyncClient,
+    AsyncCollectionReference,
+    AsyncDocumentReference,
+    AsyncTransaction,
+    DocumentSnapshot,
+)
+from pydantic import BaseModel, ConfigDict, Field, field_validator
+from pydantic.alias_generators import to_camel
+
+from genkit._ai._agents._session import (
+    SessionStore,
+    SnapshotStatusStream,
+    SnapshotSubscriber,
+    StateT,
+)
+from genkit._ai._agents._session_stores._util import (
+    SaveFn,
+    apply_save,
+    iterate_statuses,
+    require_one_selector,
+    session_id_of,
+)
+from genkit._ai._json_patch import apply_json_patch, diff_json
+from genkit._core._action import get_current_context
+from genkit._core._error import GenkitError
+from genkit._core._loop_cache import _loop_local_client
+from genkit._core._typing import (
+    AgentFinishReason,
+    GenkitRuntimeError,
+    JsonPatchOp,
+    JsonPatchOperation,
+    SessionSnapshot,
+    SessionState,
+    SnapshotStatus,
+)
+
+DEFAULT_COLLECTION = 'genkit-sessions'
+DEFAULT_PREFIX = 'global'
+logger = logging.getLogger(__name__)
+# Favor common chat workloads: enough diffs between checkpoints to keep write
+# amplification down, small enough that reconstruct stays cheap.
+DEFAULT_CHECKPOINT_INTERVAL = 25
+# Kept well under Firestore's 1 MiB/doc limit so a single shard or diff write
+# cannot be rejected for size.
+DEFAULT_SHARD_SIZE = 512 * 1024
+
+# Ceiling for shard_size: shard documents carry the chunk plus field overhead,
+# so cap below Firestore's 1 MiB document limit with headroom.
+MAX_SHARD_SIZE = 1_000_000
+
+# Firestore transactions retry internally on contention; after this many
+# attempts the failure surfaces as GenkitError(status='ABORTED').
+DEFAULT_TRANSACTION_MAX_ATTEMPTS = 5
+
+_TERMINAL_STATUSES = frozenset({
+    SnapshotStatus.COMPLETED,
+    SnapshotStatus.FAILED,
+    SnapshotStatus.ABORTED,
+    SnapshotStatus.EXPIRED,
+})
+
+
+@dataclass(eq=False)
+class _StatusSubscription:
+    """One ``on_snapshot_status_change`` call: private queue + Firestore watch."""
+
+    queue: asyncio.Queue[SnapshotStatus | None]
+    watch: Any | None = None
+    last_status: SnapshotStatus | None = None
+    closed: bool = False
+
+    __hash__ = object.__hash__
+
+
+@dataclass
+class _StatusLoopState:
+    """Per-event-loop Firestore clients + open status subscriptions."""
+
+    subscriptions: set[_StatusSubscription] = field(default_factory=set)
+    client: AsyncClient | None = None
+    sync_client: firestore.Client | None = None
+    owns_async_client: bool = False
+    owns_sync_client: bool = False
+
+
+class _ShardDoc(BaseModel):
+    """Schema for a checkpoint state shard document stored in Firestore."""
+
+    model_config = ConfigDict(extra='ignore')
+
+    chunk: bytes
+
+    @field_validator('chunk', mode='before')
+    @classmethod
+    def _coerce_chunk(cls, v: object) -> object:
+        """Coerce raw Firestore binary chunk representations to bytes.
+
+        google-cloud-firestore's gRPC deserializer returns zero-copy memoryview
+        objects for binary blob fields in DocumentSnapshot.to_dict().
+        """
+        if isinstance(v, memoryview):
+            return v.tobytes()
+        if isinstance(v, (bytes, bytearray)):
+            return bytes(v)
+        if isinstance(v, str):
+            return v.encode('utf-8')
+        return v
+
+
+class _SnapshotWriteMeta(BaseModel):
+    """Metadata written onto a snapshot doc for later reconstruction."""
+
+    model_config = ConfigDict(
+        extra='ignore',
+        populate_by_name=True,
+        alias_generator=to_camel,
+    )
+
+    kind: Literal['diff', 'checkpoint']
+    checkpoint_id: str
+    checkpoint_shard_count: int = Field(ge=1)
+    # [] means checkpoint; a missing field is corrupt, not an empty path.
+    segment_path: list[str]
+    # Wire form is a JSON string (see ``_SnapshotDoc.state_patch``); this
+    # in-memory meta keeps the parsed op list until the doc is built.
+    state_patch: list[dict[str, Any]] | None = None
+
+
+class _ParentChainMeta(BaseModel):
+    """Parent snapshot metadata required for diff calculation."""
+
+    model_config = ConfigDict(
+        extra='ignore',
+        populate_by_name=True,
+        alias_generator=to_camel,
+    )
+
+    checkpoint_id: str
+    checkpoint_shard_count: int = Field(ge=1)
+    segment_path: list[str]
+
+
+class _SnapshotDoc(BaseModel):
+    """Schema for turn snapshot document stored in Firestore."""
+
+    model_config = ConfigDict(
+        extra='ignore',
+        populate_by_name=True,
+        alias_generator=to_camel,
+    )
+
+    snapshot_id: str
+    session_id: str
+    parent_id: str | None = None
+    created_at: str
+    updated_at: str | None = None
+    status: SnapshotStatus | None = None
+    heartbeat_at: str | None = None
+    finish_reason: AgentFinishReason | None = None
+    error: GenkitRuntimeError | None = None
+    kind: Literal['diff', 'checkpoint']
+    checkpoint_id: str
+    checkpoint_shard_count: int = Field(ge=1)
+    # [] means checkpoint; a missing field is corrupt, not an empty path.
+    segment_path: list[str]
+    # Stored as a JSON string so deep nested state does not hit Firestore's
+    # ~20-level map nesting limit (checkpoints already avoid that via shards).
+    # Typed loosely so pre-release array values reach the decode error path.
+    state_patch: Any = None
+
+    def to_session_snapshot(self, state_raw: dict[str, Any] | SessionState | None = None) -> SessionSnapshot:
+        """Convert Firestore snapshot document and reconstructed state to a SessionSnapshot."""
+        state = _state_from_dict(state_raw)
+        return SessionSnapshot(
+            snapshot_id=self.snapshot_id,
+            session_id=self.session_id,
+            parent_id=self.parent_id,
+            created_at=self.created_at,
+            updated_at=self.updated_at,
+            heartbeat_at=self.heartbeat_at,
+            status=self.status,
+            finish_reason=self.finish_reason,
+            error=self.error,
+            state=state,
+        )
+
+
+class _PointerDoc(BaseModel):
+    """Schema for session pointer document stored in Firestore."""
+
+    model_config = ConfigDict(
+        extra='ignore',
+        populate_by_name=True,
+        alias_generator=to_camel,
+    )
+
+    current_snapshot_id: str = Field(min_length=1)
+    checkpoint_id: str | None = None
+    checkpoint_shard_count: int | None = Field(default=None, ge=1)
+    segment_path: list[str]
+
+
+def _status_from_doc(doc_snapshot: DocumentSnapshot) -> SnapshotStatus | None:
+    """Extract and validate the snapshot status from a Firestore document."""
+    if not doc_snapshot.exists:
+        return None
+    status_val = (doc_snapshot.to_dict() or {}).get('status')
+    if status_val is None:
+        return None
+    try:
+        out: Any = SnapshotStatus(status_val)
+    except ValueError:
+        logger.warning("Unknown SnapshotStatus '%s' in Firestore document '%s'", status_val, doc_snapshot.id)
+        return None
+    return out
+
+
+def _state_to_dict(state: SessionState | None) -> dict[str, Any]:
+    """Convert a SessionState object to a dictionary representation."""
+    if state is None:
+        return {}
+    try:
+        dumped = state.model_dump(by_alias=True, exclude_none=True, mode='json')
+    except Exception as e:
+        raise GenkitError(
+            status='INVALID_ARGUMENT',
+            message=f'session state is not storable: {e}',
+        ) from e
+    return dumped if isinstance(dumped, dict) else {}
+
+
+def _state_from_dict(data: dict[str, Any] | SessionState | None) -> SessionState | None:
+    """Parse a dictionary or SessionState object into a SessionState instance."""
+    if data is None:
+        return None
+    if isinstance(data, SessionState):
+        return data
+    try:
+        return SessionState.model_validate(data)
+    except Exception as e:
+        raise GenkitError(
+            status='DATA_LOSS',
+            message=f'FirestoreSessionStore: corrupt session state document: {e}',
+        ) from e
+
+
+def _json_dumps_utf8(value: Any) -> bytes:  # noqa: ANN401
+    """Compact UTF-8 JSON bytes, or INVALID_ARGUMENT when the value is not storable."""
+    try:
+        return json.dumps(value, separators=(',', ':'), ensure_ascii=False).encode('utf-8')
+    except (TypeError, ValueError, UnicodeEncodeError) as e:
+        raise GenkitError(
+            status='INVALID_ARGUMENT',
+            message=f'session state is not storable: {e}',
+        ) from e
+
+
+# google.api_core exception class -> gRPC status name (see _to_genkit_error).
+_GOOGLE_EXC_STATUS: dict[str, str] = {
+    'Aborted': 'ABORTED',
+    'AlreadyExists': 'ALREADY_EXISTS',
+    'Cancelled': 'CANCELLED',
+    'DataLoss': 'DATA_LOSS',
+    'DeadlineExceeded': 'DEADLINE_EXCEEDED',
+    'FailedPrecondition': 'FAILED_PRECONDITION',
+    'Forbidden': 'PERMISSION_DENIED',
+    'InternalServerError': 'INTERNAL',
+    'InvalidArgument': 'INVALID_ARGUMENT',
+    'NotFound': 'NOT_FOUND',
+    'OutOfRange': 'OUT_OF_RANGE',
+    'PermissionDenied': 'PERMISSION_DENIED',
+    'ResourceExhausted': 'RESOURCE_EXHAUSTED',
+    'ServiceUnavailable': 'UNAVAILABLE',
+    'TooManyRequests': 'RESOURCE_EXHAUSTED',
+    'Unauthenticated': 'UNAUTHENTICATED',
+    'Unauthorized': 'UNAUTHENTICATED',
+    'Unknown': 'UNKNOWN',
+}
+
+
+def _to_genkit_error(e: google_exceptions.GoogleAPICallError) -> GenkitError:
+    """Convert a Google API error into a ``GenkitError``.
+
+    The status maps 1:1 from the exception class (both vocabularies are gRPC
+    statuses); the server's message is preserved verbatim, with no diagnosis
+    added.
+    """
+    status = _GOOGLE_EXC_STATUS.get(type(e).__name__, 'UNKNOWN')
+    return GenkitError(
+        status=status,  # type: ignore[arg-type]
+        message=f'FirestoreSessionStore: Firestore rejected the transaction ({type(e).__name__}): {e.message or e}',
+    )
+
+
+class _UserCodeError(Exception):
+    """Carrier: the mutator raised; deliver its exception verbatim."""
+
+
+async def _translate_txn_errors(awaitable: Awaitable[None]) -> None:
+    """Await a transactional coroutine, converting backend failures to ``GenkitError``.
+
+    The store's public contract is that every failure arising from its own
+    serialization, storage, or the Google libraries is a ``GenkitError`` with a
+    gRPC status. Exceptions from caller-supplied mutation functions propagate
+    unchanged (tunneled past library cleanup heuristics that would otherwise
+    misclassify them). Two library quirks are recognized narrowly, both
+    surfacing as a bare ``ValueError``: retry exhaustion (chained from the
+    final ``Aborted``), and a failed rollback on a transaction that never
+    obtained an ID, which masks the original error entirely.
+    """
+    try:
+        await awaitable
+    except _UserCodeError as e:
+        cause = e.__cause__
+        assert cause is not None
+        raise cause from None
+    except ValueError as e:
+        # Private-behavior dependency, deliberately narrow: the firestore
+        # library wraps retry exhaustion as a bare ValueError ("Failed to
+        # commit transaction in N attempts.") chained from the final Aborted.
+        # If a future release changes this shape we fall through to re-raise —
+        # degraded (untyped) but never mistranslated.
+        if isinstance(e.__cause__, google_exceptions.Aborted):
+            raise _to_genkit_error(e.__cause__) from e
+        if e.__cause__ is None and 'no transaction ID' in str(e):
+            # Second shape: a failure early in the transaction lifecycle
+            # (e.g. begin under heavy load) triggers rollback on a
+            # transaction with no ID; the library raises this unchained
+            # ValueError, destroying the original error. The transaction
+            # categorically did not commit, so ABORTED (retryable) is the
+            # honest classification even though the root cause is lost.
+            raise GenkitError(
+                status='ABORTED',
+                message=(
+                    'FirestoreSessionStore: transaction failed before it could be '
+                    f'established and the original error was masked by cleanup: {e}'
+                ),
+            ) from e
+        raise
+    except GenkitError:
+        raise
+    except google_exceptions.RetryError as e:
+        cause = e.cause
+        if isinstance(cause, google_exceptions.GoogleAPICallError):
+            raise _to_genkit_error(cause) from e
+        raise GenkitError(
+            status='ABORTED',
+            message=f'FirestoreSessionStore: Firestore retries exhausted: {e}',
+        ) from e
+    except google_exceptions.GoogleAPICallError as e:
+        raise _to_genkit_error(e) from e
+
+
+def _validate_doc_id(value: str | None, name: str) -> None:
+    """Reject ids that Firestore would treat as path structure, not a document id.
+
+    Ids become single document path segments; a ``/`` re-routes the path (an
+    even segment count silently addresses a different, deeper document — a
+    watch on it registers and then never fires), and ``.``/``..`` are reserved.
+    Leading or trailing whitespace is also rejected so save and get agree on
+    what counts as a valid id. Validating up front turns these failure modes
+    into a typed error.
+    """
+    if (
+        not value
+        or value != value.strip()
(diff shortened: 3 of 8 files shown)


huangjeff5:
Addressed in commit e5b8ca923:
1. **Rebased**: Rebased onto latest `jh-py-agents`.
2. **Self-Healing Pointer Repair**: Updated `get_snapshot` to check if the snapshot returned by `read_snapshot()` is non-`None`. If `None` (e.g. deleted by TTL or manual deletion), it falls through to list session snapshots and repair the pointer automatically.
3. **Removed Defensive Copying (`copy_snapshot`)**: Removed `copy_snapshot()` as `read_snapshot()` already deserializes a fresh `SessionSnapshot` model on each read.
4. **Unit Test**: Added `test_firestore_session_store_self_healing_on_deleted_snapshot` to test self-healing on deleted snapshots.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
...
+            'updatedAt': firestore.SERVER_TIMESTAMP,
+        }
+        if not is_ambiguous:
+            payload['currentSnapshotId'] = next(iter(leaves.keys()))
+        await self.pointer_ref(session_id).set(payload)
+
+    async def save_snapshot(self, snapshot_id: str, fn: SaveFn) -> SessionSnapshot | None:
+        """Save or update a session snapshot in Firestore."""
+        async with self.lock:
+            existing = await self.read_snapshot(snapshot_id)
```
hilariie: Could we move the snapshot read, mutator invocation, snapshot write, and pointer update into the same Firestore transaction?

The local `asyncio.Lock` only serializes calls using this store instance. In a multi-instance deployment, an abort and heartbeat can both read a `PENDING` snapshot, then the heartbeat can overwrite the committed `ABORTED` status with its stale copy. This conflicts with the `SessionStore.save_snapshot` contract, which describes the operation as an atomic read-modify-write, and with the detached runtime's expectation that heartbeats cannot resurrect terminal snapshots.

Keeping the snapshot and pointer writes in one transaction would also prevent a failed pointer update from leaving a persisted snapshot that cannot be resolved by session ID. Subscriber notification should happen after the transaction commits.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
...
+            return
+        is_ambiguous = len(leaves) > 1
+        payload: dict[str, Any] = {
+            'isAmbiguous': is_ambiguous,
+            'leaves': leaves,
+            'updatedAt': firestore.SERVER_TIMESTAMP,
+        }
+        if not is_ambiguous:
+            payload['currentSnapshotId'] = next(iter(leaves.keys()))
+        await self.pointer_ref(session_id).set(payload)
```
hilariie: Could this repair use a transaction or an update-time precondition?

There is a race between listing the snapshots and this unconditional write. For example:

1. This read observes snapshot A as the current leaf.
2. A concurrent save creates snapshot B and updates the pointer to B.
3. This repair overwrites the pointer with the stale value A.

Since A still exists, subsequent session lookups return A directly and do not enter the repair path again. Re-reading the pointer transactionally before committing the repair, or conditioning the write on the pointer version originally observed, would prevent the repair from replacing a newer valid pointer.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
...
+            leaves: dict[str, str] = {}
+            if pointer and isinstance(pointer.get('leaves'), dict):
+                leaves = {
+                    str(k): str(v) for k, v in pointer['leaves'].items() if isinstance(k, str) and isinstance(v, str)
+                }
+
+            if parent_snapshot_id and parent_snapshot_id in leaves:
+                leaves.pop(parent_snapshot_id, None)
+
+            leaves[snapshot_id] = created_at
```
hilariie: Should this update distinguish between a newly created snapshot and an update to an existing snapshot?

Because updating any existing snapshot adds it to `leaves`. If the history contains A followed by B, updating A later produces leaves `{A, B}` and incorrectly marks the session as ambiguous, even though B is still the only structural leaf.

Do you think it is safer to change leaf membership only when creating a new snapshot? So for an existing snapshot, the pointer metadata would only be refreshed if that snapshot is already a leaf.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
pavelgj: +1 on this. The `save_snapshot` docstring in `_session.py` explicitly calls this an "atomically read-modify-write" and warns `fn` may be called more than once under contention, so folding the snapshot write and pointer update into one transaction (and moving `notify` after commit) is what the contract expects. This is really the main thing separating a production-grade store from the in-memory/file ones, so worth getting right here.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
...
+        await update_in_transaction(transaction)
+
+    async def list_session_snapshots(self, session_id: str) -> list[SessionSnapshot]:
+        """Query and return all snapshots belonging to a session."""
+        snaps: list[SessionSnapshot] = []
+        async for doc in self.snapshots_col().where('sessionId', '==', session_id).stream():
+            try:
+                snaps.append(SessionSnapshot.model_validate(doc.to_dict()))
+            except (ValueError, TypeError):
+                continue
```
pavelgj: Silently dropping snapshots that fail to validate makes corruption invisible: a session could quietly lose turns with no signal. Could we log at debug/warning here so it's at least diagnosable?

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
...
+        payload = snapshot.model_dump(by_alias=True, exclude_none=True, mode='json')
+        await self.snapshot_ref(snapshot.snapshot_id).set(payload)
+
+    async def read_pointer(self, session_id: str) -> str | None:
+        """Read the current snapshot ID pointer for a session from Firestore."""
+        doc = await self.pointer_ref(session_id).get()
+        if not doc.exists:
+            return None
+        current = (doc.to_dict() or {}).get('currentSnapshotId')
+        return current if isinstance(current, str) else None
```
pavelgj: Is `read_pointer` called anywhere? `get_snapshot` reads the pointer inline, so this looks like dead code. Worth dropping unless it's meant for external use.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
...
+            sid = next_snapshot.snapshot_id
+            session_id = session_id_of(next_snapshot)
+            if not session_id:
+                raise GenkitError(
+                    status='INVALID_ARGUMENT',
+                    message="FirestoreSessionStore requires 'sessionId' on the snapshot.",
+                )
+
+            await self.write_snapshot(next_snapshot)
+            parent_id = getattr(next_snapshot, 'parent_id', None)
```
pavelgj: `parent_id` is a declared field on `SessionSnapshot`, so `next_snapshot.parent_id` reads cleaner than the defensive `getattr` here.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
...
+    notify,
+    require_one_selector,
+    select_leaf,
+    session_id_of,
+    subscribe,
+)
+from genkit._core._error import GenkitError
+from genkit._core._typing import SessionSnapshot, SnapshotStatus
+
+StateT = TypeVarExt('StateT', bound=BaseModel, default=Any)
```
pavelgj: The in-memory and file stores import the shared `StateT` from `genkit._ai._agents._session` rather than redefining it. Since this definition is identical to the shared one, reusing it would keep the three stores consistent.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,342 @@
+# Copyright 2026 Google LLC
+# SPDX-License-Identifier: Apache-2.0
```
pavelgj, line 15: Could we use the full Apache license header here (and in `__init__.py` and the test file) to match the rest of the package? `_inmemory_store.py` and the `telemetry/` modules use the full block, whereas these new files use the 2-line short form.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2026 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+
+"""Firestore session store module for Google Cloud plugin."""
+
+from .firestore import FirestoreSessionStore
```
pavelgj: Just confirming intent: is `from genkit_google_cloud.session_store import FirestoreSessionStore` the intended public import path? It's not re-exported from the plugin's top-level `__init__.py`. Assuming that's deliberate, no change needed.

py/packages/genkit-google-cloud/tests/firestore_session_store_test.py
```diff
@@ -0,0 +1,409 @@
...
+        yield live_snap_doc
+
+    snapshots_stream.stream = MagicMock(return_value=mock_stream())
+
+    snap_col = MagicMock()
+    snap_doc_ref = MagicMock()
+    snap_col.document.return_value = snap_doc_ref
+    snap_doc_ref.collection.return_value = snap_col
+    snap_doc_ref.get = AsyncMock(return_value=missing_snap_snapshot)
+    snap_col.where.return_value = snapshots_stream
```
pavelgj: Minor: `missing_snap_ref` (and its `missing_snap_snapshot` wiring) is set up but never attached to anything, and `pointer_doc_ref` gets reassigned here after being built earlier in the test. Looks like leftover setup that can be trimmed.

pavelgj:
**Consistency check against the JS reference (`js/plugins/google-cloud/src/session-store/firestore.ts`)**

I compared this against the JS `FirestoreSessionStore`, which I'm treating as the reference design. Three functional divergences stand out — are these intentional for the Python store? For the first and third, if they're *not* deliberate I'd consider them blocking, since they change the durability/operability of the store rather than just its internals.

1. **No diff/checkpoint sharding → the 1 MiB document limit is a hard failure. (Blocking unless intentional.)** JS stores each turn as a JSON Patch diff anchored to periodic, sharded full-state checkpoints (`checkpointInterval` / `shardSize` / a `-shards` collection) precisely so no document approaches Firestore's 1 MiB/doc limit and sessions can grow unbounded. This PR writes each snapshot as one full document (`write_snapshot` → `model_dump` → `.set()`), so any session whose state crosses ~1 MiB fails to persist outright — silent data loss from the user's perspective. The docstring says sharding "can land later"; if that's the accepted v1 tradeoff, let's make it explicit (and ideally fail loudly with a clear error). If not, this should be addressed before merge.

2. **`snapshot_path_prefix` can't see per-call context, so tenant isolation is weaker than JS.** In JS the prefix is `(options?) => string`, derived per-request from `options.context` (e.g. the authenticated user id) — that's the entire basis of its multi-tenant isolation. Here it's `Callable[[], str]` with no args, so it can only return a static/ambient value and can't key off per-request auth (and the Python `SessionStore` protocol doesn't thread a context/options param at all). Net effect: the isolation the docstring advertises ("isolate tenants when session ids may collide across users") is meaningfully weaker than JS. Is that intentional given the current protocol? At minimum the docstring should reflect the actual guarantee.

3. **Fallback path requires a secondary index; JS deliberately needs none. (Blocking unless intentional.)** `list_session_snapshots` runs a `where('sessionId', '==', ...)` query in the repair/fallback path. JS avoids all queries by design — reconstruction is document-ID lookups (`getAll`) only, so it needs no secondary indexes and stays strongly consistent. Python deployments will need a Firestore index and inherit that query's (eventual) consistency semantics, which is an operational + correctness divergence from the reference. If this query path is intended to stay, it needs to be documented as a deployment requirement; if the intent was to match JS's index-free design, it should change.

huangjeff5:
@gemini review

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Moved the snapshot read, state patch generation, checkpoint write, and pointer update into a single Firestore transaction in save_snapshot to ensure atomic read-modify-write.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Since pointers are now updated transactionally in save_snapshot, the repair path was dropped entirely. Missing pointers just return None.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Updated the logic to only change leaf membership when creating a new snapshot or refreshing an existing leaf.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Removed.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Updated to the full license block.

py/packages/genkit-google-cloud/tests/firestore_session_store_test.py
huangjeff5: Removed.

huangjeff5:
@gemini review

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/__init__.py
huangjeff5: Re-exported FirestoreSessionStore from the top-level genkit_google_cloud package so it can be imported directly via from genkit_google_cloud import FirestoreSessionStore.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Added  calls when snapshot documents are missing, fail validation, or contain mismatched/corrupt chain metadata during reconstruction.

huangjeff5:
> * **No diff/checkpoint sharding → the 1 MiB document limit is a hard failure. (Blocking unless intentional.)** JS stores each turn as a JSON Patch diff anchored to periodic, sharded full-state checkpoints (`checkpointInterval` / `shardSize` / a `-shards` collection) precisely so no document approaches Firestore's 1 MiB/doc limit and sessions can grow unbounded. This PR writes each snapshot as one full document (`write_snapshot` → `model_dump` → `.set()`), so any session whose state crosses ~1 MiB fails to persist outright — silent data loss from the user's perspective. The docstring says sharding "can land later"; if that's the accepted v1 tradeoff, let's make it explicit (and ideally fail loudly with a clear error). If not, this should be addressed before merge.
> * **`snapshot_path_prefix` can't see per-call context, so tenant isolation is weaker than JS.** In JS the prefix is `(options?) => string`, derived per-request from `options.context` (e.g. the authenticated user id) — that's the entire basis of its multi-tenant isolation. Here it's `Callable[[], str]` with no args, so it can only return a static/ambient value and can't key off per-request auth (and the Python `SessionStore` protocol doesn't thread a context/options param at all). Net effect: the isolation the docstring advertises ("isolate tenants when session ids may collide across users") is meaningfully weaker than JS. Is that intentional given the current protocol? At minimum the docstring should reflect the actual guarantee.
> * **Fallback path requires a secondary index; JS deliberately needs none. (Blocking unless intentional.)** `list_session_snapshots` runs a `where('sessionId', '==', ...)` query in the repair/fallback path. JS avoids all queries by design — reconstruction is document-ID lookups (`getAll`) only, so it needs no secondary indexes and stays strongly consistent. Python deployments will need a Firestore index and inherit that query's (eventual) consistency semantics, which is an operational + correctness divergence from the reference. If this query path is intended to stay, it needs to be documented as a deployment requirement; if the intent was to match JS's index-free design, it should change.

Auto-resolving these comments b/c they were reviewing outdated agents code cause by this branch being based off of an older origin/main. This PR is only related to the plugin.

huangjeff5:
Addressing the architectural alignment points with JS:

1. **Diffing & Sharding:** Implemented the exact sharding and JSON patch diffing strategy from the JS SDK (`_write_shards`, periodic full-state checkpoints, and diff generation).
2. **Per-request Tenant Isolation:** Updated `snapshot_path_prefix` to accept a `context: dict[str, Any] | None` map (containing arbitrary request metadata supplied by the app developer), enabling per-request tenant isolation matching JS.
3. **No Secondary Indexes / Index-Free Lookups:** Removed `list_session_snapshots` and all collection query (`where(...)`) usages, sticking exclusively to direct document-ID lookups (`pointer_ref` -> `snapshot_ref` -> `shards_col`) for strong consistency without extra deployment index requirements.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Sounds good. Updated firestore.py to import StateT directly from genkit._ai._agents._session, staying consistent with the in-memory and file session stores.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Updated to use next_snapshot.parent_id directly.

huangjeff5:
Accidentally deleted Pavel's comment around addressing the getattr() usage to get private credentials from AsyncClient.

I found that AsyncClient ships with "self.client._to_sync_copy()" method that can be used to create a SyncClient. So we went ahead and modified the private cloning logic to use this method if it exists on AsyncClient.

py/packages/genkit-google-cloud/src/genkit_google_cloud/__init__.py
```diff
@@ -45,6 +45,7 @@
     - Cloud Monitoring: https://cloud.google.com/monitoring
 """
 
+from .session_store.firestore import FirestoreSessionStore
```
hilariie, line 59: Blocking: `genkit-google-cloud` does not declare `google-cloud-firestore` (its pyproject lists genkit, google-cloud-logging, the two OTel GCP exporters, and strenum), but this import is eager and `session_store/firestore.py` imports `google.cloud.firestore` at module level. So `import genkit_google_cloud` needs firestore even for someone who only calls `enable_google_cloud_telemetry`. It resolves in the monorepo because `genkit-vertexai` pulls firestore in, and `google-cloud-logging` does not, so a standalone install would break.

Worth noting that JS declares `@google-cloud/firestore` as a hard dependency but exports the store from `beta.ts` rather than the package root, so if the root re-export here is deliberate, it is a deliberate divergence.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,910 @@
...
+        created_at: str,
+        is_new: bool,
+        checkpoint_id: str,
+        checkpoint_shard_count: int,
+        segment_path: list[str],
+        context: dict[str, Any] | None = None,
+    ) -> None:
+        """Update the session pointer inside an already-open transaction."""
+        ref = self.pointer_ref(session_id, context)
+        snapshot = await ref.get(transaction=transaction)
```
hilariie: This read runs after the transaction already has buffered writes, so the client rejects it. `save_snapshot` buffers `transaction.set(snap_ref, ...)` at line 518 (and the shard writes before that), then `_update_pointer_in_transaction` reads the pointer here. `_helpers.get_transaction_id` raises `ReadAfterWriteError` as soon as `transaction._write_pbs` is non-empty, and `AsyncDocumentReference.get` routes through it. I reproduced it against google-cloud-firestore 2.23.0. Every branch of `rmw` reaches line 518 before this, so it should fail on any save that persists something.

The pointer is already read at line 427 and validated at 428. Could we pass that snapshot and `PointerDoc` into the helper instead of re-reading? That removes the failure and also drops a billed read per save, which is what JS version does (it reads `pointerRef` once and reuses it).

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,910 @@
...
+            refs.append(checkpoint_ref)
+        refs.extend(shard_refs)
+        refs.extend(seg_refs)
+
+        if not refs:
+            return None
+
+        by_path: dict[str, DocumentSnapshot] = {}
+        res = transaction.get_all(refs)
+        gen = await res if inspect.iscoroutine(res) else res
```
hilariie: Against a real `AsyncTransaction` this raises `TypeError: object async_generator can't be used in 'await' expression`. `AsyncTransaction.get_all` is a coroutine function whose body is `return await self._client.get_all(...)`, and `AsyncClient.get_all` is an async generator function, so awaiting it fails inside the wrapper. `inspect.iscoroutine(res)` is True here, so `await res` is the branch production takes. Reproduced on 2.23.0.

hilariie (dismissed):

pavelgj (commented):
Reviewed for consistency with the JS `FirestoreSessionStore` (`js/plugins/google-cloud/src/session-store/firestore.ts`). The core design ports over faithfully (same three-collection layout, diff-vs-checkpoint model, checkpoint-interval boundary logic, oversized-diff promotion, sharded checkpoints with stale-shard pruning, read-only reconstruct transaction, DATA_LOSS on missing shards, tenant prefixing). A few things worth calling out:

**1. Read-after-write ordering in the save transaction**

Firestore transactions require all reads to complete before any writes. In `save_snapshot` we `transaction.set(...)` the shards and the snapshot doc, and then `_update_pointer_in_transaction` does `await ref.get(transaction=transaction)` on the pointer, which is a read after those writes. The fake harness in the tests doesn't enforce the ordering, but the real backend (and the emulator) will reject this.

The pointer is also already read earlier in `save_snapshot` (`pointer_snap = await pointer_ref.get(...)`), so this second read is redundant on top of being mis-ordered. Suggest passing the already-loaded `pointer` (and `pointer_snap.exists`) into `_update_pointer_in_transaction` and dropping the second `get`, so all reads stay ahead of the writes. Worth a quick run against the Firestore emulator to confirm end-to-end, since the current tests wouldn't catch this.

**2. Branching / ambiguous-session support diverges from JS (confirm intent)**

The JS `FirestoreSessionStore` treats `session_id` lookups on a branched history as unsupported/undefined and deliberately dropped its defensive handling for simplicity (last-write-wins pointer, no leaf tracking). This store goes the other way, materializing `leaves` + `is_ambiguous` into the pointer and adding `reject_ambiguous_session` with newest-leaf resolution.

This is consistent with the Python in-memory and file stores (they share `select_leaf` / `reject_ambiguous_session`), so within Python it's the right call, not something to change in this PR. Flagging it only so it's a conscious, documented divergence rather than an accident: Python session stores support branched-session lookup, JS does not. If we want the two ecosystems to match, that's a broader contract decision worth writing down somewhere central.

**3. Corrupt/invalid docs are silently treated as "not found"**

`_reconstruct` swallows any `SnapshotDoc.model_validate` failure and returns `None`, which callers interpret as "snapshot doesn't exist." A schema-mismatched or corrupt doc therefore looks identical to a missing one, and here (unlike `_reconstruct_from`) it isn't even logged. In `save_snapshot` that's a bit sharper: a corrupt existing leaf makes `existing_recon` `None`, so the turn is treated as brand-new and writes a fresh checkpoint + advances the pointer, effectively masking the corruption. Suggest at least logging at warning here so corruption is distinguishable from absence.

**4. Smaller consistency nits vs JS (none blocking)**

- Blank ids: JS `normalize()` trims and rejects whitespace-only `snapshotId`/`sessionId`; the shared `require_one_selector` here uses `bool(...)`, so `"  "` passes and becomes a document id. Minor, and it's in shared code, but it's a behavior gap.
- `sanitize` uses `json.dumps(..., default=str)`, so non-JSON values (e.g. datetimes) get stringified instead of dropped/rejected as in JS's `JSON.parse(JSON.stringify(...))`. Slightly more permissive; fine if intended.
- `snapshot_path_prefix` receives the raw `context` dict, whereas JS passes the full `SessionStoreOptions`. Worth a one-line docstring note so a ported prefix fn doesn't surprise anyone.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
pavelgj: Read-after-write: Firestore transactions require all reads before any writes, but by the time we get here `save_snapshot` has already issued `transaction.set(...)` for the shards and the snapshot doc, so this `ref.get(transaction=...)` is a read after writes. The fake test harness doesn't enforce ordering, but the real backend/emulator will reject it. Also, the pointer was already read once in `save_snapshot` (`pointer_snap = await pointer_ref.get(...)`), so this is redundant too. Suggest passing the already-loaded `pointer`/`pointer_snap.exists` in and dropping this second get. Worth confirming against the emulator since tests won't catch it.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,910 @@
...
+        *,
+        context: dict[str, Any] | None = None,
+    ) -> tuple[SnapshotDoc, dict[str, Any]] | None:
+        snap = await self.snapshot_ref(snapshot_id, context).get(transaction=transaction)
+        if not snap.exists:
+            return None
+        try:
+            doc = SnapshotDoc.model_validate(snap.to_dict())
+        except Exception:
+            return None
```
pavelgj: This swallows any `SnapshotDoc.model_validate` failure and returns `None`, which callers read as "snapshot doesn't exist" - so a corrupt/schema-mismatched doc looks identical to a missing one, and unlike `_reconstruct_from` it isn't even logged. In `save_snapshot` a corrupt existing leaf then makes `existing_recon` `None`, so the turn is treated as brand-new and writes a fresh checkpoint + advances the pointer, masking the corruption. Suggest at least a warning log here so corruption is distinguishable from absence.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,910 @@
...
+
+    async def get_snapshot(
+        self,
+        *,
+        snapshot_id: str | None = None,
+        session_id: str | None = None,
+        context: dict[str, Any] | None = None,
+    ) -> SessionSnapshot | None:
+        """Retrieve a session snapshot by snapshot ID or session ID."""
+        require_one_selector(snapshot_id=snapshot_id, session_id=session_id)
```
pavelgj, line 853: Nit (shared code, not this PR): JS's `normalize()` trims and rejects whitespace-only `snapshotId`/`sessionId`, but `require_one_selector` -> `parse_snapshot_lookup_kw` uses `bool(...)`, so `"  "` passes and becomes a document id. Minor behavior gap vs JS.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,910 @@
...
+        out: Any = SnapshotStatus(status_val)
+    except ValueError:
+        logger.warning("Unknown SnapshotStatus '%s' in Firestore document '%s'", status_val, doc_snapshot.id)
+        return None
+    return out
+
+
+def sanitize(value: Any) -> Any:  # noqa: ANN401
+    """Drop values Firestore rejects while keeping JSON round-trip semantics."""
+    return json.loads(json.dumps(value, default=str))
```
pavelgj: Nit: `json.dumps(..., default=str)` stringifies non-JSON values (e.g. datetimes) instead of dropping/rejecting them the way JS's `JSON.parse(JSON.stringify(...))` does. Slightly more permissive; fine if intended.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,910 @@
...
+    across users.
+    """
+
+    def __init__(
+        self,
+        *,
+        client: AsyncClient | None = None,
+        sync_client: firestore.Client | None = None,
+        collection: str = DEFAULT_COLLECTION,
+        snapshot_path_prefix: Callable[[dict[str, Any] | None], str] | None = None,
```
pavelgj, line 575: Nit: `snapshot_path_prefix` receives the raw `context` dict here, whereas JS's `snapshotPathPrefix` gets the full `SessionStoreOptions`. Worth a one-line docstring note so a prefix fn ported from JS doesn't surprise anyone.

pavelgj (commented):
A few of the points from my earlier review, as inline comments for easier reference.

py/packages/genkit-google-cloud/src/genkit_google_cloud/__init__.py
huangjeff5: Good catch, added the google-cloud-firsetore dependency. 

Re: the beta namespace, I'm aware of that but decided against it because Python SDK is beta in general. Adding additional labels is a bit unnecessary.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Removed the unused branch and added a test.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Good catch. Fixed this.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Added docstring. This is intentional because snapshot_path_prefix lambda shouldn't need ALL the session store options. If eventually we need more options, just pass explicitly into the lambda.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: I looked at this carefully, JS doesn't just loads the document and assumes it's the right shape (no validation), and may fail later in unpredictable ways. Instead of just logging here, I decided to fail loudly instead.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: I removed sanitize, there's only 1 call-site, and it uses the default pydantic model dump before the sanitize. That should take care of these scenarios. That turns datetimes into isoformat strings, the same as what JS does.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: I see. I removed normalize() to do a json.dumps inline. There's only one callsite. 

parse_snapshot_lookup_kw => I made these conditions explicit now (and made sure to handle the case you pointed out). Should be easier to read at a glance now.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Fixed. Verified with real firestore.

huangjeff5:
Re: 2. Branching / ambiguous-session support diverges from JS (confirm intent)

Ack, this is intentional. Default path is same as JS (last writer wins), but we give developers the option to reject ambiguous sessions.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,929 @@
...
+            async def cleanup() -> None:
+                async with self.lock:
+                    self.subs.pop(snapshot_id, None)
+                    watch = self._watches.pop(snapshot_id, None)
+                if watch is not None:
+                    await asyncio.to_thread(watch.unsubscribe)
+
+            asyncio.run_coroutine_threadsafe(cleanup(), loop)
+
+        self._watches[snapshot_id] = ref.on_snapshot(on_snapshot)
```
hilariie: when the snapshot doc doesn't exist yet, every subscribe call lands here with a fresh watch. `subscribe()` skips registering the queue when `current is None`, so `is_first` is True on every `on_snapshot_status_change` call and this line overwrites the previous handle without unsubscribing it; each orphaned watch keeps a thread and gRPC stream alive that `close()` can't reach.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,929 @@
...
+                    parent_id = existing_doc.parent_id
+                    parent_state = None
+                    if parent_id:
+                        parent_recon = await self._reconstruct(transaction, parent_id, context=context)
+                        parent_state = parent_recon[1] if parent_recon else None
+                    candidate_patch = patch_to_json(diff_json(from_value=parent_state, to_value=new_state))
+                    if byte_length(candidate_patch) > self.shard_size:
+                        meta = self._write_checkpoint(transaction, sid, new_state, context=context)
+                    else:
+                        meta = SnapshotWriteMeta(
```
hilariie: Coverage shows 446-455 never execute, so the path that every heartbeat/abort/finalize takes on a diff turn is untested, including the oversize promotion at 452 that rewrites the doc as a checkpoint and drops its statePatch. The JS reference warns twice (firestore.ts:406, 432) that callers must only upsert the leaf because rewriting an interior snapshot invalidates descendants' diffs; that contract is neither documented nor tested here

py/packages/genkit-google-cloud/tests/firestore_session_store_test.py
```diff
@@ -0,0 +1,808 @@
...
+                    yield _doc(path=path, exists=True, data=self.docs[path], doc_id=ref.id)
+                else:
+                    yield _doc(path=path, exists=False, data=None, doc_id=ref.id)
+
+        # Production reads go through AsyncClient.get_all(transaction=...); keep
+        # transaction.get_all as a thin alias for older call sites/tests.
+        self.client.get_all = get_all
+        self.transaction.get_all = get_all
+
+        def txn_set(ref: Any, data: dict[str, Any]) -> None:  # noqa: ANN401
```
hilariie, line 220: the fake transaction applies writes immediately, never raises on read-after-write, and pins `_max_attempts = 1`, so the store's two riskiest behaviors (the real client raises ReadAfterWriteError on any read after a buffered write, and it retries the closure on Aborted) are untestable:

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,929 @@
...
+    async def on_snapshot_status_change(self, snapshot_id: str) -> asyncio.Queue[SnapshotStatus | None]:
+        """Subscribe to status changes for a session snapshot."""
+        context = get_current_context()
+        async with self.lock:
+            current = await self.read_snapshot(snapshot_id, context=context)
+            is_first = snapshot_id not in self.subs
+            q = await subscribe(subs=self.subs, snapshot_id=snapshot_id, current=current)
+            if current is not None and current.status in TERMINAL_STATUSES:
+                await q.put(None)
+                self.subs.pop(snapshot_id, None)
```
hilariie: This pops every subscriber for the snapshot but only the new queue got a None sentinel. A subscriber registered earlier, while the status was still live, is silently dropped. Its `watch_snapshot_abort` task blocks on `q.get()` forever because the watch's notify can no longer reach it.

huangjeff5: Fixed. We now notify all subscribers via `notify()` rather than just the latest queue.

py/packages/genkit-google-cloud/tests/firestore_session_store_test.py
huangjeff5: Updated the test to raise ReadAfterWriteError in the fake.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Good catch. Refactored code such that `subscribe()` registers the queue even when the doc isn't there yet, and `start_listener` unsubscribes any previous watch before overwriting. 

This makes sense from SessionStore perspective, but I believe we can't actually hit this scenario with the Agent abstraction because we always save the pending snapshot in Firestore before calling `on_snapshot_status_change`.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Added coverage for heartbeat/abort/finalize on an existing diff, including oversize promotion to a checkpoint (drops `statePatch`). Also documented on `save_snapshot` that upserts must target the current leaf. I looked carefully at JS and noticed that they document this behavior but don't have any explicit validation in code. I mirrored the behavior as JS.

huangjeff5:
Tested behavior against emulator and real Firestore and confirmed that the behavior matches in JS and Python across several different happy paths and failure scenarios. 

Removed isAmbiguous/leaves in pointer doc and matched JS's last-writer-wins semantics to simplify the codebase.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,1738 @@
...
+        The pointer is advanced for a new leaf, refreshed when the current
+        leaf is rewritten, and recreated when missing or invalid (TTL expiry,
+        corruption) — last writer wins. The document is replaced wholesale so
+        a reader never observes a partial update. Writes to other snapshots
+        leave the pointer untouched.
+
+        Callers must pass the pointer already loaded earlier in the same
+        transaction — Firestore rejects reads after any buffered writes.
+        """
+        if not (is_new or pointer is None or pointer.current_snapshot_id == snapshot_id):
```
hilariie, line 1253: When the pointer is missing or invalid, updating any existing snapshot recreates the pointer at that snapshot. For a root snapshot with a leaf descendant, deleting the pointer and performing a metadata-only update on the root makes session lookup return the root; rewriting its state can also invalidate the leaf’s patch

py/packages/genkit-google-cloud/tests/firestore_session_store_test.py
```diff
@@ -0,0 +1,2722 @@
...
+
+    _stream = await store.on_snapshot_status_change('snap-w')
+    assert len(captured_cb) == 1  # watch attached via the SYNC client
+    assert len(store._subscriptions) == 1
+    (sub,) = store._subscriptions
+    assert sub.watch is watches[0]
+
+
+@pytest.mark.asyncio
+async def test_firestore_session_store_retry_exhaustion_raises_aborted() -> None:
```
hilariie, line 1598: Could we also assert `h.commit_attempts == 2` here? Without that, this test still passes if `transaction_max_attempts` is ignored and the SDK uses its default, so it verifies error translation but not the configured retry bound.

py/packages/genkit-google-cloud/pyproject.toml
```diff
@@ -44,6 +44,7 @@ classifiers = [
 ]
 dependencies = [
   "genkit",
+  "google-cloud-firestore>=2.14.0",
```
hilariie, line 47: This package’s new root export depends on core APIs introduced in #6028, but genkit has no lower bound, and py/uv.lock does not record the new Firestore dependency for genkit-google-cloud. Could we set the core version floor and regenerate the lockfile?

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,1738 @@
...
+            existing_recon = await self._reconstruct(transaction, snapshot_id, prefix=prefix)
+            existing = self._to_snapshot(existing_recon) if existing_recon else None
+            try:
+                next_snapshot = apply_save(existing=existing, snapshot_id=snapshot_id, fn=fn)
+            except GenkitError:
+                raise
+            except BaseException as e:
+                raise _UserCodeError() from e
+            if next_snapshot is None:
+                return
```
hilariie: `committed` survives Firestore transaction retries. If the first attempt computes an update but loses its commit, and the retry observes newer data and returns None, `save_snapshot()` still returns the first attempt’s uncommitted value

py/packages/genkit-google-cloud/pyproject.toml
huangjeff5: I re-generated uv.lock. I'll set the version floor once PR 6028 is landed, in a separate release PR.

py/packages/genkit-google-cloud/tests/firestore_session_store_test.py
huangjeff5: Done.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Fixed this by returning the snapshot from the transactional callback. The SDK only keeps the attempt that committed, so a retry that no-ops can no longer return the first attempt's uncommitted value. Added a test for that race.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Ack, this is intentional - same behavior is JS's last-writer-wins behavior if pointer is gone. Leaving it as-is so the stores stay aligned.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
```diff
@@ -0,0 +1,1734 @@
...
+
+# Firestore transactions retry internally on contention; after this many
+# attempts the failure surfaces as GenkitError(status='ABORTED').
+DEFAULT_TRANSACTION_MAX_ATTEMPTS = 5
+
+_TERMINAL_STATUSES = frozenset({
+    SnapshotStatus.COMPLETED,
+    SnapshotStatus.FAILED,
+    SnapshotStatus.ABORTED,
+    SnapshotStatus.EXPIRED,
```
pavelgj: This local `_TERMINAL_STATUSES` includes `EXPIRED`, but the shared `TERMINAL_STATUSES` in `_util.py` deliberately leaves it out: per #6028, `expired` is a read-time overlay that's never persisted and never appears on the status stream. Since a Firestore watch only ever observes persisted statuses, `EXPIRED` here looks like it can't actually be hit, so it's dead-but-contradicts-the-contract and could mislead the next reader. Could we drop it to match the shared set, or add a comment if there's a path where a watch really can surface `expired`?

pavelgj (approved):
Nice work on this. The transaction atomicity + read-before-write concerns from the last round are genuinely resolved (all reads precede writes in the rmw callback, pointer + snapshot commit together, and the committed value now comes from the transactional callback so no stale-across-retry return). Test coverage is thorough (contention/retry with commit_attempts assertion, committed-value race, terminal-absorbing, notify-to-all-subscribers, aclose/close teardown including the listener-thread deadlock case). Left one minor non-blocking note about EXPIRED in _TERMINAL_STATUSES. LGTM.

py/packages/genkit-google-cloud/src/genkit_google_cloud/session_store/firestore.py
huangjeff5: Done, also shared the same constant enumerating the statuses.

PR #5758, opened 2026-07-17
test(py/plugins/anthropic): cover sampling parameter pass-through

### LINKED ISSUE #5634: [plugins] Anthropic Python: sampling-parameter guards for newest models
Parent: #5617 (Phase B, P1: item 10)

## Why

Sampling parameters (`temperature`, `top_p`, `top_k`, stop sequences, `max_tokens`) are passed through unchecked. Some of the newest models reject non-default sampling parameters under the current API, so a reasonable-looking config can fail with a raw 400 instead of a clear, local signal. Pass-through is the right default; the gap is the lack of guardrails for the models that don't accept it.

## What needs to be done

- Keep pass-through behavior for models that accept these parameters.
- Add validation and/or tests so configs that violate newest-model restrictions produce a clear signal rather than a raw API error.

## Acceptance

- Pass-through still works for supported models.
- Newest-model sampling restrictions are either validated locally or covered by tests so the behavior is explicit.

## Reference

- Current pass-through: `py/plugins/anthropic/src/genkit/plugins/anthropic/models.py:200-212`.

hilariie:
Closes #5634

Test-only change, stacked on #5720.

The issue asks for newest-model sampling restrictions to be either validated locally or made explicit with tests. This takes the test route, which matches what JS and Go already do: the newest models enforce their sampling rules server-side and the plugins forward the config unchanged. Duplicating those rules in the plugin would only drift as models and API behavior change, so instead the pass-through behavior is pinned down explicitly:

- All five sampling parameters (`temperature`, `top_p`, `top_k`, `stop_sequences`, and `max_output_tokens` as `max_tokens`) reach the SDK on both the non-streaming and streaming surfaces. The existing config test only checked `temperature`, `top_p`, and `max_tokens` on `messages.create`.
- The config test now uses `temperature=0.0`, so a truthiness check like `if params.get('temperature')` that silently drops falsy values would fail the suite.
- CamelCase aliases (`topP`, `topK`, `stopSequences`, `maxOutputTokens`) normalize to the snake_case SDK kwargs and do not leak into the request or `extra_body`.
- `claude-opus-4-8` and `claude-fable-5` forward sampling parameters unchanged, making the no-local-guards decision for the newest models explicit and keeping the plugin aligned with JS and Go.
- An empty config produces only the default `max_tokens`, with no stray sampling keys.

### DIFF
--- a/py/packages/genkit-anthropic/tests/anthropic_models_test.py
+++ b/py/packages/genkit-anthropic/tests/anthropic_models_test.py
@@ -197,18 +197,22 @@ async def test_generate_with_config() -> None:
     request = ModelRequest(
         messages=[Message(role=Role.USER, content=[Part(root=TextPart(text='Test'))])],
         config=ModelConfig(
-            temperature=0.7,
+            temperature=0.0,
             max_output_tokens=100,
             top_p=0.9,
+            top_k=40,
+            stop_sequences=['STOP'],
         ),
     )
 
     await model.generate(request)
 
     call_args = mock_client.messages.create.call_args
-    assert call_args.kwargs['temperature'] == 0.7
+    assert call_args.kwargs['temperature'] == 0.0
     assert call_args.kwargs['max_tokens'] == 100
     assert call_args.kwargs['top_p'] == 0.9
+    assert call_args.kwargs['top_k'] == 40
+    assert call_args.kwargs['stop_sequences'] == ['STOP']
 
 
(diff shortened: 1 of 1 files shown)


cabljac (approved):
Parity checks out, `stable.ts:243-258` and `toAnthropicRequest` in `go/plugins/internal/anthropic/anthropic.go:212` both forward sampling params unguarded.

PR #5795, opened 2026-07-22
feat(js/plugins/a2ui): add a2ui plugin

pavelgj:
Adds `@genkit-ai/a2ui`, a Genkit plugin that lets an agent stream **[A2UI](https://a2ui.org/)** ("Agent to UI") surfaces — rich, interactive UI rendered incrementally by the client — instead of just prose. Also adds a runnable `js/testapps/a2ui` sample (Express backend + Vite/Lit frontend).

> Status: proof-of-concept / exploration.

### How it works

A2UI travels on its own channel: a Genkit `data` part with mime type `application/a2ui+json` whose payload is an array of A2UI envelope messages (1:1 with the A2UI spec's A2A binding). The whole server integration is a single model middleware, `a2ui()`, added to an agent's `use` array. It injects the catalog's capabilities into the system prompt, then intercepts model output (streamed chunks and the final message), extracts `a2ui` fenced blocks, validates them against the catalog, and rewrites them into a2ui data parts.

### Server usage

```ts
import { genkit } from 'genkit/beta';
import { googleAI } from '@genkit-ai/google-genai';
import { a2ui, basicCatalog } from '@genkit-ai/a2ui';

const ai = genkit({ plugins: [googleAI()] });

export const uiAgent = ai.defineAgent({
  name: 'uiAgent',
  model: 'googleai/gemini-flash-latest',
  system: 'You help users. Render UI when it is clearer than prose.',
  use: [a2ui()], // <- the only A2UI-specific line
});
```

### Client usage

`@genkit-ai/a2ui/client` is browser-safe. Pull envelopes off each chunk with `a2uiEnvelopes` and feed them to a renderer (e.g. `@a2ui/lit`); send user actions (button presses, form submits) back with `actionToMessage`:

```ts
import { a2uiEnvelopes, actionToMessage } from '@genkit-ai/a2ui/client';
import { MessageProcessor } from '@a2ui/web_core/v0_9';
import { basicCatalog } from '@a2ui/lit/v0_9';
import { remoteAgent } from 'genkit/beta/client';

const chat = remoteAgent({ url: '/api/uiAgent' }).chat();
const processor = new MessageProcessor([basicCatalog], (action) =>
  runTurn({ message: actionToMessage(action) })
);

async function runTurn(input) {
  const turn = chat.sendStream(input);
  for await (const chunk of turn.stream) {
    if (chunk.text) appendProse(chunk.text);
    const envelopes = a2uiEnvelopesFromParts(chunk.raw.modelChunk.content);
    if (envelopes.length) processor.processMessages(envelopes);
  }
}

runTurn('weather in Tokyo');
```

### What's included

- **Plugin** (`js/plugins/a2ui`): `a2ui()` middleware, envelope parser/validator, the basic catalog + model-facing instructions, and the browser-safe `/client` entrypoint (`a2uiEnvelopes`, `actionToMessage`, `streamA2uiAgent`). Options: `catalog` (required), `instructions` (`'system'`), `validate` (`'strict'`), `surfaceId`, `version` (`'v0.9'`). Unit tests for the middleware and parser.
- **Testapp** (`js/testapps/a2ui`): Express server exposing the agent + a Vite/Lit chat UI that renders surfaces, with a `getWeather` demo tool and suggested prompts.

### Testing

- `pnpm -C js/plugins/a2ui test`
- Manual: `cd js/testapps/a2ui && genkit start -- pnpm start`, open http://localhost:8080 — tried weather, comparisons, and a signup form (values round-trip back to the agent).


js/plugins/a2ui/src/middleware.ts
```diff
@@ -0,0 +1,391 @@
...
+  version: z.string().optional(),
+});
+
+/** Configuration for the {@link a2ui} middleware. */
+export type A2uiOptions = z.infer<typeof A2uiOptionsSchema>;
+
+/** Type guard: is this part a text part? */
+function isTextPart(part: Part): part is TextPart {
+  return typeof (part as { text?: unknown }).text === 'string';
+}
```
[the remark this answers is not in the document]
pavelgj: Done. Added a `!!part && typeof part === 'object'` guard before reading `.text`.

js/plugins/a2ui/src/client.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/**
+ * Builds the agent input for sending a rendered surface's user action back to
+ * the agent as the next turn. The action's `name` becomes the user message so
+ * the agent can react to it; the full action object is attached as an a2ui data
+ * part for richer handling.
+ */
+export function actionToMessage(action: A2uiClientAction) {
+  const summary =
```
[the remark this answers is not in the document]
pavelgj: Done. Added a defensive check that throws if `action` is null/undefined.

js/plugins/a2ui/src/middleware.ts
```diff
@@ -0,0 +1,391 @@
...
+): GenerateRequest {
+  const text = renderCatalogInstructions(catalog);
+  const messages: MessageData[] = [...req.messages];
+  const sysIdx = messages.findIndex((m) => m.role === 'system');
+  if (sysIdx >= 0) {
+    const sys = messages[sysIdx];
+    messages[sysIdx] = {
+      ...sys,
+      content: [...sys.content, { text: '\n\n' + text }],
+    };
```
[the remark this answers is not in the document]
pavelgj: Skipping. At the model-middleware layer `MessageData.content` is schema-typed as `z.array(PartSchema)` (i.e. `Part[]`), never a bare string, so the string-splitting/`undefined` scenario can't occur here.

js/plugins/a2ui/src/middleware.ts
```diff
@@ -0,0 +1,391 @@
...
+    surfaceId: () => string;
+  }
+): GenerateResponseData {
+  const message = response.message;
+  if (!message?.content) return response;
+
+  const parser = new A2uiStreamParser(opts);
+  const newContent: Part[] = [];
+  for (const part of message.content) {
+    if (isTextPart(part)) {
```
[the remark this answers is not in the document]
pavelgj: Skipping, same reasoning as the system-message case: `response.message.content` is schema-typed as `Part[]`, not a string, and it's already guarded with `if (!message?.content) return`. The char-iteration scenario can't happen here.

js/plugins/a2ui/src/client.ts
```diff
@@ -0,0 +1,149 @@
...
+    headers: options.headers,
+    abortSignal: options.abortSignal,
+  });
+
+  for await (const chunk of response.stream) {
+    const mc = chunk?.modelChunk;
+    if (!mc?.content) continue;
+    for (const part of mc.content) {
+      if ('text' in part && typeof part.text === 'string' && part.text !== '') {
+        yield { type: 'text', text: part.text };
```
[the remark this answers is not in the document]
pavelgj: Stale, this targets an older revision. The current `streamA2uiAgent` uses `chunk.text` directly and no longer does the `in`-operator check that was flagged, so there's nothing to change here.

nan-yu:
Hi, I'm Nan from the A2UI team. We are excited to see Genkit adding native support for A2UI!

We wanted to give you a quick heads-up regarding a major refactoring and modularization effort we currently have in flight. These changes will introduce some breaking changes to the current message-processing, catalog, and inference pipelines.

To give you a clear look at where the APIs are heading, here are our active API interface:

- A2UI Core SDK: https://github.com/a2ui-project/a2ui/pull/1991
- A2UI Agent SDK: https://github.com/a2ui-project/a2ui/pull/1990

## Key Architectural Updates:

- **Refactoring a2ui-core:** We are refactoring a2ui-core to natively support multiple A2UI specification versions (like v0.9, v1.0 and more) and introducing sophisticated, deep-graph validations (e.g., checking for circular references, reachability, and duplicate IDs).
- **Modularizing a2ui-agent**: We are modularizing the a2ui-agent SDK to support multiple lightweight inference formats in order to reduce inference latency.
- **Support for Catalog Mixing**: Both a2ui-core and a2ui-inference are being updated to support catalog mixing. This will allow multiple catalogs to be dynamically combined at runtime, giving developers much greater layout flexibility.

## What this means for your implementation:

We are taking on the problem of standardizing prompt formatting, evaluations, and prompting techniques directly inside our SDKs. For example, building a native TypeScript SDK and a Dart SDK is actively on our roadmap for this quarter.

To avoid throwaway work, we suggest holding off on investing heavily in custom prompt-formatting, catalog management, or inference infrastructure on your side for now.

If you are happy to merge these plugins experimentally for now and migrate them to the updated SDK paths later, we are 100% supportive of that! We will gladly help you migrate and align with these blueprints when they are merged.

Let us know what you think, and thank you again for bringing A2UI into the Genkit ecosystem!

pavelgj:
> Hi, I'm Nan from the A2UI team. We are excited to see Genkit adding native support for A2UI!
> 
> We wanted to give you a quick heads-up regarding a major refactoring and modularization effort we currently have in flight. These changes will introduce some breaking changes to the current message-processing, catalog, and inference pipelines.
> 
> To give you a clear look at where the APIs are heading, here are our active API interface:
> 
> * A2UI Core SDK: [Update a2ui-core blueprint a2ui-project/a2ui#1991](https://github.com/a2ui-project/a2ui/pull/1991)
> * A2UI Agent SDK: [Update a2ui_agent blueprint with refined APIs a2ui-project/a2ui#1990](https://github.com/a2ui-project/a2ui/pull/1990)
> 
> ## Key Architectural Updates:
> * **Refactoring a2ui-core:** We are refactoring a2ui-core to natively support multiple A2UI specification versions (like v0.9, v1.0 and more) and introducing sophisticated, deep-graph validations (e.g., checking for circular references, reachability, and duplicate IDs).
> * **Modularizing a2ui-agent**: We are modularizing the a2ui-agent SDK to support multiple lightweight inference formats in order to reduce inference latency.
> * **Support for Catalog Mixing**: Both a2ui-core and a2ui-inference are being updated to support catalog mixing. This will allow multiple catalogs to be dynamically combined at runtime, giving developers much greater layout flexibility.
> 
> ## What this means for your implementation:
> We are taking on the problem of standardizing prompt formatting, evaluations, and prompting techniques directly inside our SDKs. For example, building a native TypeScript SDK and a Dart SDK is actively on our roadmap for this quarter.
> 
> To avoid throwaway work, we suggest holding off on investing heavily in custom prompt-formatting, catalog management, or inference infrastructure on your side for now.
> 
> If you are happy to merge these plugins experimentally for now and migrate them to the updated SDK paths later, we are 100% supportive of that! We will gladly help you migrate and align with these blueprints when they are merged.
> 
> Let us know what you think, and thank you again for bringing A2UI into the Genkit ecosystem!

Thanks for the heads up and feedback. This is definitely going out as preview and we'll have the opportunity to realign on the backend implementation once your SDKs firm up. That said, it looks like the backend SDK that you're going to ship would be considered "implementation detail" from Genkit's a2ui middleware POV, so I'm not too worried about it requiring a significant overhaul.

js/plugins/a2ui/src/middleware.ts
```diff
@@ -0,0 +1,389 @@
...
+                onChunk: (chunk: GenerateResponseChunkData) => {
+                  const transformed = transformChunk(chunk, streamParser);
+                  if (transformed) originalOnChunk(transformed);
+                },
+              }
+            : ctx;
+
+          // 3) Run downstream model, then transform the final message. The
+          //    final parse replays the same surface ids the stream minted.
+          const response = await next(request, wrappedCtx);
```
apascal07, line 266: `streamParser` is never flushed. `transformChunk` only calls `push()`, and after `next()` returns the code goes straight to `transformResponse`, which builds a fresh parser.

`drain(final=false)` always withholds the trailing `MAX_PARTIAL_FENCE` (8) chars as a possible partial fence, so the last <=8 characters of prose in every turn never reach `onChunk`. Streaming `"Hello there, friend!"` one char at a time:

```
streamed prose = "Hello there,"
withheld tail  = " friend!"
```

The final message recovers it (`transformResponse` does push + flush), but the testapp renders purely from stream deltas, so the user sees truncated prose. Same for an unterminated block at end of stream: it lands in `response.message` and never in the stream.

`parser_test.ts`'s `collect()` always calls `flush()`, which is why nothing catches this.

js/plugins/a2ui/src/parser.ts
```diff
@@ -0,0 +1,356 @@
...
+    // Guarantee the block opens with a `createSurface`, so the client always
+    // has a surface before any update targets it. Models often emit only
+    // `updateComponents`/`updateDataModel` on a follow-up (e.g. a "refresh")
+    // turn; without this the renderer would drop those updates as "surface not
+    // found". Idempotent re-creation is fine — it resets the surface.
+    const hasCreate = out.some(
+      (e) => (e as CreateSurfaceEnvelope).createSurface !== undefined
+    );
+    if (!hasCreate) {
+      out.unshift({
```
apascal07, line 285: The synthesized `createSurface` always uses the freshly-minted `surfaceId`, even when the batch's `updateComponents`/`updateDataModel` carries a different explicit id. `swapSurfaceId` only rewrites `undefined`/`''`/`SURFACE_ID`, and the model can learn a real id from `summarizeA2uiPart`, which feeds back `[UI action "refresh" on surface s1 ...]`.

Reproduced with a block whose only envelope targets `existing-surface`:

```
createSurface    { surfaceId: "NEW-UUID" }
updateComponents { surfaceId: "existing-surface" }
```

The client creates an empty surface and drops the update as "surface not found", which is the exact failure this block exists to prevent.

js/plugins/a2ui/src/parser.ts
```diff
@@ -0,0 +1,356 @@
...
+    if (!catalog) return null;
+    if (!Array.isArray(components)) {
+      return 'updateComponents.components must be an array.';
+    }
+    const known = new Set(catalog.components.map((c) => c.name));
+    const hasRoot = (components as A2uiComponent[]).some(
+      (c) => c.id === 'root'
+    );
+    if (!hasRoot) {
+      return 'component list must contain a component id "root".';
```
apascal07: This requires a `root` component in *every* `updateComponents`, so an incremental update to an existing surface is dropped under `warn` and throws under `strict`.

That's policy rather than validation: it follows from the "RE-RENDER THE WHOLE SURFACE" prompt rule, not from the catalog. A custom-catalog author reading `validate` as "check components against the catalog" won't expect it. Scoping the check to batches that also create the surface would keep partial updates viable.

js/plugins/a2ui/src/parser.ts
```diff
@@ -0,0 +1,356 @@
...
+  updateDataModel?: UpdateDataModelEnvelope['updateDataModel'];
+  deleteSurface?: DeleteSurfaceEnvelope['deleteSurface'];
+}
+
+/** Opening fence, matched case-insensitively (```a2ui). */
+const OPEN_FENCE_RE = /```[ \t]*a2ui[ \t]*\r?\n/i;
+/** The longest prefix of an opening fence, used to hold back a partial fence. */
+const MAX_PARTIAL_FENCE = '```a2ui\n'.length;
+/** Closing fence: ``` on its own (optionally indented) line, or end of text. */
+const CLOSE_FENCE_RE = /```/;
```
apascal07: This matches a fence anywhere, including inside a JSON string value. The catalog tells the model `Text` "may use inline Markdown", so a `Text` that mentions a code fence truncates the block mid-JSON and the whole surface is dropped:

```
A2UI: failed to parse envelope block as JSON: Unterminated string in JSON at position 104
```

Anchoring to line start (`/(^|\n)[ \t]*```/`) matches what `OPEN_FENCE_RE` already assumes.

js/plugins/a2ui/src/middleware.ts
```diff
@@ -0,0 +1,389 @@
...
+  validate: z.enum(['strict', 'warn', 'off']).optional(),
+
+  /**
+   * Surface id policy. Provide a fixed id to reuse for every surface. Defaults
+   * to a fresh UUID per surface.
+   */
+  surfaceId: z.string().optional(),
+
+  /** Protocol version stamped on emitted envelopes. Defaults to `'v0.9'`. */
+  version: z.string().optional(),
```
apascal07: `z.string()` here, but the parser stamps `version as SupportedVersion` (`'v0.9' | 'v0.9.1'`) with an unchecked cast. A typo like `'v0.10'` passes config validation and produces envelopes the renderer rejects at runtime. `z.enum(['v0.9', 'v0.9.1'])` makes the schema match the type.

js/plugins/a2ui/src/middleware.ts
```diff
@@ -0,0 +1,389 @@
...
+   */
+  instructions: z.enum(['system', 'none']).optional(),
+
+  /**
+   * Validate emitted envelopes against the catalog. `'warn'` (default) logs a
+   * warning and drops the offending block/envelope, keeping the rest of the
+   * turn alive; `'strict'` throws on malformed JSON or unknown components (best
+   * for development); `'off'` passes them through unchecked.
+   */
+  validate: z.enum(['strict', 'warn', 'off']).optional(),
```
apascal07, line 92: Worth stating in the doc comment that this validates component *type names* only, not props and not data-model values. `Image.url` and `Text` (inline Markdown, rendered to HTML by `@a2ui/markdown-it` in the testapp) stay fully model-controlled under `'strict'`.

That matters because generative UI moves model output into the DOM: a prompt-injected model can emit an arbitrary remote image URL or markdown the renderer may turn into HTML. `'strict'` reads like a safety boundary and isn't one. A short note here plus a trust-boundary section in the README (renderer/catalog owns prop sanitization; hosts should CSP-restrict image sources) would set the right expectation.

js/plugins/a2ui/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+ * `{ envelopes }` wrapping the array of A2UI envelopes, and whose
+ * `metadata.mimeType` is {@link A2UI_MIME_TYPE}.
+ *
+ * The array is wrapped in an object (rather than being the `data` value
+ * directly) so the payload is a `Map<String, dynamic>`-shaped object on every
+ * runtime — some (e.g. Dart) expect a data part's `data` to be an object, not a
+ * bare array.
+ */
+export interface A2uiPart {
+  data: { envelopes: A2uiEnvelope[] };
```
apascal07, line 153: `A2uiEnvelope` has no action variant, so this type is wrong for the client -> server direction. `actionToMessage` has to hand-build `{ data: { envelopes: [{ action }] } }` instead of using `a2uiPart()`, and `summarizeA2uiPart` re-narrows through a private `SummarizableEnvelope` to read it back. Adding an `ActionEnvelope` to the union (or a separate client part type) makes both directions typed and lets `a2uiPart()` be the single constructor.

js/plugins/a2ui/src/loader.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  if (!catalog || !Array.isArray(catalog.components)) {
+    throw new Error(
+      `loadCatalog(): catalog "${id}" must have a "components" array.`
+    );
+  }
+
+  // Register under the requested id. Keep the catalog's own `id` (used as the
+  // `catalogId` on surfaces) intact if present; otherwise default it to the id.
+  const registered: A2uiCatalog = { ...catalog, id: catalog.id ?? id };
```
apascal07: nit: `??` only falls back on null/undefined, so a catalog JSON file with `"id": ""` registers with an empty id and every surface then gets `catalogId: ""`. `||` covers it.

js/typedoc.json
```diff
@@ -1,6 +1,7 @@
 {
   "entryPoints": [
     "genkit",
+    "plugins/a2ui",
```
apascal07, line 4: `js/index.typedoc.md` has a matching table of plugin rows (`@genkit-ai/fetch`, `@genkit-ai/anthropic`, ...) that feeds the docs landing page. Without a row there, a2ui builds but isn't linked from anywhere.

js/plugins/a2ui/src/middleware.ts
pavelgj: Done

js/plugins/a2ui/src/parser.ts
pavelgj: Done

js/plugins/a2ui/src/parser.ts
pavelgj: Done

js/plugins/a2ui/src/middleware.ts
pavelgj: Done

js/plugins/a2ui/src/middleware.ts
pavelgj: Done

js/plugins/a2ui/src/types.ts
pavelgj: Done

js/plugins/a2ui/src/loader.ts
pavelgj: Done

js/typedoc.json
pavelgj: Done

apascal07 (approved):

js/plugins/a2ui/src/parser.ts
```diff
@@ -0,0 +1,427 @@
...
+interface RawEnvelope {
+  version?: string;
+  createSurface?: CreateSurfaceEnvelope['createSurface'];
+  updateComponents?: UpdateComponentsEnvelope['updateComponents'];
+  updateDataModel?: UpdateDataModelEnvelope['updateDataModel'];
+  deleteSurface?: DeleteSurfaceEnvelope['deleteSurface'];
+}
+
+/** Opening fence, matched case-insensitively (```a2ui). */
+const OPEN_FENCE_RE = /```[ \t]*a2ui[ \t]*\r?\n/i;
```
cabljac, line 55: With no anchoring like the closing one, if the model writes some kind of prose that mentions the fence mid-sentence

```
To render UI I emit an ```a2ui fenced block like this...
```
then could the parser block and buffer as JSON?

cabljac: Or is no anchor intentional here

cabljac: Also am i right in thinking this regex will permit some padding either side of a2ui, which might mess with the MAX_PARTIAL_FENCE line 154 ?

js/plugins/a2ui/src/parser.ts
```diff
@@ -0,0 +1,427 @@
...
+  }
+
+  /**
+   * Enforces the "RE-RENDER THE WHOLE SURFACE" protocol rule for full-surface
+   * renders: any `updateComponents` in the batch must contain a component with
+   * `id: "root"`. Returns an error message, or `null` if valid. Unlike
+   * {@link validateComponents} this is a protocol check, not a catalog check, so
+   * it runs regardless of whether a catalog is configured.
+   */
+  private validateRoot(envelopes: A2uiEnvelope[]): string | null {
```
cabljac, line 389: nit: I checked against the spec here and i think a split render (root and layout in one envelope, leaves in the next) is spec-legal but fails this check? The v0.9 spec has "One of the components in one of the component lists MUST have an id of root", which reads stream-level rather than per-message to me.  Would drop the block right?

I think this is internally consistent but maybe just worth a double check and soften the JSDoc protocol comment

cabljac (approved):
Nothing blocking that i can see. really clean. I've left 3 parser questions inline. Also checked it out and ran tests suites locally

js/plugins/a2ui/src/parser.ts
pavelgj: Anchoring is intentional: the model can open a block right after inline prose, so I rely on the `\r?\n` requirement instead (a mid-sentence mention like ```a2ui followed by text stays prose). The padding was a real bug though: `[ \t]*` made the partial-fence suffix unbounded, so the fixed holdback could leak backticks/tag into prose when a padded fence was split across chunks. Replaced the fixed `MAX_PARTIAL_FENCE` with a dynamic `PARTIAL_OPEN_FENCE_RE` that matches the actual incomplete-fence suffix at the buffer end, and added tests for split/padded fences.

js/plugins/a2ui/src/parser.ts
pavelgj: Good catch. Split renders are spec-legal, so I changed the per-envelope check to batch-level: at least one component list in the batch must declare `root`, matching the v0.9 wording ("one of the components in one of the component lists MUST have an id of `root`"). Softened the JSDoc to match, and added a passing split-render test plus a no-root-anywhere failure test.

PR #5807, opened 2026-07-23
feat(sdp): Initialize middleware and dependencies

agduan:
(1/6) This is the first PR in a stack implementing Sensitive Data Protection middleware for the `google-cloud plugin`. This PR sets up basic file structure and dependencies.

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/package.json
+++ b/js/plugins/google-cloud/package.json
@@ -54,6 +54,9 @@
     "node-fetch": "^3.3.2",
     "winston": "^3.12.0"
   },
+  "optionalDependencies": {
+    "@google-cloud/dlp": ">=3.0.0 <8.0.0"
+  },
   "peerDependencies": {
     "genkit": "workspace:^"
   },
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -14,9 +14,6 @@
  * limitations under the License.
  */
 
-export {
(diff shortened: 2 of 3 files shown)


js/plugins/google-cloud/package.json
```diff
@@ -54,6 +54,9 @@
     "node-fetch": "^3.3.2",
     "winston": "^3.12.0"
   },
+  "optionalDependencies": {
+    "@google-cloud/dlp": "^3.0.0"
+  },
```
jhuleatt, line 59: Precedent for this package seems to put everything in the `dependencies` array. Let's do the same with this dependency

jhuleatt (approved):
Please move the dlp dep out of `optionalDependencies`. Otherwise LGTM!

js/plugins/google-cloud/package.json
agduan: Advised against by Pavel, ignoring this suggestion

js/plugins/google-cloud/package.json
```diff
@@ -54,6 +54,9 @@
     "node-fetch": "^3.3.2",
     "winston": "^3.12.0"
   },
+  "optionalDependencies": {
+    "@google-cloud/dlp": "^3.0.0"
```
pavelgj: this is a very old version (6 years old)... latest is 7.0.1. Please see if you can use a range of versions covering the latest.

PR #5809, opened 2026-07-23
feat(sdp): Set up interface and options

agduan:
(2/6) Defines the configuration interfaces allowing users to configure SDP either inline or via Google Cloud Console templates.

[View clean diff](https://github.com/agduan/genkit/compare/sdp/1-deps...sdp/2-setup)

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -14,9 +14,6 @@
  * limitations under the License.
  */
 
-export {
-  FirestoreSessionStore,
-  type FirestoreSessionStoreOptions,
-} from './session-store/firestore.js';
+export interface SdpOptions {}
 
-export { sensitiveDataProtection, type SdpOptions } from './sdp/index.js';
+export function sensitiveDataProtection(options?: SdpOptions): void {}
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -14,6 +14,92 @@
  * limitations under the License.
  */
 
-export interface SdpOptions {}
+import { generateMiddleware } from 'genkit';
 
-export function sensitiveDataProtection(options?: SdpOptions): void {}
+// Option 1: Configure redaction options inline.
+export interface SdpInfoType {
+  name: string;
+  version?: string;
+}
+
+interface SdpInlineConfigBase {
+  /**
+   * Which infoTypes to inspect and replace.
+   * Default: ['CREDIT_CARD_NUMBER', 'EMAIL_ADDRESS', 'PHONE_NUMBER']
+   * All available infoTypes: https://cloud.google.com/sensitive-data-protection/docs/infotypes-reference#descriptions
+   */
+  infoTypes?: (
+    | 'CREDIT_CARD_NUMBER'
+    | 'EMAIL_ADDRESS'
+    | 'PHONE_NUMBER'
+    | 'STREET_ADDRESS'
+    | 'US_SOCIAL_SECURITY_NUMBER'
+    | 'IP_ADDRESS'
+    | 'PASSPORT'
+    | 'FINANCIAL_ACCOUNT_NUMBER'
+    | (string & {})
+    | SdpInfoType
+  )[];
+}
+
+/**
+ * The transformation method.
+ */
+export type SdpInlineConfig = SdpInlineConfigBase &
+  (
+    | {
+        /** Replaces with the type name (e.g., [EMAIL_ADDRESS]) */
+        transformation?: 'INFOTYPE';
+      }
+    | {
+        /** Replaces characters with a symbol (e.g., *****) */
+        transformation: 'MASK';
+        /** Masking configuration when transformation is 'MASK'. Defaults to '*' }
+         * Note: The masking string must be a single character.
+         */
+        maskConfig?: { maskingCharacter: string };
+      }
+    | {
+        /** Replaces with a fixed string (e.g., [REDACTED]) */
+        transformation: 'CUSTOM_STRING';
+        /** Custom string configuration when transformation is 'CUSTOM_STRING'. Defaults to '[REDACTED]' */
+        customConfig?: string;
+      }
+  );
+
+// Option 2: Create a custom config template in the Google Cloud Console.
+// Create in Google Cloud Console > Security > Sensitive Data Protection > Configuration.
+// Instructions: https://docs.cloud.google.com/sensitive-data-protection/docs/create-inspection-template
+export type SdpTemplateConfig =
+  | {
+      inspectTemplateName: string;
+      deidentifyTemplateName?: string;
+    }
+  | {
+      inspectTemplateName?: string;
+      deidentifyTemplateName: string;
+    };
+
+export interface SdpOptionsBase {
+  projectId?: string; // (Optional) Explicitly set the Google Cloud Project ID
+}
+
+export type SdpOptions = SdpOptionsBase &
+  (
+    | { templates: SdpTemplateConfig; inline?: SdpInlineConfig }
+    | { inline: SdpInlineConfig; templates?: SdpTemplateConfig }
+  );
+
+export const sensitiveDataProtection = generateMiddleware<SdpOptions>(
+  { name: 'sensitiveDataProtection' },
+  ({ config, pluginConfig }) => {
+    return {
+      generate: async (envelope, ctx, next) => {
+        // Intercept input
+        const res = await next(envelope, ctx);
+        // Intercept output
+        return res;
+      },
+    };
+  }
+);
--- a/js/plugins/google-cloud/package.json
+++ b/js/plugins/google-cloud/package.json
@@ -54,6 +54,9 @@
     "node-fetch": "^3.3.2",
     "winston": "^3.12.0"
   },
+  "optionalDependencies": {
+    "@google-cloud/dlp": ">=3.0.0 <8.0.0"
+  },
   "peerDependencies": {
     "genkit": "workspace:^"
   },
--- a/js/pnpm-lock.yaml
+++ b/js/pnpm-lock.yaml
@@ -776,6 +776,10 @@ importers:
       typescript:
         specifier: ^5.9.3
         version: 5.9.3
+    optionalDependencies:
+      '@google-cloud/dlp':
+        specifier: ">=3.0.0 <8.0.0"
+        version: 3.5.0(encoding@0.1.13)
 
   plugins/google-genai:
     dependencies:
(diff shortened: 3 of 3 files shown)


js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,70 @@
...
+import { generateMiddleware } from 'genkit';
+
+// Option 1: Configure redaction options inline.
+export interface SdpInlineConfig {
+  /**
+   * 1. Select which infoTypes to inspect and replace.
+   * Default: ['CREDIT_CARD_NUMBER', 'EMAIL_ADDRESS', 'PHONE_NUMBER']
+   * All available infoTypes: https://cloud.google.com/sensitive-data-protection/docs/infotypes-reference#descriptions
+   */
+  infoTypes?: string[];
```
jhuleatt: you can set this to be `IInfoType[]` instead of `string[]` https://docs.cloud.google.com/nodejs/docs/reference/dlp/latest/dlp/protos.google.privacy.dlp.v2.infotype

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { generateMiddleware } from 'genkit';
+
+// Option 1: Configure redaction options inline.
+export interface SdpInlineConfig {
+  /**
+   * 1. Select which infoTypes to inspect and replace.
```
jhuleatt:
```suggestion
   * Which infoTypes to inspect and replace.
```

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Default: ['CREDIT_CARD_NUMBER', 'EMAIL_ADDRESS', 'PHONE_NUMBER']
+   * All available infoTypes: https://cloud.google.com/sensitive-data-protection/docs/infotypes-reference#descriptions
+   */
+  infoTypes?: string[];
+
+  /**
+   * 2. Select the transformation method.
```
jhuleatt:
```suggestion
   * The transformation method.
```

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,70 @@
...
+   * - 'INFOTYPE' (default): Replaces with the type name (e.g., [EMAIL_ADDRESS])
+   * - 'MASK': Replaces characters with a symbol (e.g., *****)
+   * - 'CUSTOM_STRING': Replaces with a fixed string (e.g., [REDACTED])
```
jhuleatt: Above you link to infoTypes. Is there an equivalent doc for transformation methods? Maybe https://docs.cloud.google.com/sensitive-data-protection/docs/transformations-reference?

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  transformation?: 'INFOTYPE' | 'MASK' | 'CUSTOM_STRING';
+
+  /** Masking configuration when transformation is 'MASK'. Defaults to '*' } */
+  maskConfig?: { maskingCharacter: string };
+
+  /** Custom string configuration when transformation is 'CUSTOM_STRING'. Defaults to '[REDACTED]' */
+  customConfig?: string;
```
jhuleatt: since these fields are conditional on the `transformation` field, you could have a union type to enforce that each field can only be set for a certain value of `transformation`:

```ts
interface SdpInlineConfigMask extends SdpInlineConfig { /* TODO mask-specific config */ }
interface SdpInlineConfigCustomString extends SdpInlineConfig { /* TODO string-specific config */ }
interface SdpInlineConfigInfoType extends SdpInlineConfig { /* TODO infoType-specific config */ }

//... below

export interface SdpOptions {
  templates?: SdpTemplateConfig;
  inline?: SdpInlineConfigMask | SdpInlineConfigCustomString | SdpInlineConfigInfoType;

  projectId?: string; // (Optional) Explicitly set the Google Cloud Project ID
}
```

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Option 2: Create a custom config template in the Google Cloud Console.
+// Create in Google Cloud Console > Security > Sensitive Data Protection > Configuration.
+// Instructions: https://docs.cloud.google.com/sensitive-data-protection/docs/create-inspection-template
+export interface SdpTemplateConfig {
+  inspectTemplateName: string;
+  deidentifyTemplateName: string;
+}
```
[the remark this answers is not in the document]
jhuleatt: @agduan is this accurate?

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export interface SdpOptions {
+  templates?: SdpTemplateConfig;
+  inline?: SdpInlineConfig;
+
+  projectId?: string; // (Optional) Explicitly set the Google Cloud Project ID
+}
```
jhuleatt: you could again have a union type, SdpOptionsTemplate | SdpOptionsInline, to enforce that one of `templates` or `inline` is needed.

jhuleatt (changes requested):

js/plugins/google-cloud/src/sdp/index.ts
agduan: Yes, but I didn't hyperlink because the middleware only supports the three methods listed in the comments. Most of the methods in the link (crypto, bucketing, replaceDictionaryConfig) don't work. Should I still link to the list?

js/plugins/google-cloud/src/sdp/index.ts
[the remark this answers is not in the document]
agduan: Yes (although kind of a niche case). I can make this modification and add 5-10 LOC of validation to check that at least one template is non-empty?

js/plugins/google-cloud/src/sdp/index.ts
agduan: I can, but `IInfoType` is from `protos` in `@google-cloud/dlp`. Since `@google-cloud/dlp` is an optional dependency, I think directly importing IInfoType would break TypeScript builds for anyone who imports from `beta` (e.g., to use FirestoreSessionStore) but does not have dlp installed.

To avoid, I can either keep the type as-is or define a custom type for this?

js/plugins/google-cloud/src/sdp/index.ts
agduan: Applied

js/plugins/google-cloud/src/sdp/index.ts
agduan: Applied

js/plugins/google-cloud/src/sdp/index.ts
agduan: Will do! Could I do this by intersecting a base type with a union instead? Either sounds good to me.
```js
export interface SdpInfoType {
  name: string;
  version?: string;
}

// 1. Common fields that apply to all transformations
interface SdpInlineConfigBase {
  infoTypes?: (string | SdpInfoType)[];
}

// 2. The discriminated union for the transformation types
export type SdpInlineConfig = SdpInlineConfigBase & (
  | { 
      transformation?: 'INFOTYPE' 
    }
  | { 
      transformation: 'MASK'; 
      maskConfig?: { maskingCharacter: string }; 
    }
  | { 
      transformation: 'CUSTOM_STRING'; 
      customConfig?: string; 
    }
);
```

js/plugins/google-cloud/src/sdp/index.ts
agduan: Good idea! Should I instead structure the union as done below, so users can use both (e.g., a template for inspecting and inline code for de-identification)?
```ts
export type SdpOptions = SdpOptionsBase & (
  | { templates: SdpTemplateConfig; inline?: SdpInlineConfig }
  | { inline: SdpInlineConfig; templates?: SdpTemplateConfig }
);
```
^ If this sounds like too complicated of a use case, I can do your fix instead.

js/plugins/google-cloud/src/sdp/index.ts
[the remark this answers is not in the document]
jhuleatt: If at least one is required, you can define that with a union (though a validation check would be useful too, for JS users):

```ts
type SdpTemplateConfig =
  | {
      inspectTemplateName: string;
      deidentifyTemplateName?: string;
    }
  | {
      inspectTemplateName?: string;
      deidentifyTemplateName: string;
    };
```

js/plugins/google-cloud/src/sdp/index.ts
jhuleatt: If this stays an optional dependency, it's fine to keep as `string[]`.

js/plugins/google-cloud/src/sdp/index.ts
jhuleatt: Got it, no need to link to the list. If there's a specific thing preventing some methods from being used, that could be an interesting thing to note here

js/plugins/google-cloud/src/sdp/index.ts
jhuleatt: Your call! As long as it makes the conditional fields clear to the person using this middleware, I'm happy

js/plugins/google-cloud/src/sdp/index.ts
jhuleatt: Your approach sounds good to me

js/plugins/google-cloud/src/sdp/index.ts
jhuleatt: All right, turns out IInfoType doesn't even list the valid infotypes anyway. So, one better than `string[]`:

```ts
infoTypes?: Array<'CREDIT_CARD_NUMBER' | 'EMAIL_ADDRESS' | 'PHONE_NUMBER' | (string & {})>
```

You can list out a few common ones, and then `(string & {})` allows any string but stops TS from just collapsing everything to `string`.

js/plugins/google-cloud/src/sdp/index.ts
agduan: Nope, the main blockers are 1) bucketing-related methods are meant to be used on CSV data only, and 2) it's too much overhead for the sample to manage a KMS key for cryptographic methods. Leaving the comments as-is for now.

js/plugins/google-cloud/src/sdp/index.ts
agduan: Implemented my intersection + discriminated union

js/plugins/google-cloud/src/sdp/index.ts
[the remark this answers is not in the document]
agduan: Done! Going to leave out the validation check for now to avoid duplicated logic/clutter.

js/plugins/google-cloud/src/sdp/index.ts
agduan: Done! Again, I didn't add validation to check against empty objects from JS users to avoid code bloat, but I can if necessary

js/plugins/google-cloud/src/sdp/index.ts
agduan: Done, listed out a few more types!

jhuleatt (approved):

PR #5810, opened 2026-07-23
feat(sdp): Initialize DLP client

agduan:
(3/6) Implements the DLP client initialization logic, handling credentials and lazy loading.

[View diff](https://github.com/agduan/genkit/compare/sdp/2-setup...sdp/3-client)

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -14,9 +14,6 @@
  * limitations under the License.
  */
 
-export {
-  FirestoreSessionStore,
-  type FirestoreSessionStoreOptions,
-} from './session-store/firestore.js';
+export interface SdpOptions {}
 
-export { sensitiveDataProtection, type SdpOptions } from './sdp/index.js';
+export function sensitiveDataProtection(options?: SdpOptions): void {}
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
+import type { v2 } from '@google-cloud/dlp';
 import { generateMiddleware } from 'genkit';
+import { credentialsFromEnvironment } from '../auth.js';
 
 // Option 1: Configure redaction options inline.
 export interface SdpInfoType {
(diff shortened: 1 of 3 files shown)


js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  projectId?: string; // (Optional) Explicitly set the Google Cloud Project ID
+}
+
+export async function createDlpClient(
+  options: SdpOptions
+): Promise<v2.DlpServiceClient> {
+  let dlpModule;
+  try {
+    dlpModule = await import('@google-cloud/dlp');
```
jhuleatt, line 95: Do the other Cloud middleware modules do this async import, or do they just import at the top of the file? Follow their pattern

jhuleatt: Ooh, this is because it's an `optionalDependency`. See https://github.com/genkit-ai/genkit/pull/5807#discussion_r3647267879, lmk if you disagree

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  } catch (e) {
+    throw new Error(
+      'Please install the @google-cloud/dlp package to use the SDP middleware.'
+    );
+  }
+
+  const envAuth = await credentialsFromEnvironment();
+  const projectId = options.projectId || envAuth.projectId;
```
jhuleatt: if someone provides their own projectId, should they provide their own credentials too? If so, you only need to call credentialsFromEnvironment if they haven't provided their own config. In another spot, credentials are able to be passed in:

https://github.com/genkit-ai/genkit/blob/bc4c733f54ba52b53d21304334b9a7c126240898/js/plugins/google-cloud/src/index.ts#L58-L59

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  return new dlpModule.v2.DlpServiceClient({
+    credentials: envAuth.credentials as any,
```
jhuleatt: why does this need to be cast to `any`?

js/plugins/google-cloud/src/sdp/index.ts
agduan: Decided to keep it as an `optionalDependency`, resolving this suggestion

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    projectId: projectId,
+  });
```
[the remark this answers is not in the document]
agduan: Incorrect, `credentialsFromEnvironment()` will always return an object.

js/plugins/google-cloud/src/sdp/index.ts
agduan: Good catch, thanks!

js/plugins/google-cloud/src/sdp/index.ts
agduan: Type incompatibility: envAuth.credentials is typed as JWTInput (from google-auth-library), but the DLP API uses a type from an internal library for credentials. Both types have identical info, but TypeScript might throw a mismatch error since they're from different modules

jhuleatt: You can be more specific by using the `as any as T` pattern. Casting it as any so that typescript lets you cast it to the thing you want

agduan: Done. Instead of importing google-auth-library, I used a TS utility extraction for type `(ConstructorParameters<typeof v2.DlpServiceClient>[0]['credentials'])`

jhuleatt (approved):

PR #5812, opened 2026-07-23
feat(sdp): Implement direct redaction logic

agduan:
(5/6) Implements the core middleware logic: intercepts the prompt message parts, checks them against the Google Cloud DLP service, and applies the configured redaction.

[View diff](https://github.com/agduan/genkit/compare/sdp/4-config...sdp/5-redact)

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -14,9 +14,6 @@
  * limitations under the License.
  */
 
-export {
-  FirestoreSessionStore,
-  type FirestoreSessionStoreOptions,
-} from './session-store/firestore.js';
+export interface SdpOptions {}
 
-export { sensitiveDataProtection, type SdpOptions } from './sdp/index.js';
+export function sensitiveDataProtection(options?: SdpOptions): void {}
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -88,7 +88,7 @@ export interface SdpOptionsBase {
   credentials?: any; // (Optional) Explicitly set the Google Cloud credentials
 }
 
-export async function createDlpClient(
+async function createDlpClient(
   options: SdpOptions
 ): Promise<v2.DlpServiceClient> {
   let dlpModule;
@@ -123,7 +123,7 @@ export type SdpOptions = SdpOptionsBase &
   );
 
 // 1. Inspect config
-export function buildInspectConfig(
+function buildInspectConfig(
   options: SdpOptions
 ): protos.google.privacy.dlp.v2.IInspectConfig {
   const defaultInfoTypes = [
@@ -150,7 +150,7 @@ export function buildInspectConfig(
 }
 
 // 2. De-identify config
-export function buildDeidentifyConfig(
+function buildDeidentifyConfig(
   options: SdpOptions
 ): protos.google.privacy.dlp.v2.IDeidentifyConfig {
   const inlineConfig =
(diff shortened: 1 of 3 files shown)


js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,206 @@
...
+            }
+          }
+        }
+
+        // Wait for all network calls to finish in parallel
+        await Promise.all(redactionPromises);
+
+        const res = await next(envelope, ctx);
+        // Intercept output
+        return res;
```
jhuleatt, line 292: Is redacting the model response also part of this middleware?

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,206 @@
...
+                  });
+                  redactionPromises.push(promise);
+                }
+              }
+            }
+          }
+        }
+
+        // Wait for all network calls to finish in parallel
+        await Promise.all(redactionPromises);
```
jhuleatt: I think this is a little easier to follow with Array methods, but ultimately a style choice

```suggestion
if (envelope.request?.messages) {
  const redactionPromises = envelope.request.messages
    // ignore empty messages
    .filter((message) => !!message.content)
    // extract all message content into a single array
    .flatMap((message) => message.content)
    // ignore multimedia content and content that has been cleaned already
    .filter((part) => part.text && !part.metadata?.isCleaned)
    // de-identify message content
    .map(async (part): Promise<void> => {
      part.text = await sanitizeInput(client, part.text, opts, projectId);
      part.metadata = { ...part.metadata, isCleaned: true };
    });
  await Promise.all(redactionPromises);
}
```

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  };
+}
+
+export async function sanitizeInput(
+  dlp: v2.DlpServiceClient,
+  text: string,
+  options: SdpOptions,
+  projectId: string
+) {
+  const request: any = {
```
jhuleatt: Best to either pick a narrower type than `any`, or omit type assignment

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,206 @@
...
+      client: v2.DlpServiceClient;
+      projectId: string;
+    }> | null = null;
+    return {
+      generate: async (envelope, ctx, next) => {
+        const opts = options || {};
+        if (!clientPromise) {
+          clientPromise = Promise.all([
+            createDlpClient(opts),
+            credentialsFromEnvironment(),
```
jhuleatt: since `createDlpClient` calls `credentialsFromEnvironment` itself, could `createDlpClient` return all the necessary info instead of multiple calls to `credentialsFromEnvironment`?

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  const envAuth = await credentialsFromEnvironment();
+  const projectId = options.projectId || envAuth.projectId;
+
+  return new dlpModule.v2.DlpServiceClient({
+    credentials: envAuth.credentials as any,
+    projectId: projectId,
+  });
+}
```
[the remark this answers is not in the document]
agduan: Resolved from above edit

js/plugins/google-cloud/src/sdp/index.ts
agduan: Yes! Resolved from an earlier Gemini suggestion - credentials are now fetched once and passed directly into createDlpClient.

js/plugins/google-cloud/src/sdp/index.ts
agduan: Yes, looks nicer this way. Done!

js/plugins/google-cloud/src/sdp/index.ts
agduan: Whoops, it should be. Just added a block of the same (array-based) logic to sanitize the output.

js/plugins/google-cloud/src/sdp/index.ts
agduan: Omitted type assignment by using the spread operator to build it in one step

agduan: Never mind. Since we were able to import `protos`, I picked the narrower type `IDeidentifyContentRequest` and removed the clunky-looking spread operator implementation.

jhuleatt (approved):

PR #5817, opened 2026-07-27
fix: check license headers in .mjs and .cjs files

### LINKED ISSUE #5624: Repo-wide license header check only runs on Python PRs
## Summary

Missing license headers in some JS file types can merge to `main` without CI
catching them, then surface later on an unrelated PR that happens to touch
`py/**`. We just hit this: two `.mjs` files added in #5426 failed the license
check on a Python PR that did not add them.

## Root cause: two header checks with different coverage

The repo has two separate header checks, and they do not cover the same files.

1. `scripts/copyright.ts`, run on every PR via `pnpm format:check` in
   `formatter.yml` (no path filter). It only recognizes these extensions:

   ```ts
   { regex: /\.((ts)|(scss)|(js))$/, ... }  // .ts .scss .js
   { regex: /\.html$/, ... }                // .html
   { regex: /\.go$/, ... }                  // .go
   ```

   So `.mjs`, `.cjs`, `.tsx`, `.jsx`, and `.css` are never checked.

2. `bin/check_license` (addlicense), in the "Security and Compliance" job in
   `python.yml`. It scans the whole repo and does recognize `.mjs`, but the
   workflow is path filtered:

   ```yaml
   on:
     pull_request:
       paths:
         - "py/**"
         - "genkit-tools/**"
         - ".github/workflows/python.yml"
   ```

   So it only runs when a PR touches Python.

## Why #5426 slipped through

#5426 added `eslint.config.mjs` and `postcss.config.mjs` with no header.

- The always-on check (`scripts/copyright.ts`) ignores `.mjs`, so it passed.
- The check that catches `.mjs` (addlicense) never ran, because the PR was JS
  only and did not touch `py/**`.

The next PR to touch Python ran addlicense, which scans everything, and failed
on the two `.mjs` files. The failure pointed at the wrong PR and author.

## Fix options

1. Add the missing extensions (`.mjs`, `.cjs`, `.tsx`, `.jsx`, `.css`) to
   `scripts/copyright.ts` so the always-on check covers them. Smallest change.
2. Run `bin/check_license` on all PRs instead of only `py/**`, so the
   comprehensive scan is not gated behind Python.
3. Settle on one tool as the source of truth so coverage cannot drift between
   the two.

hugosmoreira:
## What changed

Extend the always-on copyright checker to cover `.mjs` and `.cjs` files.

The current regex already covers `.ts`, `.tsx`, `.js`, `.jsx`, `.css`, and
`.scss`. This change preserves that set while adding the two JavaScript module
extensions that were still only caught by the Python-gated `addlicense` check.

## Why

Without this coverage, a JavaScript-only PR can add a headerless `.mjs` or
`.cjs` file and pass the always-on formatter workflow. The missing header then
surfaces later on an unrelated Python PR because that workflow scans the whole
repository.

Fixes #5624.

## Validation

- Before the change, `scripts/copyright.ts --check` returned exit code 0 with
  untracked, headerless `.mjs` and `.cjs` fixtures present.
- After the change, the same command returned exit code 1 and named both
  fixtures.
- After removing the fixtures, `pnpm format:check` passed in an isolated
  LF-checkout worktree.
- `pnpm build` passed in the same isolated worktree, including JS libraries,
  doc snippets, test apps, and Genkit tooling.

AI assistance was used during investigation and validation. The final change
was narrowed against current `main` and exercised through the repository's own
checker.

### DIFF
--- a/scripts/copyright.ts
+++ b/scripts/copyright.ts
@@ -27,7 +27,7 @@ interface FormatType {
 const FILE_OPTS: { encoding: 'utf-8' } = { encoding: 'utf-8' };
 const FORMAT_TYPES: FormatType[] = [
   {
-    regex: /\.((tsx?)|(s?css)|(jsx?))$/,
+    regex: /\.(tsx?|s?css|[cm]?js|jsx)$/,
     header: '/**',
     body: ' *',
     footer: ' */',


hugosmoreira:
@googlebot rescan

hugosmoreira:
Updated in 9f475d2 to preserve leading hashbangs when adding headers, with regression coverage for `.js`, `.mjs`, and `.cjs`. All 27 focused cases pass on Node 22 and 24, and the PR description now records the baseline comparison and broader verification status.

The six new verification workflows currently show `action_required` with zero jobs executed. Could a maintainer approve the latest runs when convenient? The focused regression suite is also available as `pnpm test:copyright` (not automatically invoked by the unchanged workflows). Thanks!

PR #5819, opened 2026-07-27
feat(go): add qwen model support

### LINKED ISSUE #5802: feat(go/plugins/compat_oai): add dedicated Alibaba Qwen subplugin
Alibaba Qwen (DashScope) has no compat_oai subplugin. Template: #5461.

IzaakGough:
Closes: #5802 

Checklist (if applicable):


adesinah:
Nice start overall. I compared this with the JS OpenAI-compatible implementation and the current DashScope documentation. A few things should be addressed before merging:

- If `DASHSCOPE_API_KEY` is missing, the OpenAI client may fall back to `OPENAI_API_KEY` and send that key to DashScope. We should fail fast instead. Adding explicit `APIKey` and `BaseURL` fields, like the Kimi and Z.AI plugins, would also make configuration consistent.
- The models advertise dated `Versions`, but the shared adapter currently removes `version` and sends the registered model name. The selected version should become the outbound `model`.
- The multimodal metadata is incomplete. DashScope documents `qwen3.5-flash`, `qwen3.5-plus`, `qwen3.6-flash`, `qwen3.6-plus`, and `qwen3.7-plus` as supporting image/video input, but only `qwen3-vl-plus` currently has `Media: true`.
- Reasoning and provider-specific options such as `enable_thinking` should work through the shared adapter, but there are no DashScope tests proving complete, streaming, or tool-call behavior.
- Supported models should advertise JSON output where appropriate. We should be careful with `Constrained`, though: DashScope documents JSON-object mode, not general OpenAI JSON Schema support.
- Tool choice is supported with model/thinking-mode restrictions. Leaving `ToolChoice` disabled under-reports the capability, while enabling it unconditionally could allow invalid requests. This needs provider-aware handling or clearly documented validation.
- There are currently no deterministic provider tests. The live test is skipped in CI and only covers basic generation. We should add `httptest` coverage for authentication/config precedence, registration metadata, versions, reasoning, tools, JSON mode, streaming, and multimodal requests.
- The README's live-test command references tests that do not exist in this package. It would be good to standardize the entry point as `TestPluginLive`.

I'd treat the API-key fallback, version handling, multimodal metadata, and deterministic tests as the main blockers. The remaining items can be handled here or explicitly tracked as follow-ups.

IzaakGough:
@adesinah, currently working through the blocking points, and had a couple of questions:

1. **Versions → model.** 
Looks like config.version is validated but never actually overrides the outbound model, seems to be a shared-adapter bug, not dashscope-specific (openai/anthropic/gemini have it too). What would you suggest I do here? 
Happy to fix it centrally now, or ship dashscope matching its siblings for now and track the bigger fix separately, whichever you'd prefer.

2. **Tool choice.** 
No plugin currently does model-conditional tool-choice validation. Would you like me to follow zai's approach of setting ToolChoice: false and adding a README note or build out real validation?

adesinah:
> @adesinah, currently working through the blocking points, and had a couple of questions:
> 
> 1. **Versions → model.**
>    Looks like config.version is validated but never actually overrides the outbound model, seems to be a shared-adapter bug, not dashscope-specific (openai/anthropic/gemini have it too). What would you suggest I do here?
>    Happy to fix it centrally now, or ship dashscope matching its siblings for now and track the bigger fix separately, whichever you'd prefer.
> 2. **Tool choice.**
>    No plugin currently does model-conditional tool-choice validation. Would you like me to follow zai's approach of setting ToolChoice: false and adding a README note or build out real validation?

On the **Versions**, agreed. It is a shared-adapter issue rather than something introduced by DashScope. Let’s keep this PR consistent with the existing providers and track the central fix separately.

On the **Tool choice**: Yes.

adesinah (approved):
Looks good to me.

PR #5823, opened 2026-07-27
fix(py): normalize Gemini model request message roles strictly to model or user

huangjeff5:
Normalizes Gemini content message roles so non-model roles fall back to user.


MichaelDoyle (approved):
Thanks!

PR #5825, opened 2026-07-28
fix(js): set a message on ToolInterruptError

### LINKED ISSUE #5821: [JS] set a "message" on ToolInterruptError
We currently do not set a message on `ToolInterruptError` because it is not intended to be user facing. Instead, it serves as a yield mechanism to be handled by the generate loop.

However, it is totally possible to "call" a tool, that throws an interrupt, outside of generate. When this happens, it's very confusing to debug that an exception was thrown, because the Error does not conform to that of a typical TS error.

So, even though its not user facing, there are indeed scenarios where we can expect it to be visible (a) calling manually outside of a generate loop (b) calling through dev ui, etc. The error may also find its way into logs (locally and remote observability tools) where its expected to have a string representation (via message).

**Context:** we had to do some workarounds on the dev-ui to "detect" and show it properly. In some respects, this is a necessary evil. But in other respects, it was unexpected that we couldn't String(error) on it and get a reasonable result.

a2105z:
## Summary
- Set a `message` on `ToolInterruptError` so `String(error)` / logging produce a useful representation when tools are invoked outside the generate loop (manual calls, Dev UI, observability).
- Align the JS message format with the existing Go `toolInterruptError.Error()` behavior: `"tool execution interrupted"` with optional pretty-printed metadata.
- Add unit tests covering both the no-metadata and metadata message cases.

Fixes #5821

## Test plan
- [x] `js/ai` `tool_test.ts` — new coverage for `ToolInterruptError` message with and without metadata
- [x] Aligns stringification with Go `toolInterruptError.Error()` behavior
- [ ] Manual: invoke an interrupt tool outside `generate` and confirm `error.message` / `String(error)` is non-empty
- [ ] Manual: confirm Dev UI / logs show a readable interrupt error string

### DIFF
--- a/js/ai/src/tool.ts
+++ b/js/ai/src/tool.ts
@@ -523,10 +523,15 @@ export function defineInterrupt<I extends z.ZodTypeAny, O extends z.ZodTypeAny>(
  */
 export class ToolInterruptError extends Error {
   constructor(readonly metadata?: Record<string, any>) {
-    const message =
-      metadata !== undefined
-        ? `tool execution interrupted: \n\n${JSON.stringify(metadata, null, 2)}`
-        : 'tool execution interrupted';
+    let message = 'tool execution interrupted';
+    if (metadata !== undefined) {
+      try {
+        message += `: \n\n${JSON.stringify(metadata, null, 2)}`;
+      } catch {
+        // Keep construction safe if metadata has cycles / BigInt / etc.
+        message += ': \n\n[unserializable metadata]';
+      }
+    }
     super(message);
     this.name = 'ToolInterruptError';
   }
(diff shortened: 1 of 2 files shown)


a2105z:
Friendly ping: the gemini review feedback is addressed (safe metadata stringify + regression test), and CLA is passing.

Could a maintainer please approve the pending fork workflows and take a look when you have a chance? Happy to make any follow-up changes.

a2105z:
Hi @MichaelDoyle @pavelgj @shrutip90 — whenever you have a moment, could someone take a look at this PR and approve the pending fork workflows so the required checks can run?

The gemini review feedback has been addressed, and CLA is passing. Thanks so much!

a2105z:
Hi @genkit-ai/genkit-googlers @pavelgj @apascal07 @huangjeff5 @ifielker @MichaelDoyle @yesudeep @ssbushi @shrutip90 @cabljac @adesinah — circling back on this one.

It has been open since late July and is still waiting on a maintainer review. Would someone be able to take a look when you have a moment? Happy to address any feedback. Thank you!

PR #5828, opened 2026-07-28
feat(go): add z.AI model support

### LINKED ISSUE #5800: feat(go/plugins/compat_oai): add dedicated Z.ai (Zhipu / GLM) subplugin
Z.ai / GLM family has no compat_oai subplugin. Template: #5461.

adesinah:
This PR adds z.AI model support.

Closes: #5800 

Checklist (if applicable):

### DIFF
--- a/go/plugins/compat_oai/generate.go
+++ b/go/plugins/compat_oai/generate.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/internal/base"
@@ -29,12 +30,11 @@ import (
 
 // ModelGenerator handles OpenAI generation requests
 type ModelGenerator struct {
-	client     *openai.Client
-	modelName  string
-	request    *openai.ChatCompletionNewParams
-	messages   []openai.ChatCompletionMessageParamUnion
-	tools      []openai.ChatCompletionToolParam
-	toolChoice openai.ChatCompletionToolChoiceOptionUnionParam
+	client    *openai.Client
+	modelName string
+	request   *openai.ChatCompletionNewParams
+	messages  []openai.ChatCompletionMessageParamUnion
+	tools     []openai.ChatCompletionToolParam
 	// Store any errors that occur during building
 	err error
 }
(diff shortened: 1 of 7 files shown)


go/plugins/compat_oai/generate.go
```diff
@@ -154,14 +159,52 @@ func (g *ModelGenerator) WithConfig(config any) *ModelGenerator {
...
+		extraFields := make(map[string]any, len(normalizedConfig))
+		for key, value := range normalizedConfig {
+			switch key {
+			case "messages", "model", "response_format", "stream", "stream_options", "tool_choice", "tools":
+				continue
+			default:
+				extraFields[key] = value
+			}
+		}
+		openaiConfig.SetExtraFields(extraFields)
```
cabljac, line 207: Should the extraFields passthrough be scoped to zai, or is wire-through for all compat_oai providers the intent? Before, `map[string]any{"foo": 1}` was silently dropped for openai/anthropic callers; now it goes on the wire, and OpenAI 400s on unrecognized args. JS parity is the right end state, just flagging as a behaviour change riding outside the zai/ dir, worth a line in the PR description so it doesn't surprise anyone on upgrade.

go/plugins/compat_oai/generate.go
```diff
@@ -67,13 +67,18 @@ func (g *ModelGenerator) WithMessages(messages []*ai.Message) *ModelGenerator {
...
-		content := g.concatenateContent(msg.Content)
+		content := concatenateTextContent(msg.Content)
 		switch msg.Role {
 		case ai.RoleSystem:
 			oaiMessages = append(oaiMessages, openai.SystemMessage(content))
 		case ai.RoleModel:
 			am := openai.ChatCompletionAssistantMessageParam{}
 			am.Content.OfString = param.NewOpt(content)
+			if reasoning := concatenateReasoningContent(msg.Content); reasoning != "" {
+				am.SetExtraFields(map[string]any{
```
cabljac, line 78: Is sending reasoning_content back on assistant messages intended for every provider, or just GLM? For GLM it's actually required (z.ai docs say clear_thinking: false needs the full historical reasoning_content forwarded), but cross-model histories hit the same path: generate with a reasoning model, continue on openai, and the assistant message now carries a field OpenAI doesn't document. Before this PR reasoning text got concatenated into content. Just flagging as a small behaviour change, a test pinning the openai path with reasoning parts in history would lock in whichever is intended.

go/plugins/compat_oai/generate.go
```diff
@@ -29,12 +30,11 @@ import (
 
 // ModelGenerator handles OpenAI generation requests
 type ModelGenerator struct {
-	client     *openai.Client
-	modelName  string
-	request    *openai.ChatCompletionNewParams
-	messages   []openai.ChatCompletionMessageParamUnion
-	tools      []openai.ChatCompletionToolParam
-	toolChoice openai.ChatCompletionToolChoiceOptionUnionParam
```
cabljac, line 37: Good cleanup, this field was never read or written. It does look like it surfaces a pre-existing gap though: compat_oai declares ToolChoice: true for openai/anthropic models (compat_oai.go:44) but the generator never maps ai.ToolChoice into tool_choice on the wire, so WithToolChoice(ai.ToolChoiceRequired) passes core validation and silently does nothing. Do you agree? If so we should probably file a follow-up issue for the openai/anthropic side.

cabljac (approved):

PR #5832, opened 2026-07-28
fix(js/ai): search history newest-first in validateResumeAgainstHistory

huangjeff5:
History is searched newest-first so a resume matches the most recent tool request for a given `name + ref`. A resume always answers the currently paused turn (the latest interrupt), and `ref` is not guaranteed globally unique across turns. Searching from the end lands on the live request instead of a superseded one.

Refactored to use an efficient `findToolRequest` reverse loop over history without intermediate array allocations.

### DIFF
--- a/js/ai/src/agent.ts
+++ b/js/ai/src/agent.ts
@@ -1638,28 +1638,29 @@ export function validateResumeAgainstHistory(
   },
   history: MessageData[]
 ): void {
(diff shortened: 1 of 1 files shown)


huangjeff5:
Addressed in commit 095ace264. Extracted `part.toolRequest` into local variable `tr` to avoid redundant property accesses and optional chaining.

pavelgj (approved):

PR #5840, opened 2026-07-29
fix(go): bump dotprompt to unblock anthropic-sdk-go upgrades past v1.46

apascal07:
`github.com/google/dotprompt/go` is pinned to a build that requires `invopop/jsonschema` v0.13.0, which stores schema properties in `wk8/go-ordered-map`. `invopop/jsonschema` v0.14.0 switched that to `pb33f/ordered-map`, and `anthropic-sdk-go` began requiring v0.14.0 as of its v1.47.0 release.

Go resolves a single version per module across the build graph, so the two cannot coexist. Any project that depends on genkit and upgrades `anthropic-sdk-go` past v1.46.0 fails to compile, inside a file it does not own:

```
# github.com/google/dotprompt/go/dotprompt
.../dotprompt@v0.0.0-20251014011017-8d056e027254/dotprompt/picoschema.go:201:15:
cannot use orderedmap.New[string, *jsonschema.Schema]() (value of type
*"github.com/wk8/go-ordered-map/v2".OrderedMap[string, *jsonschema.Schema]) as
*"github.com/pb33f/ordered-map/v2".OrderedMap[string, *jsonschema.Schema] value
in struct literal
```

Worth being precise about where the constraint lives: this is not a floor genkit declares. `go.mod` requires `anthropic-sdk-go` v1.23.0, and under minimal version selection a consumer is free to select any higher version. The result is an effective *ceiling* that genkit never states and that only surfaces in the consumer's build, which makes it difficult to attribute back to genkit.

Bumping dotprompt to a build that already tracks `invopop/jsonschema` v0.14.0 and `pb33f/ordered-map` removes the ceiling.

## What moves

- `google/dotprompt/go` → `v0.0.0-20260708220100-73beb993ac95`
- `invopop/jsonschema` v0.13.0 → v0.14.0, and `wk8/go-ordered-map` → `pb33f/ordered-map` (both required by the new dotprompt)
- `goccy/go-yaml` v1.17.1 → v1.19.2 (also required directly by the new dotprompt)
- Assorted `golang.org/x/*` floors raised transitively by the above

The three `go/samples` modules that carry dotprompt as an indirect dependency are re-tidied so they continue to build against the local `replace`.

## Non-goal

This deliberately does **not** raise the `anthropic-sdk-go` requirement, which stays at v1.23.0. Nothing in the tree needs a newer API, and raising a library's floor forces the bump on every consumer rather than letting them opt in. The point is to stop blocking consumers who do want a newer SDK.

## Verification

Go 1.26.3, full `go test ./...` in `go/`:

- Before: 40 packages pass
- After: 40 packages pass, no new failures

The ceiling itself, checked directly:

| dotprompt | `anthropic-sdk-go` v1.61.0 | result |
| --- | --- | --- |
| current pin | `go build ./...` | fails in `picoschema.go` |
| this PR | `go build ./...` | builds, `plugins/anthropic` and `plugins/internal/anthropic` tests pass |

Each `go/samples` module touched here builds after the change.


adesinah (approved):
Looks good!

PR #5841, opened 2026-07-29
fix(go/plugins/anthropic): correct message conversion and honor tool choice

apascal07:
Fixes seven defects in the Anthropic request path, all in `plugins/internal/anthropic` and therefore shared with `vertexai/modelgarden`.

## Bugs fixed

**PDFs were sent as image blocks.** Every media part became `NewImageBlockBase64`, but Anthropic only accepts jpeg/png/gif/webp there. PDF input failed outright. Media now routes by content type: `image/*` to an image block, `application/pdf` and `text/plain` to document blocks.

**Reasoning signatures were dropped after a JSON roundtrip.** `NewReasoningPart` stores the signature as `[]byte`, and the conversion asserted `.([]byte)`. Once history is serialized and reloaded (persisted sessions, flow state) the value is a base64 `string`, the assertion failed, and the thinking block shipped with `signature: ""` — which Anthropic rejects, breaking multi-turn extended thinking. Now accepts either form, matching the helper the Gemini plugin already uses.

**Config-provided tools were silently wiped.** `req.Tools` was assigned unconditionally from the Genkit tool list. Server-side tools (web search, code execution, and friends) can only be expressed through `MessageNewParams`, so setting one and then generating dropped it. The two lists now merge.

**`OutputConfig.Effort` was clobbered by constrained JSON output.** Structured output replaced the whole `OutputConfigParam`, discarding a config-provided effort. Only `.Format` is set now.

**Empty message content panicked.** `Content[len(Content)-1]` was indexed without a length check, so a message with no parts caused an index-out-of-range. Parts are built first, and messages that yield no content blocks are skipped — Anthropic rejects empty content arrays anyway.

**`ai.ModelRequest.ToolChoice` was ignored.** The plugin advertises `ToolChoice: true` in its `Supports`, but never read the field, so `ai.WithToolChoice(...)` silently did nothing. Now mapped: auto to `auto`, required to `any`, none to `none`. Leaving it unset no longer disturbs a config-provided `tool_choice`.

**`maxTokens` was mandatory.** An unset value returned `maxTokens not set`, so a plain `Generate` call against a Claude model failed until the caller supplied config. It now falls back to `DefaultMaxOutputTokens = 4096`, matching the JS plugin's `DEFAULT_MAX_OUTPUT_TOKENS`, which every Claude model accepts. An explicit value still wins.

Two smaller things in the same code path: the data-part branch used unpadded `RawStdEncoding` where the adjacent media branch used `StdEncoding`, and empty `system: []` / `tools: []` arrays were sent on every request.

## Behavior change worth noting

Unsupported media types (for example `audio/mpeg`) now return a local error naming the accepted types, instead of silently sending a malformed image block and getting an opaque API 400. Clearer, but it is a new error path rather than a previously-working case.

## Verification

Regression tests added for all seven fixes, plus coverage for the merge and no-clobber cases. One existing case asserted the old `maxTokens not set` error and was updated to assert the default.

Go 1.26.3, full `go test ./...` in `go/`: 40 packages pass, no new failures.

## Scope

No dependency changes; this builds against the currently required `anthropic-sdk-go` v1.23.0. Prompt caching, citations, server-tool result round-tripping, and streaming fidelity are known gaps left for follow-ups.

### DIFF
--- a/go/plugins/internal/anthropic/anthropic.go
+++ b/go/plugins/internal/anthropic/anthropic.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"strings"
 
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/core/api"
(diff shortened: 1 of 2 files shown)


adesinah (approved):
The changes look good to me.

PR #5848, opened 2026-07-29
feat(genkit-tools): Update trace:get to support clean outputs

ssbushi:
- Adds tree mode
- Cleaner JSON output for programmatic access.

Usage: 

`genkit trace:get <id>`
`genkit trace:get <id> --format json`

Checklist (if applicable):


ssbushi (commented):
Addressed all feedback comments. Included fix for span status mapping, Commander option choices validation, redundant message parsing, empty message fallback, recursive formatCompactValue helper for arrays/objects, missing span attribute check, and removed dead test code.

ssbushi (commented):
Added test cases for `getSpanStatus` to cover the correct behavior of mapping status codes and states as pointed out in the review.

ssbushi (commented):
Added test cases for array bullet formatting and empty message fallback as requested.

genkit-tools/common/tests/utils/utils_test.ts
```diff
@@ -13,99 +13,76 @@
...
+import { describe, expect, it } from '@jest/globals';
 import {
-  findProjectRoot,
-  projectNameFromGenkitFilePath,
-} from '../../src/utils';
-
+  formatDuration,
+  parseAndSanitizeJson,
+  sanitizeBase64DataUrls,
+} from '../../src/utils/utils';
```
pavelgj, line 36: This rewrite drops the existing `findProjectRoot` and `projectNameFromGenkitFilePath` test suites. Both functions still ship from `utils.ts`, so this looks like a coverage regression unrelated to the trace:get feature. Could we keep the new `formatDuration` / `sanitizeBase64DataUrls` / `parseAndSanitizeJson` tests but restore the removed blocks alongside them?

genkit-tools/cli/src/commands/trace-get.ts
```diff
@@ -33,7 +51,26 @@ export const traceGet = new Command('trace:get')
           logger.error(`Trace with ID '${traceId}' not found.`);
           return;
         }
-        console.log(JSON.stringify(response, undefined, 2));
+
+        const format = (options.format || 'tree').toLowerCase();
```
pavelgj: Since the option is defined with `.choices(['tree', 'json'])`, commander rejects any other value before the action runs. That makes the final `else` (`Unknown format...`) branch unreachable, and `(options.format || 'tree').toLowerCase()` redundant given the `'tree'` default. Could simplify to:

```ts
const keepMedia = !!options.keepMedia;
if (options.format === 'json') {
  console.log(JSON.stringify(cleanTraceJson(response, keepMedia), undefined, 2));
} else {
  console.log(yellow('Hint: pass `--format json` flag to get trace data in JSON format\n'));
  console.log(formatTraceTree(response, keepMedia));
}
```

genkit-tools/cli/src/utils/trace-formatter.ts
```diff
@@ -0,0 +1,408 @@
...
+  }
+
+  return lines.join('\n');
+}
+
+/**
+ * Creates a cloned, sanitized version of a trace for pure JSON output. Base64
+ * media data inside span attributes can optionally be stripped for brevity.
+ */
+export function cleanTraceJson(trace: TraceData, keepMedia: boolean): any {
```
pavelgj: Typing nit: this returns `any`. Since it's a deep clone of the input trace, `TraceData` would be a more accurate return type and keep callers type-safe.

genkit-tools/telemetry-server/src/index.ts
```diff
@@ -346,7 +346,7 @@ export async function startTelemetryServer(params: {
 
   const host = params.host ?? '127.0.0.1';
   server = api.listen(params.port, host, () => {
-    logger.info(`Telemetry API running on http://${host}:${params.port}`);
+    logger.debug(`Telemetry API running on http://${host}:${params.port}`);
```
pavelgj: This log demotion (info -> debug) is a little out of scope for trace:get. If the goal is keeping stdout clean for programmatic/JSON consumers, `forceStderr()` (from `tools-common/utils`, already used in `mcp.ts`) might be a more targeted fix: it routes logs to stderr so they don't pollute piped stdout without losing the startup message entirely.

pavelgj (commented):
A few small things from a pass over this, mostly nits plus one coverage regression to restore. Nice feature overall.

genkit-tools/common/tests/utils/utils_test.ts
ssbushi: Good find. I mistakenly thought this test file was a new addition in this PR...

pavelgj (approved):

PR #5849, opened 2026-07-29
feat(go): add typed-config `New*Action` primitive constructors

apascal07:
Adds a typed-config constructor family for the `ai` primitives: models, embedders, evaluators, batch evaluators, background models, and retrievers. Every new function lands alongside its existing counterpart under a new name (`ai.New*Action`, `genkit.Define*Action`), so nothing breaks; the old constructors are deprecated and delegate to the new ones, leaving one code path. Builds on the `core.New*ActionOf` options-struct constructors these compose over, merged in #5862, and now targets main directly. Next in the plugin migration train: once providers move to `NewModelAction` (googlegenai in #5869, anthropic in #5874), their hand-rolled `configFromRequest` mapping and its missing validation go away.

## The framework deserializes config, not the plugin

**Before:** every provider hand-rolls the config mapping and reflects its own config schema; neither is validated.

```go
func configToMap(config any) map[string]any {
    r := jsonschema.Reflector{
        DoNotReference: true,
        ExpandedStruct: true,
        IgnoredTypes:   []any{genai.Schema{}}, // recursion-prone fields tracked by hand
    }
    ...
}

func configFromRequest(input *ai.ModelRequest) (*genai.GenerateContentConfig, error) {
    switch config := input.Config.(type) {
    case genai.GenerateContentConfig:  return &config, nil
    case *genai.GenerateContentConfig: return config, nil
    case map[string]any:               return base.MapToStruct[genai.GenerateContentConfig](config)
    case nil:                          return &genai.GenerateContentConfig{}, nil
    default:
        return nil, core.NewPublicError(core.INVALID_ARGUMENT, "Invalid configuration type ...", nil)
    }
}

fn := func(ctx context.Context, input *ai.ModelRequest, cb ai.ModelStreamCallback) (*ai.ModelResponse, error) {
    gcc, err := configFromRequest(input)
    ...
}
ai.NewModel(name, &ai.ModelOptions{
    ConfigSchema: configToMap(genai.GenerateContentConfig{}),
    ...
}, fn)
```

**After:** `Config` is inferred from the function signature and arrives converted; the schema is derived from it.

```go
fn := func(ctx context.Context, input *ai.ModelRequest, config genai.GenerateContentConfig, cb ai.ModelStreamCallback) (*ai.ModelResponse, error) {
    ...
}
ai.NewModelAction(name, opts, fn) // config schema inferred from Config
```

The framework validates the raw config against the config schema at the action boundary, converts it (exact type, `*Config`, or `map[string]any`; anything else is `INVALID_ARGUMENT`), and hands the function a shallow copy of the request whose type-erased slot carries the converted value, so the two views cannot disagree. The caller's request keeps its raw config, so one request can fan out across providers, and a pointer `Config` with no config stays `nil` rather than boxing a typed nil.

- The config JSON schema is inferred from `Config` unless `ConfigSchema` overrides it. Inferred schemas drop `required` (a config is partial by nature), and the model config slot always admits the framework-level `version` key so `Versions` keeps working. The override stays the escape hatch for types whose reflection does not match their JSON marshaling (SDK wrapper generics like `Opt[float64]`); those plugins use the new constructors and supply the schema by hand.
- The request input schema's config slot is wrapped as `anyOf [schema, null]`, because a typed-nil config marshals to JSON `null` and would otherwise fail input validation.
- Conversion runs outermost in the model's built-in middleware chain (`normalizeConfig`), so the built-in wrappers and the model function all see the typed value. Version validation moved there out of `validateSupport`: it has to run on the raw config, since converting into a `Config` with no `version` field silently drops the key.

## Constructors return concrete `*Action` types

The constructors and the `genkit.Define*Action` wrappers return `*ModelAction`, `*EmbedderAction`, `*EvaluatorAction`, `*RetrieverAction`, and `*BackgroundModelAction` instead of the interfaces. This leaves room to add methods later without the interface-growth break, and satisfies `api.Action` structurally, so a plugin `Init` can append models to its returned action slice without a type assertion. The constructors are named for the types they return, the way `NewModelRef` returns a `ModelRef`; the plain names stay occupied by the deprecated constructors. The implementation function types follow the same stem: `ModelActionFunc[Config]` is the fn a `ModelAction` is built from.

Each type embeds `core.Action` (or `core.BackgroundAction`) through an unexported package-level alias, which promotes `Desc`, `Register`, `RunJSON`, and the rest without exporting the embedded field. `Model`, `Embedder`, `Evaluator`, `Retriever`, and `BackgroundModel` are unchanged and stay the accepted-argument and lookup types; the concrete types satisfy them, and the deprecated constructors keep returning interfaces.

Those promoted methods are then redeclared on each type as forwarders. The callable surface is identical either way, but godoc cannot see through an alias into another package: without them `ModelAction` documented one method while claiming to implement `api.Action`, `BackgroundModelAction` documented none at all, and every `[ModelAction.Register]` link in the constructor docs failed to resolve. A forwarder that calls itself rather than the embedded action recurses instead of failing to compile, so each one is exercised, `RunJSON` and `RunJSONWithTelemetry` included. Each forwarder that can report an error also rejects a nil receiver with `INVALID_ARGUMENT`, matching `Generate`, `Embed`, `Evaluate`, `Retrieve`, and `RunRawMultipart`; `Name`, `Desc`, `Register`, and `SupportsCancel` cannot, so they still panic rather than report a zero value as truth.

`ToolDef` is renamed to `ToolAction` to join that family, keeping `ToolDef` as a deprecated alias. The old name read as a sibling of `ToolDefinition`, the wire type a tool advertises to the model, when it is in fact the action a tool is backed by. It is an alias rather than a deletion because nothing about the type is framework-internal: an application can hold one, so `ai.NewTool`, `NewMultipartTool`, `NewToolWithInputSchema` and their three `genkit.Define*` counterparts stay source-compatible even though their return type is now spelled `*ToolAction`.

`ToolAction` also gains `Desc`, `RunJSON`, and `RunJSONWithTelemetry`, which makes it an `api.Action` like the rest of the family. It holds its action in a named field instead of embedding one, so it never inherited them, and a plugin returns `[]api.Action`: shipping a tool from a plugin did not compile, though core's package documentation has always shown exactly that. The three forward to the action the type already holds, so they serve what the registry already serves for the tool, `RunJSON` returning the multipart envelope registered under both the tool and action keys.

## One options struct for background models

`NewBackgroundModelAction` takes the same `BackgroundModelOptions` the deprecated constructor takes: `ModelOptions` embedded plus the background extras, the Go counterpart of JS's `DefineBackgroundModelOptions`, which extends `DefineModelOptions` the same way. Embedding means every future `ModelOptions` field applies to background models without mirroring, and a plugin that already holds a `ModelOptions` drops it in wholesale. The rule on the provider surface: required functions are positional (`startFn`, `checkFn`), so the compiler enforces them and they drive type inference, and the optional `Cancel` hook lives in the options struct, so a future addition is a field rather than a signature break. `core.NewBackgroundActionOf` instead takes only `startFn` positionally, with check and cancel as fields of its generic `BackgroundActionOptions[In, Out]` (#5862): the substrate's required set is expected to evolve, while a background model is defined by its pollable check.

`ModelOptions`, `EmbedderOptions`, `EvaluatorOptions`, and `RetrieverOptions` gain a `Metadata` slot. Caller values merge into the action descriptor with the reserved keys (`type`, and the per-kind info map) stamped over them, since registry discovery depends on those. `BackgroundModelOptions` ends up with two slots, its own predating the `ModelOptions` one; both merge into the descriptor, and the top-level field wins on key conflicts.

## Registry-taking `Define*` helpers deleted

Continuing #5862's cleanup into the `ai` layer. A registry is unobtainable outside the framework, so these had no reachable external callers and are deleted rather than deprecated; the `genkit.Define*` wrappers stay as the user touchpoints.

Deleted: `DefineModel`, `DefineEmbedder`, `DefineEvaluator`, `DefineBatchEvaluator`, `DefineBackgroundModel`, `DefineRetriever`, `DefineMiddleware`, `DefineTool`, `DefineToolWithInputSchema`, `DefineMultipartTool`, `DefineResource`, plus the experimental `exp.DefineTool` and `exp.DefineInterruptibleTool`.

Kept, because the registry is a structural input rather than a registration target: `ai.DefinePrompt`, `ai.DefineDataPrompt`, `ai.DefineFormat`, `ai.DefineGenerateAction` (init bootstrap), and the exp agent constructors.

## Four behavior changes reach existing callers

Observable through the deprecated constructors:

- A `version` error surfaces before capability errors, since version validation now runs outermost instead of inside `validateSupport`.
- With `ConfigSchema` set, the request input schema's config slot is the null-tolerant `anyOf` wrapping instead of the raw schema.
- Batch evaluators advertise an inferred input schema, having previously advertised none at all.
- A background model's embedded `ModelOptions.Metadata` reaches the action descriptor; it was ignored on the background path before, and the top-level `BackgroundModelOptions.Metadata` still wins on key conflicts.

`Config` resolves to `any` on the deprecated path, which infers no schema and passes the raw config through untouched, so nothing else moves.

## API reference

```go
// Added: ai
ai.NewModelAction[Config](name, opts, fn) *ModelAction
ai.NewEmbedderAction[Config](name, opts, fn) *EmbedderAction
ai.NewEvaluatorAction[Config](name, opts, fn) *EvaluatorAction
ai.NewBatchEvaluatorAction[Config](name, opts, fn) *EvaluatorAction
ai.NewRetrieverAction[Config](name, opts, fn) *RetrieverAction
ai.NewBackgroundModelAction[Config](name, opts, startFn, checkFn) *BackgroundModelAction

ai.ModelAction, ai.EmbedderAction, ai.EvaluatorAction, ai.RetrieverAction, ai.BackgroundModelAction
ai.ModelActionFunc[Config], ai.EmbedderActionFunc[Config], ai.EvaluatorActionFunc[Config],
ai.BatchEvaluatorActionFunc[Config], ai.RetrieverActionFunc[Config], ai.BackgroundModelActionFunc[Config]
ai.ModelOptions.Metadata, ai.EmbedderOptions.Metadata,
ai.EvaluatorOptions.Metadata, ai.RetrieverOptions.Metadata

// Added: genkit
genkit.DefineModelAction, genkit.DefineEmbedderAction, genkit.DefineEvaluatorAction,
genkit.DefineBatchEvaluatorAction, genkit.DefineRetrieverAction, genkit.DefineBackgroundModelAction

ai.ToolAction.Desc, ai.ToolAction.RunJSON, ai.ToolAction.RunJSONWithTelemetry

// Renamed, deprecated alias kept
ai.ToolDef -> ai.ToolAction

// Deprecated, delegating to the new constructors
ai.NewModel, ai.NewEmbedder, ai.NewEvaluator, ai.NewBatchEvaluator,
ai.NewRetriever, ai.NewBackgroundModel
genkit.DefineModel, genkit.DefineEmbedder, genkit.DefineEvaluator,
genkit.DefineBatchEvaluator, genkit.DefineRetriever, genkit.DefineBackgroundModel
```

## Scope

- **No `Define*Action` at the `ai` layer.** Registration is one `Register` call, so a define-wrapper there would only shadow the `genkit` one applications already have.
- **No plugin migrations here.** googlegenai (#5869) and anthropic (#5874) move on top of this, so this PR reviews as pure framework surface.

## Verification

Go 1.26.3, `go build ./...` and `go test ./...` in `go/`: 41 packages with tests pass, no failures.

New coverage in `ai/config_test.go`: a typed model receives its config deserialized from each accepted shape and rejects the rest with `INVALID_ARGUMENT`; the config is normalized before the built-in wrappers run, so they see the typed value; the schema is inferred from `Config`, overridden by an explicit `ConfigSchema`, and absent for `Config=any`; the deprecated constructors pass the raw config through unchanged; and caller `Metadata` reaches the descriptor without displacing the reserved keys, for the typed model, embedder, evaluator, retriever, and background model, including the background model's two-slot merge. Embedders, evaluators, and retrievers get the same shape-by-shape config coverage. `ai/define_test.go`, `ai/exp/define_test.go`, and `plugins/middleware/define_test.go` hold test-local `New* + Register` helpers standing in for the deleted `Define*`. `ai/action_methods_test.go` runs every forwarder on all six primitives; pointing one at itself makes it fail with a stack overflow rather than passing quietly. `genkit/genkit_test.go` pins the documented plugin shape: a plugin returning `[]api.Action{model, tool}` registers both, and the tool runs through a registry lookup.

### DIFF
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -588,8 +588,6 @@ func DefineModel(g *Genkit, name string, opts *ai.ModelOptions, fn ai.ModelFunc)
 // The `name` is the identifier the model uses to request the background model. The `opts`
 // are the options for the background model. The `startFn` is the function that starts the background model.
 // The `checkFn` is the function that checks the status of the background model.
-// The `cancelFn` is optional; nil means the model does not support canceling
-// operations.
 //
 // Config is the model's typed configuration; it is usually inferred from
 // startFn's signature. See [ai.NewTypedModel] for how the request's
@@ -600,9 +598,8 @@ func DefineTypedBackgroundModel[Config any](
 	opts *ai.TypedBackgroundModelOptions,
 	startFn ai.TypedStartModelOpFunc[Config],
 	checkFn ai.CheckModelOpFunc,
-	cancelFn ai.CancelModelOpFunc,
 ) *ai.BackgroundModelAction {
-	m := ai.NewTypedBackgroundModel(name, opts, startFn, checkFn, cancelFn)
+	m := ai.NewTypedBackgroundModel(name, opts, startFn, checkFn)
 	m.Register(g.reg)
 	return m
 }
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -595,7 +595,7 @@ func DefineModel(g *Genkit, name string, opts *ai.ModelOptions, fn ai.ModelFunc)
 func DefineBackgroundModelAction[Config any](
 	g *Genkit,
 	name string,
-	opts *ai.TypedBackgroundModelOptions,
+	opts *ai.BackgroundModelOptions,
 	startFn ai.BackgroundModelActionFunc[Config],
 	checkFn ai.CheckModelOpFunc,
 ) *ai.BackgroundModelAction {
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -675,14 +675,14 @@ func LookupBackgroundModel(g *Genkit, name string) ai.BackgroundModel {
 //	}
 //
 //	fmt.Println(resp.Text()) // Might output something like "The weather in Paris is Sunny, 25°C."
-func DefineTool[In, Out any](g *Genkit, name, description string, fn ai.ToolFunc[In, Out], opts ...ai.ToolOption) *ai.ToolDef[In, Out] {
+func DefineTool[In, Out any](g *Genkit, name, description string, fn ai.ToolFunc[In, Out], opts ...ai.ToolOption) *ai.ToolAction[In, Out] {
 	t := ai.NewTool(name, description, fn, opts...)
 	t.Register(g.reg)
 	return t
 }
 
 // DefineToolWithInputSchema defines a tool with a custom input schema that can be used by models during generation,
-// registers it as a [core.Action] of type Tool, and returns an [*ai.ToolDef].
+// registers it as a [core.Action] of type Tool, and returns an [*ai.ToolAction].
 //
 // This variant of [DefineTool] allows specifying a JSON Schema for the tool's input, providing more
 // control over input validation and model guidance. The input parameter to the tool function will be
@@ -726,14 +726,14 @@ func DefineTool[In, Out any](g *Genkit, name, description string, fn ai.ToolFunc
 //		},
 //		ai.WithToolInputSchema(inputSchema),
 //	)
-func DefineToolWithInputSchema[Out any](g *Genkit, name, description string, inputSchema map[string]any, fn ai.ToolFunc[any, Out]) *ai.ToolDef[any, Out] {
+func DefineToolWithInputSchema[Out any](g *Genkit, name, description string, inputSchema map[string]any, fn ai.ToolFunc[any, Out]) *ai.ToolAction[any, Out] {
 	t := ai.NewTool(name, description, fn, ai.WithInputSchema(inputSchema))
 	t.Register(g.reg)
 	return t
 }
 
 // DefineMultipartTool defines a multipart tool that can be used by models during generation,
-// registers it as a [core.Action] of type Tool, and returns an [*ai.ToolDef].
+// registers it as a [core.Action] of type Tool, and returns an [*ai.ToolAction].
 // Unlike regular tools that return just an output value, multipart tools can return
 // both an output value and additional content parts (like images or other media).
 //
@@ -788,7 +788,7 @@ func DefineToolWithInputSchema[Out any](g *Genkit, name, description string, inp
 //	}
 //
 //	fmt.Println(resp.Text())
-func DefineMultipartTool[In any](g *Genkit, name, description string, fn ai.MultipartToolFunc[In], opts ...ai.ToolOption) *ai.ToolDef[In, *ai.MultipartToolResponse] {
+func DefineMultipartTool[In any](g *Genkit, name, description string, fn ai.MultipartToolFunc[In], opts ...ai.ToolOption) *ai.ToolAction[In, *ai.MultipartToolResponse] {
 	t := ai.NewMultipartTool(name, description, fn, opts...)
 	t.Register(g.reg)
 	return t
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -597,7 +597,7 @@ func DefineModel(g *Genkit, name string, opts *ai.ModelOptions, fn ai.ModelFunc)
 func DefineTypedBackgroundModel[Config any](
 	g *Genkit,
 	name string,
-	opts *ai.BackgroundModelOptions,
+	opts *ai.TypedBackgroundModelOptions,
 	startFn ai.TypedStartModelOpFunc[Config],
 	checkFn ai.CheckModelOpFunc,
 	cancelFn ai.CancelModelOpFunc,
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -577,18 +577,20 @@ func DefineModel(g *Genkit, name string, opts *ai.ModelOptions, fn ai.ModelFunc)
 	return m
 }
 
-// DefineTypedBackgroundModel defines a background model, registers it as
-// an [ai.BackgroundModel], and returns an [ai.BackgroundModel].
+// DefineTypedBackgroundModel defines a background model, registers it, and
+// returns the concrete [ai.BackgroundModelAction].
 //
 // The `name` is the identifier the model uses to request the background model. The `opts`
 // are the options for the background model. The `startFn` is the function that starts the background model.
 // The `checkFn` is the function that checks the status of the background model.
+// The `cancelFn` is optional; nil means the model does not support canceling
+// operations.
 //
 // Config is the model's typed configuration; it is usually inferred from
 // startFn's signature. See [ai.NewTypedModel] for how the request's
 // config is deserialized and validated.
-func DefineTypedBackgroundModel[Config any](g *Genkit, name string, opts *ai.BackgroundModelOptions, startFn ai.TypedStartModelOpFunc[Config], checkFn ai.CheckModelOpFunc) *ai.BackgroundModelAction {
-	m := ai.NewTypedBackgroundModel(name, opts, startFn, checkFn)
+func DefineTypedBackgroundModel[Config any](g *Genkit, name string, opts *ai.BackgroundModelOptions, startFn ai.TypedStartModelOpFunc[Config], checkFn ai.CheckModelOpFunc, cancelFn ai.CancelModelOpFunc) *ai.BackgroundModelAction {
+	m := ai.NewTypedBackgroundModel(name, opts, startFn, checkFn, cancelFn)
 	m.Register(g.reg)
 	return m
 }
--- a/go/genkit/genkit.go
+++ b/go/genkit/genkit.go
@@ -1475,7 +1475,9 @@ func Embed(ctx context.Context, g *Genkit, opts ...ai.EmbedderOption) (*ai.Embed
 // For retrievers that don't need to be registered (e.g., for plugin development),
 // use [ai.NewRetriever] instead.
 func DefineRetriever(g *Genkit, name string, opts *ai.RetrieverOptions, fn ai.RetrieverFunc) ai.Retriever {
-	return ai.DefineRetriever(g.reg, name, opts, fn)
+	ret := ai.NewRetriever(name, opts, fn)
+	ret.Register(g.reg)
+	return ret
 }
 
 // LookupRetriever retrieves a registered [ai.Retriever] by its provider and name.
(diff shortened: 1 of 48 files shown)


go/genkit/genkit.go
```diff
@@ -494,26 +499,30 @@ func ListTools(g *Genkit) []ai.Tool {
 	return tools
 }
 
-// DefineModel defines a custom model implementation, registers it as a [core.Action]
-// of type Model, and returns an [ai.Model] interface.
+// DefineModelAction defines a custom model implementation, registers it
+// as a [core.Action] of type Model, and returns an [ai.Model] interface.
```
pavelgj: The whole point of this PR is that the `Define*Action` wrappers return the concrete `*Action` type, and the Retriever/BackgroundModel siblings say so. This one still says "returns an [ai.Model] interface" but actually returns `*ai.ModelAction`. Suggest changing it to "...and returns the concrete [ai.ModelAction]."

go/genkit/genkit.go
```diff
@@ -1356,18 +1455,42 @@ func LookupRetriever(g *Genkit, name string) ai.Retriever {
 	return ai.LookupRetriever(g.reg, name)
 }
 
-// DefineEmbedder defines a custom text embedding implementation, registers it as a
-// [core.Action] of type Embedder, and returns an [ai.Embedder].
-// Embedders convert text documents or queries into numerical vector representations (embeddings).
+// DefineEmbedderAction defines a custom text embedding implementation,
+// registers it as a [core.Action] of type Embedder, and returns an
+// [ai.Embedder]. Embedders convert text documents or queries into numerical
```
pavelgj: Same return-type doc drift as `DefineModelAction`: this returns `*ai.EmbedderAction`, not the `[ai.Embedder]` interface. Suggest "...and returns the concrete [ai.EmbedderAction]." to match the Retriever/BackgroundModel wrappers.

go/genkit/genkit.go
```diff
@@ -1387,35 +1510,72 @@ func LookupPlugin(g *Genkit, name string) api.Plugin {
 	return g.reg.LookupPlugin(name)
 }
 
-// DefineEvaluator defines an evaluator that processes test cases one by one,
-// registers it as a [core.Action] of type Evaluator, and returns an [ai.Evaluator].
-// Evaluators are used to assess the quality or performance of AI models or flows
-// based on a dataset of test cases.
+// DefineEvaluatorAction defines an evaluator that processes test cases
+// one by one, registers it as a [core.Action] of type Evaluator, and returns
+// an [ai.Evaluator]. Evaluators are used to assess the quality or performance
```
pavelgj: Returns `*ai.EvaluatorAction`, not `[ai.Evaluator]`. Suggest "...and returns the concrete [ai.EvaluatorAction]." for consistency.

go/genkit/genkit.go
```diff
@@ -1387,35 +1510,72 @@ func LookupPlugin(g *Genkit, name string) api.Plugin {
...
+// request.
+func DefineEvaluator(g *Genkit, name string, opts *ai.EvaluatorOptions, fn ai.EvaluatorFunc) ai.Evaluator {
+	e := ai.NewEvaluator(name, opts, fn)
+	e.Register(g.reg)
+	return e
+}
+
+// DefineBatchEvaluatorAction defines an evaluator that processes the
+// entire dataset at once, registers it as a [core.Action] of type Evaluator,
+// and returns an [ai.Evaluator].
```
pavelgj: `DefineBatchEvaluatorAction` also returns `*ai.EvaluatorAction`, not `[ai.Evaluator]`. Suggest "...and returns the concrete [ai.EvaluatorAction]."

go/genkit/genkit.go
```diff
@@ -673,12 +726,14 @@ func DefineTool[In, Out any](g *Genkit, name, description string, fn ai.ToolFunc
 //		},
 //		ai.WithToolInputSchema(inputSchema),
```
pavelgj: `ai.WithToolInputSchema` isn't a real symbol here (the option is `ai.WithInputSchema`). Pre-existing, but this doc block is being touched anyway, so worth fixing the example line to `ai.WithInputSchema(inputSchema)` while here.

pavelgj (approved):

huangjeff5:
Overall looks great to me.

One quick detail in `go/ai/tools.go`\: `RunRawMultipart` guards against a nil receiver (`if t == nil`), but `RunJSON` and `RunJSONWithTelemetry` currently pass straight through to `t.action`:

```go
func (t *ToolAction[In, Out]) RunJSON(ctx context.Context, input json.RawMessage, cb core.StreamCallback[json.RawMessage]) (json.RawMessage, error) {
    return t.action.RunJSON(ctx, input, cb)
}
```

If a dynamic lookup via `LookupTool` returns `nil` (e.g., an unregistered tool name) or if an uninitialized `*ToolAction` pointer is called, `RunJSON` will trigger a nil pointer dereference panic instead of returning an error.

Would it make sense to add a receiver guard here for consistency with `RunRawMultipart` and `ModelAction.Generate`?

```diff
 func (t *ToolAction[In, Out]) RunJSON(ctx context.Context, input json.RawMessage, cb core.StreamCallback[json.RawMessage]) (json.RawMessage, error) {
+	if t == nil {
+		return nil, status.Errorf(status.ErrInvalidArgument, "ai.Tool.RunJSON: tool called on a nil tool; check that all tools are defined")
+	}
 	return t.action.RunJSON(ctx, input, cb)
 }

 func (t *ToolAction[In, Out]) RunJSONWithTelemetry(ctx context.Context, input json.RawMessage, cb core.StreamCallback[json.RawMessage]) (*api.ActionRunResult[json.RawMessage], error) {
+	if t == nil {
+		return nil, status.Errorf(status.ErrInvalidArgument, "ai.Tool.RunJSONWithTelemetry: tool called on a nil tool; check that all tools are defined")
+	}
 	return t.action.RunJSONWithTelemetry(ctx, input, cb)
 }
```

huangjeff5 (approved):

PR #5850, opened 2026-07-29
feat(go/a2ui): add A2UI plugin for streaming interactive UI

pavelgj:
Add experimental A2UI ("Agent to UI") plugin for Genkit Go, enabling agents to stream rich, interactive UI surfaces via the JSON-based A2UI protocol.

The plugin provides middleware that injects catalog capabilities into the system prompt, intercepts model output to extract and validate a2ui fenced blocks, and rewrites them into a2ui data parts using the mime type `application/a2ui+json`. This maps 1:1 onto the A2A binding, making envelopes byte-compatible with the JS and Dart plugins.

Includes a bundled basic catalog, custom catalog support via the Genkit registry, configurable validation modes, and helpers for extracting envelopes from message parts.


go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,422 @@
...
+func (r *surfaceIDReplay) replayNext() string {
+	if r.cursor < len(r.generated) {
+		id := r.generated[r.cursor]
+		r.cursor++
+		return id
+	}
+	return r.next()
+}
+
+func (r *surfaceIDReplay) reset() { r.cursor = 0 }
```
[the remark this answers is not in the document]
pavelgj: Done in 6ab2519 - added a sync.Mutex to surfaceIDReplay (with a nextLocked helper). In practice the streamed parse and the final-message parse never overlap, but the lock is cheap insurance. Verified with go test -race.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,422 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/firebase/genkit/go/ai"
+	"github.com/firebase/genkit/go/core/api"
+	"github.com/firebase/genkit/go/genkit"
+	"github.com/google/uuid"
+)
```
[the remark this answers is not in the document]
pavelgj: Done - added the sync import in 6ab2519.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,376 @@
...
+
+func newStreamParser(opts parserOptions) *streamParser {
+	if opts.validate == "" {
+		opts.validate = ValidateWarn
+	}
+	if opts.version == "" {
+		opts.version = DefaultVersion
+	}
+	return &streamParser{opts: opts}
+}
```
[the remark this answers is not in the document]
pavelgj: Done in 6ab2519 - the component-name set is now computed once in newStreamParser and cached on the parser as knownComponents.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,376 @@
...
+func (p *streamParser) validateComponents(components any) string {
+	catalog := p.opts.catalog
+	if catalog == nil {
+		return ""
+	}
+	arr, ok := components.([]any)
+	if !ok {
+		return "updateComponents.components must be an array."
+	}
+	known := catalog.componentNames()
```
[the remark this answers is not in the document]
pavelgj: Done - validateComponents now uses the cached knownComponents map (6ab2519).

go/plugins/a2ui/catalog.go
```diff
@@ -0,0 +1,307 @@
...
+`+"```", catalog.ID, rootComponent)
+}
+
+// RenderCatalogInstructions renders a catalog into model-facing instructions
+// describing the A2UI protocol and the available components. It is injected
+// into the system prompt by the middleware when Instructions is
+// InstructionsSystem.
+func RenderCatalogInstructions(catalog *Catalog) string {
+	var componentDocs strings.Builder
+	for i, c := range catalog.Components {
```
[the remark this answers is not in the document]
pavelgj: Done - added a nil-catalog guard returning "" in 6ab2519.

go/plugins/a2ui/README.md
```diff
@@ -0,0 +1,175 @@
...
+### Options
+
+`a2ui.Config` fields:
+
+| Field          | Default            | Description                                                                                       |
+| -------------- | ------------------ | ------------------------------------------------------------------------------------------------- |
+| `Catalog`      | nil                | An inline catalog (code-defined use only; not serialized). Overrides `CatalogID` when set.        |
+| `CatalogID`    | `"basic"`          | Id of a catalog registered with `LoadCatalog`. Resolved from the registry at call time.           |
+| `Instructions` | `InstructionsSystem` | Where to inject catalog capabilities. `InstructionsNone` injects nothing.                        |
+| `Validate`     | `ValidateWarn`     | Validate emitted envelopes. `ValidateWarn` logs and drops bad blocks; `ValidateStrict` returns an error; `ValidateOff` skips checking. |
```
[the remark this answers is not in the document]
pavelgj: Done - the options table now shows the string values ("system", "warn", etc.) instead of the Go constant names (6ab2519).

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,376 @@
...
+				} else {
+					keep := maxPartialFence
+					if keep > len(p.buffer) {
+						keep = len(p.buffer)
+					}
+					safeLen := len(p.buffer) - keep
+					if safeLen > 0 {
+						proseBuf += p.buffer[:safeLen]
+						p.buffer = p.buffer[safeLen:]
+					}
```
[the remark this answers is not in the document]
pavelgj: Not taking this one - the suggested code is subtly incorrect. It keeps from the *last* backtick (LastIndexByte), so for a fence split mid-opener like "hello ```a2ui" it would flush "hello ``" and retain only "`a2ui"; once reassembled with the next chunk the opener has a single backtick and no longer matches ```a2ui, so the block is missed. The constant 8-char holdback is intentional (it mirrors the JS parser) and the lag is at most a few chars until the next chunk arrives, so I'd rather keep the simple, correct version.

go/plugins/a2ui/loader.go
```diff
@@ -0,0 +1,129 @@
...
+	id := catalogID
+	if id == "" {
+		id = DefaultCatalogID
+	}
+	if g != nil {
+		if v := genkit.LookupValue(g, catalogRegistryKey(id)); v != nil {
+			if c, ok := v.(*Catalog); ok {
+				return c, nil
+			}
+			return nil, fmt.Errorf("a2ui: registry value %q is not an *a2ui.Catalog", catalogRegistryKey(id))
```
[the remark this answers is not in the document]
pavelgj: Leaving as-is. The only registration path (LoadCatalog) always stores a *Catalog, so a Catalog-by-value case would be guarding against a misuse this package never creates. The current type assertion already returns a clear error for anything unexpected.

go/ai/document.go
```diff
@@ -100,9 +101,11 @@ func NewMediaPart(mimeType, contents string) *Part {
...
 }
 
-// NewDataPart returns a Part containing raw string data.
-func NewDataPart(contents string) *Part {
-	return &Part{Kind: PartData, Text: contents}
+// NewDataPart returns a Part containing arbitrary structured data. The value is
+// serialized to the part's `data` field as-is, so it may be a string, a map, a
+// slice, or any other JSON-serializable value.
+func NewDataPart(data any) *Part {
+	return &Part{Kind: PartData, Data: data}
```
apascal07, line 115: Data parts no longer populate `Text`, but `uri.Data` still reads it (`go/plugins/internal/uri/uri.go:35`), so a data part carrying a data URI no longer converts:

```go
p := ai.NewDataPart("data:image/png;base64,aGVsbG8=")
uri.Data(p)
// "", nil, err="could not convert media part to genai part: missing file data"
```

That breaks three provider paths at runtime: `googlegenai/gemini.go:627`, `googlegenai/tools.go:57`, and `internal/anthropic/anthropic.go:378`. Since `string` satisfies `any`, existing `NewDataPart("...")` call sites still compile, so nothing in CI catches it.

The `Data any` field is the right shape (it matches `dataPart.Data` in `gen.go` and `data: z.unknown()` in JS), so the fix belongs in `uri.go`:

```go
uri := p.Text
if uri == "" {
	if s, ok := p.Data.(string); ok {
		uri = s
	}
}
```

There's no test covering `uri.Data` on a data part today, which is why this passed. Worth adding one alongside the fix.

go/ai/document.go
```diff
@@ -39,6 +39,7 @@ type Part struct {
 	Kind         PartKind       `json:"kind,omitempty"`
 	ContentType  string         `json:"contentType,omitempty"`  // valid for kind==blob
 	Text         string         `json:"text,omitempty"`         // valid for kind∈{text,blob}
+	Data         any            `json:"data,omitempty"`         // valid for kind==data
```
apascal07, line 43: With data content moved off `Text`, the `IsData()` arms in `Message.Text()` and `ModelResponseChunk.Text()` now append an empty string:

```go
// generate.go:1235 and :1366
if p.IsText() || p.IsData() {
	sb.WriteString(p.Text)
}
```

Verified: `NewModelMessage(NewTextPart("hello "), NewDataPart("world")).Text()` returns `"hello "`.

Keeping structured data out of `Text()` is the right outcome for A2UI, since you don't want envelope JSON in there. But leaving `|| p.IsData()` as a silent no-op reads as an oversight to the next reader. Better to delete it and say in a comment that data parts are deliberately excluded. The `len(m.Content) == 1` fast path at generate.go:1361 returns `Text` regardless of kind and has the same question.

go/ai/document_test.go
```diff
@@ -421,6 +447,7 @@ func TestPartClone(t *testing.T) {
 		Kind:        PartToolRequest,
 		ContentType: "application/json",
 		Text:        "body",
+		Data:        map[string]any{"dk": "dv"},
```
apascal07, line 450: The comment above says everything is populated "to catch missing fields", but `Clone` is a plain struct copy (`cp := *p`), so a new field is picked up automatically and this case can't fail.

What it does not cover is aliasing. `Clone` deliberately `maps.Clone`s `Custom` and `Metadata`, and its doc promises callers "can add or remove map keys without mutating the original". `Data` is now most often a `map[string]any`, both from `NewDataPart` in the A2UI path and from JSON decode, and it is shared by reference between clones.

Either extend the `Clone` doc to say `Data` is not deep-copied, or `maps.Clone` it when it holds a map. An assertion that mutating `clone.Data` leaves the original alone would pin whichever you pick.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,440 @@
...
+const (
+	// InstructionsSystem appends A2UI capabilities to the system prompt
+	// (default).
+	InstructionsSystem = "system"
+	// InstructionsNone injects nothing (useful if you supply your own
+	// instructions).
+	InstructionsNone = "none"
+)
+
+// Config is the configuration for the A2UI [Middleware]. Add it to a generate
```
apascal07: There is no `Middleware` identifier in this package, so this doclink does not resolve and pkg.go.dev renders it as literal `[Middleware]`. Same in `loader.go:28` and `types.go:23`. Needs to be `[ai.Middleware]` I think.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,440 @@
...
+	for _, part := range resp.Message.Content {
+		if part.IsText() {
+			// Combine the streamed-push and final-flush segments so ordering
+			// (prose before/after a block) is preserved in the aggregated
+			// message too.
+			pushed, err := parser.push(part.Text)
+			if err != nil {
+				return nil, err
+			}
+			flushed, err := parser.flush()
```
apascal07: `flush()` runs once per text part rather than once per message. If a provider ever returns a response whose content is split across multiple text parts with an a2ui block spanning the boundary, the first flush finalizes the partial block: warn mode drops it, strict mode fails the whole response.

Hoisting the flush below the loop and appending its segments at the end keeps the parser's cross-part state intact. Single-text-part responses behave identically.

go/plugins/a2ui/loader.go
```diff
@@ -0,0 +1,129 @@
...
+		return fmt.Errorf("a2ui: LoadCatalog: catalog is nil")
+	}
+	if catalog.ID == "" {
+		return fmt.Errorf("a2ui: LoadCatalog: catalog has no ID")
+	}
+	if catalog.Components == nil {
+		return fmt.Errorf("a2ui: LoadCatalog: catalog %q has no components", catalog.ID)
+	}
+	key := catalogRegistryKey(catalog.ID)
+	if genkit.LookupValue(g, key) != nil {
```
apascal07: Check-then-register is racy against `registry.RegisterValue`, which panics on a duplicate key (`go/internal/registry/registry.go:177`). Two goroutines calling `LoadCatalog` with the same id can both read nil here, and the loser panics from inside library code.

Registration is normally init-time so this is unlikely to bite, but a panic is a harsh failure mode for a plugin entry point. Since this PR is already adding `genkit.RegisterValue`, a non-panicking register-if-absent variant would close the race and let this branch go away.

Separately, on the same three lines: re-registering a *different* catalog under an existing id silently keeps the old one. It is documented, but someone iterating on a catalog JSON and re-running `LoadCatalogFile` gets the stale version back with no signal that it was ignored.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+		sid, ok := payload["surfaceId"].(string)
+		if !ok || sid == "" || sid == surfaceIDPlaceholder {
+			payload["surfaceId"] = surfaceID
+		}
+	}
+
+	if cs, ok := m["createSurface"].(map[string]any); ok {
+		swapSurfaceID(cs)
+		return Envelope{"version": version, "createSurface": cs}, nil
```
apascal07: Rebuilding the envelope as a fresh two-key map drops every other top-level field. `Envelope`'s own doc says the protocol is "open-ended and versioned" and that "the middleware only inspects a few well-known keys", but this path discards exactly the keys it does not inspect, so anything the spec adds later is silently stripped on the way out.

Copying `m` and overwriting `version` preserves them. Applies to the `updateComponents`, `updateDataModel`, and `deleteSurface` arms below as well. The JS parser does the same, so if it is deliberate it is worth a comment in both.

go/ai/document.go
pavelgj: Good catch, fixed. uri.Data now falls back to Data.(string) when Text is empty, so a data part carrying a data URI converts again. Added test cases in uri_test.go covering a data-URI data part (base64 text and image/png).

go/ai/document.go
pavelgj: Done. Dropped the `|| p.IsData()` arms in both Message.Text() and ModelResponseChunk.Text() and added a comment noting data parts are deliberately excluded (their payload is on Data, and envelope JSON must not leak into Text()). Also guarded the len(m.Content)==1 fast path so it only returns Text for a text part.

go/ai/document_test.go
pavelgj: Done. Part.Clone now maps.Clone's the Data value when it holds a map[string]any (the common case for data parts / A2UI envelopes), and the doc says nested values are still shared. Added an assertion in TestPartClone that mutating clone.Data's top-level keys leaves the original untouched.

go/plugins/a2ui/a2ui.go
pavelgj: Done. Changed all three doclinks (a2ui.go, loader.go, types.go) to [ai.Middleware] so they resolve on pkg.go.dev.

go/plugins/a2ui/a2ui.go
pavelgj: Done. Hoisted flush() below the loop in transformResponse and append its segments at the end, so a block spanning multiple text parts is no longer finalized early. Single-text-part responses are unchanged. (I also ported the JS middleware's post-next() stream flush so stream-only consumers get the withheld tail / final block.)

go/plugins/a2ui/loader.go
pavelgj: Done. Added a non-panicking genkit.RegisterValueIfAbsent (backed by registry.RegisterValueIfAbsent) and LoadCatalog now uses it, so two goroutines racing on the same id both resolve to the stored catalog and the loser no longer panics from inside the registry. On the second point: re-registering a different catalog under an existing id now logs a slog.Warn (keeping the original) instead of silently swallowing it, so re-running LoadCatalogFile on an edited JSON gives a signal.

go/plugins/a2ui/parser.go
pavelgj: Agreed, fixed. normalizeEnvelope now copies the incoming envelope and overwrites version, preserving every other top-level key, instead of rebuilding a two-key map. Applied to all four arms. I made the matching change in the JS parser so the two stay byte-compatible, and added a 'preserves forward-compatible top-level envelope keys' test in both.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,471 @@
...
+				// finalizing.
+				if final {
+					proseBuf += p.buffer
+					p.buffer = ""
+				} else {
+					keep := maxPartialFence
+					if keep > len(p.buffer) {
+						keep = len(p.buffer)
+					}
+					safeLen := len(p.buffer) - keep
```
apascal07, line 179: `safeLen` is a byte offset, so the held-back tail can start mid-rune: `proseBuf += p.buffer[:safeLen]` emits a text part containing invalid UTF-8, and `encoding/json` turns it into U+FFFD on the wire. Any non-ASCII prose whose len-8 boundary is not a rune boundary hits this, on both the stream and the final message (the final parse pushes text through the same holdback, so `"18°C"` can come out as `18�` + `�C` across two parts). Backing `safeLen` off to a rune boundary (`utf8.RuneStart`) fixes it.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,461 @@
...
+		return nil, nil
+	}
+	newChunk := *chunk
+	newChunk.Content = newContent
+	return &newChunk, nil
+}
+
+// transformResponse transforms the final response message: prose text + a2ui
+// parts.
+func transformResponse(resp *ai.ModelResponse, catalog *Catalog, validate ValidateMode, version string, surfaceID func() string) (*ai.ModelResponse, error) {
```
apascal07, line 422: `transformResponse` and `transformChunk` rebuild every text part with `ai.NewTextPart`, which drops the original part's `Metadata` and `ContentType` even when the message contains no a2ui block. That loses Gemini thought signatures: gemini.go attaches `Metadata["signature"]` to the first emitted part (often a plain text part) and reads it back from `Metadata` on the next request, and the model chain runs inside the tool loop with the transformed message fed back into the next request, so a thinking model with tools degrades or terminates with MISSING_THOUGHT_SIGNATURE. It also re-types `application/json` text parts to `plain/text`, and adjacent text parts can be fused and re-split by the shared parser. Passing through text parts that contain no fence, or copying `Metadata`/`ContentType` onto the rebuilt parts, would avoid this.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,461 @@
...
+				return nil, err
+			}
+			newContent = append(newContent, partsFromSegments(pushed)...)
+		} else {
+			newContent = append(newContent, part)
+		}
+	}
+	// Flush once per message (not per text part): finalizes the last withheld
+	// prose tail and any unterminated trailing block, preserving ordering.
+	flushed, err := parser.flush()
```
apascal07: Flushing once after the whole loop means text the parser withheld from part N is emitted after every part that follows it. A routine tool-calling turn `[Text("Checking the weather."), toolRequest]` becomes `[Text("Checking the "), toolRequest, Text("weather.")]`: the held-back tail (up to 8 bytes, or a whole unterminated block) moves behind the toolRequest in recorded history, in traces, and in the messages replayed to the model on the next loop iteration. The JS middleware flushes per text part to keep ordering.

go/ai/document.go
```diff
@@ -102,9 +109,11 @@ func NewMediaPart(mimeType, contents string) *Part {
 	return &Part{Kind: PartMedia, ContentType: mimeType, Text: contents}
 }
 
-// NewDataPart returns a Part containing raw string data.
-func NewDataPart(contents string) *Part {
-	return &Part{Kind: PartData, Text: contents}
+// NewDataPart returns a Part containing arbitrary structured data. The value is
+// serialized to the part's `data` field as-is, so it may be a string, a map, a
+// slice, or any other JSON-serializable value.
+func NewDataPart(data any) *Part {
```
apascal07, line 120: This is a breaking change to stable v1 API in a minor release, in three ways: `NewDataPart("x").Text` is now `""`, so any existing reader of `.Text` on data parts silently gets nothing (this PR had to patch `plugins/internal/uri` for exactly that; third-party code gets no such fix); a part built the old way (`Part{Kind: PartData, Text: payload}`) now marshals to `{}`, silently dropping the payload; and the `data` wire field changes from always-string to any JSON value in both directions. The JS-parity motivation makes sense, but this needs an explicit compat decision, e.g. a dual-read fallback (marshal and read `Text` when `Data` is nil). Can we make the dual-read work? Also genkit/doc.go:372 still describes `NewDataPart` as "Raw data strings".

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,471 @@
...
+	// ValidateOff passes envelopes through unchecked.
+	ValidateOff ValidateMode = "off"
+)
+
+// openFenceRE matches the opening fence (```a2ui) case-insensitively.
+var openFenceRE = regexp.MustCompile("(?i)```[ \t]*a2ui[ \t]*\r?\n")
+
+// maxPartialFence is the longest prefix of an opening fence, used to hold back a
+// partial fence between chunks.
+const maxPartialFence = len("```a2ui\n")
```
apascal07: `maxPartialFence` is 8 bytes, but `openFenceRE` accepts longer openings: "``` a2ui \n", "```a2ui\t\n", "```a2ui\r\n" (unbounded `[ \t]*` padding). When a padded fence is split across chunk boundaries the holdback keeps only 8 bytes, a leading backtick leaks out as prose, and the remaining 2-backtick tail can never match: the whole envelope JSON then streams to the client as raw prose (the final-message parse recovers it, so stream and final diverge). The holdback needs to cover the longest incomplete prefix of the fence language, e.g. a suffix-anchored partial-fence regex like the JS parser's `PARTIAL_OPEN_FENCE_RE`.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,461 @@
...
+	if instructions == "" {
+		instructions = InstructionsSystem
+	}
+	nextSurfaceID := surfaceIDFactory(c.SurfaceID)
+
+	wrapModel := func(ctx context.Context, params *ai.ModelParams, next ai.ModelNext) (*ai.ModelResponse, error) {
+		// Resolve the catalog for this turn: an explicit inline catalog wins,
+		// otherwise look CatalogID up from the registry (via the Genkit instance
+		// seeded into the context), falling back to the bundled basic catalog.
+		catalog, err := resolveCatalog(genkit.FromContext(ctx), explicitCatalog, catalogID)
```
apascal07, line 144: `genkit.FromContext(ctx)` is only seeded by the `genkit.Generate` veneers and the agents runtime; the reflection server's `runAction` does not seed it. So on JSON/Dev-UI dispatch with `use: [{name: "a2ui", config: {catalogId: ...}}]` (`GenerateActionOptions.Use`), `resolveCatalog` gets `g == nil`, skips the registry lookup, and fails with "no catalog registered under id" even though the catalog is registered in the same registry serving the request. That is exactly the path the `CatalogID` godoc advertises ("survives JSON/Dev-UI dispatch"); only the bundled basic fallback works there. Resolving through the registry the action is running on, rather than a context-seeded `*Genkit`, would cover both paths.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,461 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Share surface ids between the streamed parse and the final-message
+		// parse of this single turn, so the same surface gets the same id in
+		// both (see replayableSurfaceIDs).
+		surfaceIDs := replayableSurfaceIDs(nextSurfaceID)
```
apascal07, line 152: This `surfaceIDReplay` and the `streamParser` below are created once per `wrapModel` call but shared across every invocation of `next()`. With a retrying middleware composed inside a2ui (`WithUse(&a2ui.Config{}, &middleware.Retry{})`, and WithUse(A, B) means A wraps B), each attempt re-delivers chunks through the already-wrapped callback: a failed attempt leaves the parser mid-block and its minted ids in `generated`, so the final message replays attempt 1's surface ids onto attempt 2's text, and the streamed parts disagree with history on the surface id. Reproducible with the in-tree `Retry` and `Fallback` middleware. Creating the parser/replay state per `next()` delivery, or resetting it when a new attempt's first chunk arrives, avoids it.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,461 @@
...
+		//    fence) and any unterminated trailing block still reach the
+		//    streaming consumer. Without this, clients that render purely from
+		//    stream deltas would show truncated prose / miss a final block (the
+		//    aggregated message recovers it, but the stream would not).
+		resp, err := next(ctx, params)
+		if err != nil {
+			return resp, err
+		}
+		if origCallback != nil {
+			tail, err := streamParser.flush()
```
apascal07, line 212: Two response-discarding paths here mask model state. A flush error after `next()` succeeded returns `(nil, err)`, throwing away a committed response. And `transformResponse` parses and validates regardless of `resp.FinishReason`, so under `ValidateStrict` a blocked/aborted/interrupted turn whose partial text contains a malformed a2ui block surfaces as an untyped parse error with FinishReason, FinishMessage, and Usage gone. Core generate deliberately skips output parsing for those finish reasons ("model finished abnormally, skipping output parsing"), but a middleware error short-circuits before that guard runs. Worth gating the parse on abnormal finishes, and returning `resp` alongside the error.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,471 @@
...
+			proseBuf += p.buffer[:loc[0]]
+			p.buffer = p.buffer[loc[1]:]
+			p.inBlock = true
+			p.currentSurfaceID = p.opts.surfaceID()
+			p.hasSurfaceID = true
+			continue
+		}
+
+		// In a block: look for the closing fence.
+		closeIdx := strings.Index(p.buffer, "```")
```
apascal07: `strings.Index(p.buffer, "```")` matches a fence anywhere, including inside a JSON string. The injected instructions tell the model Text content "may use inline Markdown", so `{"text": "use ``` to fence code"}` closes the block mid-JSON: the parse fails, warn mode drops the entire surface and the JSON tail leaks to the user as prose, strict mode fails the call. The JS parser anchors its close-fence regex to line start (`(^|\n)[ \t]*` before the backticks) for exactly this reason.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,471 @@
...
+		for _, c := range arr {
+			if cm, ok := c.(map[string]any); ok {
+				if id, _ := cm["id"].(string); id == "root" {
+					hasRoot = true
+					break
+				}
+			}
+		}
+		if !hasRoot {
+			return `component list must contain a component id "root".`
```
apascal07: This requires "root" in every `updateComponents` envelope of the batch, but a full render may legally split its component list across several `updateComponents` (root in the first, leaves in a second). The JS parser checks at batch level: root must appear in at least one list. As written, a split render is silently dropped in warn mode (no UI at all) or fails the call in strict mode.

go/plugins/internal/uri/uri.go
```diff
@@ -33,6 +33,13 @@ func Data(p *ai.Part) (string, []byte, error) {
 	}
 
 	uri := p.Text
+	if uri == "" {
+		// Data parts carry their payload on Data, not Text. A data part may hold
+		// a data URI string (e.g. "data:image/png;base64,...").
+		if s, ok := p.Data.(string); ok {
```
apascal07: This fallback only rescues string `Data`, and it is one of two hand-patched consumers. A data part carrying structured `Data` (map/slice), which is what the a2ui middleware and the widened `NewDataPart` doc now produce, still fails every provider request path when the middleware is not in front of the model: googlegenai and anthropic error here with "missing file data", ollama errors "unsupported content type", compat_oai silently drops the part. Replaying an a2ui session's history into a plain Generate call hits this as an opaque media error. A single accessor on `ai.Part` (string as-is, otherwise JSON-marshal, which is what googlecloud's telemetry now hand-rolls) used by `uri.Data` and the converters would give every provider the same answer.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,471 @@
...
+	//     would reset the surface and drop the update as "surface not found" if
+	//     ids disagreed), and do NOT require a root.
+	targetID := surfaceID
+	for _, e := range out {
+		if id := envelopeSurfaceID(e); id != "" {
+			targetID = id
+			break
+		}
+	}
+	isFreshRender := targetID == surfaceID
```
apascal07, line 369: `targetID == surfaceID` as the fresh-render test collapses under a fixed `Config.SurfaceID`: every minted id equals the fixed id, so the incremental branch below is unreachable and every later-turn patch targeting the surface the model learned from an action summary is treated as a fresh render. A rootless patch is then rejected by `validateRoot` (dropped in warn, error in strict), and a rooted one gets a synthesized `createSurface` that resets the surface and wipes the client data model (entered form state). The JS parser has the same collapse, so this may be worth fixing in both runtimes, e.g. by tracking surface ids observed in inbound history instead of inferring from id equality.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,461 @@
...
+
+	// Instructions controls where the catalog's capabilities are injected.
+	// InstructionsSystem (default) appends A2UI instructions to the system
+	// prompt; InstructionsNone injects nothing.
+	Instructions string `json:"instructions,omitempty"`
+
+	// Validate controls validation of emitted envelopes against the catalog.
+	// ValidateWarn (default) logs and drops bad blocks; ValidateStrict returns
+	// an error; ValidateOff skips checking.
+	Validate ValidateMode `json:"validate,omitempty"`
```
apascal07, line 92: `Validate` (and `Version` below) accept arbitrary strings. `reject()` treats any unrecognized mode as warn, so `Config{Validate: "Strict"}` or a Dev-UI JSON typo like `"strick"` silently downgrades strict validation to log-and-drop; a `Version` typo like `"0.9"` is stamped verbatim on every envelope for the renderer to reject with no server-side signal. The JS plugin's config schema enum-validates both (`SUPPORTED_VERSIONS`, validate enum). Validating both fields in `Config.New` would fail fast instead.

go/genkit/genkit.go
```diff
@@ -426,6 +426,40 @@ func LookupAction(g *Genkit, key string) api.Action {
...
 
+// RegisterValue records an arbitrary value in the registry under the given
+// name. Values are namespaced by convention using a "/type/name" key so
+// tooling can enumerate them by type (e.g. the Dev UI's GET /api/values?type=).
+// It panics if a value with the same name is already registered.
+//
+// This is the general-purpose counterpart of the typed Define* helpers, for
+// plugins that need to publish non-action resources (e.g. catalogs) that the
+// Dev UI or other components can discover via [ListValues].
+func RegisterValue(g *Genkit, name string, value any) {
```
apascal07: Three loose ends in this new public surface before it becomes permanent v1 API: `RegisterValue` (the panicking variant) has no callers, since `LoadCatalog` uses `RegisterValueIfAbsent`; `ListValues(g)` returns the raw registry keyspace including internal keys (`/format/*`, default-model bookkeeping), so callers must know the "/type/name" convention and prefix-filter by hand, where JS exposes `listValues(type)` and the v1 HTTP endpoint requires a type filter; and the v2 reflection protocol rejects `type=a2ui-catalog` outright ("Only 'defaultModel' and 'middleware' are supported"), so the README's Dev-UI discovery story only works over the legacy v1 endpoint. Typed `(valueType, name)` wrappers plus a v2 allowlist entry would close all three.

go/plugins/a2ui/loader.go
```diff
@@ -0,0 +1,139 @@
...
+	if catalog.Components == nil {
+		return fmt.Errorf("a2ui: LoadCatalog: catalog %q has no components", catalog.ID)
+	}
+	key := catalogRegistryKey(catalog.ID)
+	// Register-if-absent closes the check-then-register race against
+	// RegisterValue (which panics on a duplicate key): concurrent LoadCatalog
+	// calls for the same id all resolve to the same stored catalog, and the
+	// losers simply observe that an entry already existed.
+	if !genkit.RegisterValueIfAbsent(g, key, catalog) {
+		if existing, ok := genkit.LookupValue(g, key).(*Catalog); ok && existing != catalog {
```
apascal07: `existing != catalog` compares pointers, so re-registering equal content from a fresh allocation warns: `RegisterBasicCatalog` twice (`BasicCatalog()` returns a new pointer per call), or `LoadCatalogFile` twice on the same file, logs "keeping the existing one" even though the doc calls this idempotent, and an actually-edited catalog produces the identical warning, so the distinction the check exists for is not made. Content equality (`reflect.DeepEqual`) or an unconditional warn would both be truthful. `TestLoadCatalogIdempotent` passes the same pointer twice, so it does not cover this.

go/ai/document.go
```diff
@@ -339,15 +348,14 @@ func (p *Part) unmarshalPartFromSchema(s partSchema) {
 		p.Kind = PartReasoning
 		p.Text = s.Reasoning
 		p.ContentType = "plain/text"
+	case s.Data != nil:
```
apascal07, line 384: A typed-nil Data value survives the nil checks asymmetrically: `NewDataPart((map[string]any)(nil))` marshals as `{"data":null}` (a nil map inside a non-nil interface is not omitted), and on unmarshal `s.Data` is nil, so this case is skipped and the part comes back as an empty text part with `IsData() == false`. An explicit `"data": null` from another runtime flips kind the same way. Worth either normalizing in `NewDataPart` or treating a present-but-null `data` key as a data part.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,471 @@
...
+	var known map[string]bool
+	if opts.catalog != nil {
+		known = opts.catalog.componentNames()
+	}
+	return &streamParser{opts: opts, knownComponents: known}
+}
+
+// push feeds a chunk of model text, returning prose + any completed blocks.
+func (p *streamParser) push(text string) ([]segment, error) {
+	p.buffer += text
```
apascal07, line 138: nit: while inside a block, each push appends to `p.buffer` (recopying the retained block) and drain rescans it from offset 0 with `strings.Index`, so a streamed block is O(n²) in its size: a 100KB envelope block arriving in 100-byte deltas rescans roughly 100MB and recopies roughly 50MB. Outside a block the buffer is capped at 8 bytes, so only the in-block path is affected. Tracking a scan offset (search only the appended text plus a 2-byte overlap) and accumulating block text in a `strings.Builder` makes it linear.

go/plugins/a2ui/loader.go
```diff
@@ -0,0 +1,139 @@
...
+// it appears alongside custom catalogs in tooling. The middleware falls back to
+// the basic catalog even without this call; register it explicitly to surface
+// it in the Dev UI. Idempotent.
+func RegisterBasicCatalog(g *genkit.Genkit) error {
+	return LoadCatalog(g, BasicCatalog())
+}
+
+// readCatalogFile reads and parses a catalog from a JSON file (without
+// registering it).
+func readCatalogFile(path string) (*Catalog, error) {
```
apascal07, line 101: nit: `base.ReadJSONFile` (go/internal/base/json.go) already does the read-plus-decode; only the a2ui-specific error wrapping and the components check need to live here.

go/plugins/a2ui/a2ui.go
```diff
@@ -0,0 +1,461 @@
...
+		} else if seg.prose != "" {
+			out = append(out, ai.NewTextPart(seg.prose))
+		}
+	}
+	return out
+}
+
+// injectInstructions appends A2UI instructions to (or creates) the system
+// message.
+func injectInstructions(req *ai.ModelRequest, catalog *Catalog) *ai.ModelRequest {
```
apascal07, line 370: go/ai has an unexported `injectInstructions` (format.go) that this re-implements with different semantics: the ai one tags the injected part with purpose metadata so tooling and later stages can recognize injected instructions and not double-inject, and it falls back to the last user message when no system message exists. Two parallel injectors will drift; widening and exporting the ai helper (parameterizing the purpose tag) would keep one behavior.

go/plugins/a2ui/parser.go
```diff
@@ -0,0 +1,471 @@
...
+// callers can `return p.reject(...)`.
+func (p *streamParser) reject(message string) ([]Envelope, error) {
+	full := "A2UI: " + message
+	switch p.opts.validate {
+	case ValidateOff:
+		return nil, nil
+	case ValidateStrict:
+		return nil, fmt.Errorf("%s", full)
+	default:
+		slog.Warn(full + " (dropping block/envelope)")
```
apascal07, line 276: nit: sibling middleware logs through `logger.Warn(ctx, msg, key, value...)` (plugins/middleware/fallback.go, skills.go), so warnings carry the span for Dev UI correlation plus structured attrs. These drops go through bare `slog.Warn` with the reason embedded in the string and no ctx (push/flush do not take one), so dropped blocks are invisible in per-span logs.

go/plugins/a2ui/types.go
```diff
@@ -0,0 +1,76 @@
...
+// updateComponents, updateDataModel, deleteSurface). It is represented as a
+// generic JSON object because the protocol is open-ended and versioned; the
+// middleware only inspects a few well-known keys.
+type Envelope = map[string]any
+
+// Component is a single entry in an A2UI adjacency list. UI is expressed as a
+// flat list of components; the tree is reconstructed via id references, with
+// exactly one component having id "root". Beyond component/id, every component
+// carries catalog-specific props, so it is represented as a generic object.
+type Component = map[string]any
```
apascal07: nit: `Component` is exported but referenced nowhere (the parser and `validateComponents` work on raw `map[string]any`). Under the v1 compatibility bar an unused export becomes permanent; either use it in the signatures that conceptually take components, or drop it and keep the doc prose in the package comment.

go/ai/document.go
```diff
@@ -48,14 +49,20 @@ type Part struct {
...
+// envelopes), that top-level map is cloned too so callers can mutate its keys
+// without disturbing the original; nested values are still shared by reference.
 func (p *Part) Clone() *Part {
 	if p == nil {
 		return nil
 	}
 	cp := *p
 	cp.Custom = maps.Clone(p.Custom)
 	cp.Metadata = maps.Clone(p.Metadata)
+	if m, ok := p.Data.(map[string]any); ok {
```
apascal07: nit: the isolation is shape-dependent: map Data gets a top-level clone, while `[]any` Data (equally valid per `NewDataPart`'s doc) gets none. A caller switching a payload from an object to an array silently loses the isolation Clone gives objects. Deep-copying Data uniformly (or a JSON round-trip) would make the guarantee shape-independent.

go/plugins/a2ui/README.md
```diff
@@ -0,0 +1,175 @@
...
+| `SurfaceID`    | fresh UUID         | Surface id policy. Provide a fixed string to reuse one id for every surface.                       |
+| `Version`      | `"v0.9"`           | Protocol version stamped on envelopes.                                                            |
+
+### Custom catalogs
+
+The bundled `BasicCatalog()` mirrors `@a2ui/web_core`'s basic catalog. To use
+your own components, register a `Catalog` with `LoadCatalog` (or `LoadCatalogFile`)
+and reference it by id. Registered catalogs live in the Genkit registry under
+the value type `a2ui-catalog`, so they are discoverable by tooling (the Dev UI's
+`GET /api/values?type=a2ui-catalog`) and shared identically across the JS, Go,
```
apascal07: "shared identically across the JS, Go, and Dart plugins" overstates it: JS's `loadCatalog` keys the registry by a caller-chosen lookup id and its middleware config field is `catalog`, while Go keys strictly by the catalog's own ID and the config field is `catalogId`. Tooling or shared middleware config that matches catalogs by key across runtimes will not line up.

go/plugins/a2ui/README.md
```diff
@@ -0,0 +1,175 @@
...
+### Options
+
+`a2ui.Config` fields:
+
+| Field          | Default            | Description                                                                                       |
+| -------------- | ------------------ | ------------------------------------------------------------------------------------------------- |
+| `Catalog`      | nil                | An inline catalog (code-defined use only; not serialized). Overrides `CatalogID` when set.        |
+| `CatalogID`    | `"basic"`          | Id of a catalog registered with `LoadCatalog`. Resolved from the registry at call time.           |
+| `Instructions` | `"system"`         | Where to inject catalog capabilities. `"none"` injects nothing.                                    |
+| `Validate`     | `"warn"`           | Validate emitted envelopes. `"warn"` logs and drops bad blocks; `"strict"` returns an error; `"off"` skips checking. |
```
apascal07: The JS README carries a "Security and the trust boundary" section: strict validation is a well-formedness check, not sanitization; model-controlled `Image.url` and Markdown text pass through; the renderer owns prop sanitization and hosts should CSP-restrict. The Go middleware has identical pass-through behavior, but neither this table nor the `Validate` godoc says so, and "strict" reads stronger than it is. Worth porting that section.

go/plugins/a2ui/parser.go
pavelgj: Fixed. The holdback is now regex-driven (partialOpenFenceRE) instead of a fixed byte count, and it only ever holds back an ASCII fence suffix, so emitted prose always ends on a rune boundary. Added TestParserNoMidRuneSplitInStreamedProse, which streams "18°C … 日本語 …" one byte at a time and asserts no U+FFFD and exact reassembly.

go/plugins/a2ui/a2ui.go
pavelgj: Fixed. transformChunk/transformResponse now pass a fence-free text part through unchanged (new partsForTextPart helper) instead of rebuilding it with NewTextPart, so Metadata (incl. Gemini thought signatures) and ContentType survive; only a part that actually contained a fence is rewritten. Added TestMiddlewarePreservesTextPartMetadata (asserts Metadata["signature"] and application/json ContentType survive).

go/plugins/a2ui/a2ui.go
pavelgj: Fixed, and this reverses the round-1 hoist you flagged. transformResponse now pushes+flushes per text part (matching the JS middleware) so withheld text lands in place, not behind a following toolRequest. The round-1 concern (a block spanning two text parts finalizing early) does not arise in practice because the aggregator coalesces adjacent text; I documented that trade-off in the function comment.

go/ai/document.go
pavelgj: Addressed the concrete breakages except the always-string case: added Part.DataString() (string as-is, else JSON-marshal) as the single accessor uri.Data and provider converters read a data part payload through, and fixed genkit/doc.go:372 ("Raw data strings" -> arbitrary structured data). We are explicitly not supporting a nil / "data":null payload — a data part with no data is not meaningful and round-trips to an empty text part, which is fine. The always-string-marshals-to-{} case is inherent to moving the payload off Text; flagging for your call on whether we also want a Text->Data dual-read fallback for old third-party code.

go/plugins/a2ui/parser.go
pavelgj: Fixed. Replaced the fixed 8-byte maxPartialFence with partialOpenFenceRE, a suffix-anchored regex mirroring the JS PARTIAL_OPEN_FENCE_RE, so the whole incomplete fence prefix (including unbounded [ \t]* padding) is held back. Added TestParserPaddedOpenFenceSplit which streams a padded "```a2ui   \n…" byte-by-byte and asserts no backticks leak to prose and the block still opens.

go/plugins/a2ui/a2ui.go
pavelgj: Not fixing in this PR — leaving as-is for now. Reaching the action-scoped registry from the model hook needs plumbing beyond genkit.FromContext (the reflection runAction path does not seed it), which is a broader change than the rest of this cleanup. The bundled basic fallback still works on that path; I would rather do the registry-threading properly as a follow-up than bolt it on here. Filed as a known limitation. Happy to prioritize if you would rather it block the PR.

go/plugins/a2ui/a2ui.go
pavelgj: Documented rather than auto-detected. Reliable per-attempt reset needs a signal the model hook does not get (chunk indices are constant within a stream for many providers), so I added a Config godoc note: put retry/fallback OUTSIDE a2ui (WithUse(retry, &a2ui.Config{})) so each attempt gets a fresh A2UI turn. Open to a sturdier mechanism if core grows an attempt marker.

go/plugins/a2ui/a2ui.go
pavelgj: Fixed both. The middleware now returns resp alongside the error on a flush or transform failure instead of (nil, err), and it skips parsing entirely on an abnormal FinishReason (blocked/aborted/interrupted/other, mirroring ai FinishReason.isAbnormal via a local isAbnormalFinish), returning the response untouched. Added TestMiddlewareSkipsParseOnAbnormalFinish (a blocked turn with a malformed partial block keeps FinishReason/FinishMessage and does not surface a parse error).

go/plugins/a2ui/parser.go
pavelgj: Fixed. The close fence is now closeFenceRE = (?m)^[ \t]*``` (line-anchored, mirroring the JS CLOSE_FENCE_RE) instead of strings.Index for "```", so a ``` inside a JSON string no longer truncates the block. Added TestParserBacktickInsideJSONDoesNotCloseBlock.

go/plugins/a2ui/parser.go
pavelgj: Fixed. validateRoot is now batch-level: root must appear in at least one updateComponents list, and a batch carrying no component lists needs no root (matching the JS parser). Added TestParserSplitRenderRootInLaterList (leaves in the first list, root in a second).

go/plugins/internal/uri/uri.go
pavelgj: Addressed via the shared accessor. uri.Data now reads through Part.DataString() (string as-is, otherwise JSON-marshal), the single answer converters should use for a data part payload, so structured Data no longer dead-ends differently per provider. Migrating each provider converter to DataString() I left as a follow-up so this PR stays scoped to a2ui + the shared primitive; the accessor is in place for them to adopt.

go/plugins/a2ui/parser.go
pavelgj: Acknowledged; not fixing here. This is a shared JS/Go design limitation (id-equality as the fresh-render test) and fixing it properly means tracking surface ids observed in inbound history in both runtimes, which is out of scope for this round. The default (per-surface UUID) path is unaffected; the collapse only bites under a fixed Config.SurfaceID. Noting as a known limitation for the cross-runtime fix.

go/plugins/a2ui/a2ui.go
pavelgj: Fixed. Config.New now enum-validates Validate, Version (against SupportedVersions), and Instructions, returning an error instead of silently downgrading a typo like "strick" to warn or stamping a bad version. Added TestConfigRejectsInvalidValidateMode / TestConfigRejectsUnsupportedVersion / TestConfigRejectsInvalidInstructions.

go/genkit/genkit.go
pavelgj: Left the value API as-is for this PR — trimming/typing it (dropping the panicking RegisterValue, a typed ListValues(valueType), and a v2 reflection allowlist entry for a2ui-catalog) touches core genkit surface beyond a2ui and deserves its own review. The a2ui plugin only needs RegisterValueIfAbsent + LookupValue, which are in use. Flagging for your call: fold the value-API cleanup in here, or do it as a follow-up before it is treated as stable?

go/plugins/a2ui/loader.go
pavelgj: Fixed. The "already registered" check now uses reflect.DeepEqual instead of pointer identity, so BasicCatalog() twice or re-loading the same file no longer warns, and an actually-different catalog under an existing id does. Message reworded to "a different catalog is already registered".

go/ai/document.go
pavelgj: Treating a nil / present-but-null data payload as unsupported rather than special-casing it: a data part with no data is not meaningful, and NewDataPart(nil) / {"data":null} round-tripping to an empty text part is acceptable. Not adding decode-time detection, since that would need either a field on the standardized wire-schema type or a raw-JSON probe in UnmarshalJSON, both more than this edge case warrants.

go/plugins/a2ui/parser.go
pavelgj: Fixed. Added a blockScan cursor so the in-block close-fence search resumes from the last line boundary instead of rescanning the whole retained block each delta, making a large streamed block close to linear. Added TestParserLargeBlockManyDeltas (~10KB block fed in 7-byte deltas).

go/plugins/a2ui/loader.go
pavelgj: Done. readCatalogFile now uses base.ReadJSONFile for the read+decode; only the a2ui-specific error wrapping and the components check remain here.

go/plugins/a2ui/a2ui.go
pavelgj: Left as a follow-up. Exporting/parameterizing ai injectInstructions (with the purpose tag + last-user-message fallback) is a change to the ai package surface that other stages depend on; I kept the a2ui-local injector for this PR to avoid widening scope. Noting the drift risk so we can unify them deliberately. Let me know if you want that pulled forward.

go/plugins/a2ui/parser.go
pavelgj: Acknowledged; not changed here. Threading ctx into push/flush so drops log through logger.Warn(ctx, …) with span correlation is a signature change through the parser hot path; I would rather do it alongside the injectInstructions/ai-helper unification as a logging pass than piecemeal. Left as a known follow-up.

go/plugins/a2ui/types.go
pavelgj: Fixed. Dropped the unused Component type export and folded its doc prose into the Envelope/package comment, so we do not lock an unused symbol into the v1 surface. The parser and validateComponents already work on raw map[string]any.

go/ai/document.go
pavelgj: Fixed. Part.Clone now clones a top-level []any Data as well as map[string]any, so the isolation guarantee is shape-independent (an object-to-array payload switch no longer silently loses it). Added TestPartCloneSliceData. Kept it a top-level (not deep) clone to match the documented "nested values still shared" contract.

go/plugins/a2ui/README.md
pavelgj: Fixed. Reworded: the wire representation (catalog + envelopes) is identical across runtimes, but the registry key differs — Go keys by the catalog ID with config field CatalogID, JS keys by a caller-chosen lookup id with field catalog — so match on the catalog id value, not the registry key.

go/plugins/a2ui/README.md
pavelgj: Done. Added a "Security and the trust boundary" section to the README (strict = well-formedness, not sanitization; Image.url and Markdown text pass through; renderer owns prop sanitization; hosts should CSP-restrict) and a matching note on the Validate godoc and the options table.

go/genkit/genkit.go
apascal07: Usually we call these `Define*` at the `genkit` package level, only exception is `RegisterAction` because it's a `genkit`-level wrapper for calling `Action.Register()`.

apascal07 (approved):

go/genkit/genkit.go
pavelgj: renamed

PR #5852, opened 2026-07-30
fix(py): Expose send/send_stream and resume/resume_stream on AgentChat client

huangjeff5:
## Description

Refactor `AgentChat` client methods to explicitly separate streaming and non-streaming calls:
- `send(...)` -> `send(...)` (non-streaming, returns `AgentResponse`) & `send_stream(...)` (streaming, returns `AgentTurn`)
- `resume(...)` -> `resume(...)` (non-streaming, returns `AgentResponse`) & `resume_stream(...)` (streaming, returns `AgentTurn`)

### Rationale
- **Cross-SDK Parity**: Cleans up syntax to align with JS, Go, and Dart SDKs which separate streaming and non-streaming chat methods.
- **Python Consistency**: Matches existing Python SDK action execution patterns (`Action.run` vs `Action.stream` / `ai.generate` vs `ai.generate_stream`).

### Implementation Details
- Shared turn construction via `_start_turn`:
  - `send_stream()` passes `chunks=CloseableQueue()`.
  - `send()` passes no queue (`chunks=None`).
  - `TurnDriver` infers streaming capabilities directly from whether `chunks` is provided.
- Updated all agent samples in `py/samples/agents/` to demonstrate correct streaming vs non-streaming usage.
- Exported `AgentInitError` on the `genkit.agent` public API surface.

## Checklist:
- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] Tested (unit tests updated and passed)
- [x] Docs & samples updated


py/packages/genkit/src/genkit/_core/_tracing.py
```diff
@@ -189,6 +189,10 @@ def record_span_outcome(
     """Write success or error attrs after the span body finishes."""
     try:
         yield
+    except GenkitInterrupt:
```
apascal07, line 193: Re-raising here skips the trailing `set_attribute(Attr.STATE, State.SUCCESS)` too, so interrupt spans end with no `genkit:state` at all. `genkit-google-cloud/telemetry/feature.py:110` logs `Unknown state` and drops the metric for those. Set `State.SUCCESS` rather than leaving it absent?

Also worth noting: JS has no equivalent carve-out in `instrumentation.ts` and `ToolInterruptError` does throw out of the tool action, so this is a Python-only divergence today.

py/packages/genkit/src/genkit/_core/_tracing.py
```diff
@@ -227,6 +231,14 @@ def run_in_new_span(
...
 
     with push_parent_path(qualified_path):
-        with tracer.start_as_current_span(name=metadata.name, links=links, attributes=start_attrs) as span:
+        # record_span_outcome owns success/error attrs so control-flow
+        # GenkitInterrupt can re-raise without OTEL painting the span red.
+        with tracer.start_as_current_span(
+            name=metadata.name,
+            links=links,
+            attributes=start_attrs,
+            record_exception=False,
```
apascal07, line 250: `record_span_outcome` only catches `Exception`, so `CancelledError`/`KeyboardInterrupt` escaping a span body now leave OTel status UNSET with no exception event, where it was ERROR before. That covers aborted and timed-out turns. Probably the behavior you want, but it's silent — worth a comment or an explicit cancellation branch.

py/packages/genkit/tests/genkit/core/run_in_new_span_test.py
```diff
@@ -210,6 +212,23 @@ def test_records_error_attributes(exporter: InMemorySpanExporter) -> None:
     assert span.status.status_code == trace_api.StatusCode.ERROR
 
 
+def test_interrupt_is_not_recorded_as_span_error(
```
apascal07: This raises a bare `Interrupt` directly inside the span. In production it comes out of a tool action, and `_tools.py:122` says those get wrapped in `GenkitError(cause=Interrupt)` — the carve-out only works because `_action.py:773` wraps *outside* the `with run_in_new_span` block. That ordering is load-bearing and untested. A case driving a real tool that raises `Interrupt` would keep a future refactor from silently undoing this.

py/packages/genkit/src/genkit/_ai/_agents/_client.py
```diff
@@ -903,6 +951,17 @@ def __init__(
...
         if init is not None:
             validate_init(init)
+            # Store-backed chats resume by snapshot/session id only. Accepting a
+            # seed state blob here would look live on the client while the server
+            # session stayed empty — so refuse up front instead of dropping it.
+            if init.state is not None and self._transport.state_management == 'server':
+                raise AgentInitError(
+                    status='FAILED_PRECONDITION',
+                    message=(
+                        "Cannot send 'state' to a server-managed agent (one with a "
```
apascal07: `init_from` bundles messages/artifacts/state into one `SessionState`, so `agent.chat(messages=[...])` on a store-backed agent hits this too and gets told it sent 'state' (verified locally). Matches the wire check at `_runtime.py:296`, so it's consistent — but naming the field the caller actually passed would save some head-scratching.

py/packages/genkit/src/genkit/_ai/_agents/_client.py
```diff
@@ -801,20 +814,49 @@ async def run(self) -> None:
...
+        if self.turn is None:
+            raise RuntimeError('TurnDriver.start() requires a chunks queue')
+        self.run_task = asyncio.create_task(self._run_background())
+        return self.turn
+
+    async def _run_background(self) -> None:
+        """Background wrapper so task exceptions stay on ``self.output`` / the stream."""
+        try:
+            await self.run()
+        except Exception:  # noqa: S110 — surfaced via self.output / stream, not this task
```
apascal07: If `on_turn_error` itself raises, `output` never resolves and `chunks` closes empty, so `await turn.response` hangs forever with nothing logged. Pre-existing shape, but this new wrapper is a natural place for a `logger.exception` fallback.

apascal07 (approved):

py/packages/genkit/src/genkit/_core/_tracing.py
huangjeff5: Good catch! Updated `record_span_outcome` to explicitly set `span.set_attribute(Attr.STATE, State.SUCCESS)` on `GenkitInterrupt` before re-raising so cloud telemetry receives a valid `genkit:state` without painting the span red.

py/packages/genkit/src/genkit/_core/_tracing.py
huangjeff5: Added an explicit `except (asyncio.CancelledError, KeyboardInterrupt):` branch to `record_span_outcome` with comments clarifying that aborts/timeouts intentionally leave `genkit:state` absent (OTel stays UNSET) so metrics treat them as unfinished work rather than errors or successes.

py/packages/genkit/tests/genkit/core/run_in_new_span_test.py
huangjeff5: Updated `test_tool_interrupt_is_not_recorded_as_span_error` to define and invoke a real `@ai.tool` with `Genkit()` that raises `Interrupt({'reason': 'needs_approval'})`, asserting that `GenkitError.cause` is an `Interrupt` and span outcome attributes/status remain `success`.

py/packages/genkit/src/genkit/_ai/_agents/_client.py
huangjeff5: Added `seeded_init_fields` helper to inspect `SessionState` and dynamically list the field(s) actually passed by the caller (e.g. `'messages'`, `'artifacts'`, or `'state'`) in the `AgentInitError` message instead of hardcoding `'state'`.

py/packages/genkit/src/genkit/_ai/_agents/_client.py
huangjeff5: Updated `_run_background` to catch `Exception`, log `logger.exception(...)`, and set `self.output.set_exception(e)` if `self.output` is not yet resolved so `await turn.response` fails immediately with the exception instead of hanging indefinitely.

PR #5806, opened 2026-07-23
feat(py): support Google AI Interactions and background models (Veo, Deep Research, Antigravity, Lyria)

huangjeff5:
Adds support for various background models on Interactions API.

Also adds support for Veo on the legacy API (Veo isn't on Interactions API).

PR #5854, opened 2026-07-30
feat(py): core support for background model operations and typed model refs

cabljac:
First slice of the #5806 split: `FinishReason.ABORTED` (typing + genkit-schema.json), a cancel-operation action kind with registry lookup, background operation lifecycle updates, generate_operation/check_operation plumbing, and a `config_schema` parameter on `model_ref`.

## Stack (split of #5806)

1. #5854 core support for background model operations and typed model refs
2. #5855 Interactions HTTP client, converters, and options
3. #5856 Deep Research, Antigravity, and Lyria interaction models
4. #5857 move Veo to background model actions
5. #5858 wire Interactions models into GoogleAI plugin
6. #5859 gemini finish_message (parallel, based on this PR)

The split originally reproduced #5806 byte-for-byte; this slice has since picked up review fixes. Work co-authored with @huangjeff5.

### Breaking changes

- `ModelRequest(config={...})` keeps the dict instead of coercing into `ModelConfig`; `ModelRequest[PluginConfig]` coerces dicts into the plugin schema and rejects mismatched config instances. Values that are neither `BaseModel` nor dict raise `TypeError`.
- Bare `ModelRequest.config` is no longer typed `ModelConfig | None`: it resolves to `Unknown | None` under pyright, so attribute access on it is unchecked. Bind `ModelRequest[YourConfig]` for typed access.

### Cross-language files

The `aborted` finish reason already exists in the JS runtime; this PR syncs the tools zod source, the generated `genkit-schema.json`, and Go's generated `go/ai/gen.go` (CI enforces all three). No hand-written Go or tools changes.

### DIFF
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -70,6 +70,7 @@ from genkit._core._typing import (
     FinishReason,
     MiddlewareRef,
     MultipartToolResponse,
+    Operation,
     Part,
     Role,
     TextPart,
@@ -747,6 +748,12 @@ async def _generate_action_turn(
                 next_fn,
             )
 
+        # Background models return an Operation to poll, not a finished message.
+        # The tool loop below assumes response.message exists, so branch out first.
+        if model.kind == ActionKind.BACKGROUND_MODEL:
+            operation = cast(Operation, model_response)
+            return ModelResponse(operation=operation, request=request)
+
         def message_parser(msg: Message) -> Any:  # noqa: ANN401
             if formatter is None:
                 return None
@@ -755,10 +762,11 @@ async def _generate_action_turn(
         # Extract schema_type for runtime Pydantic validation
         schema_type = turn_options.output.schema_type if turn_options.output else None
 
-        # Plugin returns ModelResponse directly. Framework sets request and
-        # any output format context (message_parser, schema_type) as private attrs.
+        # Plugin returns ModelResponse directly. Framework fills request when the
+        # plugin didn't already attach one (e.g. with a normalized config echo).
         response = model_response
-        response.request = request
+        if response.request is None:
+            response.request = request
         if formatter:
             response._message_parser = message_parser
         if schema_type:
@@ -1079,11 +1087,17 @@ async def resolve_parameters(
         else cast(str | None, registry.lookup_value('defaultModel', 'defaultModel'))
     )
     if not model:
-        raise Exception('No model configured.')
+        raise GenkitError(status='INVALID_ARGUMENT', message='No model configured.')
 
     model_action = await registry.resolve_model(model)
     if model_action is None:
-        raise Exception(f'Failed to to resolve model {model}')
+        hint = ''
+        if isinstance(model, str) and '/' not in model:
+            hint = f" Did you mean 'googleai/{model}' or 'vertexai/{model}'?"
+        raise GenkitError(
+            status='NOT_FOUND',
+            message=f"Failed to resolve model '{model}'.{hint}",
+        )
 
     # Resolve tools up front to fail fast on invalid caller-supplied tool names or
     # duplicate short names before running side effects or middleware.
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -1091,12 +1091,12 @@ async def resolve_parameters(
 
     model_action = await registry.resolve_model(model)
     if model_action is None:
-        hint = ''
+        message = f"Failed to resolve model '{model}'."
         if isinstance(model, str) and '/' not in model:
-            hint = f" Did you mean 'googleai/{model}' or 'vertexai/{model}'?"
+            message += f" Ensure the model name includes a provider prefix (e.g., 'googleai/{model}' or 'vertexai/{model}')."
         raise GenkitError(
             status='NOT_FOUND',
-            message=f"Failed to resolve model '{model}'.{hint}",
+            message=message,
         )
 
     # Resolve tools up front to fail fast on invalid caller-supplied tool names or
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -1093,7 +1093,7 @@ async def resolve_parameters(
     if model_action is None:
         message = f"Failed to resolve model '{model}'."
         if isinstance(model, str) and '/' not in model:
-            message += ' Ensure the model name includes a plugin prefix.'
+            message += ' Ensure the model name includes the plugin namespace.'
         raise GenkitError(
             status='NOT_FOUND',
             message=message,
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -1093,9 +1093,7 @@ async def resolve_parameters(
     if model_action is None:
         message = f"Failed to resolve model '{model}'."
         if isinstance(model, str) and '/' not in model:
-            message += (
-                f" Ensure the model name includes a provider prefix (e.g., 'googleai/{model}' or 'vertexai/{model}')."
-            )
+            message += ' Ensure the model name includes a plugin prefix.'
         raise GenkitError(
             status='NOT_FOUND',
             message=message,
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -765,8 +765,7 @@ async def _generate_action_turn(
         # Extract schema_type for runtime Pydantic validation
         schema_type = turn_options.output.schema_type if turn_options.output else None
 
-        # Plugin returns ModelResponse directly. Framework fills request when the
-        # plugin didn't already attach one (e.g. with a normalized config echo).
+        # Attach the request to the response if the model plugin did not already include one.
         response = model_response
         if response.request is None:
             response.request = request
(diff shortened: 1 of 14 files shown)


py/packages/genkit/src/genkit/_core/_action.py
```diff
@@ -488,10 +488,18 @@ async def run(
...
         if self._input_type is not None and not skip_validation:
             try:
-                input = self._input_type.validate_python(input)
+                try:
+                    input = self._input_type.validate_python(input)
+                except ValidationError:
+                    if not isinstance(input, BaseModel):
+                        raise
+                    input = self._input_type.validate_python(input.model_dump(mode='python'))
             except ValidationError as e:
```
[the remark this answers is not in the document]
cabljac: The premise doesn't hold on the current code: validate_python succeeds without raising for both dict configs and cross-schema BaseModel configs (verified empirically), so the except path is a rare fallback, not per-call overhead. The suggested upfront model_dump would instead copy the full request - including long message histories - on every dict-config call, which is worse for exactly the cases cited. Keeping the retry form.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -1079,11 +1087,17 @@ async def resolve_parameters(
...
     if not model:
-        raise Exception('No model configured.')
+        raise GenkitError(status='INVALID_ARGUMENT', message='No model configured.')
 
     model_action = await registry.resolve_model(model)
     if model_action is None:
-        raise Exception(f'Failed to to resolve model {model}')
+        hint = ''
+        if isinstance(model, str) and '/' not in model:
+            hint = f" Did you mean 'googleai/{model}' or 'vertexai/{model}'?"
```
hilariie: googleai and vertexai in core will misdirect anyone on ollama, openai or anthropic. The registry knows which namespaces are actually registered, could we suggest those instead?

py/packages/genkit/src/genkit/_core/_action.py
```diff
@@ -695,8 +695,16 @@ def _validate_input(self, input: InputT | None) -> InputT | None:
...
+        # GenerationCommonConfig into an action typed as ModelRequest[PluginConfig]),
+        # dump to plain data and retry so the action's schema wins.
         try:
-            return self._input_type.validate_python(input)
+            try:
+                return self._input_type.validate_python(input)
+            except ValidationError:
+                if not isinstance(input, BaseModel):
+                    raise
+                return self._input_type.validate_python(input.model_dump(mode='python'))
```
hilariie: I instrumented _validate_input and ran the case the comment describes, bare ModelRequest with a dict config into an action typed ModelRequest[PluginConfig]:

    retries: 0 | got output_format: json | config: PluginConfig(temperature=0.5)

Pydantic takes it on the first pass, so the retry costs a nested try/except on every action call and I couldn't find a path that reaches it. Is there a case I'm missing?

Also, when it does fire on a ModelRequest it drops output config. GenkitModel.model_dump defaults by_alias=True, so _serialize_for_spec swaps the flat fields for a nested output key nothing reads back:

    r.model_dump(mode='python').keys()                          # ['config', 'messages', 'output']
    ModelRequest.model_validate(r.model_dump()).output_format   # None

Pre-existing gap in _serialize_for_spec, but this would be the first caller to hit it.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -311,7 +318,7 @@ async def wrapped_check(op: Operation, ctx: ActionRunContext) -> Operation:
         # Capture cancel in local scope for the nested function
         cancel_fn = cancel
 
-        async def wrapped_cancel(op: Operation, ctx: ActionRunContext) -> Operation:
+        async def wrapped_cancel(op: Operation, _: ActionRunContext) -> Operation:
```
hilariie, line 295: Cancel has no test coverage and this PR touches the signature

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -255,6 +257,27 @@ def gemini_model(request: ModelRequest[GeminiConfig]) -> ModelResponse:
...
+        if isinstance(v, BaseModel):
+            return v
+        if isinstance(v, dict):
+            args = cls.__pydantic_generic_metadata__['args']
+            if args:
+                schema = args[0]
+                if isinstance(schema, type) and issubclass(schema, BaseModel):
+                    return schema.model_validate(v)
+            return v
+        raise TypeError(f'config must be a BaseModel or dict, got {type(v).__name__}')
```
hilariie: mode='plain' replaces the whole core schema for the field, so the serializer and the JSON schema go along with the validator. A few things fall out of that.

Schema is gone. A/B on two otherwise identical generic models:

    plain    config prop: {"default": null, "title": "Config"}
    before   config prop: {"anyOf": [{"$ref": "#/$defs/Cfg"}, {"type": "null"}], "default": null}

_background.py:300 runs to_json_schema on ModelRequest[config_schema], so the empty one is what lands in inputSchema.

Type isn't enforced either. `ModelRequest[PluginConfig](config=SomeOtherConfig())` is accepted, so request.config.thinking_summaries raises AttributeError at use rather than failing validation at the boundary.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -64,7 +64,9 @@
 
 # TypeVars for generic types
 OutputT = TypeVar('OutputT', default=object)
-ConfigT = TypeVar('ConfigT', bound=ModelConfig, default=ModelConfig)
+# No default — callers/actions bind ModelRequest[TheirConfig]. A ModelConfig
+# default would reject plugin config instances on bare ModelRequest.
+ConfigT = TypeVar('ConfigT', bound=BaseModel)
```
hilariie: Dropping default=ModelConfig is a typing break for external plugins. Bare ModelRequest resolves the missing type arg to Unknown, not BaseModel, so request.config.temperature doesn't start failing pyright, it just stops being checked at all. Nothing in tree hits it though, but it might be worth a line in the breaking changes list next to the runtime coercion

hilariie (changes requested):
Overall, the CL looks good but there are two things we may want to change before this
lands, both in the config validation, and they interact:

1. mode='plain' on _validate_config replaces more than the validator. The JSON schema for config comes back as {"default": null, "title": "Config"} (define_background_model feeds exactly that into _override_input_schema), the serializer goes with it so every traced call logs a PydanticSerializationUnexpectedValue warning, and the annotation no longer matches what the field accepts at runtime, which is why pyright fails on this PR's own tests under just lint (CI only checks packages/*/src). The dev UI does still get the real schema via customOptions, same as JS, so the inputSchema loss is mostly a regression against main's $ref. I provided more details inline.

2. The _validate_input retry never fires for the case its own comment describes. I patched it to count and ran that flow, 0 retries. Pydantic revalidates instances that share the generic origin, so a bare ModelRequest is taken on the first pass, and when the config is already a model instance it passes through untouched, so the action's schema doesn't win there either. When the retry does fire on a ModelRequest (it is reachable, via a same-shape foreign BaseModel, just not from this path) the model_dump round-trip silently drops output config.

These two are coupled: if 1 moves off mode='plain', wrong-typed instance configs start
failing first-pass validation, the retry in 2 becomes the thing that rescues them, and
its output drop goes live. Whatever the fix is, it probably wants to land as one change.

Everything else is comment-level and left inline.

py/packages/genkit/src/genkit/_ai/_generate.py
cabljac: Agreed. 5cd304bbd builds the hint from a new `Registry.plugin_names()`; no plugins, no hint. Tests cover both.

py/packages/genkit/src/genkit/_core/_action.py
cabljac: Right on both counts. After 286e2250b the stricter union makes cross-schema instance configs fail first-pass, so the retry now rescues exactly the flow its comment describes, and output unnesting makes the model_dump round-trip lossless. Round-trip test plus an e2e through Action.run pin it.

py/packages/genkit/src/genkit/_core/_background.py
cabljac: Added in 7217039c3: handler invocation with action key stamped, registry lookup under `ActionKind.CANCEL_OPERATION`, and no-handler passthrough (JS parity).

py/packages/genkit/src/genkit/_core/_model.py
cabljac: Fixed in 286e2250b (details in top-level reply). Your A/B now returns the $ref, and that's what lands in inputSchema. Warnings gone (tested under `simplefilter('error')`), and `config=SomeOtherConfig()` fails validation at the boundary.

py/packages/genkit/src/genkit/_core/_model.py
cabljac: Good catch, added to the breaking changes list. Bare `ModelRequest.config` is `Unknown | None` under pyright, so access silently stops being checked. Parametrized access stays fully typed.

cabljac:
Both config issues fixed in 286e2250b as one change, as you suggested.

1. Dropped `mode='plain'`. Config is now a left_to_right union (`dict | SerializeAsAny[ConfigT] | None`) with a before-validator doing only the dict coercion, so the core schema survives: $ref back in inputSchema, no serializer warnings, mismatched instances rejected at the boundary. A TYPE_CHECKING split keeps `config: ConfigT | None` for type checkers, which also clears the pyright errors on this PR's tests.

2. With the stricter union, cross-schema instance configs now fail first-pass, so the retry is the live rescue path its comment describes. The output drop is fixed by unnesting the wire `output` key on validation; round-trip test covers your repro.

Namespace hint, cancel coverage, and the breaking-changes note are in follow-up commits. Can you take another look?

hilariie (approved):
LGTM

PR #5855, opened 2026-07-30
feat(py/plugins/google-genai): raw Interactions HTTP client, converters, and options

cabljac:
Second slice of the #5806 split (stacked on #5854): minimal async HTTP client for the Google AI Interactions API, request/response converters between Genkit and Interactions wire types, ClientOptions resolution, shared error-mapping utilities, and the google-genai >=2.14.0,<3 bump.

### Breaking changes

- `google-genai` dependency floor moves from `>=1.63.0` to `>=2.14.0,<3` (major version bump) for the Interactions SDK types. Downstream projects pinned below 2.x will fail to resolve.

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/client.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/client.py
@@ -1,0 +1,257 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Raw HTTP helpers for the Google AI Interactions API."""
+
+from __future__ import annotations
+
+import json
+from collections.abc import Mapping
+from datetime import datetime, timezone
+from email.utils import parsedate_to_datetime
+from typing import cast
+
+import httpx
+from genkit_google_genai._interactions.options import ClientOptions
+from google.genai.interactions import Interaction
+
+from genkit import GenkitError
+from genkit._core._error import ErrorResponseMetadata, StatusName
+from genkit._core._logger import get_logger
+from genkit.plugin_api import GENKIT_CLIENT_HEADER, get_cached_client
+
+logger = get_logger(__name__)
+
+DEFAULT_API_VERSION = 'v1beta'
+DEFAULT_BASE_URL = 'https://generativelanguage.googleapis.com'
+API_REVISION = '2026-05-20'
+# Interactions creates (esp. Antigravity/Lyria) can run far longer than the
+# shared client's 60s default, so this transport opts out of a wall clock.
+CACHE_KEY = 'googleai-interactions'
+NO_TIMEOUT = httpx.Timeout(None)
+
+
+def google_ai_url(
+    resource_path: str,
+    *,
+    client_options: ClientOptions | None = None,
+) -> str:
+    """Build a Google AI REST URL for the given resource path."""
+    opts = client_options or ClientOptions()
+    api_version = opts.api_version or DEFAULT_API_VERSION
+    base_url = (opts.base_url or DEFAULT_BASE_URL).rstrip('/')
+    return f'{base_url}/{api_version}/{resource_path}'
+
+
+def headers(*, api_key: str, client_options: ClientOptions | None) -> dict[str, str]:
+    """Build request headers; api key and Genkit client attribution win over custom."""
+    custom = dict((client_options or ClientOptions()).custom_headers or {})
+    # Not allowed in user settings — same as the JS Google AI client.
+    custom.pop('x-goog-api-key', None)
+    custom.pop('x-goog-api-client', None)
+    return {
+        **custom,
+        'Content-Type': 'application/json',
+        'x-goog-api-client': GENKIT_CLIENT_HEADER,
+        'Api-Revision': API_REVISION,
+        'x-goog-api-key': api_key,
+    }
+
+
+def timeout_seconds(client_options: ClientOptions | None) -> float | None:
+    """Convert ClientOptions.timeout (milliseconds) to httpx seconds."""
+    opts = client_options or ClientOptions()
+    if opts.timeout is not None and opts.timeout >= 0:
+        return opts.timeout / 1000.0
+    return None
+
+
+def parse_retry_after_ms(value: str | None) -> float | None:
+    """Read a Retry-After header, which may be a delay in seconds or an HTTP date."""
+    if not value or not value.strip():
+        return None
+    try:
+        seconds = float(value)
+    except ValueError:
+        seconds = -1.0
+    if seconds >= 0:
+        return seconds * 1000
+    try:
+        retry_at = parsedate_to_datetime(value)
+    except (TypeError, ValueError, OverflowError):
+        return None
+    if retry_at.tzinfo is None:
+        retry_at = retry_at.replace(tzinfo=timezone.utc)
+    return max(0.0, (retry_at - datetime.now(timezone.utc)).total_seconds() * 1000)
+
+
+def status_for_http_code(status_code: int) -> StatusName:
+    """Map an HTTP status onto the Genkit error status callers switch on."""
+    match status_code:
+        case 429:
+            return 'RESOURCE_EXHAUSTED'
+        case 400:
+            return 'INVALID_ARGUMENT'
+        case 401:
+            return 'UNAUTHENTICATED'
+        case 403:
+            return 'PERMISSION_DENIED'
+        case 404:
+            return 'NOT_FOUND'
+        case 499:
+            return 'CANCELLED'
+        case 500:
+            return 'INTERNAL'
+        case 503:
+            return 'UNAVAILABLE'
+        case _:
+            return 'UNKNOWN'
+
+
+async def create_interaction(
+    api_key: str,
+    body: dict[str, object],
+    client_options: ClientOptions | None = None,
+) -> Interaction:
+    """POST /interactions and return the parsed Interaction."""
+    url = google_ai_url('interactions', client_options=client_options)
+    return await request(
+        method='POST',
+        url=url,
+        api_key=api_key,
+        client_options=client_options,
+        json_body=body,
+    )
+
+
+async def get_interaction(
+    api_key: str,
+    interaction_id: str,
+    client_options: ClientOptions | None = None,
+) -> Interaction:
+    """GET /interactions/{id} and return the parsed Interaction."""
+    url = google_ai_url(f'interactions/{interaction_id}', client_options=client_options)
+    return await request(
+        method='GET',
+        url=url,
+        api_key=api_key,
+        client_options=client_options,
+    )
+
+
+async def cancel_interaction(
+    api_key: str,
+    interaction_id: str,
+    client_options: ClientOptions | None = None,
+) -> Interaction:
+    """POST /interactions/{id}/cancel and return a cancelled Interaction."""
+    url = google_ai_url(f'interactions/{interaction_id}/cancel', client_options=client_options)
+    try:
+        await request(
+            method='POST',
+            url=url,
+            api_key=api_key,
+            client_options=client_options,
+        )
+        # Successful cancel often surfaces as CANCELLED; normalize both paths.
+        raise GenkitError(status='CANCELLED', message='successfully cancelled')
+    except GenkitError as error:
+        if error.status == 'CANCELLED':
+            return Interaction.model_validate({'id': interaction_id, 'status': 'cancelled'})
+        raise
+
+
+async def request(
+    *,
+    method: str,
+    url: str,
+    api_key: str,
+    client_options: ClientOptions | None,
+    json_body: dict[str, object] | None = None,
+) -> Interaction:
+    """Issue one Interactions HTTP call and parse the Interaction body."""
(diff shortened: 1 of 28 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -0,0 +1,714 @@
...
+        raise ValueError('Media part missing contentType')
+    block_type = next((name for prefix, name in MEDIA_CONTENT_TYPES if content_type.startswith(prefix)), None)
+    if block_type is None:
+        raise ValueError(f'Unsupported media type: {content_type}')
+
+    block: dict[str, Any] = {'type': block_type, 'mime_type': content_type}
+    # Inline data URLs travel as base64; anything else is a reference the API fetches.
+    url = part.media.url
+    if url.startswith('data:'):
+        block['data'] = url[url.index(',') + 1 :]
```
cabljac: I think `url[url.index(',') + 1:]` can throw a bare `ValueError: substring not found` if a data URL arrives without a comma. The other failure modes in this function all raise descriptive errors, so a malformed media part could plausibly be a confusing one to debug. Something like:

```python
if ',' not in url:
    raise ValueError('Malformed data URL for media part: missing payload separator')
block['data'] = url.split(',', 1)[1]
```

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -233,7 +233,7 @@ async def _generate_multimodal(self, request: EmbedRequest) -> EmbedResponse:
         if api_client is None or not hasattr(api_client, 'async_request'):
             raise RuntimeError(
                 'Multimodal embedding relies on google-genai client internals that are '
-                'unavailable in the installed google-genai version; install google-genai>=1.63.0.'
+                'unavailable in the installed google-genai version; install google-genai>=2.3.0.'
```
cabljac: This message suggests `google-genai>=2.3.0` but the floor this PR sets in pyproject is `>=2.14.0,<3`, so the suggested install can't satisfy the guard. The test in googlegenai_embedder_test.py asserts the same string. Both should move to 2.14.0, unless 2.3.0 is the real feature floor for the private transport.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -0,0 +1,714 @@
...
+            root = part.root
+            if isinstance(root, ToolResponsePart) and root.tool_response and not root.tool_response.ref:
+                matched_id = generated_ids.pop(0) if generated_ids else None
+                if matched_id:
+                    root.tool_response.ref = matched_id
+                else:
+                    root.tool_response.ref = f'genkit-orphan-id-{next_id_counter}'
+                    next_id_counter += 1
+
+    return new_messages
```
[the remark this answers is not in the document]
cabljac: Confirmed reproducible, but this is a line-for-line port of ensureToolIds in the JS plugin (interaction-converters.ts), which has the same ordering assumption and documents it. Genkit's own tool loop always copies the request ref onto the response, so a mixed-ref history has to be hand-built. Keeping JS parity in this port; a fix would want to land cross-language rather than fork the behavior here.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/client.py
```diff
@@ -0,0 +1,257 @@
...
+            url=url,
+            api_key=api_key,
+            client_options=client_options,
+        )
+        # Successful cancel often surfaces as CANCELLED; normalize both paths.
+        raise GenkitError(status='CANCELLED', message='successfully cancelled')
+    except GenkitError as error:
+        if error.status == 'CANCELLED':
+            return Interaction.model_validate({'id': interaction_id, 'status': 'cancelled'})
+        raise
```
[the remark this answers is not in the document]
cabljac: Intentional, matching the JS client byte-for-byte: the API surfaces a successful cancellation as a CANCELLED error, and a 200 OK means a no-op cancel (race between finish and cancel), so raising on the 200 path unifies the exit logic for both. Returning the 200 body with status stamped to 'cancelled' would mislabel a possibly-finished interaction.

py/packages/genkit-google-genai/src/genkit_google_genai/models/interactions_utils.py
```diff
@@ -0,0 +1,215 @@
...
+            'ABORTED',
+            'OUT_OF_RANGE',
+            'UNIMPLEMENTED',
+            'INTERNAL',
+            'UNAVAILABLE',
+            'DATA_LOSS',
+        }
+        if candidate in valid:
+            return cast(StatusName, candidate)
+    return status_for_http_code(int(error.code or 0))
```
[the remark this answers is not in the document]
cabljac: Fixed in 46a1fd783: the conversion now falls back to 0 on TypeError/ValueError, same pattern this file already uses for retry-after codes. Test added with a non-numeric code.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
cabljac: Fixed in 46a1fd783: missing separator now raises a descriptive ValueError, with a test.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
cabljac: Fixed in 46a1fd783: message and test both say 2.14.0 now. 2.3.0 predates the pyproject floor so there's no install that could hit it.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/client.py
```diff
@@ -0,0 +1,257 @@
...
+    client_options: ClientOptions | None = None,
+) -> str:
+    """Build a Google AI REST URL for the given resource path."""
+    opts = client_options or ClientOptions()
+    api_version = opts.api_version or DEFAULT_API_VERSION
+    base_url = (opts.base_url or DEFAULT_BASE_URL).rstrip('/')
+    return f'{base_url}/{api_version}/{resource_path}'
+
+
+def headers(*, api_key: str, client_options: ClientOptions | None) -> dict[str, str]:
```
cabljac, line 59: The reserved-header strip is case-sensitive: custom.pop('x-goog-api-key') misses X-Goog-Api-Key, and unlike the JS Headers object a dict keeps both casings, so the wire request carries two api-key headers (verified via MockTransport). The existing strip test passes even with the pop lines deleted, because the explicit keys are spread after **custom and overwrite same-case duplicates anyway. Building through httpx.Headers (or lowercasing before the pops) fixes both, and the test should assert the reserved key is actually absent.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -0,0 +1,716 @@
...
+
+def with_signature(step: StepParam, metadata: dict[str, Any]) -> StepParam:
+    """Carry a thought signature from part metadata back onto its step."""
+    signature = metadata.get(THOUGHT_SIGNATURE)
+    if signature is not None:
+        cast(dict[str, Any], step)['signature'] = signature
+    return step
+
+
+def to_function_call_step(request: ToolRequest) -> FunctionCallStepParam:
```
cabljac, line 262: Non-dict tool input is silently replaced with {}: ToolRequest(input='Paris') goes to the wire as arguments: {}, so on the next turn the model sees its own call with empty args. Scalar input schemas are legal in Genkit, and the JS equivalent throws here (zod parse) rather than dropping data. We should either box it like to_function_result_step does, or raise.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/client.py
```diff
@@ -0,0 +1,257 @@
...
+
+logger = get_logger(__name__)
+
+DEFAULT_API_VERSION = 'v1beta'
+DEFAULT_BASE_URL = 'https://generativelanguage.googleapis.com'
+API_REVISION = '2026-05-20'
+# Interactions creates (esp. Antigravity/Lyria) can run far longer than the
+# shared client's 60s default, so this transport opts out of a wall clock.
+CACHE_KEY = 'googleai-interactions'
+NO_TIMEOUT = httpx.Timeout(None)
```
cabljac, line 44: httpx.Timeout(None) also disables the connect timeout, so with no ClientOptions.timeout set (the default), an unreachable host parks the coroutine forever with no DEADLINE_EXCEEDED. The comment only argues for dropping the wall clock; httpx.Timeout(None, connect=10.0) keeps that intent.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/options.py
```diff
@@ -0,0 +1,69 @@
...
+        if not overrides:
+            return self
+        if isinstance(overrides, dict):
+            overrides = ClientOptions.model_validate(overrides)
+        # Field names (not aliases) — model_copy(update=...) keys off Python attrs.
+        update = overrides.model_dump(exclude_none=True)
+        return self.model_copy(update=update) if update else self
+
+    @classmethod
+    def from_metadata(cls, metadata: dict[str, Any] | None) -> Self:
```
cabljac, line 58: Two trust issues on the check/cancel path: from_metadata accepts whatever dict comes back on the operation, including a baseUrl, and google_ai_url never quotes the interaction id, so 'interactions/../../v1/models' resolves to a different endpoint once httpx normalizes it (verified). In a server flow the operation comes from the client, so a caller can point the plugin's authenticated request at an arbitrary host or path. Minimum fix: quote(interaction_id, safe='') and ignore or validate baseUrl from metadata.

cabljac:
Did a deeper adversarial pass over this slice and found a handful of real issues. The blocking ones are inline (api key persistence, header casing, scalar tool args). Lower priority, happy to fold into this PR or a follow-up: unrecognized interaction statuses leave done unset so polling never terminates, malformed custom parts escape as bare AttributeError, Retry-After: inf parses to infinity, annotations land in metadata as SDK objects rather than plain data, and custom/raw on ModelResponse share one dict.

py/packages/genkit-google-genai/src/genkit_google_genai/models/interactions_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        try:
+            code = int(raw)
+        except (TypeError, ValueError):
+            continue
+        if code > 0:
+            return code
+    return None
+
+
+def client_options_for_operation(
```
cabljac, line 140: Flagging a credential-handling concern with persisting resolved auth here, since the later slices pick this up. Raised with the core team separately, happy to take the fix once there's a direction.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -0,0 +1,716 @@
...
+def to_content_step(role: InteractionRole, content: list[ContentParam]) -> StepParam:
+    """Wrap a turn's inline content in the step for whoever produced it."""
+    if role == 'model':
+        model_step: ModelOutputStepParam = {'type': 'model_output', 'content': content}
+        return model_step
+    user_step: UserInputStepParam = {'type': 'user_input', 'content': content}
+    return user_step
+
+
+def to_interaction_steps(messages: list[Message]) -> list[StepParam]:
```
hilariie, line 364: Blocking: this history conversion is lossy. Mixed parts are reordered because inline content is buffered until the end, preserved `unknownStep` values are dropped on replay, and signed multimodal thoughts are reconstructed differently. Stateless Interactions history must replay model-generated steps exactly as received.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -0,0 +1,716 @@
...
+    if isinstance(step, CodeExecutionCallStep):
+        return [from_code_execution_call(step)]
+    if isinstance(step, CodeExecutionResultStep):
+        return [from_code_execution_result(step)]
+    if isinstance(step, ThoughtStep):
+        return [from_thought_step(step)]
+    if isinstance(step, FunctionCallStep):
+        return [from_function_call_step(step)]
+    if isinstance(step, FunctionResultStep):
+        return [from_function_result_step(step)]
```
hilariie, line 559: `function_result` is client input echoed by Interactions, similar to `user_input`. Keeping it here places the caller's `ToolResponsePart` inside a `role="model"` response and can duplicate that result when the history is sent again.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -0,0 +1,716 @@
...
+        'name': request.name,
+        'arguments': request.input if isinstance(request.input, dict) else {},
+        'id': request.ref or '',
+    }
+
+
+def to_function_result_step(response: ToolResponse) -> FunctionResultStepParam:
+    """Compile a Genkit tool response into a function_result step."""
+    output = response.output
+    if isinstance(output, (str, dict, list)):
```
hilariie, line 275: This treats every list as multimodal function-result content, but the wire only accepts lists of text/image content blocks. A normal tool result such as `[1, 2]` is therefore emitted as an invalid `function_result`. Could we validate content-block lists and box ordinary JSON arrays in an object, with tests for both cases?

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -0,0 +1,716 @@
...
+
+
+def from_code_execution_result(step: CodeExecutionResultStep) -> Part:
+    """Convert a code_execution_result step to a Genkit custom part."""
+    result = step.result
+    return server_tool_part(
+        CODE_EXECUTION_RESULT,
+        {
+            'output': result if isinstance(result, str) else json.dumps(result),
+            'outcome': 'OUTCOME_OK',
```
hilariie, line 451: Failed tool results become successful as result failure state is lost during this conversion, even when `is_error=True`

hilariie (changes requested):

PR #5856, opened 2026-07-30
feat(py/plugins/google-genai): Deep Research, Antigravity, and Lyria interaction models

cabljac:
Third slice of the #5806 split: Deep Research background model with cancel support, Antigravity and Lyria foreground models on the Interactions transport, and a registry of known models with capability metadata. Rewrites lyria.py from the legacy Vertex-only LyriaModel. Model-level tests included here; plugin wiring and its integration tests land in the next slice.

### Breaking changes

- `LyriaModel` is removed from `genkit_google_genai.models.lyria`; Lyria is now served by the Interactions-backed implementation (`create_lyria_action`). The package-level exports `LyriaConfig` and `LyriaVersion` are unchanged.

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/antigravity.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/antigravity.py
@@ -14,11 +14,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Google AI Interactions Lyria audio model action."""
+"""Google AI Interactions Antigravity model action."""
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Literal
 
 from pydantic import BaseModel, ConfigDict
 from pydantic.alias_generators import to_camel
@@ -33,8 +33,8 @@ from genkit_google_genai._interactions.converters import (
     split_system_instruction,
     to_interaction_steps,
 )
-from genkit_google_genai._interactions.options import ClientOptions, ResponseModality
-from genkit_google_genai.models.interactions_registry import lyria_model_info
+from genkit_google_genai._interactions.options import ClientOptions
+from genkit_google_genai.models.interactions_registry import antigravity_model_info
 from genkit_google_genai.models.interactions_utils import (
     calculate_api_key,
     client_overrides_from_config,
@@ -44,11 +44,18 @@ from genkit_google_genai.models.interactions_utils import (
     require_interaction_steps,
 )
 
-CREATE_OPTION_KEYS = ('response_modalities',)
+DEFAULT_ENVIRONMENT: dict[str, str] = {'type': 'remote'}
+
+CREATE_OPTION_KEYS = (
+    'previous_interaction_id',
+    'store',
+    'environment',
+    'response_modalities',
+)
 
 
-class LyriaConfig(BaseModel):
-    """Google AI Interactions Lyria model configuration."""
+class AntigravityConfig(BaseModel):
+    """Antigravity model configuration."""
 
     # Per-request options arrive camelCased (apiKey).
     model_config = ConfigDict(extra='allow', populate_by_name=True, alias_generator=to_camel)
@@ -58,21 +65,24 @@ class LyriaConfig(BaseModel):
     # Milliseconds — applied to the HTTP call, not the create body.
     timeout: float | None = None
     custom_headers: dict[str, str] | None = None
-    response_modalities: list[ResponseModality] | None = None
+    previous_interaction_id: str | None = None
+    store: bool | None = None
+    environment: str | dict[str, Any] | None = None
+    response_modalities: list[Literal['text', 'image']] | None = None
 
 
-def create_lyria_action(
+def create_antigravity_action(
     name: str,
     *,
     plugin_api_key: str | None,
     client_options: ClientOptions,
-) -> Action[ModelRequest[LyriaConfig], ModelResponse, Never]:
-    """Build a foreground model action for Interactions Lyria."""
+) -> Action[ModelRequest[AntigravityConfig], ModelResponse, Never]:
+    """Build a foreground model action for Antigravity."""
     version = extract_version(name)
-    info = lyria_model_info(version)
+    info = antigravity_model_info(version)
 
-    async def run(request: ModelRequest[LyriaConfig], _: ActionRunContext) -> ModelResponse:
-        config = request.config or LyriaConfig()
+    async def run(request: ModelRequest[AntigravityConfig], _: ActionRunContext) -> ModelResponse:
+        config = request.config or AntigravityConfig()
         api_key = calculate_api_key(plugin_api_key, config.api_key)
         merged_options = client_options.merge(
             client_overrides_from_config(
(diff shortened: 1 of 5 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/models/deep_research.py
```diff
@@ -0,0 +1,281 @@
...
+    if config.google_search:
+        tools.append({'type': 'google_search'})
+    if config.url_context:
+        tools.append({'type': 'url_context'})
+    if config.code_execution:
+        tools.append({'type': 'code_execution'})
+    if config.file_search is not None:
+        tools.append({'type': 'file_search', **config.file_search.model_dump(exclude_none=True)})
+    for mcp_server in config.mcp_servers or []:
+        tools.append({'type': 'mcp_server', **mcp_server.model_dump(exclude_none=True)})
```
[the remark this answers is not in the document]
cabljac: The suggestion would break these requests rather than fix them. This body goes over a raw HTTP client verbatim, and the whole create payload is snake_case (agent_config, previous_interaction_id, user_input, mime_type in the content blocks), so camelCasing only these two nested blocks would produce a mixed-casing body. The SDK's own FileSearchDict/McpServerDict types also take snake_case keys. The alias_generator on these models handles the input direction (camelCased request config), not serialization.

py/packages/genkit-google-genai/src/genkit_google_genai/models/deep_research.py
```diff
@@ -0,0 +1,281 @@
...
+        info=info,
+        config_schema=DeepResearchConfig,
+    )
+
+
+def build_tools(request: ModelRequest[DeepResearchConfig], config: DeepResearchConfig) -> list[dict[str, Any]]:
+    """Build Interactions API tool configurations for Deep Research."""
+    tools: list[dict[str, Any]] = []
+    if request.tools:
+        tools.extend(dict(to_interaction_tool(tool_def)) for tool_def in request.tools)
```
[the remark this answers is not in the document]
cabljac: Correct that the copy is a no-op (to_interaction_tool builds a fresh dict per call). No functional difference; will drop it in the next tidy commit on this slice.

py/packages/genkit-google-genai/src/genkit_google_genai/models/antigravity.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+        require_interaction_steps(steps)
+        create_kwargs: dict[str, Any] = {
+            'agent': version,
+            'input': steps,
+            **create_options,
+            **passthrough,
+        }
+        # Default missing environment to remote; the API rejects unsupported values.
+        create_kwargs.setdefault('environment', DEFAULT_ENVIRONMENT)
```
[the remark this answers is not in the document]
cabljac: Nothing mutates the environment dict downstream today (the body goes straight to JSON serialization), but the shared-reference risk is real and the fix is cheap. Will take this.

py/packages/genkit-google-genai/src/genkit_google_genai/models/deep_research.py
hilariie, line 152: Could we add tests for the `file_search` and `mcp_servers` branches here, plus `response_format_from_request`? None of the three are exercised in interactions_models_test.py.

These are the only places where the wire keys depend on `model_dump` being called without `by_alias`. Both config models set `alias_generator=to_camel`, so the correct `file_search_store_names` and `allowed_tools` only happen by omission. If someone adds `by_alias=True` later, it silently sends camelCase and no test fails.

py/packages/genkit-google-genai/src/genkit_google_genai/models/deep_research.py
```diff
@@ -0,0 +1,281 @@
...
+    # Per-request options arrive camelCased (apiKey). Without accepting those
+    # names, the override silently disappears before check/cancel can reuse it.
+    model_config = ConfigDict(extra='allow', populate_by_name=True, alias_generator=to_camel)
+    api_key: str | None = None
+    base_url: str | None = None
+    api_version: str | None = None
+    # Milliseconds — applied to the HTTP call, not the create body.
+    timeout: float | None = None
+    custom_headers: dict[str, str] | None = None
+    thinking_summaries: Literal['auto', 'none'] | None = None
```
hilariie, line 109: The JS config accepts `thinkingSummaries: 'AUTO' | 'NONE'` and lowercases before building the request body. This accepts only `'auto' | 'none'`, and the same split applies to `visualization` and to `response_modalities` across all three models.

This means that a config copied from the JS samples can raise a ValidationError in Python.

py/packages/genkit-google-genai/src/genkit_google_genai/models/lyria.py
```diff
@@ -14,188 +14,103 @@
...
+        modalities = create_options.get('response_modalities') or ['audio', 'text']
+        system_instruction, turns = split_system_instruction(request.messages or [])
+        steps = to_interaction_steps(ensure_tool_ids(turns))
+        create_kwargs: dict[str, Any] = {
+            'model': version,
+            'input': steps,
+            'response_modalities': modalities,
+            **passthrough,
+        }
+        if system_instruction:
```
hilariie, line 97: This path has no JS counterpart. JS Lyria passes messages straight to `toInteractionSteps` with no system handling, so sending `system_instruction` and skipping the empty input guard is new behavior here. If this divergence from JS is intentional, there should be a test for the system-only case to pin the intent, since it is the one path that can reach the API with empty input.

py/packages/genkit-google-genai/src/genkit_google_genai/models/deep_research.py
```diff
@@ -0,0 +1,281 @@
...
+    info = KNOWN_DEEP_RESEARCH_MODELS.get(clean) or DEEP_RESEARCH_INFO
+    # Prefer a version-specific label over the shared catalog default.
+    if info.label is None or info.label == DEEP_RESEARCH_INFO.label:
+        info = ModelInfo(label=f'Google AI - {clean}', supports=info.supports)
+    return model_ref(
+        name=clean,
+        namespace='googleai',
+        info=info,
+        config_schema=DeepResearchConfig,
+    )
```
hilariie, line 134: This reproduces the labeling that `deep_research_model_info` already does. Every entry in `KNOWN_DEEP_RESEARCH_MODELS` has either `label=None` or `label == DEEP_RESEARCH_INFO.label`, so the condition on line 127 is always true and the result is the same `ModelInfo` the registry helper returns.

Could this just call `deep_research_model_info(version)`? It keeps the labeling rule in the one file the registry docstring points people at.

hilariie (commented):
The breaking changes section says `LyriaConfig` and `LyriaVersion` are unchanged, but both changed substantially. `LyriaVersion` swaps `LYRIA_002` for `LYRIA_3_CLIP` and `LYRIA_3_PRO`, and `LyriaConfig` replaces all four of its fields (`negative_prompt`, `seed`, `sample_count`, `location`) with the Interactions client options. Only the exported names are the same.

Worth noting that neither was reachable through any plugin before this stack, so nothing functional regresses, but the release notes get built from this section.

PR #5857, opened 2026-07-30
refactor(py/plugins/google-genai): move Veo to background model actions

cabljac:
Fourth slice of the #5806 split: rewrites veo.py around create_veo_background_action on the shared background-model plumbing, wires Veo through BACKGROUND_MODEL/CHECK_OPERATION resolution in both GoogleAI and VertexAI (dropping the plain MODEL registration), and updates the google-genai-media sample to poll generate_operation directly. Touches only the Veo hunks of google.py; the Interactions wiring hunks follow in the next slice.

### Breaking changes

- Veo is now background-only on both GoogleAI and VertexAI: `ai.generate(model='vertexai/veo-...')` no longer resolves; use `ai.generate_operation(...)` + `ai.check_operation(...)` (see the updated google-genai-media sample).
- `VeoModel` and `VeoConfigSchema` are removed from `genkit_google_genai.models.veo`; use `create_veo_background_action` and `VeoConfig`.

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/google.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/google.py
@@ -97,8 +97,8 @@ from genkit_google_genai.models.imagen import (
     vertexai_image_model_info,
 )
 from genkit_google_genai.models.veo import (
-    VeoConfigSchema,
-    VeoModel,
+    VeoConfig,
+    create_veo_background_action,
     is_veo_model,
     veo_model_info,
 )
@@ -284,7 +284,7 @@ def _create_embedder_action(
         ),
     )
 
-    async def _run(request: Any) -> Any:  # noqa: ANN401
+    async def run(request: Any) -> Any:  # noqa: ANN401
         embedder = Embedder(
             version=clean_name,
             client=client_getter(),
@@ -295,7 +295,7 @@ def _create_embedder_action(
     action = Action(
         kind=ActionKind.EMBEDDER,
         name=full_name,
-        fn=_run,
+        fn=run,
         metadata=action_metadata.metadata,
     )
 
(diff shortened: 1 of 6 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/google.py
```diff
@@ -881,7 +853,25 @@ async def resolve(self, action_type: ActionKind, name: str) -> Action | None:
             Action object if found, None otherwise.
         """
         if action_type == ActionKind.MODEL:
+            prefix = VERTEXAI_PLUGIN_NAME + '/'
+            clean_name = name.replace(prefix, '') if name.startswith(prefix) else name
+            if is_veo_model(clean_name):
+                return None
             return self._resolve_model(name)
+        elif action_type == ActionKind.BACKGROUND_MODEL:
```
hilariie, line 861: None of VertexAI's three new Veo branches has coverage. I can delete the `MODEL` guard, the `BACKGROUND_MODEL` branch, or the `CHECK_OPERATION` branch and the suite stays green. The GoogleAI equivalents are all pinned: removing the `MODEL` guard at line 499 fails both new end-to-end tests, and the other two have `test_googleai_resolve_background_model` / `test_googleai_resolve_check_operation`.

Worth noting these branches only fire for a Veo model discovery didn't return, since `init()` already registers start and check for known ones and `resolve_action` hits the registry before `plugin.resolve`. That's the case they exist for, and it's untested on Vertex. Could we mirror the GoogleAI tests here?

py/packages/genkit-google-genai/src/genkit_google_genai/models/veo.py
```diff
@@ -106,280 +75,118 @@ class VeoConfigSchema(BaseModel):
...
-        op.output = {
-            'finishReason': 'stop',
-            'message': {
-                'role': 'model',
-                'content': content,
-            },
-        }
-
+    response = operation.response or operation.result
+    if response is not None and extract_video_uris(response):
```
hilariie, line 152: When Veo filters a generation, the operation finishes with `done: true`, no error, and empty `generated_videos`, with the explanation in `rai_media_filtered_count` / `rai_media_filtered_reasons`. As written, this returns `done=True, output=None, error=None`, so the caller gets nothing to debug from.

hilariie (changes requested):

PR #5859, opened 2026-07-30
feat(py/plugins/google-genai): finish_message for image models and effective config echo

cabljac:
Parallel slice of the #5806 split (based on #5854, independent of the Interactions chain): normalizes finish reasons (NO_IMAGE, IMAGE_OTHER), synthesizes finish_message when image models return no media so empty results are not silent successes, propagates finish_message onto candidates, echoes the normalized effective config on response.request, and drops the retired gemini-2.5-flash-image-preview registration.

### Breaking changes

- `gemini-2.5-flash-image-preview` is no longer registered as a model (the `GEMINI_2_5_FLASH_IMAGE_PREVIEW` enum members remain, but the model name stops resolving). Use `gemini-2.5-flash-image`.

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
@@ -157,6 +157,7 @@ from pydantic import BaseModel, ConfigDict, Field, WithJsonSchema
 from genkit import (
     Constrained,
     GenkitError,
+    MediaPart,
     Message,
     ModelConfig,
     ModelInfo,
@@ -183,9 +184,19 @@ def _to_dict(obj: JsonAny) -> JsonAny:  # noqa: ANN401
     return obj.model_dump() if isinstance(obj, BaseModel) else obj
 
 
+def _finish_reason_name(fr: Any) -> str | None:  # noqa: ANN401
+    """Normalize a google-genai finish reason to an uppercase name string."""
+    if fr is None:
+        return None
+    name = getattr(fr, 'name', fr)
+    if isinstance(name, str):
+        return name.upper()
+    return str(name).upper() if name is not None else None
+
+
 def _to_finish_reason(fr: Any) -> FinishReason:  # noqa: ANN401
     """Map a google-genai finish reason onto Genkit's FinishReason."""
-    fr_name = getattr(fr, 'name', fr) if fr is not None else None
+    fr_name = _finish_reason_name(fr)
     if fr_name == 'STOP':
         return FinishReason.STOP
     if fr_name == 'MAX_TOKENS':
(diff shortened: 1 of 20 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -201,11 +212,42 @@ def _to_finish_reason(fr: Any) -> FinishReason:  # noqa: ANN401
...
+    if _content_has_media(content):
+        return None
+    if fr_name == 'IMAGE_SAFETY':
+        return 'Image blocked by safety filters.'
+    if fr_name == 'IMAGE_OTHER':
+        return 'Image generation stopped (IMAGE_OTHER). Try revising the prompt.'
+    if fr_name == 'NO_IMAGE':
+        return 'No image was returned. Try a clearer image prompt (for example, "draw a red circle on white").'
+    # Ambiguous prompts sometimes come back as STOP/OTHER/UNKNOWN with empty parts.
+    return 'No image was returned. Try a clearer image prompt (for example, "draw a red circle on white").'
```
hilariie, line 248: I think this guard is broader than the comment on the fallback intends. The comment says these come back with empty parts, but the check is `not _content_has_media(content)`, which is also true of a non-empty text-only reply. `GEMINI_IMAGE_SUPPORTS` marks these models `multiturn=True` with `TEXT, IMAGE` modalities, so a text-only turn (the model asking a clarifying question, for example) is a normal outcome, and it now comes back carrying `finish_message="No image was returned. Try a clearer image prompt..."`.

The new `test_generate_echoes_normalized_effective_config` actually constructs this exact case (image model, `finish_reason='STOP'`, a single text part) but only asserts on `response.request.config`, so it passes without noticing. The empty-`candidates` path (prompt-level block) reaches the same fallback.

Could we gate the synthesized message on the reasons that actually mean no image was produced, and leave `STOP` with text content alone?

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -201,11 +212,42 @@ def _to_finish_reason(fr: Any) -> FinishReason:  # noqa: ANN401
...
         return FinishReason.BLOCKED
-    if fr_name in ('OTHER', 'MALFORMED_FUNCTION_CALL', 'MISSING_THOUGHT_SIGNATURE'):
+    # NO_IMAGE: request accepted but no image bytes. IMAGE_OTHER: generation
+    # stopped for a non-safety reason (often prompt/content policy adjacent).
+    if fr_name in (
+        'OTHER',
+        'MALFORMED_FUNCTION_CALL',
+        'MISSING_THOUGHT_SIGNATURE',
+        'NO_IMAGE',
+        'IMAGE_OTHER',
```
hilariie, line 222: `IMAGE_PROHIBITED_CONTENT` and `IMAGE_RECITATION` are left unmapped, so they fall through to `FinishReason.UNKNOWN` and then pick up the generic "try a clearer image prompt" text, even though the non-image `PROHIBITED_CONTENT` and `RECITATION` both map to `BLOCKED` in the tuple above.

Both members exist as far back as the declared floor (`google-genai>=1.63.0`), so this is not a new-SDK concern. The practical effect is that a genuine content-policy block on an image model never surfaces as `blocked`, and the caller gets prompt-quality advice instead of a policy explanation.

Could we add both to the blocked tuple?

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1624,9 +1663,46 @@ async def generate(self, request: ModelRequest, ctx: ActionRunContext) -> ModelR
...
+            normalized = self._normalize_config_to_dict(request.config)
+            if normalized:
+                effective.update(normalized)
+
+        if request_cfg is not None and request_cfg.response_modalities:
+            effective['response_modalities'] = list(request_cfg.response_modalities)
+        elif is_tts_model(model_name):
+            effective['response_modalities'] = ['AUDIO']
+        elif is_image_model(model_name):
+            effective['response_modalities'] = ['TEXT', 'IMAGE']
```
hilariie, line 1697: These two fallbacks look unreachable from the only caller. `generate()` already sets `request_cfg.response_modalities` unconditionally for TTS and image models before dispatching, so the first branch always wins by the time `effective_config` runs. `test_generate_echoes_injected_modalities_when_config_omitted` passes through that first branch too.

PR #5862, opened 2026-07-30
feat(go): add options-struct `New*ActionOf` constructors and delete registry-taking `Define*` helpers

apascal07:
Reworks the `core` action constructor surface as a non-breaking addition for the next minor: options-struct constructors under new names (`New*ActionOf`), the old constructors deprecated and delegating to them, and the registry-taking `Define*` helpers deleted outright (no reachable external callers). Base PR of a train: the typed-config constructors (`ai.New*Action`, #5849) build on it, then plugin migrations.

## Options-struct action constructors

**Before:** metadata and the input schema are positional; output and stream schemas have no override at all.

```go
core.NewAction[In, Out](name, atype, metadata map[string]any, inputSchema map[string]any, fn)
core.NewStreamingAction[In, Out, Stream](name, atype, metadata, inputSchema, fn)
core.NewBackgroundAction[In, Out](name, atype, metadata, startFn, checkFn, cancelFn)
core.NewBidiAction[In, Out, Stream, Init](name, atype, opts, fn)
```

**After:** the action type leads (matching `ResolveActionFor`) and one options struct covers all schema slots, each inferred from its type parameter when nil.

```go
type core.ActionOptions struct {
    Description  string
    Metadata     map[string]any
    InputSchema  map[string]any
    OutputSchema map[string]any
    StreamSchema map[string]any
}

type core.BackgroundActionOptions[In, Out] struct {
    Description  string
    Metadata     map[string]any
    InputSchema  map[string]any
    OutputSchema map[string]any
    Check        CheckOpFunc[Out]  // required
    Cancel       CancelOpFunc[Out] // optional
}

core.NewActionOf[In, Out](atype, name, opts, fn)
core.NewStreamingActionOf[In, Out, Stream](atype, name, opts, fn)
core.NewBackgroundActionOf[In, Out](atype, name, opts, startFn)
core.NewBidiActionOf[In, Out, Stream, Init](atype, name, opts, fn)
```

Options structs hold descriptor data, and a single-function action's implementation stays positional. The background bundle's lifecycle functions past start ride in its options instead: `BackgroundActionOptions[In, Out]` (no stream slot) carries `Check`, required and validated at construction since polling is currently the only way to resolve a pending operation, and `Cancel`, optional. Adding a lifecycle function, or relaxing which ones are required once other resolvers exist, is then a field and validation change rather than a signature break; `In` is reserved for future input-typed functions. `NewBackgroundActionOf` also no longer leaks the start action's schemas into its check/cancel sub-actions, which share one inferred `Operation` schema. The bidi constructor already took an options struct, so `NewBidiActionOf` only reorders arguments; `BidiActionOptions` gains `Description`. `core.BackgroundActionDef` is renamed to `BackgroundAction` (consistent with `Action`, `BidiAction`, `Flow`), old name kept as a deprecated alias. Every caller in the tree moved.

`BackgroundAction` also stops exporting the start action it embeds. The field was named `Action` and therefore reachable as `bg.Action` from any package; it now embeds through an unexported alias, the way the `ai` primitives embed `core.Action`, so every promoted method is unchanged and only the field goes away. This is the one compile-level break here, on a field with no callers in the tree and whose type is documented for internal use. `base.SchemaMapFor` lands here for reuse by the typed-config PR.

## Explicit schema options apply to tools; inference helpers do not

For a tool, `In` and `Out` are the schema-inference mechanism and the model produces only the tool's input, so the type/enum inference helpers and the generation-steering output options make no sense on a tool; explicit schemas do, standing in for a type parameter of `any`. The option interfaces encode the split at compile time: the four schema options now return `InputSchemaOption`/`OutputSchemaOption`, which embed `ToolOption`, and no other input or output option compiles against a tool constructor.

```go
genkit.DefineTool(g, "search", "Searches the index",
    func(ctx *ai.ToolContext, input any) (any, error) { ... },
    ai.WithInputSchema(inSchema),
    ai.WithOutputSchema(outSchema))
```

A custom output schema requires `Out` to be `any` (the guard the input side already had) and becomes the tool's advertised output schema. Multipart tools, and the `exp` constructors that wrap them, accept the options too: the schema describes the logical output (the envelope's `output` field) while the wire format stays the envelope. Middleware-contributed tools now capture their definitions after registration, so named schemas resolve; on main they send the model a raw `{"$ref": "genkit:..."}` map. `WithInputType` no longer compiles against tool constructors; it never worked there.

## Registry-taking Define helpers deleted

A registry is unobtainable outside the framework, so the registry-taking `Define*` helpers in `core` had no reachable external callers and are deleted outright rather than deprecated: `DefineAction`, `DefineStreamingAction`, `DefineBackgroundAction`, `DefineBidiAction`, `DefineFlow`, `DefineStreamingFlow`, `DefineSchema`, `DefineSchemasFor`. The genkit wrappers remain the touchpoints, now composing `New* + Register`. `DefineSchemasFor`'s type walk moves into `genkit.DefineSchemasFor`, where its only reachable caller was; the guard panics name the genkit entry point.

To make `New* + Register` equivalent to the old `Define*`, the `"dynamic"` marker is now registration-derived: `Register` strips it on definition-time registration only, into a copy of the metadata map, which may be caller-owned or shared. Per-request child registration (detached tools during `Generate`) leaves metadata untouched, so detached tools keep their marker, matching JS.

## Bugs the migration surfaced

**Background models never advertised or validated their config schema.** `ai.NewBackgroundModel` built the input schema with `ConfigSchema` spliced into `properties.config`, exactly as `NewModel` does, then dropped it: the flat `core.NewBackgroundAction` had no input schema parameter. The options struct has the slot, so it is now passed and config is validated. Intended behavior change: a map config with keys outside the plugin's schema (googlegenai's generated schemas set `additionalProperties: false`) now fails with `INVALID_ARGUMENT` instead of the unknown keys being silently dropped.

**`BackgroundModelOptions.Metadata` was declared but never read.** It now seeds the metadata map, with the built-in `type` and `model` keys stamped over it. The model label also doubles as the description on the start and check actions, matching JS; a label defaulted from the model name yields to an explicit `Metadata["description"]`, an explicit label wins.

**Veo actions were rebuilt by hand instead of using the model they already had.** `googlegenai.resolveAction` called `newVeoModel`, which returns a fully-formed `ai.BackgroundModel`, then threw it away and re-wrapped its `Start` and `Check` in bare `core.NewAction` calls with nil metadata and nil schemas: no model metadata in the Dev UI, and the middleware `NewBackgroundModel` installs never ran.

**Before:** two hand-rolled actions, each re-deriving the operation key the framework already sets.

```go
case api.ActionTypeBackgroundModel:
    return createVeoBackgroundAction(client, name, provider)
case api.ActionTypeCheckOperation:
    return createVeoCheckAction(client, name, provider)

// ...

func createVeoBackgroundAction(client *genai.Client, name, provider string) api.Action {
    veoModel := newVeoModel(client, name, GetModelOptions(name, provider))
    actionName := api.NewName(provider, name)
    return core.NewAction(actionName, api.ActionTypeBackgroundModel, nil, nil,
        func(ctx context.Context, input *ai.ModelRequest) (*core.Operation[*ai.ModelResponse], error) {
            op, err := veoModel.Start(ctx, input)
            if err != nil {
                return nil, err
            }
            op.Action = api.KeyFromName(api.ActionTypeBackgroundModel, actionName)
            return op, nil
        })
}
```

**After:** return the bundle.

```go
case api.ActionTypeBackgroundModel, api.ActionTypeCheckOperation:
    if mt != ModelTypeVeo {
        return nil
    }
    opts := GetModelOptions(name, provider)
    return newVeoModel(client, name, opts).(api.Action)
```

Registering a background model registers its component actions together, and the registry looks up the requested key after registering, so one value serves both cases; that contract is documented on `api.DynamicPlugin.ResolveAction`. The `api.Action` assertion is pinned at compile time on the `ai` side, so breaking the embedded-action chain fails the build instead of panicking during resolution.

## Deprecations

`core.NewAction`, `NewStreamingAction`, `NewBackgroundAction`, and `NewBidiAction` are deprecated, delegating to their `Of` counterparts; unlike the `Define*` helpers they are reachable by plugin authors, so they get a normal cycle. When they go in the next major, the `Of` names collapse back to the plain ones, so migrating now is the last rename.

## Scope

- **No `Define*Of` surface.** A parallel family would only be `New* + Register` under a second name; the genkit wrappers stay the registering touchpoint.
- **`checkFn` is not positional even though it is required today.** A positional parameter encodes required-forever, and the background requirement is expected to evolve toward "at least one resolver" (wait, subscribe), which only a field plus construction-time validation can express. `ActionOptions` and `BidiActionOptions` stay non-generic: a single-function action customizes by wrapping its function, so no field there needs a type parameter, and bidi is experimental, so its options can still change when a real need appears.
- **The `ai` and `genkit` constructor surfaces are unchanged.** They move in #5849; this PR's `ai`/`genkit` changes are call-site migrations plus the fixes above.
- **Action registration stays unsynchronized.** `Register` documents the register-before-sharing contract; making it safe for concurrent registration (including a pre-existing registry-reference race) is deferred to a follow-up.

## Verification

Go 1.26.3, `go build ./...`, `go vet ./...`, and `go test ./...` in `go/`: 41 packages with tests pass, no failures; `go test -race` over `core`, `ai`, and `ai/exp` passes. New tests pin each schema slot's inference and override, the `"dynamic"` marker semantics, the tool schema options (multipart advertisement, last-wins repeats, concrete-`Out` panics in `ai` and `exp` alike), the middleware fix (fails without the reordering, with the raw `$ref` reaching the model), both background model fixes, and the Veo resolution paths.


huangjeff5 (commented):
Hi Alex (@apascal07), overall this is a great cleanup! Moving to `ActionOptions` makes action construction so much cleaner and aligns Go nicely with the other SDKs.

One quick question on `newAction` in `go/core/action.go`:

```go
description := opts.Description
if description == "" {
    if d, ok := opts.Metadata["description"].(string); ok {
        description = d
    }
}
```

If a caller specifies both `opts.Description` AND passes a `Metadata` map containing `"description"`, `action.Desc().Description` gets the new `opts.Description`, but `action.Desc().Metadata` still retains the old `"description"` string.

Could that cause metadata divergence for telemetry/tracing tools that inspect `Metadata["description"]`, or is keeping `Metadata` untouched intentional here?

apascal07:
Intentional, and the fallback is one-way on purpose. `Description` is the field tooling reads (reflection API, Dev UI, `Desc().Description`); `Metadata["description"]` is the legacy vehicle from the flat constructors, which had no description parameter, so the fallback exists to keep those callers working rather than to establish a second source of truth.

Nothing in the framework reads the metadata key apart from that fallback, and JS never had it: `defineAction` takes `description` and `metadata` as independent fields and passes metadata through verbatim. So a telemetry or tracing consumer that works across SDKs is already reading the top-level field.

What core deliberately doesn't do is stamp the resolved description back over the metadata key. That map is caller-owned and can be shared between actions (the same hazard that makes `Register` clone before stripping the `dynamic` marker), so overwriting it would silently discard a value someone set on purpose. Redundant-but-stale is a better failure mode than deleted.

Pinned in 49e38de: `ActionOptions.Description` now says the fallback is one-way and that metadata reaches the descriptor as given, and the precedence test asserts the metadata key is untouched when both are set.

go/ai/option.go
```diff
@@ -319,14 +328,14 @@ func WithInputType(input any) InputOption {
...
 
-// WithInputSchema manually provides a schema map for the prompt's input.
-func WithInputSchema(schema map[string]any) InputOption {
+// WithInputSchema manually provides a schema map for the input.
+func WithInputSchema(schema map[string]any) InputSchemaOption {
 	return &inputOptions{InputSchema: schema}
 }
 
 // WithInputSchemaName sets a pre-registered schema by name for the input.
 // The schema will be resolved lazily at execution time using [DefineSchema].
```
pavelgj: This godoc link is stale now that `core.DefineSchema` is deleted. From the `ai` package `[DefineSchema]` doesn't resolve to anything, so it'll render as a dead link. Probably want `genkit.DefineSchema` (or `registry.RegisterSchema`) here.

```go
// The schema will be resolved lazily at execution time using [genkit.DefineSchema].
```

go/ai/tools.go
```diff
@@ -295,8 +295,31 @@ func applyStrictMetadata(metadata map[string]any, strict *bool) {
...
+	if schema != nil {
+		metadata["originalOutputSchema"] = schema
+	}
+}
+
+// requireAnyTypeParam panics unless the type parameter T is an interface type
+// (in practice 'any'). The tool constructors call it before honoring an
+// explicit schema option: the custom schema stands in for a type parameter of
+// 'any', and a concrete T would silently disagree with the advertised schema.
+// requirement is the leading clause of the panic message, e.g.
```
pavelgj: Tiny nit: the sentence starts with the lowercase param name `requirement`, which reads a bit awkwardly in godoc. Maybe reword so it opens with a capital:

```go
// The requirement argument is the leading clause of the panic message, e.g.
// "WithInputSchema requires In".
```

go/plugins/googlegenai/veo_test.go
```diff
@@ -508,3 +510,66 @@ func TestCheckVeoOperationStructure(t *testing.T) {
...
+
+		desc := action.Desc()
+		if desc.Metadata["model"] == nil {
+			t.Errorf("start action is missing model metadata, got %v", desc.Metadata)
+		}
+	})
+
+	t.Run("non-veo models do not resolve as background models", func(t *testing.T) {
+		for _, atype := range []api.ActionType{api.ActionTypeBackgroundModel, api.ActionTypeCheckOperation} {
+			if action := resolveAction(client, googleAIProvider, atype, "gemini-2.5-flash"); action != nil {
```
pavelgj: Prefer `gemini-flash-latest`. (Also update the string in the `t.Errorf` right below.)

pavelgj (approved):

go/ai/option.go
apascal07: Used the full import path, `[github.com/firebase/genkit/go/genkit.DefineSchema]`: `ai` can't import `genkit` (that's the cycle), so the short `genkit.DefineSchema` form wouldn't resolve either. Gave `WithOutputSchemaName` the same pointer while there.

PR #5866, opened 2026-07-30
feat(sdp): Add unit tests

agduan:
(6/6) Adds unit tests for the Sensitive Data Protection (SDP) middleware.

[View diff](https://github.com/agduan/genkit/compare/sdp/5-redact...sdp/6-tests)

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-cloud/package.json
+++ b/js/plugins/google-cloud/package.json
@@ -54,6 +54,9 @@
     "node-fetch": "^3.3.2",
     "winston": "^3.12.0"
   },
+  "optionalDependencies": {
+    "@google-cloud/dlp": ">=3.0.0 <8.0.0"
+  },
   "peerDependencies": {
     "genkit": "workspace:^"
   },
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -14,9 +14,6 @@
  * limitations under the License.
  */
 
-export {
-  FirestoreSessionStore,
-  type FirestoreSessionStoreOptions,
-} from './session-store/firestore.js';
+export interface SdpOptions {}
 
-export { sensitiveDataProtection, type SdpOptions } from './sdp/index.js';
+export function sensitiveDataProtection(options?: SdpOptions): void {}
--- a/js/plugins/google-cloud/src/sdp/index.ts
+++ b/js/plugins/google-cloud/src/sdp/index.ts
@@ -88,7 +88,7 @@ export interface SdpOptionsBase {
   credentials?: any; // (Optional) Explicitly set the Google Cloud credentials
 }
 
-export async function createDlpClient(
+async function createDlpClient(
   options: SdpOptions
 ): Promise<v2.DlpServiceClient> {
   let dlpModule;
@@ -123,7 +123,7 @@ export type SdpOptions = SdpOptionsBase &
   );
 
 // 1. Inspect config
-export function buildInspectConfig(
+function buildInspectConfig(
   options: SdpOptions
 ): protos.google.privacy.dlp.v2.IInspectConfig {
   const defaultInfoTypes = [
@@ -150,7 +150,7 @@ export function buildInspectConfig(
 }
 
 // 2. De-identify config
-export function buildDeidentifyConfig(
+function buildDeidentifyConfig(
   options: SdpOptions
 ): protos.google.privacy.dlp.v2.IDeidentifyConfig {
   const inlineConfig =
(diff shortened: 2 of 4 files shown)


jhuleatt (approved):

js/plugins/google-cloud/tests/sdp_test.ts
```diff
@@ -0,0 +1,216 @@
...
+        inspectConfig: {
+          infoTypes: [
+            { name: 'US_SOCIAL_SECURITY_NUMBER' },
+            { name: 'EMAIL_ADDRESS', version: 'latest' },
+          ],
+        },
+      })
+    );
+  });
+});
```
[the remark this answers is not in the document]
agduan: Resolved, added this test

js/plugins/google-cloud/package.json
```diff
@@ -54,6 +54,9 @@
     "node-fetch": "^3.3.2",
     "winston": "^3.12.0"
   },
+  "optionalDependencies": {
+    "@google-cloud/dlp": "^3.0.0"
```
pavelgj: this is a very old version (6 years old)... latest is 7.0.1. Please see if you can use a range of versions covering the latest.

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,296 @@
...
+  } else {
+    request.inspectConfig = buildInspectConfig(options);
+    request.deidentifyConfig = buildDeidentifyConfig(options);
+  }
+
+  const [response] = await dlp.deidentifyContent(request);
+  return response.item?.value || text;
+}
+
+export const sensitiveDataProtection = generateMiddleware<SdpOptions>(
```
pavelgj: Using the generic type parameter isn't the correct way to type config on a generate middleware. Config must be declared via a Zod `configSchema` passed in the options. Beyond type-safety, the schema is required for critical Dev UI use cases: it's what gets attached to the middleware metadata so the UI can generate a config form.

Follow the `retry`/`tool-approval` pattern: define `SdpOptionsSchema` in Zod, pass it as `configSchema`, and type the export as `GenerateMiddleware<typeof SdpOptionsSchema>`. The inline-XOR-template shape maps to `z.union([...])`, and `z.infer` gives you back the existing `SdpOptions` type.

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,296 @@
...
+                projectId
+              );
+              part.metadata = { ...part.metadata, isCleaned: true };
+            });
+
+          // Wait for all network calls to finish in parallel
+          await Promise.all(redactionPromises);
+        }
+
+        const res = await next(envelope, ctx);
```
pavelgj, line 308: In streaming mode, chunks are emitted to the caller before this post-`next()` output redaction runs, so streamed model output won't be redacted. Fine to resolve via documentation for now, but let's call it out explicitly since a redaction middleware silently skipping streamed output is surprising.

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,296 @@
...
+  if (options.templates) {
+    request.inspectTemplateName = options.templates.inspectTemplateName;
+    request.deidentifyTemplateName = options.templates.deidentifyTemplateName;
+  } else {
+    request.inspectConfig = buildInspectConfig(options);
+    request.deidentifyConfig = buildDeidentifyConfig(options);
+  }
+
+  const [response] = await dlp.deidentifyContent(request);
+  return response.item?.value || text;
```
pavelgj: `|| text` falls back to the original whenever DLP returns a falsy value. If a part legitimately de-identifies to an empty string, this restores the sensitive original. Prefer a nullish check:

```suggestion
  return response.item?.value ?? text;
```

pavelgj (commented):
Took a pass over the SDP middleware (reviewing this PR since it carries the full feature). Overall it reads well and the config surface is nicely thought out. A few things inline.

One general item: this middleware needs to be documented in the plugin README, and a testapp exercising it would be really valuable for both users and manual verification. Right now there's no README coverage or sample for `sensitiveDataProtection`.

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,296 @@
...
+
+        const { client, projectId } = await clientPromise;
+
+        // Intercept input
+        if (envelope.request?.messages) {
+          const redactionPromises = envelope.request.messages
+            // ignore empty messages
+            .filter((message) => !!message.content)
+            // extract all message content into a single array
+            // @ts-ignore - flatMap creates a flat array of content parts
```
pavelgj: Rather than suppressing this, can we type it? `content` is `Part[]`, so annotating the callback param (or the `flatMap` result) should remove the need for `@ts-ignore`.

js/plugins/google-cloud/src/sdp/index.ts
```diff
@@ -0,0 +1,296 @@
...
+      deidentifyTemplateName?: string;
+    }
+  | {
+      inspectTemplateName?: string;
+      deidentifyTemplateName: string;
+    };
+
+export interface SdpOptionsBase {
+  projectId?: string; // (Optional) Explicitly set the Google Cloud Project ID
+  credentials?: any; // (Optional) Explicitly set the Google Cloud credentials
```
pavelgj: Minor: `any` loses type safety here. The DLP client constructor exposes a credentials type you can reuse, e.g. `NonNullable<ConstructorParameters<typeof v2.DlpServiceClient>[0]>['credentials']` (already referenced below at lines 112-114).

js/plugins/google-cloud/src/sdp/index.ts
agduan: Done, I converted all config interfaces to Zod schemas. Thanks, it looks much cleaner!

js/plugins/google-cloud/src/sdp/index.ts
agduan: Done, removed the `@ts-ignore` and annotated with 'Part[]'.

js/plugins/google-cloud/src/sdp/index.ts
agduan: I added a comment right above the `export const sensitiveDataProtection` statement: `**Warning:** In streaming mode, chunks are streamed from the model before redaction runs and will not be redacted.`

js/plugins/google-cloud/package.json
agduan: Replaced with `">=3.0.0"`. Let me know if I should use a more specific union instead.

js/plugins/google-cloud/package.json
```diff
@@ -54,6 +54,9 @@
     "node-fetch": "^3.3.2",
     "winston": "^3.12.0"
   },
+  "optionalDependencies": {
+    "@google-cloud/dlp": ">=3.0.0"
```
pavelgj: upper bound is usually a good idea:

```suggestion
    "@google-cloud/dlp": ">=3.0.0 <8.0.0"
```

pavelgj:
Alex, you mentioned you pushed the README changes, but I don't see it here... could you please doublecheck?

agduan:
@pavelgj I pushed the README changes in a separate PR here: #6073

PR #5874, opened 2026-07-30
refactor(go/plugins): migrate to the typed-config constructors for Anthropic, streamline others

apascal07:
Migrates the Anthropic plugin and the Vertex AI Model Garden Claude models to the typed-config constructors from #5849, the fourth PR of the typed-config train, after #5869. Both surfaces are backed by the same `plugins/internal/anthropic` package, so they move together and the hand-rolled config type switch is gone.

Making the config a typed value gives it a schema the framework validates on every request and the dev UI renders as a form, and pulling on that turned up two more things. The schema offered fields the plugin silently overwrites, and the errors explaining as much arrived as 500s with their causes detached. Correcting what a plugin knows about a model turned out to be impossible through the entry points built for it. So this also reaches `go/core`, `go/genkit`, and `googlegenai`.

**Before:** the plugin re-implements what the framework now does.

```go
func configFromRequest(input *ai.ModelRequest) (*anthropic.MessageNewParams, error) {
    switch config := input.Config.(type) {
    case anthropic.MessageNewParams:  return &config, nil
    case *anthropic.MessageNewParams: return config, nil
    case map[string]any:              return base.MapToStruct[anthropic.MessageNewParams](config)
    case nil:                         return &anthropic.MessageNewParams{}, nil
    default:                          return nil, fmt.Errorf("unexpected config type: %T", input.Config)
    }
}
ai.NewModel(name, meta, fn)
```

**After:** the type parameter is inferred from the function and the framework fills it.

```go
ai.NewModelAction(api.NewName(provider, name), &opts,
    func(ctx context.Context, input *ai.ModelRequest, config anthropic.MessageNewParams,
        cb ai.ModelStreamCallback) (*ai.ModelResponse, error) {
        return Generate(ctx, client, provider, apiModel, input, config, cb)
    })
```

The config travels by value from the model function down, so `toAnthropicRequest` amends the request's own copy rather than the caller's struct. The config schema is reflected once at package level instead of per model, which matters because `ListActions` built one per discovered model. A request carrying a config type the model was not defined with now fails with the framework's error rather than the plugin's `unexpected config type: %T`.

## One constructor for both plugins

`ant.DefineModel` and the Anthropic plugin's own `newModel` built the same action from the same options with the same generate closure, so the two collapse into `ant.NewModel`. The old name registered nothing despite the `Define` prefix, which is what #5849 removed elsewhere in the framework.

It returns the concrete `*ai.ModelAction`, so `ListActions`, `ResolveAction`, and the Model Garden `Init` no longer assert their way back to `api.Action`. Two options fields the old shared constructor dropped, `Stage` and `Metadata`, now reach the descriptor because the options struct is passed through whole rather than copied field by field.

Labels are the caller's to set: `ant.NewModel` fills one in only when the caller leaves it empty, deriving `"Anthropic - <name>"` or `"Vertex AI - <name>"` from the provider. The old constructor overwrote whatever it was given with `"<Provider>-<model id>"`, so the labels curated in the model catalog never reached the dev UI. Model Garden Claude models now show `Claude Opus 4.7` rather than `Vertex AI-claude-opus-4-7`, matching how the Llama models in the same plugin present theirs.

## Capabilities come from plugin config, not a registration call

Correcting what a plugin knows about a model is what `DefineModel` and its successors were for, and none of them could do it. `Init` registers nothing in any of these plugins, so every model arrives through `ResolveAction`, which built capabilities from the package-level catalog and never saw what the registration call was given. An override only won if it ran before the model's first use, and it could not reliably run at all: resolving a name registers it, and registering the same name afterwards panics. `IsDefinedModel` existed to guard exactly that race.

Capabilities move onto the plugin struct, where both paths read them:

```go
&googlegenai.GoogleAI{
    Models: map[string]ai.ModelOptions{
        "gemini-not-yet-released": {Supports: &ai.ModelSupports{Multiturn: true, Tools: true}},
    },
    Embedders: map[string]ai.EmbedderOptions{
        "text-embedding-not-yet-released": {Dimensions: 256},
    },
}
```

Ordering stops mattering, nothing panics, and the guard has nothing left to guard. `Anthropic`, `GoogleAI` and `VertexAI` all take `Models`; the two Gemini plugins also take `Embedders`. One catalog feeds both `ListActions` and `ResolveAction`, so Gemini, Imagen, Veo, Claude and the embedders all honor an entry no matter which path describes them first.

**Entries overlay rather than replace.** A zero-value field keeps what the plugin resolves, so pinning one capability does not mean restating the label and the config schema, which a model needs set to work at all. This is why the merge is a helper in `plugins/internal` and not a struct assignment: every field of both options structs distinguishes its zero value from a meaningful one, so "not specified" needs no extra signal.

An entry also makes an ID a known one, so the deprecated `DefineModel` stops rejecting a model described only in config.

## Typed config at the call site

`ModelRef` pairs a model with its config, so the config is a `*sdk.MessageNewParams` the compiler checks rather than an `any` the model validates at runtime:

```go
ai.WithModel(anthropic.ModelRef("claude-opus-4-5", &sdk.MessageNewParams{MaxTokens: 1024}))
```

This package and the Anthropic SDK are both named `anthropic`, so one of them needs an import alias.

Every entry point in the package now takes the model ID bare or provider-prefixed. `api.NewName` concatenates, so a prefixed ID previously produced `anthropic/anthropic/claude-opus-4-5` and named a model that resolves nowhere.

## The config schema the dev UI renders

A typed config is only as good as the schema advertised for it, and the raw SDK params struct makes a poor one. The plugin now curates it, the way `googlegenai` does.

**Fields a Genkit primitive owns are refused, naming the option to use instead.** `messages` and `model` were overwritten unconditionally while the request was built, the system prompt whenever the request carried system messages, and the output format whenever constrained generation was on, so setting any of them dropped the caller's value in silence. A custom function tool set through the config is worse than dropped: it reaches the model, which then calls a tool the Genkit runtime has no handler for. Server-side tools stay config-only, since the config is the only way to express them.

**Descriptions ride along.** The SDK types carry Go doc comments but no JSON Schema descriptions, so every field rendered bare. Curated text now covers each one, including that `temperature`, `top_k` and `top_p` are deprecated on Claude Opus 4.7 and later, which reject a value rather than ignoring it.

**The `any` junk field is gone.** The reflector emitted one on every object at every depth. It comes from the `param.APIObject` the SDK embeds in each params struct, which holds an anonymous `any` for the raw decoded message. It is machinery, not a request field, and the dev UI was rendering it throughout.

## Hiding a field no longer opens the object it lives in

Both plugins now hide a managed field by **replacing it with the permissive `true` schema rather than deleting it**, which fixes a real hole in the `googlegenai` schema.

Hiding has to survive input validation. A hidden field must still reach the plugin's own check, since that check is what names the primitive to use instead. Deleting the property fails the value as an unknown one, so `googlegenai` paid for the delete by setting `additionalProperties: true` on the parent to let it back through. That price is not confined to the hidden field: it drops unknown-field rejection for every other property of the same object. The root of `GenerateContentConfig` hides five fields and `tools[]` hides one, so both were wide open, and `maxOutputToken` or `max_output_tokens` was accepted and silently ignored. Only `thinkingConfig` and the Imagen and Veo configs, which hide nothing, still caught a typo.

Replacing hides just as well, since the dev UI renders only properties whose type it recognizes and a typeless one is skipped either way. The value validates, reaches the plugin check, and `additionalProperties` stays `false`.

The anthropic schema was built this way from the start, so this is `googlegenai` catching up rather than two changes.

## Errors along the way

Four fixes to the error plumbing, all found by driving the new config surface from the dev UI.

**Config rejections were 500s.** Both plugins build these with `errors.New`, and an unclassified error is `UNKNOWN`, which `status.Of` reports as `INTERNAL`. Telling a caller they used the wrong option reached them as a server fault. Both now use `status.ErrInvalidArgument`, so the dev UI and any HTTP transport answer 400. Not `ErrInvalidInput`, which means a value failed the action's input schema; these pass the schema and are refused on what they mean.

**The reflection server panicked on any unclassified error.** `handleRunAction` wrote the run's trace ID straight into the error envelope, which is a pointer allocated only from a stack or trace the error itself carried. A provider SDK rejecting a call returns its own error type, which arrives here as a plain error with neither, so the handler dereferenced nil and killed the connection mid-response instead of answering. Both the streaming and non-streaming paths did it. Pre-existing, and the common case for a failing request rather than a corner.

**Ten status errors formatted their cause instead of wrapping it.** `%v` and `%w` produce the same message, and `status.Errorf` takes its cause from the wrapping the format string performs, so `%v` returns an error carrying no cause at all. Schema resolution, input validation, and stream chunk failures all buried the error that explained them. Nothing else in the framework does this: 33 other call sites already use `%w`.

That class needed a test, because a call that drops its cause looks correct at the call site and stays correct in any test of that call site. `TestErrorArgsUseTheWrapVerb` type-checks every package in the module and reads each `status.Errorf` and `status.PublicErrorf` call, resolving the type of each argument and requiring `%w` wherever one implements `error`. About a second, and it fails with the file and line of every offender. It found an eleventh the survey of `%v` missed, since `convertKeepText` used `%s`.

**Invalid action input now reports through the sentinel.** The two input paths in `core/action.go` were the last callers of `NewSchemaValidationError`, deprecated since the status package landed, while their neighbors in the same function already used `ErrInvalidSchema` and `ErrInvalidOutput`.

## Constrained generation follows Anthropic's Structured Outputs list

Every curated model claimed `ConstrainedSupportAll`, and so did the fallback for models resolved dynamically. That list is narrower than the catalog, and the request path sends `output_config` only when the model claims support. A model claiming it wrongly has the request rejected after Genkit has already dropped the schema instructions from the prompt, so the failure mode is a hard error on a request that would otherwise have worked.

Models resolved dynamically no longer claim it, since a model the list does not name may predate the feature. Every curated model is on the list and keeps `ConstrainedSupportAll`, which leaves one capability set rather than two.

The Claude 5 generation (`claude-fable-5`, `claude-opus-5`, `claude-sonnet-5`) joins the curated set on the same terms.

## Bugs fixed

**`claude-opus-4-1` was retired on 2026-08-05 and still curated.** A retired ID resolves, registers, and advertises capabilities like any other, then fails every request with a 404 naming the model rather than the list that offered it. Nothing offline tells the difference, which is how it stayed. `TestSupportedModelsAreServableLive` now sends one 16-token request per curated entry, reading `supportedModels` directly so a model added to the catalog is covered without touching the test. Ten requests in parallel, about three seconds.

**A shared config had its tool list scribbled over by concurrent requests.** `toAnthropicRequest` appends the Genkit tools onto the config's `Tools` rather than assigning, so that server-side tools set through the config survive. The config it receives is a shallow copy, so that slice header still points at the caller's backing array, and a config hoisted into a package-level var or a `ModelRef` is shared by every request made with it. The append wrote into that array's spare capacity, so two concurrent requests raced over the same slots and one request's tools surfaced in another's. Clipping before the append forces an allocation. The test that claimed to pin this compared only lengths, which an in-place append leaves untouched while overwriting the slots past them.

**`IsDefinedModel` answered true for every model Anthropic serves.** It looked up through `genkit.LookupModel`, which resolves dynamically, so it asked the plugin to resolve the very model the caller was checking for, registered it as a side effect, and reported it defined. It now looks the action up without resolving, and is deprecated: with capabilities in config there is no registration left for it to guard.

**A nil `ModelOptions` panicked.** The pointer was dereferenced unguarded. A nil now takes the capabilities the plugin resolves by name, curated for a known model and the Claude defaults for the rest, rather than an empty `ModelOptions` advertising a model that supports neither tools nor multiturn nor a system role.

**`Anthropic.DefineModel` handed back a model closed over a zero-value client** when the plugin had not been initialized. That failed much later with `base url is not set`, pointing nowhere near the cause. It refuses now.

## Public API surface

```go
// Added: plugin config fields
anthropic.Anthropic.Models     map[string]ai.ModelOptions
googlegenai.GoogleAI.Models    map[string]ai.ModelOptions
googlegenai.GoogleAI.Embedders map[string]ai.EmbedderOptions
googlegenai.VertexAI.Models    map[string]ai.ModelOptions
googlegenai.VertexAI.Embedders map[string]ai.EmbedderOptions

// Added
anthropic.ModelRef(id string, config *sdk.MessageNewParams) ai.ModelRef

// Deprecated, behavior unchanged
(*anthropic.Anthropic).DefineModel        ->  Anthropic.Models
(*googlegenai.GoogleAI).DefineModel       ->  GoogleAI.Models
(*googlegenai.VertexAI).DefineModel       ->  VertexAI.Models
(*googlegenai.GoogleAI).DefineEmbedder    ->  GoogleAI.Embedders
(*googlegenai.VertexAI).DefineEmbedder    ->  VertexAI.Embedders
(*googlegenai.GoogleAI).IsDefinedEmbedder ->  nothing left to guard
(*googlegenai.VertexAI).IsDefinedEmbedder ->  nothing left to guard
anthropic.IsDefinedModel                  ->  nothing left to guard
anthropic.Model                           ->  ai.WithModelName, ModelRef, or genkit.LookupModel

// Removed from go/plugins/googlegenai, added after go/v1.11.0 and never released
(*GoogleAI).RegisterModel, (*VertexAI).RegisterModel
(*GoogleAI).RegisterEmbedder, (*VertexAI).RegisterEmbedder
(*GoogleAI).IsDefinedModel, (*VertexAI).IsDefinedModel

// Removed from go/core
core.SchemaValidationError, core.NewSchemaValidationError
core.ReflectionError, core.ReflectionErrorDetails, core.ToReflectionError
```

## Breaking changes

Two removals in `go/core`, both of types that were already deprecated and neither with a plausible caller outside the framework.

`SchemaValidationError` embedded `*GenkitError`, unwrapped straight back to it, and its constructor was one call to `status.Errorf` with `ErrInvalidInput`. Nothing builds one now that the action paths report through the sentinel, and everything the type could be matched by, `errors.Is(err, status.ErrInvalidInput)` still matches.

The reflection error envelope described how the reflection API answers a failed run. That API is how the dev UI talks to a running app, so the envelope was never something an application would build or read, and nothing outside `go/genkit` ever did. It moves there unexported, beside the handler that writes it. Wire format unchanged.

The `googlegenai` removals are not breaking. `RegisterModel`, `RegisterEmbedder` and `IsDefinedModel` landed on main in #5869, after `go/v1.11.0`, and no release contains them. `IsDefinedEmbedder` did ship and is only deprecated.

Three plugin behavior changes are user-visible and intended: a config the model was not defined with is rejected by the framework rather than the plugin, with a different message; the constrained-generation narrowing above; and config fields a Genkit primitive owns are now an error rather than silently dropped.

## What this deliberately does not do

- **No `ModelRef` for Model Garden Claude.** `vertexai/modelgarden` names its models differently and defines them from a fixed catalog at `Init`, so it needs its own pass rather than a copy of this one.
- **No coercion of dev UI form values.** Numeric config fields arrive as strings from the dev UI and fail validation (#6009). That is an Angular value-accessor bug in the UI, and papering over it in Go would be an unbounded new coercion rule with no precedent in `internal/base`.
- **No shared schema-curation helper.** `googlegenai` and `anthropic` now have the same `configOverrides` struct, the same path parser and the same hiding rule, in two copies. Lifting it out is worth doing once a third plugin needs it, not on this branch.
- **`VertexAI.Init` still requires credentials to describe a model.** It builds the genai client eagerly, so config-driven metadata, which needs no network, is gated behind the whole auth stack. Untouched here; it is why the catalog test has to supply throwaway ADC.

## Verification

Go 1.26.3.

Offline, the ten packages this touches (`ai`, `ai/exp`, `core`, `core/status`, `genkit`, `plugins/anthropic`, `plugins/googlegenai`, `plugins/internal`, `plugins/internal/anthropic`, `plugins/vertexai/modelgarden`): 772 tests, 2060 including subtests, no failures. Run with `HOME` pointed at an empty directory, so nothing passes on the strength of a developer's gcloud login.

Live against the Anthropic API: 2 tests, 26 including subtests, 46s. The suite moved from Sonnet 4.5 to Haiku 4.5, which supports every capability it exercises (tools, streaming, extended thinking, media, native constrained generation), taking it from 138s to 39s at a fraction of the cost. Requests carry no temperature, which is deprecated on Claude 4.7 and later and returns 400 when set.

New coverage pins a config entry reaching resolution rather than only registration, for models, embedders, Veo and both Gemini backends; the overlay keeping the fields an entry leaves unset; the config schema reaching the request input schema; everything the schema advertises deserializing into `MessageNewParams`; the curated labels surviving to the descriptor for both providers; the tool append leaving the caller's array untouched; bare and prefixed IDs naming the same model; which curated models advertise native structured output; every managed field being both hidden and rejected with the classification asserted; an unknown field on each `googlegenai` object that hides something now being rejected, which is what the closed schema buys and what fails on the old one; the reflection handler answering rather than panicking on an unclassified error; and every `status.Errorf` in the module wrapping its cause.


huangjeff5 (approved):

pavelgj (approved):

adesinah (approved):
Excellent work, @apascal07 
The change look good to me. I spent lot of time understanding the value in migrating to a typed-config constructors, then tests which pass reliably. I ran the PR against recently last released version, everything seems perfect.

PR #5875, opened 2026-07-30
feat(go): relax option merging to accumulate or last-win

apascal07:
Every option in `ai/option.go` returned an error when it was set more than once, or combined with a sibling variant (`WithMessages` with `WithMessagesFn`, `WithModel` with `WithModelName`, and so on). That made options impossible to compose: a caller could not build a request up from several helpers, and `GenerateData` could not inject its output type without conflicting with a caller-supplied `WithOutputSchema`.

Options now follow the standard Go functional-options pattern, merging left to right under two rules, and applying them never fails.

```go
// Before: rejected with "cannot set messages more than once"
// After:  messages accumulate in call order
genkit.Generate(ctx, g,
    ai.WithMessages(history...),
    ai.WithMessages(turn...),
)
```

**Options carrying multiple items accumulate** across repeats and across their variants: `WithMessages`/`WithMessagesFn`, `WithTools`, `WithResources`, `WithUse`, `WithMiddleware`, `WithDocs`/`WithTextDocs`, `WithDataset`, `WithToolResponses`, `WithToolRestarts`.

**Options filling a single slot take the last one set**: `WithConfig`, `WithModel`/`WithModelName`, `WithSystem`/`WithPrompt`, the output schema and format options, `WithStreaming`, `WithInput*`, `WithStrictSchema`, and the tool restart/respond options. The input configuration is one such slot as a whole: schema and default input replace together, so overriding `WithInputType` with `WithInputSchema` or `WithInputSchemaName` does not leave the old type's defaults behind to be rendered against the new schema.

A zero value does not fill a slot, so an earlier non-zero value cannot be un-set by a later `WithMaxTurns(0)` or `WithConfig(nil)`; the package doc states this. Only genuinely invalid arguments still fail, and they panic at the call site (a type `WithInputType` cannot turn into a schema). The unexported `apply*` methods drop their `error` return accordingly.

## Stream wrappers chain a caller's WithStreaming

`GenerateStream`, `GenerateDataStream`, `Prompt.ExecuteStream`, and `DataPrompt.ExecuteStream` attach an internal callback to feed their iterator. Appending it after the caller's options would let it win the last-win slot and silently discard a caller-supplied `WithStreaming` (previously a loud conflict error). The wrappers instead chain the two: the caller's callback runs first, then the iterator's, and both receive every chunk.

`DataPrompt.Execute` and `ExecuteStream` apply their typed input argument after caller options, so it wins the input slot over an `ai.WithInput` passed in opts; the docs on both methods and on `WithInput` state that precedence.

## Not a breaking change

**Compile-time.** Every `apply*` method is unexported and takes unexported parameter types, so `ai.GenerateOption` and its siblings were only ever implementable inside `package ai`. No consumer can have implemented them. Every exported constructor keeps its signature. This was verified by building a package outside `genkit/go` that holds each option interface, constructs each one, directly constructs the two exported option structs (`ai.RestartOptions`, `ai.RespondOptions`), and embeds one in a user type.

**Runtime.** The change is a relaxation, not a redefinition. "Valid today" means each slot was set at most once, and with exactly one setter both merge rules reduce to that same one value. The only inputs whose behavior changes are the ones that return an error today.

Two deltas worth naming:

- `GenerateData` now **prepends** its inferred `WithOutputType` instead of appending it, so a caller-supplied `WithOutputSchema` wins the schema slot while typed extraction keeps working. This matches what `GenerateDataStream` and `DefineDataPrompt` already did. It changes one currently-valid combination: `GenerateData[T](..., ai.WithOutputFormat("text"))` previously ended up as JSON because the injected option applied last; the caller's format now wins, and the godoc warns that a non-JSON format breaks typed extraction.
- A few genuinely-bad inputs report a different error. `WithTools(a), WithTools(a)` now accumulates and trips the existing duplicate check, so the message moves from `cannot set tools more than once` to `duplicate tool "a"`, surfacing at request time rather than at option application.

## Examples

Build a request in pieces, exercising both rules:

```go
opts := []ai.GenerateOption{
    ai.WithSystem("You are a helpful assistant."),
    ai.WithTools(searchTool),
}

opts = append(opts, ai.WithTools(adminTools...))   // Concatenates onto searchTool.
opts = append(opts, ai.WithSystem("Be terse."))    // Fills the same slot, so this wins.

resp, err := genkit.Generate(ctx, g, append(opts, ai.WithPrompt("What should I pack for Tokyo?"))...)
```

Keep typed extraction while supplying your own schema:

```go
// The caller's schema wins; the value is still parsed into Recipe.
recipe, _, err := genkit.GenerateData[Recipe](ctx, g,
    ai.WithPrompt("Invent a recipe."),
    ai.WithOutputSchemaName("StrictRecipe"),
)
```

## Scope

**Options outside `go/ai` keep rejecting repeats.** `genkit.Init` options (`WithPlugins`, `WithDefaultModel`, `WithPromptDir`), `ai/exp` agent options, and the firebase/localstore plugin options are untouched. Whether those should follow, particularly whether `WithPlugins` should accumulate, is a separate design question.

## Notes

Message accumulation composes the `MessagesFn` closures and concatenates into a fresh slice (`slices.Concat`) rather than appending in place. `ai.WithMessages(history...)` hands the caller's slice straight through, so appending would land in that array's spare capacity, invisible until the caller appends to their own history and silently rewrites messages already produced for the request.

Reviewing the touched functions also fixed two pre-existing defects riding along: `Retrieve` called with no document option now returns an error instead of panicking on `Documents[0]`, and prompt definition/execution, `Retrieve`, and `Embed` now guard ref-carried configs with the same typed-nil check `Generate` already had, so a ref holding a `(*T)(nil)` config no longer clobbers a real one.

Unrelated to the option change: the Tool Interrupts example in `go/README.md` called `Generate` inside the loop over `resp.Interrupts()` while reassigning `resp`, so with two or more interrupts the later ones resumed against a history that had already advanced past them. It now collects the restart parts and resumes once, matching the shape the interruptible-tools section further down already used.

## Testing

Go 1.26.3: `go build ./...` and `go vet ./...` clean across the module, `go test ./...` passes all 41 packages with tests with no failures, and `go test -race ./ai/` passes.

New coverage: collection options accumulate (messages across both variants, tools, docs, resources, middleware, tool responses/restarts, dataset), single-value options take last-win, the input config replaces as one slot with stale defaults cleared, `GenerateData` reaches the model with a caller-supplied schema while still extracting into the typed output, and `Retrieve` with no documents errors instead of panicking. Two of the tests are negative-checked, verified to fail against the unfixed code: the aliasing test appends to the caller's own history after the composed messages are materialized, and the streaming test asserts a caller's `WithStreaming` and the iterator both observe every chunk.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -416,6 +416,11 @@ if terse {
 response, _ := genkit.Generate(ctx, g, append(opts, ai.WithPrompt("What should I pack for Tokyo?"))...)
 ```
 
+Tool names must be unique across the merged list: repeating the same tool in
+two helpers is rejected when the request runs. These rules cover a single
+options list; APIs that layer two lists, like a prompt's define-time options
+against `Execute`-time options, document their own precedence.
+
 ### Generate Structured Data
 
 Get type-safe JSON output that maps directly to your Go structs:
--- a/go/README.md
+++ b/go/README.md
@@ -395,7 +395,26 @@ Options compose, so you can build a request up from several helpers. Options
 carrying multiple items (`ai.WithMessages`, `ai.WithTools`, `ai.WithDocs`,
 `ai.WithUse`) accumulate across repeats, while single-value options
 (`ai.WithConfig`, `ai.WithModelName`, `ai.WithSystem`) take the last one set.
-Repeating an option is never an error.
+Repeating an option is never an error, so a request can be assembled in pieces:
+
+```go
+opts := []ai.GenerateOption{
+    ai.WithModelName("googleai/gemini-flash-latest"),
+    ai.WithSystem("You are a helpful assistant."),
+    ai.WithTools(searchTool, weatherTool),
+}
+
+if isAdmin {
+    // Appends to the tools above; the model sees all of them.
+    opts = append(opts, ai.WithTools(adminTools...))
+}
+if terse {
+    // Fills the same slot as the system prompt above, so this one wins.
+    opts = append(opts, ai.WithSystem("You are a terse assistant. One sentence."))
+}
+
+response, _ := genkit.Generate(ctx, g, append(opts, ai.WithPrompt("What should I pack for Tokyo?"))...)
+```
 
 ### Generate Structured Data
 
--- a/go/README.md
+++ b/go/README.md
@@ -391,6 +391,12 @@ text, _ := genkit.GenerateText(ctx, g,
 fmt.Println(text)
 ```
 
+Options compose, so you can build a request up from several helpers. Options
+carrying multiple items (`ai.WithMessages`, `ai.WithTools`, `ai.WithDocs`,
+`ai.WithUse`) accumulate across repeats, while single-value options
+(`ai.WithConfig`, `ai.WithModelName`, `ai.WithSystem`) take the last one set.
+Repeating an option is never an error.
+
 ### Generate Structured Data
 
 Get type-safe JSON output that maps directly to your Go structs:
--- a/go/README.md
+++ b/go/README.md
@@ -561,18 +561,24 @@ resp, _ := genkit.Generate(ctx, g,
     ai.WithTools(transferTool),
 )
 
-if resp.FinishReason == ai.FinishReasonInterrupted {
-    for _, interrupt := range resp.Interrupts() {
-        meta, _ := ai.InterruptAs[TransferInterrupt](interrupt)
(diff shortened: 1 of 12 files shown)


huangjeff5 (approved):
LGTM! The accumulate vs last-win option rules make option composition much cleaner in Go.

Issue #5820, opened 2026-07-27
[plugins] port Amazon Bedrock to Python


cabljac:
Port of the **Amazon Bedrock** plugin to Python, using the Go plugin
(**[`genkit-ai/aws-bedrock-go-plugin`](https://github.com/genkit-ai/aws-bedrock-go-plugin)**)
as the source of truth. Same slice discipline as the Go train
(genkit-ai/aws-bedrock-go-plugin#136).

## Landing strategy

Work lands as small PRs into a shared base branch, `py-bedrock-plugin`,
rather than straight to `main`. The base branch rebases onto `main`
periodically; when the train completes, one PR takes the assembled plugin
to `main`. Reviewers get small diffs during development and the full
plugin at the end.

Plugin lives at `py/plugins/amazon-bedrock`, package
`genkit-plugin-amazon-bedrock`, module `genkit.plugins.amazon_bedrock`.

## Port structure (Go file → Python module)

| Go file | Responsibility | Python module |
|---|---|---|
| `bedrock.go` | plugin init, model/embedder definition, cache-point part | `plugin.py` |
| `types.go` | config, roles, finish reasons, reasoning part ctor | `config.py`, `typing.py` |
| `models.go` | capability registry, inference-profile prefix stripping | `model_info.py` |
| `generate.go` | Converse request/response conversion | `converters.py` + `models.py` |
| `stream.go` | ConverseStream block-indexed reassembly | `models.py` |
| `embed.go` | Titan/Cohere/Nova embedders, batching, concurrency | `embedders.py` |
| `image.go` | Titan Image / SD / Nova Canvas via InvokeModel | `image.py` |
| `rerank.go` | Cohere Rerank via InvokeModel | `rerank.py` |

## Key decisions

- **Namespace `bedrock`**, matching Go, so model refs read
  `bedrock/anthropic.claude-...` in every language.
- **Plain `boto3` + thread bridge, behind an internal transport seam.**
  aioboto3 rejected: aiobotocore pins botocore to a narrow patch range, so
  the plugin would dictate boto3 versions for every user. A thread bridge
  (`asyncio.to_thread` for unary calls, a pump thread for `ConverseStream`)
  keeps the event loop unblocked at LLM concurrency levels. Same call as
  Strands, pydantic-ai (pydantic/pydantic-ai#1222), and langchain-aws.
  All boto3 calls go through one transport module so we can swap to AWS's
  async SDK when it matures.
[...]

PR #5876, opened 2026-07-31
feat(py/plugins/amazon-bedrock): add Converse non-streaming generate path

hilariie:
Slice 3 of #5820. Non-streaming Converse generate path for the Amazon Bedrock Python plugin, ported from `genkit-ai/aws-bedrock-go-plugin` (`generate.go`, `types.go`). Stacked on #5699.

## What lands

- `transport.py`: every boto3 call goes through here. One shared client built lazily under a lock, each call pushed onto a worker thread with `asyncio.to_thread` so the loop stays free for the seconds-to-minutes a generate takes. Keeping boto3 behind one module means we can swap in AWS's async SDK later without touching converters or models.
- `converters.py`: pure request/response conversion. Roles, system extraction, media, tools and tool choice, cache points, stop reasons, token usage, tool-input coercion, reasoning parts.
- `models.py`: `BedrockModel.generate` plus AWS error to `GenkitError` status mapping.
- `plugin.py`: `resolve()` and `list_actions()` wired to real model actions.
- `samples/amazon-bedrock-sample`: Dev UI sample with text, structured output, tool calling, and reasoning flows, over Nova Lite, Llama 3.3, and DeepSeek R1.

Region resolution now fails loudly instead of defaulting to `us-east-1`, same as the Go plugin. A silent default sends traffic, and data, to a region nobody picked.

## Reasoning round-trips

Converse `reasoningContent` blocks become Genkit `ReasoningPart`s, and Bedrock-originated signed reasoning is replayed verbatim on follow-up turns. Verified live both ways: DeepSeek R1 (unsigned reasoning, not replayed) and Claude Sonnet 4.5 extended thinking (signed reasoning, replayed and accepted by the service, follow-up computes correctly).

Two wire facts reviewers should know:

- The Converse `signature` is a string on the wire. Go stores `[]byte` only because Go marshals bytes to base64 in JSON; base64-decoding it in Python corrupts the replay, and botocore rejects bytes for that field anyway.
- `redactedContent` is bytes, stored base64-encoded in part metadata because raw bytes crash `model_dump_json` during tracing on non-UTF-8 blobs. Wire-shape handling is unit-tested; Anthropic's documented magic-string trigger returns ordinary `reasoningText` through Bedrock, so no live trigger for a redacted block is known.

Both metadata keys are Bedrock-specific (`bedrockReasoningSignature`, `bedrockRedactedContent`) so reasoning from another provider never gets replayed into a Bedrock request.

## Divergences from Go

1. Timeouts are botocore `read_timeout`/`connect_timeout`/`max_pool_connections` (3600/60/50) rather than Go's whole-call deadline, which kills long generations mid-flight.
2. `resolve()` is lazy and serves any model ID, inference profile or ARN. `init()` returns `[]`, the Bedrock catalogue can't be enumerated.
3. AWS error codes map to `GenkitError` statuses. Go has no mapping.
4. `request.tool_choice` is a fallback; the Bedrock config wins because it can name a specific tool the core enum can't express.
5. Explicit `temperature=0.0` is sent, not dropped as unset.
6. `AWS_REGION` is read directly. botocore only started honoring it in 1.41, above our floor, so leaving it to the SDK chain would silently ignore the var our own error message tells you to set.
7. Go injects a default `maxTokens` for Claude models. We don't: the requirement and the per-family values couldn't be verified here, and guessing one silently caps output. Callers set `maxTokens` themselves.
8. Streaming requests work but arrive as one chunk until ConverseStream lands next slice.
9. boto3 floor is `>=1.37.24`, first release with cache points and `reasoningContent`.

Structured output needs `output_instructions=True` passed alongside `output_format` and `output_schema`. `supports.constrained` is `NONE` because Converse has no constrained-decoding mode, and core's json format injects no instructions on its own, so without the flag the schema never reaches the model. The sample and its README both show the working call.

## Lockfile

`uv.lock` refreshed with uv 0.12.0, what CI's unpinned `setup-uv` installs today. That drops the `exclude-newer`/`exclude-newer-span` options block this branch inherited (no released uv accepts `exclude-newer-span`, main's lock doesn't carry the block, and `uv lock --check` fails while it's there) and records the plugin's real version. Rest of the churn is 0.12.0 marker normalization, no dependency versions changed.

## Tests

159 unit tests, no AWS, plus 3 opt-in live tests (`BEDROCK_LIVE_TESTS=1`). 

Ran Nova Lite for text and the undocumented-tool round trip, DeepSeek R1 for reasoning, Llama 3.3 for structured output.

`tests/request_validation_test.py` runs botocore's own `ParamValidator` over built requests against the bedrock-runtime service model. Several Converse fields are `NonEmptyString` (`min: 1`) and botocore enforces that client-side, so an empty tool description, system text or `toolUseId` fails before it reaches AWS, as `ParamValidationError` (a `BotoCoreError`, not a `ClientError`). Go never hits this, its SDK does no client-side length check. The corpus covers every request-shaping branch the builder has: undocumented tool, blank system prompt, empty assistant text, tool round trip, media and cache points, dropped reasoning, and inference config on two model families.

## Dev UI

Screenshot 2026-07-31 at 14 08 19
Screenshot 2026-07-31 at 14 09 07
Screenshot 2026-07-31 at 14 10 36
Screenshot 2026-08-03 at 17 22 20

Declared models resolve, the plugin's `customOptions` schema (Max tokens, Tool choice) renders in the model runner, and the sample registers five flows and one tool.

### DIFF
--- a/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
+++ b/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
@@ -47,8 +47,8 @@ class BedrockConfig(ModelConfig):
     )
 
     max_tokens: int | None = None
-    """Maximum tokens to generate. When unset, the plugin leaves the field
-    unset except for Claude models, where Bedrock requires a value."""
+    """Maximum tokens to generate. When unset, the field is left unset and
+    the service applies its own default cap."""
 
     tool_choice: str | None = None
     """Tool choice mode: ``auto``, ``required``/``any``, ``none``, or a tool name."""
--- a/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
+++ b/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
@@ -24,10 +24,14 @@ from pydantic.alias_generators import to_camel
 from genkit import ModelConfig
 
 DEFAULT_MAX_RETRIES = 3
-# Socket read timeout, not a whole-call deadline: Bedrock generations can
+# Socket read timeout: the gap botocore tolerates between two reads, which
+# resets on every byte received. Generous because Bedrock generations can
 # legitimately run for many minutes (Nova allows 60-minute inference).
 DEFAULT_READ_TIMEOUT = 3600.0
 DEFAULT_CONNECT_TIMEOUT = 60.0
+# Whole-call deadline. Unlike the read timeout this does not reset, so a
+# connection that dribbles a byte at a time still ends.
+DEFAULT_TOTAL_TIMEOUT = 3600.0
 # The botocore default of 10 pooled connections throttles LLM concurrency.
 DEFAULT_MAX_POOL_CONNECTIONS = 50
 
--- a/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
+++ b/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
@@ -35,9 +35,8 @@ DEFAULT_MAX_POOL_CONNECTIONS = 50
 class BedrockConfig(ModelConfig):
     """Per-call configuration for Bedrock models.
 
-    Mirrors the Go plugin's ``Config`` surface. Unknown keys are tolerated for
-    forward compatibility but only the declared fields (and
-    ``additional_model_request_fields``) reach the Converse API.
+    Unknown keys are tolerated for forward compatibility, but only the declared
+    fields (and ``additional_model_request_fields``) reach the Converse API.
     """
 
     model_config = ConfigDict(
--- a/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
+++ b/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
@@ -24,15 +24,20 @@ from pydantic.alias_generators import to_camel
 from genkit import ModelConfig
 
 DEFAULT_MAX_RETRIES = 3
-DEFAULT_REQUEST_TIMEOUT = 30.0
+# Socket read timeout, not a whole-call deadline: Bedrock generations can
+# legitimately run for many minutes (Nova allows 60-minute inference).
+DEFAULT_READ_TIMEOUT = 3600.0
+DEFAULT_CONNECT_TIMEOUT = 60.0
+# The botocore default of 10 pooled connections throttles LLM concurrency.
+DEFAULT_MAX_POOL_CONNECTIONS = 50
 
 
 class BedrockConfig(ModelConfig):
     """Per-call configuration for Bedrock models.
 
-    Mirrors the Go plugin's ``Config`` surface. Unknown keys are allowed and
-    forwarded so callers can use provider-specific options without a plugin
-    release.
+    Mirrors the Go plugin's ``Config`` surface. Unknown keys are tolerated for
+    forward compatibility but only the declared fields (and
+    ``additional_model_request_fields``) reach the Converse API.
     """
 
     model_config = ConfigDict(
@@ -40,6 +45,10 @@ class BedrockConfig(ModelConfig):
         populate_by_name=True,
         extra='allow',
     )
+
+    max_tokens: int | None = None
+    """Maximum tokens to generate. When unset, the plugin leaves the field
+    unset except for Claude models, where Bedrock requires a value."""
 
     tool_choice: str | None = None
     """Tool choice mode: ``auto``, ``required``/``any``, ``none``, or a tool name."""
@@ -58,5 +67,5 @@ class ModelDefinition(BaseModel):
     name: str
     """Bedrock model ID, e.g. ``anthropic.claude-sonnet-4-5-20250929-v1:0``."""
 
-    type: Literal['chat', 'image', 'embedding'] = 'chat'
+    type: Literal['chat', 'text', 'image', 'embedding'] = 'chat'
     """Routes generate calls: chat/text via Converse, image via InvokeModel."""
--- a/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
+++ b/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
@@ -45,10 +45,6 @@ class BedrockConfig(ModelConfig):
         extra='allow',
     )
 
-    max_tokens: int | None = None
-    """Maximum tokens to generate. When unset, the field is left unset and
-    the service applies its own default cap."""
-
     tool_choice: str | None = None
     """Tool choice mode: ``auto``, ``required``/``any``, ``none``, or a tool name."""
 
--- a/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/converters.py
+++ b/py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/converters.py
@@ -57,18 +57,6 @@ REDACTED_CONTENT_METADATA_KEY = 'bedrockRedactedContent'
 CACHE_POINT_CUSTOM_KEY = 'bedrockCachePointType'
 DEFAULT_CACHE_POINT_TYPE = 'default'
 
-# Bedrock requires maxTokens for Claude models; newer families allow more.
-DEFAULT_CLAUDE_MAX_TOKENS = 4096
-DEFAULT_EXTENDED_CLAUDE_MAX_TOKENS = 8192
-EXTENDED_CLAUDE_MAX_TOKEN_PATTERNS = (
-    'claude-3-5',
-    'claude-3-7',
-    'claude-4',
-    'claude-haiku-4',
-    'claude-sonnet-4',
-    'claude-opus-4',
-)
-
 IMAGE_FORMATS = {
     'image/png': 'png',
     'image/jpeg': 'jpeg',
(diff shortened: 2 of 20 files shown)


py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/plugin.py
```diff
@@ -0,0 +1,201 @@
...
+        Args:
+            action_type: The kind of action to resolve.
+            name: The namespaced action name.
+
+        Returns:
+            Action object if resolvable, None otherwise.
+        """
+        if action_type != ActionKind.MODEL:
+            return None
+        model_id = name.removeprefix(f'{BEDROCK_PLUGIN_NAME}/')
```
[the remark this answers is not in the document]
hilariie: Added the guard in `3995bf538`, though as defense-in-depth rather than a fix for the described scenario.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/converters.py
```diff
@@ -0,0 +1,755 @@
...
+
+def _reasoning_block_to_part(block: dict[str, Any]) -> Part | None:  # noqa: ANN401
+    reasoning_text = block.get('reasoningText')
+    if reasoning_text is not None:
+        # Both wire shapes occur: {'text': ...} and a bare string.
+        if isinstance(reasoning_text, str):
+            text, signature = reasoning_text, None
+        else:
+            text = reasoning_text.get('text') or ''
+            signature = reasoning_text.get('signature')
```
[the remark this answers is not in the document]
hilariie: Addressed in `3995bf538`

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/transport.py
```diff
@@ -0,0 +1,143 @@
...
+        else:
+            region = self._region or env_region or session.region_name
+        if not region:
+            raise GenkitError(message=NO_REGION_MESSAGE, status='FAILED_PRECONDITION')
+
+        return session.client(
+            'bedrock-runtime',
+            region_name=region,
+            config=Config(
+                retries={'max_attempts': self._max_retries, 'mode': 'standard'},
```
cabljac: This always sets an explicit `mode` and `max_attempts`, and botocore treats both as overriding everything, including the env vars. With `AWS_MAX_ATTEMPTS=10 AWS_RETRY_MODE=adaptive` set, the client still comes out `{'mode': 'standard', 'total_max_attempts': 4}`. Same for `read_timeout`, `connect_timeout` and `max_pool_connections`, including via an injected session.

So if someone has set those env vars, we quietly ignore them. Adaptive is the mode you'd want under Bedrock throttling, and they'd get standard instead without knowing.

Could we only set the keys the caller hasn't already?

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/config.py
```diff
@@ -41,6 +46,10 @@ class BedrockConfig(ModelConfig):
         extra='allow',
     )
 
+    max_tokens: int | None = None
```
cabljac: `maxTokens` and core's `maxOutputTokens` are two public keys for the same knob, and `maxTokens` wins at `converters.py:195-199`. The comment at `converters.py:197` calls `max_output_tokens` the legacy key, but `maxOutputTokens` is the canonical Genkit one and no other plugin redeclares it.

Raising it now because removing `maxTokens` after release breaks configs.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/transport.py
```diff
@@ -0,0 +1,143 @@
...
+        with self._lock:
+            if self._client is None:
+                self._client = self._build_client()
+            return self._client
+
+    async def ensure_client(self) -> None:
+        """Builds the client off-loop so init fails fast on config errors."""
+        await asyncio.to_thread(self.client)
+
+    async def converse(self, **kwargs: Any) -> dict[str, Any]:  # noqa: ANN401
```
cabljac, line 150: `converse()` / `_converse_sync()` has no test. `transport_test.py` covers region resolution, single client build and the botocore config, and `models_test.py:44` stubs the transport itself, so the boto3 bridge is never exercised. That is the path every non-streaming generate takes, in the module this PR describes as the single chokepoint for boto3.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/models.py
```diff
@@ -0,0 +1,131 @@
...
+    """Structural contract for the transport seam (see ``transport.py``)."""
+
+    async def converse(self, **kwargs: Any) -> dict[str, Any]:  # noqa: ANN401
+        """Calls the Converse API and returns the raw response dict."""
+        ...
+
+
+# AWS error codes → Genkit statuses; anything unlisted maps to UNKNOWN.
+_ERROR_CODE_STATUS: dict[str, StatusName] = {
+    'ThrottlingException': 'RESOURCE_EXHAUSTED',
```
cabljac, line 51: The error mapping is one of the bigger wins over Go here, but it has no coverage: `TooManyRequestsException`, `ServiceQuotaExceededException`, `UnrecognizedClientException`, `ExpiredTokenException`, `ModelNotReadyException` and `ModelErrorException` appear in no test, in either casing, nor do `PartialCredentialsError`, `NoRegionError` or `ConnectTimeoutError`.

Also `ThrottlingException` maps to `RESOURCE_EXHAUSTED` with no `retry_after_ms`. `genkit-anthropic/models.py:102` parses `Retry-After` into milliseconds, and Bedrock returns the header at `error.response['ResponseMetadata']['HTTPHeaders']`.

cabljac (commented):
This looks good overall. Ran the converters against the Go implementation fairly closely and the mapping holds up: stop reasons, role handling, all the content block types and media decoding all line up. `request_validation_test.py` running botocore's own `ParamValidator` over the built requests is a nicer approach than anything the other plugins do, and the reasoning signature round-trip is well handled and well explained.

One thing worth digging into, then a couple of small ones.

Timeout first. `config.py:27-29` already flags this as a socket read timeout rather than a whole-call deadline, so this is more about the consequence than the choice: as far as I can tell nothing else caps total duration. `read_timeout` restarts on every socket read, so a server that emits any byte inside each window never trips it, and I couldn't find an `asyncio.timeout` anywhere in the package. Cancelling the `await` at `transport.py:113` won't stop the boto3 call either, since the work is already on a worker thread.

Worth checking on your side, but if that holds, a hung or slow-dribbling connection would pin a worker thread until the process dies, and with `max_pool_connections=50` a handful of those could stop the app serving without anything showing up in the logs.

Definitely not asking for Go's 30s, that kills long generations mid-stream on their side. Something configurable that defaults to finite would cover it.

---

Separate and much smaller: a fair few comments explain behaviour by referring to the Go plugin (`matching Go`, `like Go`, `Ported from`), mostly in `converters.py`. Makes complete sense while porting, but once this is merged the reader won't have the Go source to hand, so the reasoning goes with it. Most are a single clause away already. The region docstring at `transport.py:90` is the one I'd point at as the model: the rationale there is the best comment in the diff and stands entirely on its own once `Matching the Go plugin,` comes off the front.

`transport.py:123` is the opposite case and worth keeping, it's the only thing justifying the manual `AWS_REGION` lookup. The one risk is that nothing pins it: if the botocore floor moves past 1.41 the comment goes stale, the branch becomes dead, and the suite would still pass. `pyproject.toml:40` only declares `boto3>=1.37.24` and both `transport.py` and `models.py` import `botocore` directly, so it may want declaring there regardless, which would give the version claim a home.

Tiny thing, `main.py:20` mentions three models and there are four now.

py/packages/genkit-amazon-bedrock/tests/converters_test.py
```diff
@@ -0,0 +1,842 @@
...
+
+def test_configured_max_tokens_is_sent() -> None:
+    request = user_text_request(config=BedrockConfig(max_tokens=32))
+    kwargs = build_converse_request('anthropic.claude-3-haiku-20240307-v1:0', request)
+    assert kwargs['inferenceConfig'] == {'maxTokens': 32}
+
+
+@pytest.mark.parametrize('model_id', ['us.anthropic.claude-sonnet-4-5-20250929-v1:0', 'amazon.nova-lite-v1:0'])
+def test_no_max_tokens_is_injected_for_any_model(model_id) -> None:
+    # Verified live: Converse accepts Claude requests without maxTokens and
```
cabljac: nit: this comment sits in a unit test that never calls the service, `live_test.py:130` already records the same fact in the place that does. Worth dropping this copy.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/converters.py
```diff
@@ -0,0 +1,715 @@
...
+
+    Args:
+        model_id: Bedrock model ID, inference-profile ID, or ARN.
+        request: The Genkit model request.
+
+    Returns:
+        Keyword arguments for the Converse call.
+    """
+    if request is None:
+        raise GenkitError(message='bedrock: model request is nil', status='INVALID_ARGUMENT')
```
cabljac: nit: "nil" in user-facing error text here and at `:697`.

hilariie:
Thanks, all addressed:

- [fa1aa45](https://github.com/genkit-ai/genkit/pull/5876/commits/fa1aa45764fa2085d6e221b2a0ecc66276924117) comments rewritten to stand on their own without the Go source
- [669d145](https://github.com/genkit-ai/genkit/pull/5876/commits/669d14571cad4ac5ac60b733277cce8d5579ba7d) `maxTokens` dropped, core `maxOutputTokens` is the only key now
- [8f9b0d1](https://github.com/genkit-ai/genkit/pull/5876/commits/8f9b0d1360e05b2e38081d9506cffa2fbf18d90e) `Retry-After` surfaced as `retry_after_ms`, plus full coverage of the error mapping
- [5f6152c](https://github.com/genkit-ai/genkit/pull/5876/commits/5f6152c76804c81da0fdbbf260d8885d19e4d659) retries and timeouts now defer to the ambient AWS config per key, and `converse` is covered directly
- [40cd24d](https://github.com/genkit-ai/genkit/pull/5876/commits/40cd24defbd34c7b423b7a23dc2f53f36ad3e2bf) `botocore` declared

On the timeout, I added `total_timeout` and finite by default at 3600s. It frees the caller, but boto3 gives no way to abort an in-flight call, so the worker thread still runs until the socket timeouts end it.

PR #5879, opened 2026-07-31
fix(py): correct PyPI package READMEs and metadata for renamed plugins

jeffdh5:
## Summary
- Fix install commands and prose still pointing at `genkit-plugin-*` (and the misspelled `genkit-plugins-google-genai`) so PyPI project pages install the new package names
- Correct `genkit-flask` summary from "Firebase" to "Flask"
- Add install/usage sections to stub READMEs (`flask`, `google-cloud`, `vertexai`, `openai`, `anthropic`)
- Drop broken Changelog URLs (only `genkit-ollama` has a CHANGELOG file) and use absolute GitHub sample links that work on PyPI

## Test plan
- [ ] Spot-check PyPI long descriptions after next publish / TestPyPI dry-run build
- [ ] Confirm `pip install` commands in each package README match `project.name` in `pyproject.toml`
- [ ] Confirm no relative `../../samples/...` links remain in package READMEs

### DIFF
--- a/py/packages/genkit-anthropic/README.md
+++ b/py/packages/genkit-anthropic/README.md
@@ -5,7 +5,30 @@
 >
 > **Preview** — This plugin is in preview and may have API changes in future releases.
 
-This Genkit plugin provides a set of tools and utilities for working with Anthropic.
+Anthropic Claude model provider for Genkit.
+
+## Installation
+
+```bash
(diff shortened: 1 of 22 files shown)


jeffdh5:
Closing so Jeff can land this via patch instead.

PR #5882, opened 2026-08-01
feat(go/ai): type prompt content functions and fix the message slots

### LINKED ISSUE #4108: [Go] Bring Prompts / Dotprompt into parity with TS
There is a lot to untangle here. Some general observations:

- [ ] `WithSystem` should render a single message or fail
- [ ] `WithPrompt` (aka user) should render a single message or fail
  - Note this is where dotprompt is currently passed, so we need to make some decision here on how to disambiguate
- [ ] `WithMessageFn` should not render templates at all
  - Currently it renders part by part as templates 
  - This adds significant complexity and merge logic which should be removed
- [ ] `MessageFn` should be passed Message[] (aka history) and be allowed to inject where appropriate
- [ ] Dotprompt should be passed Message[] (aka history) and be allowed to inject where appropriate. 
  - aka {{history}} helper should work
- [ ] In general, if provided, `MessageFn` or Dotprompt should be "in charge" of rendering history

apascal07:
Fixes #4108. Stacks on #5875, which relaxes option merging to accumulate or last-win. Every option added here follows its rules.

Reworks how a prompt assembles its messages. Three defects motivate it.

**The `any` argument to a content function had no stable type.** Its concrete Go type depended on how the prompt was invoked, so a function written against the prompt's input type worked in-process and panicked everywhere else:

| Call path | Type the function received |
|---|---|
| `p.Execute(ctx, ai.WithInput(HelloInput{...}))` | `HelloInput` |
| `p.Execute(ctx)`, default input from `WithInputType` | `map[string]any` |
| Dev UI, flows over HTTP, any reflection-API call | `map[string]any` |
| `genkit.Generate(ctx, g, ai.WithPromptFn(...))` | always `nil` |

`go/samples/prompts/main.go` demonstrated the pattern with a type assertion, so the documented usage panicked on two of the three prompt paths the moment anyone opened the Dev UI.

**Computed content was recompiled as a template.** Whatever a content function returned, and the text of every `WithMessages` value, was fed back through dotprompt. A user asking "how do I write `{{#if}}` in handlebars" failed the whole request with `Parse error on line 1: Expecting OpenEndBlock, got: 'EOF'`, text that merely resembled a placeholder was silently substituted, and MCP-fetched prompt messages became a template-injection surface.

**The conversation had no slot.** A `.prompt` file's body was loaded into `WithPrompt`, the slot for the final user turn, because Go had no string form for a multi-turn conversation. `{{history}}` therefore never worked, no prompt could decide where a caller's conversation goes, and prompt-backed agents spliced their session's messages in by hand after the whole rendered prompt.

## Content functions declare their input type

**Before:** the function receives `any`, and its concrete type depends on the caller.

```go
genkit.DefinePrompt(g, "greet",
	ai.WithInputType(HelloInput{Theme: "pirate"}),
	ai.WithPromptFn(func(ctx context.Context, input any) (string, error) {
		// panics from the Dev UI and on the default-input path,
		// where input is a map[string]any
		return "Hello, my name is " + input.(HelloInput).UserName, nil
	}),
)
```

**After:** the function declares the type and Genkit converts whatever arrived.

```go
genkit.DefinePrompt(g, "greet",
	ai.WithInputType(HelloInput{Theme: "pirate"}),
	ai.WithPromptFn(func(ctx context.Context, input HelloInput) (string, error) {
		return "Hello, my name is " + input.UserName, nil
	}),
)
```

`Generate` has no prompt input and passes the zero value rather than `nil`, so the same function is safe there.

The guarantee reaches the Go types inside the declared type, not just the type itself. A function declaring a loose `map[string]any` saw an `int64` for an integer field when the value arrived over the wire, because the reflection API normalizes against the input schema, and a `float64` when an in-process caller passed a struct, because nothing normalized that round trip. `WithInputType` had the same gap building its default input. Both normalize now, so the field has one type on every path.

## An input that cannot reach its content function is classified

The conversion can fail: an input that is neither the declared type nor decodable into it means the function is never called. That is a caller-supplied value going wrong at runtime, so it gets a sentinel rather than a string to match on.

```go
// ai/errors.go. Also matches status.ErrInvalidArgument.
ErrInputTypeMismatch = status.ErrInvalidArgument.Subtype("input type mismatch")
```

It is attached in `coerceFn`, the single point every typed content function's input passes through, so it covers all four slots and both ways the conversion fails (an unrelated struct, and a value that will not decode). The underlying error stays in the chain. It cannot be attached where it is raised: `internal/base` produces it, `core/status` imports `internal/base`, and users cannot import `internal/base` at all.

## Strings are templates, structured content is not

JS compiles `system`, `prompt`, and `messages` only in their **string** form. `Part`, `Part[]` and `MessageData[]` values, and every resolver result, go through `normalizeParts` untouched ([prompt.ts:598](https://github.com/genkit-ai/genkit/blob/main/js/ai/src/prompt.ts#L598)):

```ts
} else if (typeof options.messages === 'string') {
    promptCache.messages = await registry.dotprompt.compile(options.messages);
    ...
} else {
    messages.push(...options.messages);   // MessageData[]: verbatim
}
```

Go recompiled the structured forms and the function results, which JS never does, so this is a parity fix rather than a Go-specific policy. All three string forms now exist in Go and all three are compiled; everything else is content the caller already produced. The call sites do not change, only what they do: `ai.WithMessages(resp.History()...)`, the dominant usage across samples and plugins, used to fail outright on a turn containing `{{`.

Templating had been supplying message copies as a side effect, so removing it made the copying explicit: `renderMessages` clones every message it splices in, from any source. Messages from `WithMessages` live on the prompt and are reused by every execution, and messages from a content function or from execute-time history typically alias a session or the caller's own slice. Middleware stamping metadata is enough to corrupt either one.

## The prompt owns its conversation

JS loads a `.prompt` body into `messages` ([prompt.ts:911](https://github.com/genkit-ai/genkit/blob/main/js/ai/src/prompt.ts#L911)); Go had nowhere to put it. Four defects followed.

**`{{history}}` never worked.** Dotprompt places the caller's conversation at that marker, and inserts it before the template's final user message when the template does not say, but only if the render is handed the messages. Nothing ever passed them, so the marker always expanded to nothing.

```go
// The string form of messages, which Go was missing.
func WithMessagesTemplate(text string, args ...any) PromptOption
```

The `.prompt` loader uses it now, and history reaches it.

**Ownership was inverted.** Because no template could place history, `Prompt.Execute` spliced the caller's messages in at a fixed position, directly after the leading system messages and ahead of everything the prompt itself declared. A prompt that declares its conversation now places them itself, with `{{history}}` in a template or `HistoryFromContext` in a function; one that declares none still gets them in the middle, exactly where they went before. A `.prompt` file lands in the first case but keeps the old order anyway, since dotprompt inserts history before the template's final user message when the template does not say. That is why every existing `.prompt` file test passed unchanged.

**The single-message slots were not enforced.** `WithSystem` and `WithPrompt` each fill exactly one message whose role belongs to the slot, since the surrounding conversation is positioned against them. A `{{role}}` block that split the text took over the conversation; one that produced a single message had its role stamped back to the slot's, so `WithSystem("{{role \"user\"}}hello")` quietly became a system message. Both are now an error naming `WithMessagesTemplate`, as is a marker that merely restates the slot's own role: it means the same thing, that the author is writing turns. JS throws only on the split ([prompt.ts:751](https://github.com/genkit-ai/genkit/blob/main/js/ai/src/prompt.ts#L751)) and discards the role otherwise, so this is a deliberate tightening past parity that rejects only templates whose text already contradicted what they produced.

**Agents assembled their own messages.** `agentLoop` rendered the prompt and then appended `sess.Messages()` itself, so an agent's prompt could never place its conversation: `{{history}}` expanded to nothing and `HistoryFromContext` returned nil on every turn, which put trimming and summarizing an agent's history out of reach. It also mis-ordered a prompt with a user turn, producing `[system, user_prompt, ...history]` where JS pins `[system, ...history, user_prompt]` ([agent_test.ts:3022](https://github.com/genkit-ai/genkit/blob/main/js/ai/tests/agent_test.ts#L3022)).

The session's messages now go to `Render` on a scoped context and the prompt places them, matching what JS has been doing all along ([agent.ts:1498](https://github.com/genkit-ai/genkit/blob/main/js/ai/src/agent.ts#L1498)), internal tag name included. Because the stored conversation is what the prompt placed, an agent that compacts its history compacts its session too, rather than recomputing the compaction from a growing transcript every turn.

History keeps its identity across the render. `convertToPartPointers` handles only text and media, so sending messages through dotprompt and back would have turned tool requests and resources into nil parts and failed schema validation. The stand-ins are marked as history on the way in, and the originals are restored once dotprompt has decided where they go.

## Multi-part content no longer needs a template escape

Before this, the only way to get media out of a content function was to emit a `{{media url=...}}` string and rely on the re-templating defect to expand it.

```go
genkit.Generate(ctx, g,
	ai.WithPromptParts(
		ai.NewTextPart("What is in this image?"),
		ai.NewMediaPart("image/png", imageURL),
	),
)
```

```go
func WithSystemParts(parts ...*Part) PromptingOption
func WithPromptParts(parts ...*Part) PromptingOption
func WithSystemPartsFn[In any](fn func(context.Context, In) ([]*Part, error)) PromptingOption
func WithPromptPartsFn[In any](fn func(context.Context, In) ([]*Part, error)) PromptingOption

// Input-driven document selection, the counterpart of the JS DocsResolver.
// DefinePrompt only, since that is the only place the documents are not yet
// known when the option is constructed.
func WithDocsFn[In any](fn func(context.Context, In) ([]*Document, error)) PromptOption
```

`ai.WithDocs` now also applies to `DefinePrompt`, which it previously did not.

## The conversation reaches content functions

```go
// The conversation the prompt was handed, for one that manages its own.
// Counterpart of the history argument JS hands a messages resolver.
func HistoryFromContext(ctx context.Context) []*ai.Message

// Attach a conversation for the next Render to place. Prompt.Execute does this
// itself; this is for callers pairing Render with GenerateWithRequest, which is
// how the agent runtime drives a prompt.
func NewHistoryContext(ctx context.Context, messages []*ai.Message) context.Context
```

A context accessor rather than an extra function argument, since `ctx` is already threaded. `HistoryFromContext` is the function-form counterpart of `{{history}}`: a prompt using `WithMessagesFn` owns the conversation and reads it here to summarize, truncate, or reorder rather than only prepend. `Prompt.Execute` scopes the context to the render alone, so the history does not ride the generate loop into tools and nested prompts.

## The new options merge; one combination is refused

#5875 removes the "cannot set X more than once" errors from `ai/option.go`, so each option added here needed a merge rule rather than a conflict.

**The system and prompt slots hold one message each**, so their four setters share a slot and the last one set wins:

```go
// The parts win the slot outright. They are not appended to the text.
ai.WithSystem("You are terse."),
ai.WithSystemParts(ai.NewTextPart("You are verbose."), logoPart),
```

The losing field is cleared rather than left behind, since both feed the same message and the renderer has to read them in some fixed order. Before this, a function beat text no matter which came first, silently.

**Documents accumulate**, so `WithDocsFn` adds to `WithDocs` and `WithTextDocs` rather than replacing them. Fixed documents resolve first, then computed ones.

**The conversation follows both rules, with one combination refused.** `WithMessages` and `WithMessagesFn` accumulate; `WithMessagesTemplate` is a slot whose last setter wins. Giving a template alongside either of the others in one `DefinePrompt` call panics:

```
ai.DefinePrompt: "chat" sets both WithMessagesTemplate and WithMessages/WithMessagesFn,
which have no meaningful combination: the template is the whole conversation.
```

A template lays the conversation out, down to where `{{history}}` puts the caller's, so separately supplied messages have no position relative to it: before it they are dropped, after it they land past the template's own final user turn, where few-shot examples are useless. Anything with a sensible position is already a `{{role}}` block in the template. This is a wiring mistake rather than two values competing for one slot, which is the class #5875 stopped rejecting, so it panics at definition where the offending call is, as `DefinePrompt` already does for a missing name and bad middleware.

The pairing that matters is protected by the type rather than a check: `WithMessagesTemplate` is a `PromptOption`, so it can never reach `Prompt.Execute`, which is where a caller's conversation comes from and how it reaches `{{history}}`.

## Source compatible; the runtime changes are the point

Nothing exported was removed or renamed, and `PromptFn` / `MessagesFn` still exist. The generic retrofit is source compatible: a function with the old `func(context.Context, any)` signature still satisfies the option with `In` inferred as `any`. The one construct that breaks is using `ai.WithPromptFn` as an un-instantiated function value, which no code in this repo does. `DocumentOption` gained a method, but all of its methods are unexported, so no external package can have implemented it.

Runtime impact is real and intended:

- Results of `WithSystemFn` / `WithPromptFn` / `WithMessagesFn`, and the text of `WithMessages` values, are no longer rendered as templates. This reaches exactly one class of input, text containing handlebars syntax that was previously expanded; text with no braces behaves identically. `go/samples/menu/s03.go` is the in-repo case: it built its chat preamble as a `{{#each menuData}}` template inside `WithMessages`, and moves to `WithMessagesTemplate`, which is what it always was.
- A content function returning `""` produces no message, where it previously produced an empty one that some providers reject.
- A `{{role}}` marker in `WithSystem` or `WithPrompt` is an error, including one that produced a single message and was previously overridden in silence. `.prompt` files are unaffected, since their bodies moved to the conversation slot.
- A prompt that declares messages no longer has execute-time history spliced in ahead of them, and places it with `{{history}}` or `HistoryFromContext` instead. A prompt that declares none is unchanged.
- The same rule reaches prompt-backed agents. A prompt with a user turn sees `[system, ...history, user_prompt]` rather than the conversation trailing it, and one that declares static `WithMessages` and never places the conversation no longer has one, which is what JS does too. `{{history}}` and `HistoryFromContext` work inside an agent for the first time.
- Loose targets receive schema-normalized values, so an integer field arrives as `int64` where an in-process caller previously saw `float64`. Normalization also drops null-valued keys, matching wire behavior.
- Via `base.ConvertTo`, three public lookups (`ai.ResumedValue`, `ai.OriginalInputAs`, `exp/tool.OriginalInput`) report `false` for an unrelated struct instead of `true` with a zeroed value.

## What this does not do

- **A template places history but cannot inspect it.** The messages handed to dotprompt are stand-ins carrying roles and the text of text parts; media, tool requests, and resources appear as empty text. The originals are restored after placement, so what reaches the model is intact, but a template that tries to read history sees text only.
- **Execute-time `WithDocs` replaces rather than accumulates.** When an execution supplies its own documents, `buildRequest` skips the prompt's `WithDocsFn` entirely rather than resolving and discarding it, since the retrieval it exists for costs a query.
- **No new session-state accessor.** A content function on an agent's prompt reaches the live session with `exp.SessionFromContext`, which hands back the concrete state with no round trip, and templates reach the same values as `{{@state}}` without naming a type at all.
- **The `{{role}}` check is a regex over template source.** Matching the rendered messages cannot work, since dotprompt's default role is `user`. A rendered role that disagrees with the slot is caught as a fallback, but a marker constructed at render time is not.

## Testing

Go 1.26.3, full `go test ./...` in `go/`: 41 packages pass, no failures. `-race` clean across `ai/...`, `genkit/...`, and `core/...` (10 packages), and `GOMAXPROCS=1 -count=3` clean on `ai/exp/...`. `TestAgentConformance`, which drives the shared `tests/specs/agent.yaml`, passes 10 consecutive runs.

`go/ai/prompt_fn_test.go` adds 26 test functions, `go/ai/exp/agent_history_test.go` 3, and `go/internal/base/convert_test.go` 3. They cover both call paths that used to panic, both ways the input conversion fails, and every placement form across multiple turns, asserting the model request and what the session is left holding. The copying tests fail when the clone is removed.

Two obsolete tests became coverage of the new rules: the text-versus-parts conflict test now asserts that the later option holds the slot and the earlier field is cleared, and the `WithMessages` normalization test became `TestConversationMergeSemantics`, which pins accumulation before and after a template and the template's own last-win.

`ai/exp` state-transform fixtures now build their errors with sentinels, and two additionally assert `errors.Is` on `AgentOutput.Error`. The runtime rebuilds a failure onto that field rather than passing the value through, and nothing pinned that the classification survived it.

### DIFF
--- a/go/ai/option.go
+++ b/go/ai/option.go
@@ -56,9 +56,9 @@ import (
 // options) document their own precedence.
 //
 // Applying options therefore never fails on a "set more than once" conflict.
-// The only failures are genuinely invalid arguments (for example a type that
-// WithInputType cannot turn into a schema), which panic at the call site where
-// the mistake is.
+// What does fail is a genuinely invalid argument (a type that WithInputType
+// cannot turn into a schema) or the refused combination above, and both panic
+// at the call site where the mistake is rather than deferring to the request.
 
 // PromptFn is a function that generates a prompt from a prompt's input.
 //
--- a/go/ai/option.go
+++ b/go/ai/option.go
@@ -349,16 +349,17 @@ func WithMessages(messages ...*Message) CommonGenOption {
 // template where it was first declared.
 //
 // Compiling the template needs a prompt, so this applies to [DefinePrompt]
-// only. Passing it to [Generate] or to [Prompt.Execute] is an error: use
-// [WithMessages] or [WithMessagesFn] there to supply the conversation the
-// prompt's own template will place.
-func WithMessagesTemplate(text string, args ...any) CommonGenOption {
+// only. That is why it returns a [PromptOption] rather than a
+// [CommonGenOption]: passing it to [Generate] or [Prompt.Execute] does not
+// compile. Use [WithMessages] or [WithMessagesFn] there to supply the
+// conversation the prompt's own template will place.
+func WithMessagesTemplate(text string, args ...any) PromptOption {
 	if len(args) > 0 {
 		// Assigning avoids a compile-time warning about non-constant text.
 		t := text
 		text = fmt.Sprintf(t, args...)
 	}
-	return &commonGenOptions{MessagesText: &text}
+	return &promptOptions{commonGenOptions: commonGenOptions{MessagesText: &text}}
 }
 
 // WithMessagesFn adds messages produced by fn at request time, placed between
--- a/go/ai/option.go
+++ b/go/ai/option.go
@@ -37,17 +37,15 @@ import (
 //   - Single-value options take the last one set. WithConfig, WithModel /
 //     WithModelName, WithSystem, WithPrompt, the output-schema options, and the
 //     like each fill one slot, so the final call wins and earlier ones are
-//     overwritten rather than rejected. A slot is shared by every option that
-//     fills it: WithSystem, WithSystemParts, WithSystemFn, and WithSystemPartsFn
-//     are four ways to write one system message, so the last of them wins and
-//     the others are discarded, not merged into it.
+//     overwritten rather than rejected. Every option that fills a slot shares
+//     it: WithSystem, WithSystemParts, WithSystemFn, and WithSystemPartsFn are
+//     four ways to write one system message, so they do not merge.
 //
 // One combination is refused rather than merged. WithMessagesTemplate lays out
-// the whole conversation, down to where {{history}} puts the caller's, so there
-// is no position relative to it that separately supplied messages could
-// meaningfully take. Giving it in the same DefinePrompt call as WithMessages or
-// WithMessagesFn panics, saying so. Repeating the template alone is fine and
-// takes the last one, as any other slot does.
+// the whole conversation, down to where {{history}} puts the caller's, so
+// separately supplied messages have no position relative to it. Passing it to
+// DefinePrompt alongside WithMessages or WithMessagesFn panics. Repeating the
+// template alone is an ordinary slot.
 //
 // A zero value does not fill a slot: WithMaxTurns(0), WithToolChoice(""), or
 // WithConfig(nil) is a no-op, so an earlier non-zero value cannot be un-set by
@@ -62,16 +60,14 @@ import (
 
 // PromptFn is a function that generates a prompt from a prompt's input.
 //
-// The input argument is untyped. Prefer passing a function with a concrete
-// input type to [WithPromptFn] or [WithSystemFn], which coerces the input for
-// you and works on every invocation path.
+// The input is untyped. [WithPromptFn] and [WithSystemFn] take a function with
+// a concrete input type and convert for you.
 type PromptFn = func(context.Context, any) (string, error)
 
 // MessagesFn is a function that generates messages from a prompt's input.
 //
-// The input argument is untyped. Prefer passing a function with a concrete
-// input type to [WithMessagesFn], which coerces the input for you and works on
-// every invocation path.
+// The input is untyped. [WithMessagesFn] takes a function with a concrete input
+// type and converts for you.
 type MessagesFn = func(context.Context, any) ([]*Message, error)
 
 // appendMessagesFn composes two message-producing functions so their outputs
@@ -103,10 +99,8 @@ func appendMessagesFn(existing, next MessagesFn) MessagesFn {
 	}
 }
 
-// appendDocsFn composes two document-producing functions so their outputs
-// concatenate in call order, backing the accumulate semantics of [WithDocsFn]
-// the same way [appendMessagesFn] does for messages. Either side may be nil,
-// in which case the other is returned unwrapped.
+// appendDocsFn is [appendMessagesFn] for documents, backing the accumulate
+// semantics of [WithDocsFn].
 func appendDocsFn(existing, next DocsFn) DocsFn {
 	if existing == nil {
 		return next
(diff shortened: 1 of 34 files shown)


go/ai/option.go
```diff
@@ -84,6 +99,87 @@ func appendMessagesFn(existing, next MessagesFn) MessagesFn {
...
+// the same type whichever way they were reached. A nil raw value yields the
+// zero value of In, which is what [Generate] passes.
+func coerceFn[In, Out any](fn func(context.Context, In) (Out, error)) func(context.Context, any) (Out, error) {
+	return func(ctx context.Context, raw any) (Out, error) {
+		in, err := base.ConvertToExact[In](raw)
+		if err != nil {
+			var zero Out
+			// base cannot classify this itself: the same failure is only a
+			// caller's bad input once a content function disagrees with it.
+			return zero, status.Errorf(ErrInputTypeMismatch, "content function input: %w", err)
```
pavelgj: Nit: this wraps `ErrInputTypeMismatch` with the same slot-agnostic message for all four content-function slots, so a prompt with several typed functions can't tell which one rejected the input. Could thread the option name through (like `renderSingleMessage` does) so the error names the offending slot. Not blocking.

pavelgj (approved):

PR #5883, opened 2026-08-01
fix(go/ai): make `ModelResponse.History` nil-safe and stop aliasing request messages

apascal07:
Fixes three panic paths and a slice-aliasing defect in `ModelResponse.History`, found while verifying #4683.

The nil guard dereferenced the receiver it was checking: `if mr == nil` fell through to `return mr.Request.Messages`. `Request` is nil-able independently of the receiver and is nil in shipped code, since the Veo background model stashes its request in `Metadata["inputRequest"]` and never assigns it, so `op.Output.History()` panicked on any Veo operation. A response with neither field set panicked back in the guard.

Separately, `append(mr.Request.Messages, mr.Message)` reused spare capacity in `Request.Messages` when it had any, writing the response message into a slot the caller still owned. The result aliased anything else pointing into that array, and two calls on one response returned slices sharing the element each had just written.

`History` now returns early on a nil receiver, treats a nil `Request` as no prior messages, and builds the result with `slices.Clone`, which also preserves the old nil return for the empty case.

`TestModelResponseHistory` covers seven cases: five panic or fail on `main`, two pin behavior that already worked. Go 1.26.3, full `go test ./...` in `go/`: 40 packages pass, 0 fail.

Populating `Request` on every model path is a plugin-contract change and belongs with the `compat_oai` streaming gap in #5239.

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -1131,12 +1131,21 @@ func (mr *ModelResponse) Text() string {
 }
 
 // History returns messages from the request combined with the response message
-// to represent the conversation history.
+// to represent the conversation history. The result is always freshly
+// allocated, so callers may retain or append to it without disturbing
+// Request.Messages.
 func (mr *ModelResponse) History() []*Message {
-	if mr == nil || mr.Message == nil {
-		return mr.Request.Messages
+	if mr == nil {
+		return nil
 	}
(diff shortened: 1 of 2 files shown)


huangjeff5 (approved):
LGTM! Excellent catch on the slice aliasing bug and the nil-safety edge cases.

PR #5910, opened 2026-08-03
fix(go): complete partial JSON by nesting order, not by brace counts

apascal07:
Rewrites `CompleteJSON` as a scanner. It completed truncated documents by counting braces and brackets independently, which produced invalid JSON for the most common streaming shape and for four other classes of truncated tail. `parseJSON` in `go/ai/format.go` discards the resulting parse error and returns `nil`, so a bad completion never surfaces as an error: the chunk is silently dropped. The streaming example in `go/README.md` reads `result.Chunk.Ingredients[0].Name`, an object holding an array of objects, which yielded no chunk at all until the array closed.

The replacement keeps a stack of open containers and a rollback offset, the last position at which the document could legally be closed. Closers pop LIFO, and a tail that cannot be finished without inventing data rewinds to the rollback offset rather than being closed in place. The scanner also tracks what the grammar allows at each position, so input that is malformed rather than truncated rewinds the same way. The result is always valid JSON.

## Bugs fixed

**Containers were closed by type instead of by nesting order.** All open brackets were appended before all open braces, so `[{"a": 1` became `[{"a": 1]}`. Counters cannot recover the order the containers were opened in. This hit every array of objects and every object holding one.

**A dangling escape swallowed the closing quote.** A chunk ending mid-escape had a quote appended to it, so `{"a": "x\` became `{"a": "x\"}` and the string stayed open. Half-arrived `\uXXXX` sequences failed the same way.

**A key whose value had not arrived became a bare key.** `{"a": 1, "b"` became `{"a": 1, "b"}`, and `{"a": 1, "b":` became `{"a": 1, "b":}`. Both now drop the unfinished member.

**Truncated numbers and keywords were emitted raw**, giving `{"a": tr}`, `{"a": 1.}`, and `{"a": -}`. Keywords now complete: no two JSON keywords share a first letter, so `tr` can only have been going to become `true`. Numbers drop the member instead, since `1.` gives no way to know what follows.

**A multi-byte rune split across a chunk boundary decoded to `U+FFFD`.** Those trailing bytes are legal inside a JSON string, so this one passes `json.Valid` and is invisible to a validity check. It shows up only as a replacement character in the decoded value.

## Malformed input rewinds rather than completing in place

Truncation is only half of what reaches this function. A document can also be structurally wrong, and completing that in place produces invalid JSON just as surely: `{"a": 1, {` became `{"a": 1, {}}`, `[1, 2,]` became itself, and `{:}`, `{"a"}`, `{,}`, and `[1 2]` likewise. Tracking the grammar turns all of these into a rewind to the last well-formed offset, which is what `partial-json` does with them.

Raw control characters and unrecognized escapes inside strings are the same class. A literal newline is illegal inside a JSON string and a closing quote further along cannot make the preceding bytes legal, so the string ends at the offending byte and `{"a": "line1<newline>line2", "b": 2}` completes to `{"a": "line1"}`.

## Verification

Go 1.26.3, `go test ./...` in `go/`: 41 packages with tests pass, no failures.

`FuzzCompleteJSONArbitrary` takes its input as given and asserts only that the result parses: 15.7M executions, no failures. `FuzzCompleteJSON` holds the input to being a genuine prefix of a valid document, which is what lets the corpus tests check content rather than mere parseability: 7.8M executions, no failures. The second found one real problem, in the test rather than the code, since `1E0007000` is valid JSON that overflows `float64` and the assertion had to be `json.Valid` rather than unmarshalability. That input is kept as a seed.

Two property tests carry more weight than the table-driven cases: every byte prefix of a corpus document completes to valid JSON, and every completed prefix is a *snapshot* of the finished document, so no chunk reports something the finished document goes on to contradict. The second is what catches the `U+FFFD` case.

Behavior was diffed against `partial-json`, the package `js/ai/src/extract.ts` uses, at all 259 byte-cut points of five documents: no differences once key order is canonicalized. The keyword completion above was the one divergence before this change.

### DIFF
--- a/go/ai/format_test.go
+++ b/go/ai/format_test.go
@@ -249,6 +249,36 @@ func TestJSONFormatter(t *testing.T) {
 		}
 	})
 
+	t.Run("ParseChunk streams an array of objects", func(t *testing.T) {
+		// Completing the containers by type rather than by nesting order turned
+		// `[{"name": "Jo` into `[{"name": "Jo"]}`, which failed to parse and
+		// left every chunk of an array of objects empty until the array closed.
+		handler, _ := jsonFormatter{}.Handler(nil)
+		sfh := handler.(StreamingFormatHandler)
+
+		chunks := []string{`[{"name": "Jo`, `hn"}, {"nam`, `e": "Jane"}]`}
+		want := []any{
+			[]any{map[string]any{"name": "Jo"}},
+			// The second object has been opened but holds nothing yet: its key
+			// is still arriving, so there is no member to report.
+			[]any{map[string]any{"name": "John"}, map[string]any{}},
+			[]any{map[string]any{"name": "John"}, map[string]any{"name": "Jane"}},
+		}
+
+		for i, chunk := range chunks {
+			got, err := sfh.ParseChunk(&ModelResponseChunk{
+				Content: []*Part{NewTextPart(chunk)},
+				Index:   0,
+			})
+			if err != nil {
+				t.Fatalf("ParseChunk(%q) error = %v", chunk, err)
+			}
+			if diff := cmp.Diff(want[i], got); diff != "" {
+				t.Errorf("ParseChunk(%q) mismatch (-want +got):\n%s", chunk, diff)
+			}
+		}
+	})
+
 	t.Run("ParseMessage validates JSON and creates JSON part", func(t *testing.T) {
 		handler, _ := jsonFormatter{}.Handler(schema)
 
--- a/go/internal/base/extract.go
+++ b/go/internal/base/extract.go
@@ -17,6 +17,7 @@ package base
 import (
 	"encoding/json"
 	"strings"
+	"unicode/utf8"
 )
 
 // ExtractJSON extracts JSON from string with lenient parsing rules.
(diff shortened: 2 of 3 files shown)


huangjeff5 (commented):
Hey @apascal07! Great rewrite of `CompleteJSON` — moving to a stack-based scanner is a huge improvement over brace counting.

While testing edge cases, I noticed a couple of input shapes and wanted to check if this behavior is intended or if `CompleteJSON` should cut back in these cases:

### 1. Opening container where an object key is expected
When an object is expecting a key (`expectKey == true`), encountering `{` or `[` currently pushes a nested container:
- `{"a": 1, {` ➔ completes to `{"a": 1, {}}` (fails `json.Valid`)
- `{"a": 1, [` ➔ completes to `{"a": 1, []}` (fails `json.Valid`)
- `{{` ➔ completes to `{{}}` (fails `json.Valid`)

*(Should this cut back to `s[:safe]` like numbers/keywords do when `expectKey` is true?)*

### 2. Closed containers with trailing commas
When a container is explicitly closed with `]` or `}` but carries a trailing comma inside:
- `[1, 2,]` ➔ completes to `[1, 2,]` (fails `json.Valid`)
- `{"a": 1,}` ➔ completes to `{"a": 1,}` (fails `json.Valid`)

*(Is this intended as malformed input that `CompleteJSON` leaves alone, or should trailing commas before `]` / `}` be trimmed?)*

pavelgj (approved):

apascal07:
Both should cut back, and there were more of them than the two: `{:}`, `{"a"}`, `{,}`, and `[1 2]` were also completing to invalid JSON. Same cause, which is that the scanner only checked what may legally come next in some token positions and not others.

`e38387686` tracks what the grammar allows at every position and rewinds to the last well-formed offset on the first violation:

| input | before | now |
|---|---|---|
| `{"a": 1, {` | `{"a": 1, {}}` | `{"a": 1}` |
| `[1, 2,]` | `[1, 2,]` | `[1, 2]` |
| `{"a"}` | `{"a"}` | `{}` |
| `[1 2]` | `[1 2]` | `[1]` |

That matches `partial-json` on all of them except `[1 2]`, where JS drops the whole container to `[]` and this keeps the `1`.

Raw control characters and unrecognized escapes inside strings came with it, since they were the same class of invalid output. A literal newline is illegal inside a JSON string and a closing quote further along cannot make the preceding bytes legal, so the string now ends at the offending byte: `{"a": "line1<newline>line2", "b": 2}` gives `{"a": "line1"}`.

The contract is now that the result is always valid JSON whatever the input, asserted directly by a new `FuzzCompleteJSONArbitrary` (15.7M executions, clean). The existing fuzz target only cut *valid* documents at arbitrary offsets, which is why none of this surfaced there. Parity with `partial-json` is unchanged at 259/259 cut points.

PR #5911, opened 2026-08-03
fix(go/ai/exp): scope resume validation to the pending model response

apascal07:
Fixes a resume-validation hole in `ValidateResumeAgainstHistory`, mirroring JS [#5832](https://github.com/genkit-ai/genkit/pull/5832) on the ordering half and going further on scope.

The function flattened the tool requests from every model message in session history into one slice, then matched each resume entry against the first hit. `ensureToolRequestRefs` only fills in empty refs, and only within a single response, so a provider-supplied ref recurs across turns and the first hit is the oldest occurrence rather than the pending one. A client restarting the interrupted call had its input compared against a settled turn's request: a faithful restart was rejected as "modified inputs", while a restart carrying an input forged from that older call validated clean.

Ordering alone is not the whole fix. `handleResumeOption` resolves exactly the tool requests of `messages[len-1]` and requires it to be a model message carrying at least one, so an entry aimed at an earlier turn resolves to nothing downstream. Validation now searches only the last model message, newest-first within it, and reports an entry naming an earlier turn as stale rather than as a tool request that was never made. Trailing non-model messages are skipped rather than read as nothing-pending, since `AgentInput` permits a turn message alongside a resume payload and that combination must keep failing on generate's own `FAILED_PRECONDITION`.

## Behavior change worth noting

A client that resends an accumulated resume payload each turn, valid entries for the pending call plus settled ones from earlier turns, now gets `INVALID_ARGUMENT` where the stale entries were previously accepted and silently ignored downstream. JS still searches the whole history, so the two diverge here.

## Verification

Go 1.26.3, `go test ./...` in `go/`: 41 packages pass, no failures. Conformance harness 40/40, including the forged-restart and unknown-tool specs in `tests/specs/agent.yaml`. `GOMAXPROCS=1 go test ./ai/exp/ -race -count=2` clean.

`TestValidateResumeAgainstHistory_RecurringRefs` pins both original symptoms and every subtest fails on `main`. `TestValidateResumeAgainstHistory` needed its fixture retargeted, since it asserted cross-turn matching as correct.

### DIFF
--- a/go/ai/exp/agent.go
+++ b/go/ai/exp/agent.go
@@ -2498,12 +2498,14 @@ func validateUserMessage(m *ai.Message) error {
 }
 
 // ValidateResumeAgainstHistory ensures every restart and respond entry on a
-// resume payload references a tool request the model actually issued, so a
+// resume payload references a tool request that is actually pending, so a
 // caller cannot drive a tool the model never asked for and interrupted on.
 // For restart entries it additionally checks the input is unchanged from the
 // original request, preventing a client from forging tool inputs on the
 // interrupted call. The whole history is searched (every model message), not
-// just the last turn. On a violation it returns an INVALID_ARGUMENT error.
+// just the last turn, and newest-first: when a tool name + ref recurs across
+// turns, the entry is validated against the most recent one rather than a
+// stale earlier call. On a violation it returns an INVALID_ARGUMENT error.
 //
 // The prompt-backed agent loop ([DefineAgent]) calls this automatically. A
 // custom agent ([DefineCustomAgent]) that accepts an [AgentInput.Resume] from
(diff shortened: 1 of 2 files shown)


huangjeff5 (approved):
LGTM! Scoping resume validation to the pending model response cleanly fixes recurring ref collisions across turns.

go/ai/exp/agent.go
```diff
@@ -2498,12 +2498,21 @@ func validateUserMessage(m *ai.Message) error {
...
-// resume payload references a tool request the model actually issued, so a
+// resume payload references a tool request that is actually pending, so a
 // caller cannot drive a tool the model never asked for and interrupted on.
 // For restart entries it additionally checks the input is unchanged from the
 // original request, preventing a client from forging tool inputs on the
-// interrupted call. The whole history is searched (every model message), not
-// just the last turn. On a violation it returns an INVALID_ARGUMENT error.
+// interrupted call.
+//
+// Only the most recent model message in history is searched, and within it
```
pavelgj, line 2507: this is different from JS, but I think I'm OK with this... it's actually a requirement for anthropic that tool request is followed by a tool response, so this tracks. I'm OK with Go being stricter than JS... we can move JS in this direction too.

pavelgj (approved):

go/ai/exp/agent.go
apascal07: Ack, this was intentional.

PR #5921, opened 2026-08-04
feat(ollama): detect model capabilities dynamically

adesinah:
This PR introduces detection of model capabilities dynamically via /api/show. 
Here is the initial contributor's PR #5114

Checklist (if applicable):

### DIFF
--- a/go/plugins/ollama/ollama.go
+++ b/go/plugins/ollama/ollama.go
@@ -164,8 +164,16 @@ func (o *Ollama) listLocalModels(ctx context.Context) ([]ollamaLocalModel, error
 }
 
 func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.ModelOptions) ai.Model {
-	// Detect capabilities before acquiring the lock to avoid holding it
-	// during HTTP I/O.
+	// Check the init guard first under a brief lock — before any I/O — so
+	// that a forgotten Init() panics immediately rather than after a timeout.
+	o.mu.Lock()
+	if !o.initted {
+		o.mu.Unlock()
+		panic("ollama.Init not called")
+	}
+	o.mu.Unlock()
+
+	// Detect capabilities outside the lock to avoid holding it during HTTP I/O.
 	var modelOpts ai.ModelOptions
 	if opts != nil {
 		modelOpts = *opts
@@ -184,11 +192,9 @@ func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.M
 		}
 	}
 
+	// Re-acquire lock for the registration step.
 	o.mu.Lock()
 	defer o.mu.Unlock()
-	if !o.initted {
-		panic("ollama.Init not called")
-	}
 	meta := &ai.ModelOptions{
 		Label:        "Ollama - " + model.Name,
 		Supports:     modelOpts.Supports,
--- a/go/plugins/ollama/ollama.go
+++ b/go/plugins/ollama/ollama.go
@@ -164,13 +164,9 @@ func (o *Ollama) listLocalModels(ctx context.Context) ([]ollamaLocalModel, error
 }
 
 func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.ModelOptions) ai.Model {
-	o.mu.Lock()
-	defer o.mu.Unlock()
-	if !o.initted {
-		panic("ollama.Init not called")
-	}
+	// Detect capabilities before acquiring the lock to avoid holding it
+	// during HTTP I/O.
 	var modelOpts ai.ModelOptions
-
 	if opts != nil {
 		modelOpts = *opts
 	} else {
@@ -178,12 +174,20 @@ func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.M
 		// /api/show. This replaces the hardcoded allowlist approach so that
 		// newly released models (e.g. gemma4) work automatically without
 		// code changes. Falls back to the static list for older servers.
-		caps := o.getModelCapabilities(context.Background(), model.Name)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(o.Timeout)*time.Second)
+		defer cancel()
+		caps := o.getModelCapabilities(ctx, model.Name)
 		modelOpts = ai.ModelOptions{
 			Label:    model.Name,
 			Supports: modelSupportsFromCapabilities(caps, model.Name),
 			Versions: []string{},
 		}
+	}
+
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if !o.initted {
+		panic("ollama.Init not called")
 	}
 	meta := &ai.ModelOptions{
 		Label:        "Ollama - " + model.Name,
@@ -425,6 +429,10 @@ func (o *Ollama) ListActions(ctx context.Context) []api.ActionDesc {
 		if strings.Contains(name, "embed") {
 			continue
 		}
+		// Check for context cancellation before each potentially slow HTTP call.
+		if ctx.Err() != nil {
+			break
+		}
 		// Query each model's actual capabilities from the Ollama server.
 		caps := o.getModelCapabilities(ctx, name)
 		supports := modelSupportsFromCapabilities(caps, name)
@@ -442,7 +450,9 @@ func (o *Ollama) ResolveAction(atype api.ActionType, name string) api.Action {
 		return nil
 	}
 	// Query the model's actual capabilities from the Ollama server.
-	caps := o.getModelCapabilities(context.Background(), name)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(o.Timeout)*time.Second)
+	defer cancel()
+	caps := o.getModelCapabilities(ctx, name)
 	supports := modelSupportsFromCapabilities(caps, name)
 	model := o.newModel(name, ai.ModelOptions{Supports: supports})
 	if action, ok := model.(api.Action); ok {
--- a/go/plugins/ollama/ollama.go
+++ b/go/plugins/ollama/ollama.go
@@ -80,10 +80,64 @@ type ollamaTagsResponse struct {
 	Models []ollamaLocalModel `json:"models"`
 }
 
-// ollamaLocalModel represents a locally available Ollama model.
+// ollamaLocalModel represents a locally available Ollama model from /api/tags.
 type ollamaLocalModel struct {
 	Name  string `json:"name"`
 	Model string `json:"model"`
+}
+
+// ollamaShowResponse represents the response from POST /api/show.
+type ollamaShowResponse struct {
+	Capabilities []string `json:"capabilities"`
+}
+
+// getModelCapabilities calls POST /api/show to retrieve the model's capabilities.
+// Returns nil if the endpoint is unavailable or the model doesn't report capabilities.
+func (o *Ollama) getModelCapabilities(ctx context.Context, modelName string) []string {
+	body, err := json.Marshal(map[string]string{"model": modelName})
+	if err != nil {
+		return nil
+	}
+	req, err := http.NewRequestWithContext(ctx, "POST", o.ServerAddress+"/api/show", bytes.NewReader(body))
+	if err != nil {
+		return nil
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := o.client.Do(req)
+	if err != nil {
+		return nil
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil
+	}
+	var showResp ollamaShowResponse
+	if err := json.NewDecoder(resp.Body).Decode(&showResp); err != nil {
+		return nil
+	}
+	return showResp.Capabilities
+}
+
+// modelSupportsFromCapabilities derives ModelSupports from capabilities reported
+// by the Ollama /api/show endpoint. Falls back to the static allowlists when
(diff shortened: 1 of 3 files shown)


go/plugins/ollama/ollama.go
```diff
@@ -386,7 +487,15 @@ func (o *Ollama) ResolveAction(atype api.ActionType, name string) api.Action {
 	if atype != api.ActionTypeModel {
 		return nil
 	}
-	model := o.newModel(name, ai.ModelOptions{Supports: &defaultOllamaSupports})
+	// Query the model's actual capabilities from the Ollama server.
+	ctx, cancel := o.modelCapabilitiesContext(context.Background())
```
apascal07: `registry.ResolveAction` holds the global `resolveMu` across this plugin call (`go/internal/registry/registry.go:231-254`), so this HTTP round trip serializes dynamic resolution for every plugin, not just ollama. Against a server that accepts the connection but never responds, I measured `ResolveAction` returning after the full 5s ceiling with all other plugins' resolution blocked behind it. This path did no I/O before.

`context.Background()` also means a cancelled caller cannot abort it.

go/plugins/ollama/ollama.go
```diff
@@ -373,7 +462,19 @@ func (o *Ollama) ListActions(ctx context.Context) []api.ActionDesc {
...
 			continue
 		}
-		model := o.newModel(name, ai.ModelOptions{Supports: &defaultOllamaSupports})
+		// Check for context cancellation before each potentially slow HTTP call.
+		if ctx.Err() != nil {
+			break
+		}
+		// Query each model's actual capabilities from the Ollama server.
+		capabilityCtx, cancel := o.modelCapabilitiesContext(ctx)
+		caps, detected := o.getModelCapabilities(capabilityCtx, name)
```
apascal07: This is one `/api/show` per model, serially, on a path the Dev UI polls (`GET /api/actions` via `listResolvableActions`), with no caching between polls. Measured with 20 local models at 50ms latency: 20 sequential requests, 1.06s wall clock, repaid on every poll.

`/api/tags` returns a `digest` per model that would key a cache. `ollamaLocalModel` currently decodes only `Name` and `Model`.

go/plugins/ollama/ollama.go
```diff
@@ -80,12 +82,85 @@ type ollamaTagsResponse struct {
...
 
+// ollamaShowResponse represents the response from POST /api/show.
+type ollamaShowResponse struct {
+	Capabilities *[]string `json:"capabilities"`
+}
+
+// getModelCapabilities calls POST /api/show to retrieve the model's capabilities.
+// The boolean reports whether the server returned a capabilities field, allowing
+// callers to distinguish an explicitly empty list from an unavailable endpoint.
+func (o *Ollama) getModelCapabilities(ctx context.Context, modelName string) ([]string, bool) {
```
apascal07: All six failure paths here return `nil, false` without logging: marshal, request construction, transport, non-200, decode, and nil `Capabilities`. A model that should report `Tools: true` and silently doesn't leaves the user nothing to debug from, and the fallback makes the failure look like a successful detection. `ListActions` logs its `/api/tags` failure via `slog.Error` immediately below.

go/plugins/ollama/ollama.go
```diff
@@ -110,29 +185,43 @@ func (o *Ollama) listLocalModels(ctx context.Context) ([]ollamaLocalModel, error
...
 	if opts != nil {
 		modelOpts = *opts
 	} else {
-		// Check if the model supports tools (must be a chat model and in the supported list)
-		supportsTools := model.Type == "chat" && slices.Contains(toolSupportedModels, model.Name)
+		// Query the Ollama server for the model's actual capabilities via
+		// /api/show. This replaces the hardcoded allowlist approach so that
+		// newly released models (e.g. gemma4) work automatically without
+		// code changes. Falls back to the static list for older servers.
+		ctx, cancel := o.modelCapabilitiesContext(context.Background())
```
apascal07: `DefineModel` was pure registration and now blocks on a network round trip per call. Against a server that accepts but does not respond that is 5s per model, so an app defining ten models stalls for 50s inside `genkit.Init`.

Passing non-nil `opts` skips detection entirely, which is the escape hatch, but it is neither documented nor tested.

go/plugins/ollama/ollama.go
```diff
@@ -373,7 +462,19 @@ func (o *Ollama) ListActions(ctx context.Context) []api.ActionDesc {
...
-		model := o.newModel(name, ai.ModelOptions{Supports: &defaultOllamaSupports})
+		// Check for context cancellation before each potentially slow HTTP call.
+		if ctx.Err() != nil {
+			break
+		}
+		// Query each model's actual capabilities from the Ollama server.
+		capabilityCtx, cancel := o.modelCapabilitiesContext(ctx)
+		caps, detected := o.getModelCapabilities(capabilityCtx, name)
+		cancel()
+		supports := &defaultOllamaSupports
```
apascal07: `ListActions` and `ResolveAction` fall back to `defaultOllamaSupports` (everything true) while `DefineModel` falls back to `modelSupportsFromStaticLists` (mostly false), for the same "server did not report capabilities" condition. Preserving each path's prior behavior is defensible, but the divergence is invisible at both call sites, and now that the static path matches base names it is the better fallback for both.

go/plugins/ollama/ollama.go
```diff
@@ -80,12 +82,85 @@ type ollamaTagsResponse struct {
...
+		SystemRole: true,
+		Tools:      slices.Contains(caps, "tools"),
+		Media:      slices.Contains(caps, "vision") || slices.Contains(caps, "audio"),
+	}
+}
+
+// modelSupportsFromStaticLists preserves the fallback used by explicitly
+// defined models when the server does not report capabilities.
+func modelSupportsFromStaticLists(modelName string) *ai.ModelSupports {
+	baseName, _, _ := strings.Cut(modelName, ":")
```
apascal07: Matching the base name before the tag changes results even when `/api/show` is unavailable: `mistral:7b` and `qwen2.5:7b` go from `Tools: false` to `true`, `bakllava:7b` from `Media: false` to `true`. The old code compared `model.Name` against the lists verbatim.

This is an improvement, but it is a behavior change on servers with no capability endpoint, which the title and description don't cover.

go/plugins/ollama/ollama.go
```diff
@@ -80,12 +82,85 @@ type ollamaTagsResponse struct {
...
+}
+
+// modelSupportsFromCapabilities derives ModelSupports from capabilities reported
+// by the Ollama /api/show endpoint.
+func modelSupportsFromCapabilities(caps []string) *ai.ModelSupports {
+	return &ai.ModelSupports{
+		Multiturn:  true,
+		SystemRole: true,
+		Tools:      slices.Contains(caps, "tools"),
+		Media:      slices.Contains(caps, "vision") || slices.Contains(caps, "audio"),
```
apascal07: `concatImages` filters parts to `image/*` and drops the rest, so a model reporting `audio` but not `vision` advertises media input that the plugin then discards silently. Either drop `audio` from this mapping or carry audio parts through `concatImages`.

go/plugins/ollama/ollama.go
```diff
@@ -373,7 +462,19 @@ func (o *Ollama) ListActions(ctx context.Context) []api.ActionDesc {
 		if strings.Contains(name, "embed") {
 			continue
 		}
-		model := o.newModel(name, ai.ModelOptions{Supports: &defaultOllamaSupports})
+		// Check for context cancellation before each potentially slow HTTP call.
+		if ctx.Err() != nil {
```
apascal07: On cancellation this returns the models processed so far with no error and no log, so a truncated list is indistinguishable from a complete one. The Dev UI would render a partial model list as if it were the full set.

go/plugins/ollama/ollama.go
```diff
@@ -80,12 +82,85 @@ type ollamaTagsResponse struct {
...
+
+// getModelCapabilities calls POST /api/show to retrieve the model's capabilities.
+// The boolean reports whether the server returned a capabilities field, allowing
+// callers to distinguish an explicitly empty list from an unavailable endpoint.
+func (o *Ollama) getModelCapabilities(ctx context.Context, modelName string) ([]string, bool) {
+	body, err := json.Marshal(map[string]string{"model": modelName})
+	if err != nil {
+		return nil, false
+	}
+	req, err := http.NewRequestWithContext(ctx, "POST", o.ServerAddress+"/api/show", bytes.NewReader(body))
```
apascal07: nit: a `ServerAddress` with a trailing slash produces `//api/show`. The live-test helper added in this PR does `strings.TrimRight(*serverAddress, "/")`; production code doesn't.

apascal07 (changes requested):

adesinah:
@apascal07 thanks for the review. I have addressed the comments.

go/plugins/ollama/ollama.go
```diff
@@ -386,7 +554,12 @@ func (o *Ollama) ResolveAction(atype api.ActionType, name string) api.Action {
 	if atype != api.ActionTypeModel {
 		return nil
 	}
-	model := o.newModel(name, ai.ModelOptions{Supports: &defaultOllamaSupports})
+	supports := &defaultOllamaSupports
+	if cached, ok := o.cachedModelCapabilities(name, ""); ok {
```
apascal07, line 576: `/api/tags` reports tagged names, so ListActions caches under e.g. `moondream:latest`. This lookup (and the one in DefineModel) uses the caller-supplied name, which is typically untagged (`genkit.LookupModel(g, "ollama/moondream")`), so the exact-match lookup always misses and resolution falls back to the permissive defaults. Repro: a tags response with `{Name: "moondream:latest"}` and `/api/show` reporting `["completion","vision"]`, then `ResolveAction(api.ActionTypeModel, "moondream")` reports `tools=true`.

Two consequences:
- Detected capabilities only surface through ListActions; LookupModel and DefineModel behave exactly as before this PR.
- TestLiveDynamicDiscovery fails against a capabilities-reporting server: the resolved-action assertion expects `tools=false` for moondream but gets the fallback `true`. `sameLiveModelName` papers over this same mismatch on the test side.

Suggest normalizing the cache key on both write and read with `strings.TrimSuffix(name, ":latest")`, which matches how Ollama resolves untagged references, plus a unit test where `/api/tags` returns a `:latest` name and resolution uses the bare name.

go/plugins/ollama/ollama.go
```diff
@@ -80,15 +83,91 @@ type ollamaTagsResponse struct {
...
-	Name  string `json:"name"`
-	Model string `json:"model"`
+	Name   string `json:"name"`
+	Model  string `json:"model"`
+	Digest string `json:"digest"`
+}
+
+// ollamaShowResponse represents the response from POST /api/show.
+type ollamaShowResponse struct {
+	Capabilities *[]string `json:"capabilities"`
```
apascal07: nit: a plain `[]string` behaves identically here: absent and `null` both decode to nil, `[]` to a non-nil empty slice. The pointer only adds a dereference.

go/plugins/ollama/ollama.go
```diff
@@ -110,29 +189,39 @@ func (o *Ollama) listLocalModels(ctx context.Context) ([]ollamaLocalModel, error
 }
 
 func (o *Ollama) DefineModel(g *genkit.Genkit, model ModelDefinition, opts *ai.ModelOptions) ai.Model {
+	// Check the init guard first under a brief lock — before any I/O — so
```
apascal07: The stated reason no longer holds: this function does no I/O. The reason the lock cannot be held across the block below is that `cachedModelCapabilities` acquires `o.mu`, so a single lock/defer would self-deadlock. Worth stating that here so a future cleanup does not reinstate it.

go/plugins/ollama/ollama.go
```diff
@@ -366,14 +496,52 @@ func (o *Ollama) ListActions(ctx context.Context) []api.ActionDesc {
...
+	supports := make([]*ai.ModelSupports, len(filtered))
+	var wg sync.WaitGroup
+	for i, m := range filtered {
+		if cached, ok := o.cachedModelCapabilities(m.Name, m.Digest); ok {
+			cachedSupports := cached.supports
+			supports[i] = &cachedSupports
+			continue
+		}
+		wg.Add(1)
+		go func(i int, m ollamaLocalModel) {
```
apascal07, line 534: nit: one goroutine per uncached model with no bound, so the first discovery fires every `/api/show` request at once. Harmless against a local server, but a shared or remote Ollama instance with a large model library gets the full burst. A small concurrency cap would bound it.

adesinah:
@apascal07 I have addressed the comments.

go/plugins/ollama/ollama.go
```diff
@@ -366,14 +505,61 @@ func (o *Ollama) ListActions(ctx context.Context) []api.ActionDesc {
 		return nil
 	}
 
-	var actions []api.ActionDesc
+	filtered := make([]ollamaLocalModel, 0, len(models))
 	for _, m := range models {
-		name := m.Name
 		// Filter out embedding models (following JS: !m.model.includes('embed'))
-		if strings.Contains(name, "embed") {
+		if strings.Contains(m.Name, "embed") {
```
apascal07, line 511: `/api/show` reports an `embedding` capability, which is a more reliable signal than this substring match: it catches embedders whose name lacks `embed`, and it avoids filtering out a generative model whose name happens to contain it.

Not for this PR. The substring match is deliberately mirroring JS (`!m.model.includes('embed')`), and switching would also invert the order here, since filtering on capabilities means probing every model before filtering rather than after. Noting it because the capability data is now available in this function.

apascal07 (approved):

PR #5928, opened 2026-08-04
fix(mcp): retrieve MCP server output schema

adesinah:
This PR introduces retrieval of MCP server output schema. 
Here is the initial contributor's PR #3751

Checklist (if applicable):


go/ai/tools.go
```diff
@@ -379,6 +379,46 @@ func NewToolWithInputSchema[Out any](name, description string, inputSchema map[s
...
+	if schema.ValidationOutput != nil {
+		validationOutput = schema.ValidationOutput
+	}
+	validatedFn := func(ctx *ToolContext, input In) (Out, error) {
+		output, err := fn(ctx, input)
+		if err != nil {
+			return output, err
+		}
+		if err := base.ValidateValue(output, validationOutput); err != nil {
+			return base.Zero[Out](), core.NewError(core.INTERNAL, "invalid output from tool %q: %v", name, err)
```
cabljac: nit: should this use the status package from #5829, like action.go:333?

```go
status.Errorf(status.ErrInvalidOutput, "invalid output from action %q: %v", a.desc.Key, err)
```

Not sure if the intent is for tool output validation to classify as INTERNAL or ErrInvalidOutput, happy either way. Non-blocking.

go/plugins/mcp/tools_test.go
```diff
@@ -161,3 +164,153 @@ func TestPrepareToolArguments(t *testing.T) {
...
+			server: &ServerRef{Error: "protocol version negotiation failed"},
+		}
+		_, _, err := client.fetchToolsPage(t.Context(), "")
+		if err == nil || !strings.Contains(err.Error(), "protocol version negotiation failed") {
+			t.Fatalf("fetchToolsPage() error = %v, want initialization error", err)
+		}
+	})
+}
+
+func ParseMapToStruct[T any](t *testing.T, v any) T {
```
cabljac: nit: other helpers in this file (asMap, eqStr, reqContains) are unexported, parseMapToStruct for consistency.

cabljac (approved):
lgtm, two nits

go/ai/tools.go
adesinah: yes good point. I think this would be better as a separate PR.

apascal07:
Let's either hold off on this until https://github.com/genkit-ai/genkit/pull/5862 is merged or rebase this on top of it now. We don't want to introduce `NewToolWithSchema`, we already have coverage for that in the API redesign.

adesinah:
> Let's either hold off on this until #5862 is merged or rebase this on top of it now. We don't want to introduce `NewToolWithSchema`, we already have coverage for that in the API redesign.

sounds good.

adesinah:
@apascal07 this is good for re-review now.

go/plugins/mcp/tools.go
```diff
@@ -84,22 +116,45 @@ func (c *GenkitMCPClient) createTool(mcpTool mcp.Tool) (ai.Tool, error) {
...
-	} else {
-		tool = ai.NewTool(
-			namespacedToolName,
-			mcpTool.Description,
-			toolFunc,
-		)
+		opts = append(opts, ai.WithInputSchema(inputSchema))
+	}
+	if hasOutputSchema {
+		opts = append(opts, ai.WithOutputSchema(outputSchema))
```
apascal07, line 141: The advertised schema describes `structuredContent`, but the tool func returns the whole `CallToolResult`. Against a server declaring `{"type":"object","properties":{"Weather":{"type":"string"}},"required":["Weather"]}`, `Definition().OutputSchema` reports exactly that while `RunRaw` returns:

```json
{"content":[{"text":"Sunny","type":"text"}],"structuredContent":{"Weather":"Sunny"}}
```

That schema is what reaches the model, and it is what `handleResumedToolRequest` validates a resumed tool response against (`go/ai/generate.go:1697`), so a resume carrying the real tool output fails validation.

`ai.NewMultipartTool` fits here: `WithOutputSchema` on a multipart tool is documented to describe the envelope's output field, so putting `structuredContent` in `Output` and the content parts in `Content` makes the advertised schema true. That is also what the JS plugin does in `processMultipartResult`.

go/plugins/mcp/tools.go
```diff
@@ -84,22 +116,45 @@ func (c *GenkitMCPClient) createTool(mcpTool mcp.Tool) (ai.Tool, error) {
...
+	if !ok || result == nil {
+		return fmt.Errorf("expected *mcp.CallToolResult, got %T", output)
+	}
+	// MCP tool-level errors are model-visible results, not successful structured
+	// output. They commonly omit structuredContent and must not be schema-validated.
+	if result.IsError {
+		return nil
 	}
-	return tool, nil
+	return base.ValidateValue(result.StructuredContent, outputSchema)
```
apascal07, line 155: Servers that declare an `outputSchema` but return text-only content now fail the whole generate turn. `mcp.NewToolResultText` produces exactly that shape, and `ValidateValue` marshals the nil `StructuredContent` to `null`:

```
error calling tool test_getWeather: invalid output from tool "test_getWeather":
data did not match expected schema: - (root): Invalid type. Expected: object, given: null
```

Those servers are out of spec, but they exist and the call used to work. Worth either skipping validation with a warning when `StructuredContent` is nil, or parsing JSON out of the text content first and validating that, which is what the JS `processResult` path does.

go/go.sum
```diff
@@ -293,6 +293,8 @@ github.com/mailru/easyjson v0.9.0 h1:PrnmzHw7262yW8sTBwxi1PdJA3Iw/EKBa8psRf7d9a4
 github.com/mailru/easyjson v0.9.0/go.mod h1:1+xMtQp2MRNVL/V1bOzuP3aP8VNwRW55fQUto+XFtTU=
 github.com/mark3labs/mcp-go v0.29.0 h1:sH1NBcumKskhxqYzhXfGc201D7P76TVXiT0fGVhabeI=
 github.com/mark3labs/mcp-go v0.29.0/go.mod h1:rXqOudj/djTORU/ThxYx8fqEVj/5pvTuuebQ2RC7uk4=
+github.com/mark3labs/mcp-go v0.42.0 h1:gk/8nYJh8t3yroCAOBhNbYsM9TCKvkM13I5t5Hfu6Ls=
```
apascal07, line 294: `go mod tidy` was not run: the v0.29.0 lines are still here. `go mod tidy -diff` shows them as the only delta.

go/plugins/mcp/tools.go
```diff
@@ -139,12 +201,71 @@ func (c *GenkitMCPClient) fetchToolsPage(ctx context.Context, cursor mcp.Cursor)
 		},
 	}
 
-	result, err := c.server.Client.ListTools(ctx, listReq)
+	// Decode the tools/list response here instead of through mcp-go's typed
+	// ListTools result. Its ToolOutputSchema only models a subset of JSON Schema
+	// and would otherwise discard extension and unsupported keywords.
+	response, err := c.server.Transport.SendRequest(ctx, transport.JSONRPCRequest{
```
apascal07, line 213: This duplicates `Client.sendRequest`, down to the `transport.NewError` wrapping, and will drift as mcp-go changes. The reason for the bypass is real (`ToolOutputSchema` only models `$defs`, `type`, `properties`, `required`), so a comment pointing at an upstream issue would mark this as removable rather than permanent.

One difference worth noting: mcp-go guards on `c.initialized`, and `c.server.Error != ""` is not equivalent. A client built with `Disabled: true` has an empty `Error` and an uninitialized session. `GetActiveTools` catches that today via `IsEnabled()`, but the guard here reads as if it stands on its own.

go/plugins/mcp/tools.go
```diff
@@ -139,12 +201,71 @@ func (c *GenkitMCPClient) fetchToolsPage(ctx context.Context, cursor mcp.Cursor)
...
+}
+
+// toolWithRawOutputSchema retains the exact schema sent by the MCP server.
+// mcp.Tool's default decoder only keeps the JSON Schema fields represented by
+// mcp.ToolOutputSchema.
+type toolWithRawOutputSchema struct {
+	mcp.Tool
+}
+
+func (t *toolWithRawOutputSchema) UnmarshalJSON(data []byte) error {
```
apascal07: `inputSchema` loses the same keywords through the same typed decode, and this is the one place that could keep it. `{"type":"object","description":"...","additionalProperties":false,"anyOf":[...]}` reaches the model as `{"type":"object","properties":{...}}`, and a union `"type":["object","null"]` fails the decode for the entire page. `mcp.Tool` has `RawInputSchema` for this. Fine as a follow-up if you want to hold this PR to output schemas.

go/plugins/mcp/tools.go
```diff
@@ -139,12 +201,71 @@ func (c *GenkitMCPClient) fetchToolsPage(ctx context.Context, cursor mcp.Cursor)
...
+		return err
+	}
+
+	var tool mcp.Tool
+	if err := json.Unmarshal(toolData, &tool); err != nil {
+		return err
+	}
+	if hasOutputSchema && !bytes.Equal(bytes.TrimSpace(rawOutputSchema), []byte("null")) {
+		tool.RawOutputSchema = append(json.RawMessage(nil), rawOutputSchema...)
+		tool.OutputSchema = mcp.ToolOutputSchema{}
```
apascal07: nit: dead. `outputSchema` is deleted from `fields` before the decode, so `tool.OutputSchema` is already zero here.

apascal07 (commented):
`go/samples/mcp-git-pr-explainer` pins mcp-go v0.29.0 and resolves genkit through `replace ../../`, so it stops building after the bump: `go build ./...` there fails with `updates to go.mod needed`. CI does not catch it, since go.yml runs `go test -C go ./...` and samples.yml is JS-only and scoped to `samples/**`.

Issue #5899, opened 2026-08-03
feat(go/plugins/compat_oai): image, TTS, and STT model support


cabljac:
Go compat_oai is chat and embeddings only. JS and Python both cover image (dall-e-3, gpt-image-1), TTS (tts-1, tts-1-hd, gpt-4o-mini-tts), and STT (whisper-1, gpt-4o-transcribe, gpt-4o-mini-transcribe). JS reference: js/plugins/compat-oai/src/openai/{dalle,tts,stt,whisper}.ts plus the shared image.ts/audio.ts schemas. Biggest single gap in the Go plugin; fine to land image and audio as separate PRs under this issue.

Part of #5903.

GtechGovind:
I’d like to implement the full Go compat_oai modality gap in separate reviewable PRs, starting with image generation (dall-e-3 and gpt-image-1), followed by TTS and STT. I’ll follow the existing JS/Python schemas and behavior, keep each PR focused, add offline unit tests, run the repository formatting/build checks, and link each PR here.

PR #5937, opened 2026-08-04
feat(go/compat-oai): add image generation

GtechGovind:
Adds the first Go modality slice from #5899: shared OpenAI-compatible Images API support plus OpenAI registration for `dall-e-3` and `gpt-image-1`.

The implementation follows the canonical JavaScript plugin behavior: exact model-specific config schemas and defaults, version overrides, base64/URL media conversion, raw/request preservation, streaming rejection, and image-aware dynamic discovery/resolution. It also adds deterministic HTTP tests, OpenAI registration/schema tests, README documentation, and an image-generation sample flow.

Part of #5899.

Checklist (if applicable):
- [x] Tested (`go test ./plugins/compat_oai/... ./samples/compat_oai/openai`, `go vet ./...`, repository-wide `go test ./...`)
- [x] Docs updated


go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,164 @@
...
+		if part != nil && part.IsText() {
+			promptBuilder.WriteString(part.Text)
+		}
+	}
+	prompt := promptBuilder.String()
+	if prompt == "" {
+		return openai.ImageGenerateParams{}, fmt.Errorf("image generation requires text in the first message")
+	}
+
+	params := openai.ImageGenerateParams{Model: modelName, Prompt: prompt}
```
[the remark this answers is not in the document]
GtechGovind: The repository pins github.com/openai/openai-go v1.8.2. In that version ImageGenerateParams.Prompt is string and Model is openai.ImageModel; only N, OutputCompression, and User use param.Opt. The go doc output confirms the checked-in shape, and both the focused compat_oai suite and repository-wide Go suite compile and pass. The suggested param.NewOpt assignments do not type-check against v1.8.2, so the current code is correct for the repository dependency.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	// DALL-E defaults to short-lived URLs. Match the other Genkit runtimes by
+	// requesting durable inline media unless the caller explicitly opts out.
+	if params.ResponseFormat == "" && !strings.Contains(string(params.Model), "gpt-image") {
+		params.ResponseFormat = openai.ImageGenerateParamsResponseFormatB64JSON
+	}
+	// GPT Image always returns base64 and rejects response_format.
+	if strings.Contains(string(params.Model), "gpt-image") {
+		params.ResponseFormat = ""
+	}
```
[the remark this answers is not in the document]
GtechGovind: This is the same SDK-version mismatch. In pinned openai-go v1.8.2, Prompt, Model, and ResponseFormat are plain fields, not param.Field values; Present and Value do not exist on them. The current comparisons and assignments compile under the repository dependency and are covered by TestImageGenerateParams and TestGenerateImage.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,164 @@
...
+	if config.N != 0 {
+		params.N = param.NewOpt(config.N)
+	}
+	if config.OutputCompression != 0 {
+		params.OutputCompression = param.NewOpt(config.OutputCompression)
+	}
+	if config.User != "" {
+		params.User = param.NewOpt(config.User)
+	}
+}
```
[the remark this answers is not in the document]
GtechGovind: For pinned openai-go v1.8.2, Background, Moderation, OutputFormat, Quality, ResponseFormat, Size, and Style are plain enum fields with omitzero behavior. Only N, OutputCompression, and User are optional wrappers, and the implementation already uses param.NewOpt for those. Empty plain enum values are omitted by the generated SDK marshaler; the HTTP request-body tests verify the emitted payload.

GtechGovind:
I pushed a focused correction for pinned openai-go v1.8.2 behavior: GPT Image now exposes webp rather than web, DALL-E-only style is removed from GPT Image requests and schema, and base64 media parts preserve requested PNG, JPEG, or WebP content types. Fresh focused tests and vet pass. This is separate from the review suggestion about param.Field, which still does not apply to the pinned SDK version.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,182 @@
...
+	OutputFormat      openai.ImageGenerateParamsOutputFormat   `json:"output_format,omitempty"`
+	Quality           openai.ImageGenerateParamsQuality        `json:"quality,omitempty"`
+	ResponseFormat    openai.ImageGenerateParamsResponseFormat `json:"response_format,omitempty"`
+	Size              openai.ImageGenerateParamsSize           `json:"size,omitempty"`
+	Style             openai.ImageGenerateParamsStyle          `json:"style,omitempty"`
+	User              string                                   `json:"user,omitempty"`
+}
+
+func imageGenerateParams(modelName string, input *ai.ModelRequest) (openai.ImageGenerateParams, error) {
+	if len(input.Messages) == 0 || input.Messages[0] == nil {
```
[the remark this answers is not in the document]
GtechGovind: Fixed in 19a86a3aa. imageGenerateParams now rejects a nil model request and the invalid-prompt regression table covers it.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,182 @@
...
+func generateImage(
+	ctx context.Context,
+	client *openai.Client,
+	modelName string,
+	input *ai.ModelRequest,
+	cb func(context.Context, *ai.ModelResponseChunk) error,
+) (*ai.ModelResponse, error) {
+	if cb != nil {
+		return nil, fmt.Errorf("streaming mode not supported for image generation")
+	}
```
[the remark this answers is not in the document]
GtechGovind: Fixed in 19a86a3aa. generateImage now returns a clear initialization error for a nil client, with regression coverage.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,182 @@
...
+	params, err := imageGenerateParams(modelName, input)
+	if err != nil {
+		return nil, err
+	}
+	result, err := client.Images.Generate(ctx, params)
+	if err != nil {
+		return nil, err
+	}
+	return imageResponse(result, input, params.OutputFormat), nil
```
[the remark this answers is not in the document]
GtechGovind: Fixed in 19a86a3aa. A successful nil Images result is now rejected before response conversion. The regression uses the pinned SDK with an endpoint returning JSON null, which reproduces the nil result.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,191 @@
...
+		if part != nil && part.IsText() {
+			promptBuilder.WriteString(part.Text)
+		}
+	}
+	prompt := promptBuilder.String()
+	if prompt == "" {
+		return openai.ImageGenerateParams{}, fmt.Errorf("image generation requires text in the first message")
+	}
+
+	params := openai.ImageGenerateParams{Model: modelName, Prompt: prompt}
```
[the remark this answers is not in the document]
GtechGovind: No change. This repository pins github.com/openai/openai-go v1.8.2, where ImageGenerateParams.Model is openai.ImageModel and Prompt is string, not param.Field. The exact current head compiles and the full compat_oai test suite passes.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,191 @@
...
+
+	params.Prompt = prompt
+	if params.Model == "" {
+		params.Model = modelName
+	}
+	// DALL-E defaults to short-lived URLs. Match the other Genkit runtimes by
+	// requesting durable inline media unless the caller explicitly opts out.
+	if params.ResponseFormat == "" && !strings.Contains(string(params.Model), "gpt-image") {
+		params.ResponseFormat = openai.ImageGenerateParamsResponseFormatB64JSON
+	}
```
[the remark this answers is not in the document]
GtechGovind: No change. In pinned openai-go v1.8.2, ResponseFormat and Model are direct enum values, not param.Field values. Direct comparison and conversion are the correct API for this repository version.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,191 @@
...
+	// GPT Image always returns base64 and rejects response_format.
+	if strings.Contains(string(params.Model), "gpt-image") {
+		params.ResponseFormat = ""
+		// Style is a DALL-E 3 option and is rejected by GPT Image models.
+		params.Style = ""
+	}
```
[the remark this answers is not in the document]
GtechGovind: No change. The pinned v1.8.2 definitions use direct enum fields for Model, ResponseFormat, and Style. The proposed Value access would not compile against go.mod; current code and tests do compile.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,191 @@
...
+	if config.Version != "" {
+		params.Model = config.Version
+	}
+	params.Background = config.Background
+	params.Moderation = config.Moderation
+	params.OutputFormat = config.OutputFormat
+	params.Quality = config.Quality
+	params.ResponseFormat = config.ResponseFormat
+	params.Size = config.Size
+	params.Style = config.Style
```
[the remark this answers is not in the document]
GtechGovind: No change. The pinned v1.8.2 SDK uses direct values for these image fields; only N, OutputCompression, and User use param.Opt. The proposed param.F API targets a different SDK version and would break this repository build.

go/plugins/compat_oai/openai/openai.go
```diff
@@ -157,8 +161,73 @@ var (
 			},
 		},
 	}
+
+	supportedImageModels = map[string]ai.ModelOptions{
+		openaiGo.ImageModelDallE3: {
+			Label:        "OpenAI DALL-E 3",
+			Supports:     &compat_oai.ImageGeneration,
+			ConfigSchema: imageConfigSchema(openaiGo.ImageModelDallE3),
+			Versions:     []string{openaiGo.ImageModelDallE3},
```
cabljac, line 93: `version` can't be reached through `ai.Generate`. `validateSupport` (go/ai/generate.go:185) runs `validateVersion`, which rejects any `config.version` not in `ModelOptions.Versions`, so the only value that passes here is `"dall-e-3"` and that's a no-op. Verified against a stub: `Version: "dall-e-3-2026-01-01"` fails with `supported versions: [dall-e-3]` and no HTTP request is made.

Dropping `Versions` doesn't help, an empty list rejects too, so the dynamic `ListActions`/`ResolveAction` paths are broken the same way. And openai-go v1.8.2 documents no dated image snapshots, only dall-e-2, dall-e-3, gpt-image-1.

I'd drop `version` from `ImageGenerationConfig` and the schema rather than ship a field that can't work. Note that diverges from JS, image.ts:64 routes it the same way but has no version validation.

go/plugins/compat_oai/image_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/firebase/genkit/go/ai"
+	"github.com/openai/openai-go"
+	"github.com/openai/openai-go/option"
+)
+
+func TestImageGenerateParams(t *testing.T) {
+	request := &ai.ModelRequest{
+		Messages: []*ai.Message{ai.NewUserTextMessage("a cat wearing a hat")},
+		Config: map[string]any{
+			"version":         "dall-e-3-2026-01-01",
```
cabljac: Passes only because line 45 calls `imageGenerateParams` directly. Through `genkit.Generate` this config fails at schema validation first (`temperature` isn't in `imageConfigSchema`, which sets `additionalProperties: false`), then at `validateVersion` once you drop it. So the `model: "dall-e-3-2026-01-01"` expectation at line 58 asserts something no caller can reach.

A test through `genkit.Generate` against the httptest server at line 143 would put both checks in the loop.

go/plugins/compat_oai/openai/openai.go
```diff
@@ -242,9 +326,38 @@ func (o *OpenAI) Embedder(g *genkit.Genkit, name string) ai.Embedder {
...
+		descriptions[i] = o.DefineImageModel(name, opts).(api.Action).Desc()
+	}
+	return descriptions
 }
 
 func (o *OpenAI) ResolveAction(atype api.ActionType, name string) api.Action {
+	if atype == api.ActionTypeModel && isImageModel(name) {
+		opts, ok := supportedImageModels[name]
+		if !ok {
+			opts = ai.ModelOptions{
```
cabljac: This fallback omits `Versions`, as does the byte-identical one in `ListActions` at 337-343, so any model not in `supportedImageModels` rejects every `version` value including its own name.

`Versions: []string{name}` isn't the fix though, it just turns the error into a silent no-op, since `generateImage` is already called with that name. Worth being consistent with whatever you decide for `version` overall.

go/plugins/compat_oai/openai/openai.go
```diff
@@ -242,9 +326,38 @@ func (o *OpenAI) Embedder(g *genkit.Genkit, name string) ai.Embedder {
...
+				Label:        "OpenAI - " + name,
+				Stage:        ai.ModelStageStable,
+				Supports:     &compat_oai.ImageGeneration,
+				ConfigSchema: imageConfigSchema(name),
+			}
+		}
+		descriptions[i] = o.DefineImageModel(name, opts).(api.Action).Desc()
```
cabljac, line 570: nit: builds a full model action per `ListActions` call just to read `Desc()`, including four middlewares and an uncached `InferSchemaMap`. Building the `api.ActionDesc` directly would be cheaper. This block is also byte-identical to the one in `ResolveAction`, so a helper would stop them drifting.

go/plugins/compat_oai/openai/openai.go
```diff
@@ -157,8 +161,73 @@ var (
...
+			Label:        "OpenAI GPT Image 1",
+			Supports:     &compat_oai.ImageGeneration,
+			ConfigSchema: imageConfigSchema(openaiGo.ImageModelGPTImage1),
+			Versions:     []string{openaiGo.ImageModelGPTImage1},
+		},
+	}
 )
 
+func isImageModel(name string) bool {
+	return strings.Contains(name, "dall-e") || strings.Contains(name, "gpt-image")
```
cabljac, line 105: `dall-e-2` matches here and then gets the DALL-E 3 schema: `1792x1024`, `1024x1792`, `quality: standard|hd`, `style`. Per the pinned SDK it takes only `256x256`/`512x512`/`1024x1024` (image.go:530-536), `standard` only (image.go:519), and no `style` (image.go:538-539).

Not cosmetic, `ConfigSchema` is validated on every Run, so `512x512` is rejected locally while `1792x1024` is offered and would 400. Since the match is `strings.Contains`, every future `dall-e-*` name inherits it, so loosening the fallback beats special-casing dall-e-2.

go/plugins/compat_oai/openai/openai.go
```diff
@@ -157,8 +161,73 @@ var (
...
+}
+
+func imageConfigSchema(name string) map[string]any {
+	properties := map[string]any{
+		"n": map[string]any{
+			"type":    "integer",
+			"minimum": 1,
+			"maximum": 10,
```
cabljac, line 113: nit: `dall-e-3` only accepts `n: 1` (openai-go v1.8.2 image.go:478-479, where 1-10 is the general range with dall-e-3 called out as the exception). Inherited from the JS `ImageGenerationCommonConfigSchema`, which also allows 10 (image.ts:52), so parity is fair. Since this schema is built per model name it could be exact.

cabljac (commented):
Thanks for this. Reviewed properly and it merges cleanly onto current main including the typed-config refactors. Shared-layer changes are additive, so no blast radius on kimi/zai/dashscope/anthropic, and it tracks the JS image.ts design closely.

One blocker: the `version` config field can't be exercised through `ai.Generate` on any image model, static or dynamic. Everything else is schema accuracy and a couple of small things.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,191 @@
...
+func imageResponse(
+	result *openai.ImagesResponse,
+	input *ai.ModelRequest,
+	format openai.ImageGenerateParamsOutputFormat,
+) *ai.ModelResponse {
+	contentType := imageContentType(format)
+	response := &ai.ModelResponse{
+		FinishReason: ai.FinishReasonStop,
+		Message:      &ai.Message{Role: ai.RoleModel},
+		Raw:          result,
```
cabljac: nit: `Raw: result` keeps `b64_json`, and `ModelResponse` goes into the `genkit:output` span attribute untruncated, so the image is serialised twice per span and again on the parent. JS does the same (image.ts:102) and veo.go too, so parity. Stripping `b64_json` from the `Raw` copy would halve it. Not suggesting you drop `Raw`, that'd break image_test.go:172.

go/plugins/compat_oai/image.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+}
+
+func imageResponse(
+	result *openai.ImagesResponse,
+	input *ai.ModelRequest,
+	format openai.ImageGenerateParamsOutputFormat,
+) *ai.ModelResponse {
+	contentType := imageContentType(format)
+	response := &ai.ModelResponse{
```
cabljac, line 138: nit: gpt-image-1 returns token usage in `result.Usage` (`InputTokens`/`OutputTokens`/`TotalTokens`), which maps onto `ai.GenerationUsage` almost directly. Needs an int64 to int cast and a `result.JSON.Usage.Valid()` guard, otherwise DALL-E responses report a zeroed usage since `Usage` is a value struct. JS drops it too, so parity, but it would make usage reporting work for image flows.

go/plugins/compat_oai/openai/openai.go
GtechGovind: Fixed in a17d21519 and carried onto current main in 5d2fa2139. I removed `version` from `ImageGenerationConfig`, its schema, and request mapping. Image model selection now comes only from the action/model ID, so the unreachable no-op override is no longer advertised.

go/plugins/compat_oai/openai/openai.go
GtechGovind: Fixed in a17d21519. The DALL-E 3 schema now caps `n` at 1, with regression coverage pinning that model-specific limit.

go/plugins/compat_oai/image.go
GtechGovind: Fixed in a17d21519. GPT Image usage is mapped to `ai.GenerationUsage` only when `result.JSON.Usage.Valid()` is true, including the required int64-to-int conversion. DALL-E responses therefore do not get a misleading zero usage object.

go/plugins/compat_oai/image_test.go
GtechGovind: Fixed in a17d21519. The unreachable `version`/`temperature` expectations were removed, and an httptest-backed regression now calls `genkit.Generate` through the registered DALL-E 3 model so schema validation, model resolution, request conversion, and response parsing are all exercised end to end.

go/plugins/compat_oai/image.go
GtechGovind: Fixed in a17d21519. The response keeps a shallow raw response copy for compatibility but clears `B64JSON` from its copied data entries, so the media payload is not serialized twice into tracing output. The original response remains untouched while media parts are built.

go/plugins/compat_oai/openai/openai.go
GtechGovind: Fixed in a17d21519. DALL-E 2 now has its exact size set (`256x256`, `512x512`, `1024x1024`), `standard` quality only, and no `style`. DALL-E 3 remains exact, while unknown future DALL-E IDs use a deliberately looser string schema rather than inheriting DALL-E 3 constraints.

go/plugins/compat_oai/openai/openai.go
GtechGovind: Addressed in a17d21519 and the current-main integration in 5d2fa2139. Static and dynamic image options now share `imageModelOptions`, while `ListActions` uses the refactored `compat_oai.ListModelActions` path, which caches deterministic descriptors rather than rebuilding every generic descriptor on each reflection poll.

go/plugins/compat_oai/openai/openai.go
GtechGovind: Fixed consistently with the main finding in a17d21519. The `version` option was removed for both curated and dynamically discovered image models instead of adding a silent self-version fallback.

PR #5938, opened 2026-08-04
feat(go/compat-oai): add audio models

GtechGovind:
Adds the second Go modality slice from #5899: shared OpenAI-compatible text-to-speech, transcription, and Whisper translation support.

The OpenAI plugin now registers `tts-1`, `tts-1-hd`, `gpt-4o-mini-tts`, `whisper-1`, `gpt-4o-transcribe`, and `gpt-4o-mini-transcribe`. The implementation follows the canonical JavaScript schemas and routing, supports binary speech media plus JSON/text/SRT/VTT transcription results, sends multipart audio correctly, classifies future audio models, and explicitly rejects unsupported streaming. Whisper `translate` routes to the translations API.

Deterministic tests cover request payloads, multipart fields, response conversion, model metadata/schemas, chunking strategies, translation, discovery/resolution, and streaming rejection. Documentation includes TTS, transcription, and translation examples.

Part of #5899.

Checklist (if applicable):
- [x] Tested (`go test ./...`, `go vet ./...`, fresh focused tests with `-count=1`)
- [x] Docs updated


go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,450 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("read speech response: %w", err)
+		}
+
+		format := config.ResponseFormat
+		if format == "" {
+			format = openai.AudioSpeechNewParamsResponseFormatMP3
+		}
+		contentType := responseFormatMediaTypes[format]
+		dataURI := fmt.Sprintf("data:%s;base64,%s", contentType, base64.StdEncoding.EncodeToString(audio))
```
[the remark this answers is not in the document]
GtechGovind: Fixed in da10d177a. Unknown or future speech response formats now fall back to application/octet-stream, and the response media/data URI test covers the fallback.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,450 @@
...
+		"audio/ogg":  ".ogg",
+		"audio/flac": ".flac",
+		"audio/webm": ".webm",
+	}
+	extension := extensions[contentType]
+	if extension == "" {
+		extension = ".mp3"
+	}
+	return "input" + extension
+}
```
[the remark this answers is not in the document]
GtechGovind: Fixed in da10d177a. audioFilename now strips MIME parameters, normalizes case and whitespace, and maps common MP3, M4A, WAV, OGG, FLAC, and WebM aliases. Table-driven tests cover the variants.

go/plugins/compat_oai/openai/openai.go
```diff
@@ -157,8 +159,55 @@ var (
...
+	}
 )
 
+func speechConfigSchemaWithoutSpeed() map[string]any {
+	schema := core.InferSchemaMap(compat_oai.SpeechConfig{})
+	if properties, ok := schema["properties"].(map[string]any); ok {
+		delete(properties, "speed")
+	}
+	return schema
+}
```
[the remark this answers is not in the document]
GtechGovind: Fixed in da10d177a. speechConfigSchemaWithoutSpeed now clones the top-level schema and properties map before deletion. A regression test verifies deriving the GPT-4o Mini TTS schema does not mutate the full speech schema.

GtechGovind:
da10d177a also replaces full dynamic action descriptors, uses JSON-only defaults and cloned schemas for GPT transcription models (including versioned IDs), validates audio media, separates translation and transcription parameter construction, and expands the voice schema to every voice in pinned openai-go v1.8.2. Fresh compat_oai and repository-wide tests plus vet pass.

GtechGovind:
7b732c0e4: transcription now rejects remote and unsupported audio inputs before API calls; GPT transcription metadata is JSON-only while Whisper remains text and JSON; gpt-4o-mini-tts now exposes and forwards instructions without exposing that option on legacy TTS schemas. Regression tests and README guidance were added. Focused compat_oai tests, vet, and diff checks pass.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,559 @@
...
+		return openai.AudioTranscriptionNewParams{}, fmt.Errorf("invalid chunking strategy: %w", err)
+	}
+	params := openai.AudioTranscriptionNewParams{
+		File:                   file,
+		Model:                  model,
+		ChunkingStrategy:       chunkingStrategy,
+		Include:                config.Include,
+		ResponseFormat:         config.ResponseFormat,
+		TimestampGranularities: config.TimestampGranularities,
+	}
```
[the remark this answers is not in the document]
GtechGovind: Fixed in c8dca2006. ResponseFormat is now initialized once from the resolved format.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,559 @@
...
+	// The pinned OpenAI SDK supports only JSON responses for GPT transcription
+	// models, which is intentionally stricter than the current canonical JS schema.
+	schema = maps.Clone(schema)
+	properties := maps.Clone(schema["properties"].(map[string]any))
+	schema["properties"] = properties
+	responseFormat := maps.Clone(properties["response_format"].(map[string]any))
+	properties["response_format"] = responseFormat
+	responseFormat["enum"] = []any{string(openai.AudioResponseFormatJSON)}
+	responseFormat["default"] = string(openai.AudioResponseFormatJSON)
```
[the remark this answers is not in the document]
GtechGovind: Fixed in c8dca2006. Schema traversal now uses checked assertions and returns unexpected structures unchanged; regression coverage includes nil, missing, and wrong-typed nested schema values.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,571 @@
...
+		model := id
+		if config.Version != "" {
+			model = config.Version
+		}
+		params := openai.AudioSpeechNewParams{
+			Input:          req.Messages[0].Text(),
+			Model:          model,
+			Voice:          voice,
+			ResponseFormat: config.ResponseFormat,
+		}
```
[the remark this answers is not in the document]
GtechGovind: Fixed in 3f4267edd. Speech requests now reject empty and whitespace-only text before any API call, with regression coverage.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	return params, nil
+}
+
+func toChunkingStrategy(value any) (openai.AudioTranscriptionNewParamsChunkingStrategyUnion, error) {
+	var result openai.AudioTranscriptionNewParamsChunkingStrategyUnion
+	if value == nil {
+		return result, nil
+	}
```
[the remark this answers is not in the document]
GtechGovind: Fixed in 3f4267edd. Already typed chunking strategy values and pointers now use a direct fast path; the regression verifies a typed VAD pointer is preserved.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,571 @@
...
+		if config == nil {
+			return zero, nil
+		}
+		return *config, nil
+	case map[string]any:
+		return base.MapToStruct[T](config)
+	default:
+		return zero, fmt.Errorf("unexpected config type: %T", config)
+	}
+}
```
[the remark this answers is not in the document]
GtechGovind: No change. Repository action input arrives as JSON, then Action.RunJSON normalizes it into a structured ModelRequest; JSON object config reaches parseAudioConfig as map[string]any, which is already supported. json.RawMessage, byte slices, and JSON strings are not part of that action contract. Adding speculative coercion for arbitrary direct Go values would broaden behavior without a runtime path or repository precedent.

cabljac (commented):
Thanks for this. Went through it properly. The shared package change is a new file and the siblings (kimi/zai/dashscope/anthropic) delegate to the unchanged shared methods, so I don't think there's any blast radius there, though ListActions and ResolveAction in the openai subpackage are rewritten rather than extended. The json-only restriction for the GPT transcription models matches the pinned SDK as far as I can see, and the tests are strong.

Main thing I'd want to sort before merge: json is the only output format the transcription models advertise, and it looks like it's the one that fails through genkit.Generate. Default text calls are fine, and speech is unaffected.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,584 @@
...
+
+	// TranscriptionSupports describes speech-to-text model capabilities.
+	TranscriptionSupports = ai.ModelSupports{
+		Media:  true,
+		Output: []string{"text", "json"},
+	}
+
+	gptTranscriptionSupports = ai.ModelSupports{
+		Media:  true,
+		Output: []string{"json"},
```
cabljac: `gptTranscriptionSupports` advertises `output: ["json"]`, but json seems to be the one output format that doesn't work end to end. I ran it through `genkit.Generate` with `ai.WithOutputFormat("json")` and got `model failed to generate output matching expected schema: message is not a valid JSON`. As far as I can tell `transcriptionResponse` (audio.go:424) returns the bare transcript and `jsonHandler.ParseMessage` then rejects it (go/ai/format.go:360 via go/ai/generate.go:465).

Looks like the same applies to `TranscriptionSupports` and whisper-1. Speech is unaffected and default text calls are fine, so it's opt-in rather than every call. Worth noting `Supports.Output` doesn't appear to be enforced anywhere, it reads as metadata only (go/ai/generate.go:165).

Root cause looks like `transcriptionResponseFormat` (audio.go:473) piping genkit's `output.Format` straight into OpenAI's `response_format`, which are two different vocabularies that happen to share the token `json`. If that's right, I'd advertise `text` and keep `response_format=json` at the wire level.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,584 @@
...
+		}
+		contentType := responseFormatMediaTypes[format]
+		if contentType == "" {
+			contentType = "application/octet-stream"
+		}
+		dataURI := fmt.Sprintf("data:%s;base64,%s", contentType, base64.StdEncoding.EncodeToString(audio))
+		return &ai.ModelResponse{
+			Message:      ai.NewModelMessage(ai.NewMediaPart(contentType, dataURI)),
+			FinishReason: ai.FinishReasonStop,
+			Raw:          audio,
```
cabljac: `Raw: audio` is a `[]byte`, so it marshals as base64, and the whole `ModelResponse` gets serialised into the `genkit:output` span attribute (go/core/tracing/tracing.go:400-401, field at go/ai/gen.go:348). I marshalled a representative response and the same base64 string came out twice, once in the media data URI and once in `raw`, so a 3 KB clip produced roughly 8 KB of JSON.

The JS speech runner sets `raw` to the already-drained HTTP `Response` rather than the payload (audio.ts:165), though its transcription path does store the parsed body (:332), so the divergence looks specific to speech. Might be worth dropping `Raw` here, or storing format and byte count instead, unless there's a reason you want the payload.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,584 @@
...
+		opts.Versions = []string{id}
+	}
+
+	return ai.NewModel(api.NewName(provider, id), &opts, func(
+		ctx context.Context,
+		req *ai.ModelRequest,
+		cb func(context.Context, *ai.ModelResponseChunk) error,
+	) (*ai.ModelResponse, error) {
+		if cb != nil {
+			return nil, errors.New("streaming is not supported for speech models")
```
cabljac, line 124: Question rather than a finding. Hard-failing on a stream callback diverges from JS, where the speech runner is handed the streaming context but destructures only `{ abortSignal }` and never emits (audio.ts:209-219, transcription at 378-394), so a streaming caller there gets an empty chunk stream plus the complete final response rather than an error.

`ai.ModelSupports` (go/ai/gen.go:280-301) has no streaming flag, same on the JS side, so a caller can't easily discover this up front and these models will error in any streaming context rather than degrading.

`TestAudioModelsRejectStreaming` (audio_test.go:191) asserts it deliberately across all three `Define*Model` functions (audio.go:124, 215, 249), so I assume this is considered rather than accidental. Was JS parity ruled out here, or is the hard error load-bearing for something?

go/plugins/compat_oai/audio_test.go
```diff
@@ -0,0 +1,577 @@
...
+	plugin := newAudioPlugin(t, func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = io.WriteString(w, `{"text":"Hello JSON","language":"en"}`)
+	})
+	model := plugin.DefineTranscriptionModel("test", "gpt-4o-transcribe", ai.ModelOptions{})
+	resp, err := model.Generate(context.Background(), &ai.ModelRequest{
+		Messages: []*ai.Message{ai.NewUserMessage(
+			ai.NewMediaPart("audio/wav", "data:audio/wav;base64,YXVkaW8="),
+		)},
+		Output: &ai.ModelOutputConfig{Format: "json"},
```
cabljac, line 302: I think this passes for the wrong reason. `model.Generate` is just `m.Action.Run` (go/ai/generate.go:932-938), and format resolution and parsing live in `GenerateWithRequest` at :289 and :462-470, which only `genkit.Generate` reaches, so setting `Output` here hands the model a field it ignores and never runs the formatter. That looks like what's hiding the json issue flagged on `gptTranscriptionSupports`.

Probably worth one end-to-end test per model kind (speech, transcription, whisper) through `genkit.Generate`. `TestWhisperModelTranslation` (openai/audio_test.go:228) inits the plugin but calls the action directly at :285, so I don't think it covers this either.

go/plugins/compat_oai/audio.go
```diff
@@ -0,0 +1,584 @@
...
+	PrefixPaddingMS   int     `json:"prefix_padding_ms,omitempty"`
+	SilenceDurationMS int     `json:"silence_duration_ms,omitempty"`
+	Threshold         float64 `json:"threshold,omitempty" jsonschema:"minimum=0,maximum=1"`
+}
+
+// TranscriptionConfig configures an OpenAI-compatible speech-to-text request.
+type TranscriptionConfig struct {
+	Temperature float64 `json:"temperature,omitempty"`
+	// ChunkingStrategy is either the string "auto" or a server VAD configuration object.
+	ChunkingStrategy       any                           `json:"chunking_strategy,omitempty"`
```
cabljac, line 77: `chunking_strategy` as `any` seems to infer to `"chunking_strategy": true`, the always-true schema, so nothing gets validated. I tried `"garbage"`, `42`, and an out-of-range threshold and all three passed. JS models it as a union of the literal `auto` and a server-VAD object (audio.ts:65-76).

`TranscriptionChunkingStrategy` already defines the object half but doesn't appear to be used outside tests, and this file already post-processes inferred schemas in `jsonOnlyTranscriptionConfigSchema` (:514), so a `oneOf`/`anyOf` there would probably restore parity. I checked gojsonschema handles it, including rejecting the values above. Would need applying to the Whisper schema too, which currently skips post-processing. Non-blocking.

go/plugins/compat_oai/openai/openai.go
```diff
@@ -242,9 +332,71 @@ func (o *OpenAI) Embedder(g *genkit.Genkit, name string) ai.Embedder {
...
+	if opts, ok := supportedSpeechModels[name]; ok {
+		return opts, "speech", true
+	}
+	if opts, ok := supportedTranscriptionModels[name]; ok {
+		return opts, "transcription", true
+	}
+	if opts, ok := supportedWhisperModels[name]; ok {
+		return opts, "whisper", true
+	}
+	if strings.Contains(name, "tts") {
```
cabljac, line 607: nit: these substring checks run before delegating to the shared resolver, so I think any future chat model whose name contains `tts`, `whisper` or `transcribe` would get re-typed as an audio model. Mirrors JS (openai/index.ts:89-114) so not asking you to diverge, just flagging that the override now sits in front of the shared path rather than beside it.

go/plugins/compat_oai/audio.go
GtechGovind: Fixed in 2923fc8ad and carried onto current main in 2a20c4376. Transcription models now advertise text output while retaining the provider-required JSON wire format for GPT transcription. When Genkit explicitly requests JSON, the response message contains valid JSON for the formatter. An end-to-end `genkit.Generate(..., ai.WithOutputFormat("json"))` regression covers the original failure.

go/plugins/compat_oai/audio_test.go
GtechGovind: Fixed the formatter coverage gap in 2923fc8ad. The JSON transcription regression now goes through `genkit.Generate`, so format resolution and `ParseMessage` run and would reproduce the original failure. Direct model tests remain for lower-level speech and Whisper request routing.

go/plugins/compat_oai/audio.go
GtechGovind: No code change here; the hard error remains deliberate. Returning a complete final response after emitting zero stream chunks makes a requested streaming contract appear to succeed while silently degrading. Until model capabilities can advertise non-streaming support, failing immediately is clearer and keeps all three audio model kinds consistent. The explicit regression remains to pin that behavior.

go/plugins/compat_oai/audio.go
GtechGovind: Fixed in 2923fc8ad. Both transcription and Whisper schemas now expose `chunking_strategy` as `auto` or a closed `server_vad` object with a bounded threshold. Runtime conversion also rejects invalid strings, scalars, object types, and out-of-range thresholds, with regression coverage.

go/plugins/compat_oai/audio.go
GtechGovind: Fixed in 2923fc8ad. Speech responses no longer store the audio byte slice in `Raw`; they retain only `content_type` and `byte_length` metadata. The payload remains once in the media data URI, avoiding duplicate base64 serialization in trace attributes.

go/plugins/compat_oai/openai/openai.go
GtechGovind: No code change, as suggested. The substring classification remains aligned with the canonical JS provider and is limited to the audio override before the shared resolver. If the OpenAI chat catalog introduces an overlapping ID, this should be replaced with an explicit catalog/classifier update rather than speculatively diverging here.

PR #5949, opened 2026-08-04
fix(cli): wait for action registration in ephemeral runtimes

pavelgj:
Add waitForActionKeys utility that polls the runtime until target actions are registered before dispatching. This closes a race condition where SDKs (notably Go) register the runtime with the CLI during initialization but define their actions slightly later, causing "action not found" errors.

Wire up waitForActionKeys in flow:run, flow:batchRun, eval:flow, and eval:run commands.

### DIFF
--- a/genkit-tools/cli/src/commands/eval-flow.ts
+++ b/genkit-tools/cli/src/commands/eval-flow.ts
@@ -212,9 +212,19 @@ export const evalFlow = new Command('eval:flow')
         }
       };
 
+      // Wait for the target flow to register. If specific evaluators were
+      // requested, wait for those too. When none are specified we cannot know
+      // the keys ahead of time, so we skip them and let discovery handle it.
+      const waitForActionKeys = [`/flow/${flowName}`];
+      if (options.evaluators) {
+        waitForActionKeys.push(
+          ...options.evaluators.split(',').map((k) => `/evaluator/${k}`)
+        );
+      }
+
       await runWithManager(projectRoot, runAction, {
         runtimeCommand,
-        waitForActionKeys: [`/flow/${flowName}`],
+        waitForActionKeys,
       });
     }
   );
--- a/genkit-tools/cli/src/commands/eval-flow.ts
+++ b/genkit-tools/cli/src/commands/eval-flow.ts
@@ -212,7 +212,10 @@ export const evalFlow = new Command('eval:flow')
         }
       };
 
-      await runWithManager(projectRoot, runAction, { runtimeCommand });
+      await runWithManager(projectRoot, runAction, {
+        runtimeCommand,
+        waitForActionKeys: [`/flow/${flowName}`],
+      });
     }
   );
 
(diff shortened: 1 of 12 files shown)


genkit-tools/cli/src/utils/manager-utils.ts
```diff
@@ -220,12 +220,79 @@ export async function waitForRuntime(
...
+ * with the CLI during initialization but define their actions slightly later.
+ * If we dispatch a runAction before the target action is registered, the
+ * runtime returns an "action not found" error. Polling until the actions
+ * appear closes that race. On timeout we proceed anyway and let the real
+ * error surface.
+ */
+export async function waitForActionKeys(
+  manager: BaseRuntimeManager,
+  keys: string[],
+  timeoutMs: number = 30000
```
apascal07: 30s seems like a long time... if you mistype the name, you'll be sitting there for a while not realizing. What if it just compares the length of `listActions()` like once per second over 5 secs and if it doesn't change (which resets the counter) then call it quits?

pavelgj: Good point. Done.

genkit-tools/cli/src/commands/eval-flow.ts
```diff
@@ -212,7 +212,10 @@ export const evalFlow = new Command('eval:flow')
         }
       };
 
-      await runWithManager(projectRoot, runAction, { runtimeCommand });
+      await runWithManager(projectRoot, runAction, {
+        runtimeCommand,
+        waitForActionKeys: [`/flow/${flowName}`],
```
ssbushi: This command too has the option to specify evaluators...

ssbushi (approved):

genkit-tools/cli/src/utils/manager-utils.ts
```diff
@@ -220,12 +220,107 @@ export async function waitForRuntime(
...
+ * To avoid making a mistyped action name block for the full timeout, we watch
+ * the number of registered actions: while it keeps changing, registration is
+ * still in progress; once it settles (stops changing for `stableForMs`) without
+ * the target appearing, we stop waiting and let the subsequent runAction
+ * surface the real "not found" error.
+ */
+export async function waitForActionKeys(
+  manager: BaseRuntimeManager,
+  keys: string[],
+  options: WaitForActionKeysOptions = {}
```
ssbushi: nit: Can we move the defaults in here, for the sake of readability?

genkit-tools/cli/src/commands/eval-flow.ts
pavelgj: done

genkit-tools/cli/src/utils/manager-utils.ts
pavelgj: done

apascal07 (approved):

PR #5950, opened 2026-08-04
fix(go/plugins/anthropic): map multipart tool response content to tool_result blocks

apascal07:
Multipart tool responses were dropped in the Anthropic request path. `toAnthropicParts` marshaled `ToolResponse.Output` into a single text block and never read `ToolResponse.Content`, so the parts a multipart tool attaches through `DefineMultipartTool` or `tool.AttachParts` never reached the model. A tool that returned an image and no structured output sent Claude a lone `"null"` text block. `vertexai/modelgarden` shares this conversion and had the same behavior.

Anthropic's `tool_result` carries one `content` array rather than Gemini's separate `response` and `parts` fields, so the structured output leads as a text block and the content parts follow as `text`, `image`, or `document` blocks, reusing the plugin's existing content-type mapping (`image/*`, `application/pdf`, `text/plain`):

```json
{"type":"tool_result","tool_use_id":"toolu_1","is_error":false,"content":[
  {"type":"text","text":"{\"result\":\"ok\"}"},
  {"type":"text","text":"here is the chart"},
  {"type":"image","source":{"type":"base64","media_type":"image/png","data":"..."}}]}
```

Parts Anthropic cannot carry inside a `tool_result` (reasoning, nested tool requests, audio or video media) now fail with `ErrInvalidPart` or `ErrUnsupportedByModel` rather than being silently dropped. Plain tool responses serialize byte-identically to before.

## Verification

Six cases added to `TestToAnthropicParts`: output plus content, content with no output, an unsupported media type, an unsupported part kind, a tool response part with no tool response, and a nil part inside the content. Each fails on main. Go 1.26.3, full `go test ./...` in `go/`: 41 packages pass, no failures.

### DIFF
--- a/go/plugins/anthropic/anthropic_live_test.go
+++ b/go/plugins/anthropic/anthropic_live_test.go
@@ -254,6 +254,40 @@ func TestAnthropicLive(t *testing.T) {
 			t.Fatal("expected a response but nothing was returned")
 		}
 	})
+	t.Run("multipart tool", func(t *testing.T) {
+		m := anthropicPlugin.Model(g, "claude-sonnet-4-5-20250929")
+		img64, err := fetchImgAsBase64()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		tool := genkit.DefineMultipartTool(g, "getImage", "returns a misterious image",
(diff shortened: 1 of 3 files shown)


adesinah (approved):
Looks pretty good.

PR #5967, opened 2026-08-06
fix(go): redact inline media data from trace attributes

### LINKED ISSUE #5725: [Go] Large span attributes for images (media parts)
**Describe the bug**
When including media parts in GenerateData, the attributes added to the span include the base64 encoded image (maybe cuttoff, not sure if our infra/otel/...). This is an issue of itself because it can incur large infra costs, and additionally this can lead to a memory footprint of the application that is far too large.

**To Reproduce**
Use `genkit.GenerateData` with `ai.NewMediaPart` in the messages.

**Expected behavior**
Not the base64 encoded image. Maybe the source (name, url, ...) would be enough. Just keep the memory usage and size of attribute reasonable.

**Screenshots**

Image

Image

 - OS: all
 - Version: v1.10.0

** Go version
go version go1.26.5 darwin/arm64 (but also in our containerized deployed application)

mikemikimike:
## Summary

- Redact inline base64 data URLs from Go tracing input, init, and output attributes.
- Preserve the media URI metadata and remote media URLs.
- Add an OTEL regression test using an actual ai.ModelRequest with media parts.

Closes #5725

## Validation

- go test ./core/tracing
- go test ./core/...
- go vet ./core/tracing
- git diff --check

Google CLA is required for this contribution.


MichaelDoyle:
Thanks @mikemikimike. A couple things.

1. We do still need to preserve the base 64 for local development. So this needs to be either gated by GENKIT_ENV!=dev, or optional some other way.

2. We're actually heading toward a path of allowing more custom control of telemetry. Starting with Dart and Python, followed by Go and TS. So anything done here will be short lived, as an FYI.

PR #5969, opened 2026-08-06
feat(py): make ModelRef generic over ModelRefConfigT

huangjeff5:
### What this does
Makes ModelRef a typed, immutable reference to a model, generic over its config schema. A plugin can now export refs like gemini_25_flash: ModelRef[GeminiConfig], and users who pass a config of the wrong shape find out immediately — at construction, with a clear error — instead of deep inside a request, or worse, silently.

### For SDK users
```python
from genkit.model import model_ref

ref = model_ref(
    'gemini-2.5-flash',
    namespace='googleai',
    config_schema=GeminiConfig,
    config=GeminiConfig(temperature=0.7),
)
```
Your editor autocompletes `ref.config` when the ref is ModelRef[GeminiConfig], and rejects a dict passed in as config.
Passing a config that isn't an instance of config_schema raises GenkitError with status INVALID_ARGUMENT at construction time, naming the expected and actual types.

Refs are frozen: mutating a field raises FrozenInstanceError, so a ref shared across your codebase can't be changed out from under you. Two refs with the same fields compare equal.

### For plugin authors
Export typed refs alongside your models. ModelRefConfigT is covariant and bound to pydantic.BaseModel, so a ModelRef[GeminiConfig] is assignable wherever a ModelRef[BaseModel] is accepted.

### Breaking changes
model_ref() now requires a config_schema keyword argument, and all arguments after name are keyword-only. Before → after: model_ref('m', namespace='ns') → model_ref('m', namespace='ns', config_schema=MyConfig).
config must be an instance of that schema (previously a plain dict).
ModelRef is now an unhashable frozen dataclass rather than a mutable Pydantic model; use its fields, not .model_dump().

### Notes
This is the foundation PR of a stack extracted from #5854 / the original #5965: it deliberately changes no call sites. Accepting ModelRef at generate(), define_prompt(), and agent surfaces lands separately (#5977), on top of a small helpers PR that carries the config-merge semantics.

### Test plan
py/packages/genkit/tests/genkit/ai/model_ref_test.py covers construction, namespace prefixing, typed-config validation (including the INVALID_ARGUMENT path), frozen-ness, equality, and the missing-kwarg TypeError.

### DIFF
--- a/py/packages/genkit/src/genkit/_ai/_model.py
+++ b/py/packages/genkit/src/genkit/_ai/_model.py
@@ -33,6 +33,7 @@ from genkit._core._model import (
     Message,
     ModelConfig,
     ModelRef,
+    ModelRefConfigT,
     ModelRequest,
     ModelResponse,
     ModelResponseChunk,
@@ -67,16 +68,23 @@ def model_action_metadata(
 
 def model_ref(
     name: str,
+    *,
+    config_schema: type[ModelRefConfigT],
     namespace: str | None = None,
     info: ModelInfo | None = None,
     version: str | None = None,
-    config: dict[str, object] | None = None,
-) -> ModelRef:
+    config: ModelRefConfigT | None = None,
+) -> ModelRef[ModelRefConfigT]:
     """Create a ModelRef, optionally prefixing name with namespace."""
-    # Logic: if (options.namespace && !name.startsWith(options.namespace + '/'))
     final_name = f'{namespace}/{name}' if namespace and not name.startswith(f'{namespace}/') else name
 
-    return ModelRef(name=final_name, info=info, version=version, config=config)
+    return ModelRef[ModelRefConfigT](
+        name=final_name,
+        config_schema=config_schema,
+        info=info,
+        version=version,
+        config=config,
+    )
 
 
 def define_model(
--- a/py/packages/genkit/src/genkit/_ai/_model.py
+++ b/py/packages/genkit/src/genkit/_ai/_model.py
@@ -78,7 +78,7 @@ def model_ref(
     """Create a ModelRef, optionally prefixing name with namespace."""
     final_name = f'{namespace}/{name}' if namespace and not name.startswith(f'{namespace}/') else name
 
-    return ModelRef[ModelRefConfigT](
+    return ModelRef(
         name=final_name,
         config_schema=config_schema,
         info=info,
--- a/py/packages/genkit/src/genkit/_core/_model.py
+++ b/py/packages/genkit/src/genkit/_core/_model.py
@@ -74,13 +74,26 @@ ModelRefConfigT = TypeVar('ModelRefConfigT', bound=BaseModel, covariant=True)
 
 @dataclass(frozen=True, kw_only=True)
 class ModelRef(Generic[ModelRefConfigT]):
-    """Frozen reference to a model, optionally tied to a config schema."""
+    """Frozen reference to a model tied to a config schema."""
 
     name: str
     config_schema: type[ModelRefConfigT]
     info: ModelInfo | None = None
     version: str | None = None
     config: ModelRefConfigT | None = None
+
+    # Config schemas are Pydantic models and usually aren't hashable, so a
+    # generated __hash__ would work when config is None and blow up once one is
+    # set — a nasty footgun if someone keys a cache or set by ModelRef.
+    __hash__ = None  # type: ignore[assignment]
+
+    def __post_init__(self) -> None:
+        # Catch dicts / wrong types at construction so plugins don't discover
+        # the mismatch later when they touch typed config fields.
+        if self.config is not None and not isinstance(self.config, self.config_schema):
+            raise TypeError(
+                f'config must be an instance of {self.config_schema.__name__}, got {type(self.config).__name__}'
+            )
 
 
 class Message(MessageData):
--- a/py/packages/genkit/src/genkit/_core/_model.py
+++ b/py/packages/genkit/src/genkit/_core/_model.py
@@ -75,7 +75,12 @@ ModelRefConfigT = TypeVar('ModelRefConfigT', bound=BaseModel, covariant=True)
 
 @dataclass(frozen=True, kw_only=True)
 class ModelRef(Generic[ModelRefConfigT]):
-    """Frozen reference to a model tied to a config schema."""
+    """Handle for a model tied to a config schema.
+
+    Fields cannot be rebound. config and info are copied at construction so later
+    mutations of the caller's objects don't change the ref; the copies themselves
+    stay ordinary mutable Pydantic models.
+    """
 
     name: str
     config_schema: type[ModelRefConfigT]
@@ -114,6 +119,12 @@ class ModelRef(Generic[ModelRefConfigT]):
                 status='INVALID_ARGUMENT',
                 message=(f'{self.name}: info must be an instance of {ModelInfo.__module__}.ModelInfo, got {actual}'),
             )
+        # Callers often keep the config/info they passed in. Copy so later
+        # mutations of those objects don't change the ref's defaults.
+        if self.config is not None:
+            object.__setattr__(self, 'config', self.config.model_copy(deep=True))
+        if self.info is not None:
+            object.__setattr__(self, 'info', self.info.model_copy(deep=True))
 
 
 class Message(MessageData):
--- a/py/packages/genkit/src/genkit/_core/_model.py
+++ b/py/packages/genkit/src/genkit/_core/_model.py
@@ -82,9 +82,8 @@ class ModelRef(Generic[ModelRefConfigT]):
     version: str | None = None
     config: ModelRefConfigT | None = None
 
-    # Config schemas are Pydantic models and usually aren't hashable, so a
-    # generated __hash__ would work when config is None and blow up once one is
-    # set — a nasty footgun if someone keys a cache or set by ModelRef.
+    # Explicitly opt out of hashing: Pydantic configs are unhashable, so an
+    # auto-generated __hash__ would fail once set.
     __hash__ = None  # type: ignore[assignment]
 
     def __post_init__(self) -> None:
--- a/py/packages/genkit/src/genkit/_core/_model.py
+++ b/py/packages/genkit/src/genkit/_core/_model.py
@@ -88,10 +88,22 @@ class ModelRef(Generic[ModelRefConfigT]):
     __hash__ = None  # type: ignore[assignment]
 
     def __post_init__(self) -> None:
-        # Catch dicts / wrong types at construction so plugins don't discover
-        # the mismatch later when they touch typed config fields.
-        if self.config is not None and not isinstance(self.config, self.config_schema):
-            expected = f'{self.config_schema.__module__}.{self.config_schema.__name__}'
+        # Schema first: isinstance(config, schema) raises TypeError if schema
+        # isn't a class — the JSON-schema dict a plugin would paste in from
+        # action metadata.
+        schema = self.config_schema
+        if not isinstance(schema, type) or not issubclass(schema, BaseModel):
+            got = (
+                f'{schema.__module__}.{schema.__name__}'
+                if isinstance(schema, type)
+                else f'{type(schema).__module__}.{type(schema).__name__}'
+            )
+            raise GenkitError(
+                status='INVALID_ARGUMENT',
+                message=f'{self.name}: config_schema must be a BaseModel subclass, got {got}',
+            )
+        if self.config is not None and not isinstance(self.config, schema):
+            expected = f'{schema.__module__}.{schema.__name__}'
             actual = f'{type(self.config).__module__}.{type(self.config).__name__}'
             raise GenkitError(
                 status='INVALID_ARGUMENT',
--- a/py/packages/genkit/src/genkit/_core/_model.py
+++ b/py/packages/genkit/src/genkit/_core/_model.py
@@ -91,10 +91,11 @@ class ModelRef(Generic[ModelRefConfigT]):
         # Catch dicts / wrong types at construction so plugins don't discover
         # the mismatch later when they touch typed config fields.
         if self.config is not None and not isinstance(self.config, self.config_schema):
-            config_type = type(self.config).__name__
+            expected = f'{self.config_schema.__module__}.{self.config_schema.__name__}'
+            actual = f'{type(self.config).__module__}.{type(self.config).__name__}'
             raise GenkitError(
                 status='INVALID_ARGUMENT',
-                message=f'config must be an instance of {self.config_schema.__name__}, got {config_type}',
+                message=f'{self.name}: config must be an instance of {expected}, got {actual}',
             )
 
 
--- a/py/packages/genkit/src/genkit/_core/_model.py
+++ b/py/packages/genkit/src/genkit/_core/_model.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 
 from collections.abc import Callable, Sequence
 from copy import deepcopy
+from dataclasses import dataclass
 from functools import cached_property
 from typing import Any, ClassVar, Generic, cast
 
@@ -47,6 +48,7 @@ from genkit._core._typing import (
     MediaPart,
     MessageData,
     MiddlewareRef,
+    ModelInfo,
     ModelResponseChunk as ModelResponseChunkSchema,
     Operation,
     Part,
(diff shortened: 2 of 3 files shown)


py/packages/genkit/src/genkit/_ai/_model.py
```diff
@@ -67,16 +68,23 @@ def model_action_metadata(
...
     final_name = f'{namespace}/{name}' if namespace and not name.startswith(f'{namespace}/') else name
 
-    return ModelRef(name=final_name, info=info, version=version, config=config)
+    return ModelRef[ModelRefConfigT](
+        name=final_name,
+        config_schema=config_schema,
+        info=info,
+        version=version,
+        config=config,
+    )
```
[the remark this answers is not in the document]
huangjeff5: Intentionally avoided allowing dict[str, Any]. model_ref is a helper on the plugin author side, so the ergonomic benefits are relatively low compared to the maintenance overhead.

py/packages/genkit/tests/genkit/ai/model_ref_test.py
```diff
@@ -0,0 +1,107 @@
...
+    assert isinstance(ref, ModelRef)
+    assert ref.name == 'googleai/gemini-pro-latest'
+    assert ref.config_schema is CustomConfig
+    assert ref.config is config
+    assert ref.config is not None
+    assert ref.config.temperature == 0.7
+    assert ref.config.top_p == 0.9
+
+
+def test_model_ref_with_bare_base_model_schema() -> None:
```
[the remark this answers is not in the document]
huangjeff5: Not fixing for same reason as #1. Don't want to support union types unless it improves ergonomics on the app developer's API surface.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +67,20 @@
 # TypeVars for generic types
 OutputT = TypeVar('OutputT', default=object)
 ConfigT = TypeVar('ConfigT', bound=ModelConfig, default=ModelConfig)
+# Bound to BaseModel so ModelRef is always parameterized with a concrete Pydantic config schema.
+# Covariant so ModelRef[GeminiConfig] is assignable to ModelRef[BaseModel] or ModelRef[Any].
+ModelRefConfigT = TypeVar('ModelRefConfigT', bound=BaseModel, covariant=True)
```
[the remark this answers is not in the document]
huangjeff5: Skipping default=BaseModel on purpose.

model_ref() / ModelRef construction is a plugin-author API — they always pass a concrete config_schema (e.g. GeminiConfig). App developers don't construct ModelRef; they use a model string or a helper like google_ai.model(...) that returns a ready-made ref.

A TypeVar default only makes bare ModelRef mean ModelRef[BaseModel] in annotations. That isn't an expected app path, and we'd rather keep the parameterization explicit (ModelRef[BaseModel] when you mean "any schema").

py/packages/genkit/src/genkit/_ai/_model.py
```diff
@@ -67,16 +68,23 @@ def model_action_metadata(
...
-    config: dict[str, object] | None = None,
-) -> ModelRef:
+    config: ModelRefConfigT | None = None,
+) -> ModelRef[ModelRefConfigT]:
     """Create a ModelRef, optionally prefixing name with namespace."""
-    # Logic: if (options.namespace && !name.startsWith(options.namespace + '/'))
     final_name = f'{namespace}/{name}' if namespace and not name.startswith(f'{namespace}/') else name
 
-    return ModelRef(name=final_name, info=info, version=version, config=config)
+    return ModelRef[ModelRefConfigT](
```
apascal07: Subscripting buys nothing here beyond extra work. `_GenericAlias.__call__` instantiates the origin, then attempts `result.__orig_class__ = self`, which raises `FrozenInstanceError` on a frozen dataclass and is swallowed; I confirmed the returned instance has no `__orig_class__`. Plain `ModelRef(...)` infers the same `ModelRef[ModelRefConfigT]` from `config_schema`.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +67,20 @@
...
 
 
-class ModelRef(BaseModel):
-    """Reference to a model with configuration."""
+@dataclass(frozen=True, kw_only=True)
```
apascal07, line 76: `frozen=True` with the default `eq=True` generates a `__hash__`, but Pydantic models are not hashable by default, so `hash(ref)` raises `TypeError: unhashable type` as soon as `config` is set (it succeeds when `config is None`). Refs look hashable to anything that wants to key a dict or dedupe a set, which is plausible once generate/prompt wiring caches them, and the error points at the config class rather than at this decision.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +67,20 @@
...
+    """Frozen reference to a model, optionally tied to a config schema."""
 
     name: str
-    config_schema: object | None = None
-    info: object | None = None
+    config_schema: type[ModelRefConfigT]
+    info: ModelInfo | None = None
     version: str | None = None
-    config: dict[str, object] | None = None
+    config: ModelRefConfigT | None = None
```
apascal07, line 89: Separate from the coercion thread above (agreed on not taking `dict` in the signature): a dict is currently accepted and stored silently at runtime, since the dataclass validates nothing. Statically-checked plugin code is safe, but a ref built from JSON or dynamic model resolution gets a `dict` in a field typed `ModelRefConfigT` and fails later at attribute access. Raising in `__post_init__` when `config` is not an instance of `config_schema` keeps the narrow signature and makes the failure land here.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +67,20 @@
...
+# Bound to BaseModel so ModelRef is always parameterized with a concrete Pydantic config schema.
+# Covariant so ModelRef[GeminiConfig] is assignable to ModelRef[BaseModel] or ModelRef[Any].
+ModelRefConfigT = TypeVar('ModelRefConfigT', bound=BaseModel, covariant=True)
 
 
-class ModelRef(BaseModel):
-    """Reference to a model with configuration."""
+@dataclass(frozen=True, kw_only=True)
+class ModelRef(Generic[ModelRefConfigT]):
+    """Frozen reference to a model, optionally tied to a config schema."""
```
apascal07: nit: `config_schema` is required, so "optionally tied to a config schema" says the opposite.

py/packages/genkit/tests/genkit/ai/model_ref_test.py
```diff
@@ -0,0 +1,107 @@
...
+        namespace='googleai',
+        version='001',
+        info=info,
+    )
+
+    assert ref.name == 'googleai/veo-2'
+    assert ref.version == '001'
+    assert ref.info is info
+    assert ref.info.supports is not None
+    assert ref.info.supports.multiturn is True
```
apascal07, line 148: Pyright errors on both lines: `assert ref.info is info` does not narrow `ModelInfo | None`, so `.supports` is `reportOptionalMemberAccess`. CI only runs pyright over `packages/*/src`, so it stays green, but an `assert ref.info is not None` clears it.

py/packages/genkit/tests/genkit/ai/model_ref_test.py
```diff
@@ -0,0 +1,107 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for ModelRef and model_ref()."""
+
+from dataclasses import FrozenInstanceError
+
+import pytest
+from pydantic import BaseModel
+
+from genkit._core._typing import ModelInfo, Supports
```
apascal07: `ModelInfo` and `Supports` are re-exported from `genkit.model`, which is where a plugin author would import them from.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Good call—added __post_init__ to ModelRef so passing an unvalidated dict raises TypeError right at creation.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Fixed, updated docstring.

py/packages/genkit/tests/genkit/ai/model_ref_test.py
huangjeff5: Updated import to genkit.model.

py/packages/genkit/tests/genkit/ai/model_ref_test.py
huangjeff5: Added assert ref.info is not None to clear the optional member access warning.

huangjeff5:
@gemini review

py/packages/genkit/src/genkit/_ai/_model.py
huangjeff5: Good catch, removed that.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Makes sense. Opted out of hash with `__hash__ = None`.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +68,34 @@
...
+    config: ModelRefConfigT | None = None
+
+    # Explicitly opt out of hashing: Pydantic configs are unhashable, so an
+    # auto-generated __hash__ would fail once set.
+    __hash__ = None  # type: ignore[assignment]
+
+    def __post_init__(self) -> None:
+        # Catch dicts / wrong types at construction so plugins don't discover
+        # the mismatch later when they touch typed config fields.
+        if self.config is not None and not isinstance(self.config, self.config_schema):
```
apascal07: `config_schema` itself is never checked. `model_ref('m', config_schema={'type': 'object'}, config=SomeConfig())` raises `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union` out of this line, and with `config=None` the bad schema is stored silently and only surfaces when someone later attaches a config. The old field type was `object | None`, so a schema that isn't a `BaseModel` subclass is exactly what a migrating caller has on hand.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +68,34 @@
...
+            config_type = type(self.config).__name__
+            raise GenkitError(
+                status='INVALID_ARGUMENT',
+                message=f'config must be an instance of {self.config_schema.__name__}, got {config_type}',
```
apascal07: nit: two identically named classes from different modules produce `config must be an instance of GeminiConfig, got GeminiConfig`. Qualifying both names with `__module__` and naming the ref makes it identifiable when it's one of forty built at import time.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +68,34 @@
...
-    """Reference to a model with configuration."""
+@dataclass(frozen=True, kw_only=True)
+class ModelRef(Generic[ModelRefConfigT]):
+    """Frozen reference to a model tied to a config schema."""
 
     name: str
-    config_schema: object | None = None
-    info: object | None = None
+    config_schema: type[ModelRefConfigT]
+    info: ModelInfo | None = None
```
apascal07, line 87: `info` also tightened, from `object | None` to `ModelInfo | None`, but nothing enforces it now that Pydantic is gone: `ModelRef(name=..., config_schema=..., info={'not': 'a ModelInfo'})` is accepted and fails later at `info.supports`. Same failure mode `config` now guards against, and plugins here already carry model info as plain dicts (`ollama_model_info` returns `dict[str, object]`, `google.py` passes `model_ref.model_dump(by_alias=True)` as `info`), so a dict reaching this field is the likely path. Either check it alongside `config`, or say in the docstring that `config` is the only field validated at runtime.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +68,34 @@
...
-class ModelRef(BaseModel):
-    """Reference to a model with configuration."""
+@dataclass(frozen=True, kw_only=True)
+class ModelRef(Generic[ModelRefConfigT]):
+    """Frozen reference to a model tied to a config schema."""
 
     name: str
-    config_schema: object | None = None
-    info: object | None = None
+    config_schema: type[ModelRefConfigT]
```
apascal07, line 86: `config_schema` went from defaulting to `None` to always holding a class object, which takes a ref out of anything that serializes. A Pydantic model with a `ModelRef[X]` field constructs fine, but `model_dump()` yields `'config_schema': <class 'X'>`, and `dataclasses.asdict` does the same, so neither result is JSON-serializable. Worth settling before #5977: if a ref reaches `generate()` and the options land in a trace or the reflection API, that's a TypeError at serialization rather than at construction.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -65,16 +68,34 @@
 # TypeVars for generic types
 OutputT = TypeVar('OutputT', default=object)
 ConfigT = TypeVar('ConfigT', bound=ModelConfig, default=ModelConfig)
+# Bound to BaseModel so ModelRef is always parameterized with a concrete Pydantic config schema.
+# Covariant so ModelRef[GeminiConfig] is assignable to ModelRef[BaseModel] or ModelRef[Any].
+ModelRefConfigT = TypeVar('ModelRefConfigT', bound=BaseModel, covariant=True)
```
apascal07, line 73: `ModelRefConfigT` isn't re-exported from `genkit.plugin_api` alongside `ModelRef`, so a plugin author writing anything generic over the schema (`def with_defaults(ref: ModelRef[ModelRefConfigT]) -> ModelRef[ModelRefConfigT]`) has to import from `genkit._core._model` or redeclare the TypeVar with matching variance.

py/packages/genkit/src/genkit/_ai/_model.py
```diff
@@ -67,16 +68,23 @@ def model_action_metadata(
...
 def model_ref(
     name: str,
+    *,
+    config_schema: type[ModelRefConfigT],
     namespace: str | None = None,
     info: ModelInfo | None = None,
     version: str | None = None,
-    config: dict[str, object] | None = None,
-) -> ModelRef:
+    config: ModelRefConfigT | None = None,
```
apascal07, line 76: `ModelRefConfigT` is solved from both `config_schema` and `config`, so a mismatch isn't a static error: pyright infers `ModelRef[AConfig | BConfig]` for `model_ref('m', config_schema=AConfig, config=BConfig())` and reports nothing (same result with an invariant TypeVar). `__post_init__` is the only thing that catches it. What does hold statically is `ref.config.<field>` and rejection of a `dict`, so the PR description's "your editor type-checks config against the plugin's schema" is worth narrowing before it becomes release-note material.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: It's not clear that this is something developers actually want to do. The main use case I'm aware of is enabling plugin authors to expose typed ModelRef factories - usually one per model family. In that situation, developers might write `ModelRef[GeminiConfig]` as the return output annotation. However I haven't encountered a use case where they need to reference `ModelRefConfigT` directly.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Hmm I went back and forth on this one... I think it makes sense to check it alongside config. Because even though technically there are static warnings against using a dict, they're just warnings and you can still write the wrong code without a runtime check, and this is an axes where 3P's contribute code regularly.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Good catch. Updated.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: ModelRef is only passed in at the veneer layer, and then unwrapped before it touches the generate action (see _aio.py l1077-1085):

```
if isinstance(model, ModelRef): 
  resolved = resolve_model_ref(model=model, config=normalized_config)
  ...
  model=resolved.name,
```

`GenerationActionOptions` is the generate action's input and what actually gets trace/dumped. `GenerateActionOptions.model` is still `str | None`.

py/packages/genkit/src/genkit/_ai/_model.py
huangjeff5: Makes sense. Updated description.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Makes sense. Note that there are no existing users of ModelRef right now (the veneer layers don't accept it, not part of the plugin/sample surface either), so let's not worry about migrating callers.

However, I do agree there should be a runtime check here because static warnings on the IDE can get ignored in Python (they are warnings only since Python doesn't have a compilation step). Added that.

apascal07 (approved):

PR #5972, opened 2026-08-06
feat(js): Allow options for checkOperation and cancelOperation

ifielker:
Also scrub auth and secrets from context logging.

Checklist (if applicable):

### DIFF
--- a/js/ai/src/model.ts
+++ b/js/ai/src/model.ts
@@ -17,6 +17,7 @@
 import {
   ActionFnArg,
   BackgroundAction,
+  BackgroundActionFnArg,
   GenkitError,
   MiddlewareWithOptions,
   Operation,
@@ -276,8 +277,6 @@ export function defineModel<
 }
 
 /** Options for defining a background model that processes requests asynchronously. */
-import { BackgroundActionFnArg } from '@genkit-ai/core';
-
 export type DefineBackgroundModelOptions<
   CustomOptionsSchema extends z.ZodTypeAny = z.ZodTypeAny,
 > = DefineModelOptions<CustomOptionsSchema> & {
--- a/js/ai/src/model.ts
+++ b/js/ai/src/model.ts
@@ -276,6 +276,8 @@ export function defineModel<
 }
 
 /** Options for defining a background model that processes requests asynchronously. */
+import { BackgroundActionFnArg } from '@genkit-ai/core';
+
 export type DefineBackgroundModelOptions<
   CustomOptionsSchema extends z.ZodTypeAny = z.ZodTypeAny,
 > = DefineModelOptions<CustomOptionsSchema> & {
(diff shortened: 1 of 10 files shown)


js/ai/src/model.ts
```diff
@@ -276,17 +276,21 @@ export function defineModel<
 }
 
 /** Options for defining a background model that processes requests asynchronously. */
+import { BackgroundActionFnArg } from '@genkit-ai/core';
```
pavelgj: This import landed in the middle of the file, which detaches the JSDoc comment right below from `DefineBackgroundModelOptions` (it now reads as documenting the import). Could we move `BackgroundActionFnArg` into the existing `@genkit-ai/core` import block at the top instead?

```ts
import {
  ActionFnArg,
  BackgroundAction,
  BackgroundActionFnArg,
  GenkitError,
  ...
} from '@genkit-ai/core';
```

js/ai/src/model.ts
```diff
@@ -276,17 +276,21 @@ export function defineModel<
...
 export type DefineBackgroundModelOptions<
   CustomOptionsSchema extends z.ZodTypeAny = z.ZodTypeAny,
 > = DefineModelOptions<CustomOptionsSchema> & {
   start: (
     request: GenerateRequest<CustomOptionsSchema>
   ) => Promise<Operation<GenerateResponseData>>;
   check: (
-    operation: Operation<GenerateResponseData>
+    operation: Operation<GenerateResponseData>,
+    options?: BackgroundActionFnArg<unknown>
```
pavelgj, line 289: Small consistency thing: here the callback options are `options?: BackgroundActionFnArg<unknown>` (optional + `unknown`), but in core `BackgroundActionParams` they're `options: BackgroundActionFnArg<z.infer<S>>` (required + generic). Might be worth aligning these so the types match end to end.

pavelgj (approved):

js/ai/src/model.ts
ifielker: I caught that 3 times, but I guess the fourth one snuck past me. Thanks! Fixed.

js/ai/src/model.ts
ifielker: We looked into aligning them, but I think they actually need to stay different because they serve two distinct layers of the framework:

1. __`BackgroundActionParams`__ is the internal Core primitive. We made `options` strictly required here because the core `action()` wrapper guarantees it will always synthesize and inject that object (to pass mandatory `trace` data).
2. __`DefineBackgroundModelOptions`__ is the public API for plugin authors. We intentionally made `options` optional here to avoid breaking backwards compatibility for existing plugins that haven't updated their `check`/`cancel` callbacks to accept a second argument. Additionally, we use `unknown` instead of `z.infer<S>` because background models don't stream chunks, meaning there is no streaming generic `S` available at the model layer to infer.

js/ai/src/model.ts
```diff
@@ -338,18 +342,18 @@ export function backgroundModel<
...
           if (!options.cancel) {
             throw new GenkitError({
               status: 'UNIMPLEMENTED',
               message: 'cancel not implemented',
             });
           }
-          return options.cancel(op);
+          return options.cancel(op, opts);
         }
       : undefined,
```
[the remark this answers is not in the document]
ifielker: pre-existing

js/ai/src/model.ts
```diff
@@ -276,17 +276,21 @@ export function defineModel<
 }
 
 /** Options for defining a background model that processes requests asynchronously. */
+import { BackgroundActionFnArg } from '@genkit-ai/core';
+
 export type DefineBackgroundModelOptions<
```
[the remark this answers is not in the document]
ifielker: fixed

PR #5976, opened 2026-08-06
feat(go/plugins/compat_oai): rebuild the core on typed configs and migrate the openai plugin

apascal07:
Rebuilds the `compat_oai` core on the typed-config constructors and migrates the `openai` plugin. Base of a seven-PR train: `anthropic`, `dashscope`, `kimi`, and `zai` migrate in their own PRs stacked on this one, then two new plugins (`deepseek`, `xai`). Stacked on #6027 and built on its shared helpers (`SchemaOverrides`, `ModelOptions.Overlay`, `LookupOverride`, `ProviderLabel`, `WalkSubschemas`, `core/status`). This PR previously carried the whole family in one diff; the sub-plugin PRs now hold their pieces.

Until now every plugin in the family took `any` as config and reconciled it at request time: a wrong field surfaced as the provider's error, and the Dev UI had no schema to render.

## The plugin and its config, at a glance

```go
// The base plugin every sub-plugin embeds.
type OpenAICompatible struct {
	Opts       []option.RequestOption
	Provider   string
	APIKey     string
	BaseURL    string
	ListModels func(ctx context.Context, client *openai.Client) ([]string, error)
}

// The contract a sub-plugin model config implements; the last two methods
// come from embedding RequestConfig.
type ChatConfig interface {
	ApplyToChatCompletion(params *openai.ChatCompletionNewParams)
	RequestAPIKey() string
	RequestExtra() map[string]any
}

// Embedded by every sub-plugin config: the settings Genkit owns.
type RequestConfig struct {
	APIKey  string         `json:"-"`
	Version string         `json:"version,omitempty"`
	Extra   map[string]any `json:"extra,omitempty"`
}

// The embedder config, shared by the family.
type EmbeddingConfig struct {
	Dimensions     int                                     `json:"dimensions,omitempty"`
	EncodingFormat openai.EmbeddingNewParamsEncodingFormat `json:"encodingFormat,omitempty"`
	APIKey         string                                  `json:"-"`
	User           string                                  `json:"user,omitempty"`
	Extra          map[string]any                          `json:"extra,omitempty"`
}

// The openai plugin. Its model config is the SDK's own ChatCompletionNewParams.
type OpenAI struct {
	APIKey    string
	Opts      []option.RequestOption
	Models    map[string]ai.ModelOptions
	Embedders map[string]ai.EmbedderOptions
}

type TextEmbeddingConfig = compat_oai.EmbeddingConfig
```

## Each plugin declares its own config fields

No shared generation struct: the providers disagree (DeepSeek takes neither penalty, Kimi's K-series drops `temperature` too, Z.ai caps `temperature` at 1 where OpenAI allows 2, `maxOutputTokens` is `max_tokens` on some wires and `max_completion_tokens` on others), and Go cannot subtract an embedded field, so a shared struct would advertise fields a provider rejects. Each plugin declares what its provider's API reference lists and maps it onto the request:

```go
type ChatConfig struct {
    compat_oai.RequestConfig

    Temperature     *float64 `json:"temperature,omitempty" jsonschema:"minimum=0,maximum=2" jsonschema_description:"Controls the degree of randomness in token selection, from 0 to 2."`
    MaxOutputTokens int      `json:"maxOutputTokens,omitempty" jsonschema:"minimum=1" jsonschema_description:"Maximum number of tokens to generate, sent as the API's max_tokens."`
}

func (c ChatConfig) ApplyToChatCompletion(params *openai.ChatCompletionNewParams) {
    c.ApplyVersion(params)
    if c.Temperature != nil {
        params.Temperature = openai.Float(*c.Temperature)
    }
    if c.MaxOutputTokens > 0 {
        params.MaxTokens = openai.Int(int64(c.MaxOutputTokens))
    }
}
```

The `jsonschema` tags mirror the provider's documented hard limits, so an out-of-range value fails with the field named before it is sent and billed, and the descriptions become Dev UI tooltips. `RequestConfig` exposes `ApplyVersion` rather than the interface method on purpose: a plugin that inherited a full `ApplyToChatCompletion` and forgot its own would compile and silently drop every field it declared. Field names are camelCase and identical across plugins, pinned by a conformance test that grows an entry per PR; `openai` is exempt because its config is the SDK type.

**Before:** every config from every plugin went through one request-time reconciliation path.

```go
case map[string]any:
    normalizedConfig := make(map[string]any, len(cfg))
    for key, value := range cfg {
        normalizedConfig[key] = value
    }
    for source, target := range map[string]string{
        "frequencyPenalty": "frequency_penalty",
        "logProbs":         "logprobs",
        "presencePenalty":  "presence_penalty",
        "stopSequences":    "stop",
        "topLogProbs":      "top_logprobs",
        "topP":             "top_p",
    } {
        if value, ok := normalizedConfig[source]; ok {
            normalizedConfig[target] = value
            delete(normalizedConfig, source)
        }
    }
    for _, key := range []string{
        "apiKey", "maxOutputTokens", "topK", "version", "visualDetailLevel",
    } {
        delete(normalizedConfig, key)
    }
    openaiConfig, err = base.MapToStruct[openai.ChatCompletionNewParams](normalizedConfig)
    ...
    extraFields := make(map[string]any, len(normalizedConfig))
    for key, value := range normalizedConfig {
        if _, standard := chatCompletionParamFields[key]; standard {
            continue
        }
        extraFields[key] = value
    }
    openaiConfig.SetExtraFields(extraFields)
```

**After:** the type parameter is the contract, and the framework validates every request against its schema.

```go
compat_oai.NewChatModel[ChatConfig](&p.openAICompatible, model, opts)
```

The before path survives behind the deprecated `DefineModel` until the train's last migration (below).

## Any model resolves; `Models` corrects the catalog

`ListChatActions`/`ResolveChatAction` (and the `ListModelActions`/`ResolveModelAction` SDK-config counterparts) take a `modelOptions func(id string) ai.ModelOptions`: curated IDs keep their capabilities, anything else resolves with generic defaults. The plugin's `Models` map (keys bare or provider-prefixed) overlays corrections onto every path that describes a model, same shape as `googlegenai` and native `anthropic`, and no `Register*` surface exists:

```go
&openai.OpenAI{Models: map[string]ai.ModelOptions{
    "gpt-4o": {Supports: &ai.ModelSupports{Multiturn: true, Tools: true}},
}}
```

Two fixes fell out: `validateVersion` rejected any `version` pin on a model that declared no versions (an empty list now means unconstrained), and `openai`'s dynamic resolution used the base's generic defaults instead of its own catalog.

## Fields Genkit owns are hidden, and rejected by name

The SDK-config models advertised all of `ChatCompletionNewParams`, including the five fields Genkit builds from the request (`messages`, `tools`, `tool_choice`, `functions`, `function_call`). On #6027's `SchemaOverrides` they are now hidden from the advertised schema (replaced by the permissive `true` schema, as `googlegenai` and native `anthropic` do), and a value that arrives anyway is refused by name:

```
tools must be set using Genkit feature: ai.WithTools()
```

`response_format` joins the managed set (a config or extra carrying it would marshal over what Genkit computed and break structured output), and so does `n` (the response path reads `Choices[0]` only, so extra candidates are billed and dropped; `googlegenai` hides `candidateCount` for the same reason). Every visible field carries a description from OpenAI's API reference; `model` stays visible as the version pin. `WithParams` and the deprecated `WithConfig` still clear managed fields for callers driving `ModelGenerator` directly.

## Undeclared provider fields ride `extra`

```go
ai.WithConfig(map[string]any{
    "temperature": 0.2,
    "extra": map[string]any{"safe_mode": true}, // undeclared, sent verbatim
})
```

Every curated config inherits `extra` from `RequestConfig`: forwarded verbatim at the top level of the body, colliding key wins (openai-go's documented `SetExtraFields` precedence, and the JS config spread's). Managed fields are rejected inside it too, and `clearManagedFields` sweeps them out of the SDK extras so they cannot resurrect at marshal time. The SDK-typed models carry no `extra`: their schema is reflected from the linked openai-go, so bumping the SDK widens it.

## The OpenAI catalog was three generations stale

12 curated models become 24: the GPT-5.6 (`sol`/`terra`/`luna`), 5.5, 5.4, 5.2, 5.1, and 5 families plus `o3` and `o4-mini` arrive; `o1-preview`/`o1-mini` (shut down in 2025) and the retired GPT-4 snapshots leave, and three wrong capability sets are corrected against the catalog. Constrained generation is claimed only where OpenAI documents `json_schema`, from the `gpt-4o-mini` and `gpt-4o-2024-08-06` snapshots on. `-pro` tiers stay out (Responses API only). Seven entries carry announced 2026-10-23 shutdowns in comments; still live, so they stay.

## Streamed generations reported zero tokens

Nothing in the family set `stream_options.include_usage`, so every stream reported empty usage (the JS adapter has always sent it). It is requested once in `generateStream`, and the final usage chunk is kept whole because the SDK accumulator drops `completion_tokens_details` and any provider field it does not model. The response path also gains mappings the later PRs pin live: `prompt_cache_hit_tokens` as the cache-breakdown fallback, `end_turn` under `FinishReasonStop`, `insufficient_system_resource` under `FinishReasonOther`, provider metadata on `Raw`.

## The pre-typed model path stays until the last migration

`OpenAICompatible.DefineModel` keeps its released untyped behavior (map configs, unknown keys forwarded verbatim, nothing validated) and is deprecated pointing at `NewModel`: the sub-plugins still on it route thinking flags through those extra keys, and `main` must stay green at every prefix of the train. `TestDefineModelKeepsConfigUntyped` pins the split end to end: the same undeclared key rides to the wire from a `DefineModel` model and fails the boundary schema on a `NewModel` one.

## Not a breaking change

Everything `go/v1.11.0` shipped from this tree keeps its signature: `openai.DefineModel`, `.DefineEmbedder`, `.EmbedderRef`, `.Model`, `.Embedder`, the base's `DefineEmbedder`/`IsDefinedModel`/`IsDefinedEmbedder`, and `ModelGenerator.WithConfig` are deprecated in place with released behavior. One source break, in unkeyed literals: `openai.TextEmbeddingConfig` becomes an alias of `compat_oai.EmbeddingConfig` and gains `APIKey` and `User`.

```go
openai.TextEmbeddingConfig{256, "float"}   // compiled against v1.11.0, no longer does
```

Keyed literals, the wire format, embedding, and assertions are unchanged; this is the [Go 1 compatibility](https://go.dev/doc/go1compat) struct-literal carve-out, and `go vet` flags unkeyed literals of imported structs. Runtime tightens on `openai` only: managed config fields fail loudly instead of misbehaving silently, `encodingFormat` takes only the documented `float`/`base64`, and streams gain the token counts they left at zero.

## Public API surface

```go
// compat_oai: added
type ChatConfig interface { ApplyToChatCompletion(*openai.ChatCompletionNewParams); RequestAPIKey() string; RequestExtra() map[string]any }
type RequestConfig struct { APIKey string `json:"-"`; Version string; Extra map[string]any }
type EmbeddingConfig struct { Dimensions int; EncodingFormat ...; APIKey string `json:"-"`; User string; Extra map[string]any }
func (RequestConfig) ApplyVersion(*openai.ChatCompletionNewParams)
func (RequestConfig) RequestAPIKey() string
func (RequestConfig) RequestExtra() map[string]any
func AddExtraFields(*openai.ChatCompletionNewParams, map[string]any)
func (*ModelGenerator) WithOutputFormats([]string) *ModelGenerator
func ActionName(provider, id string) string
func DefaultModelOptions() ai.ModelOptions
func ModelOptionsFor(provider, id string, curated map[string]ai.ModelOptions, dynamic ai.ModelOptions, models map[string]ai.ModelOptions) ai.ModelOptions
func NewChatModel[Config ChatConfig](*OpenAICompatible, id string, ai.ModelOptions) *ai.ModelAction
func ListChatActions[Config ChatConfig](context.Context, *OpenAICompatible, func(id string) ai.ModelOptions) []api.ActionDesc
func ResolveChatAction[Config ChatConfig](*OpenAICompatible, api.ActionType, id string, func(id string) ai.ModelOptions) api.Action
func ListModelActions(context.Context, *OpenAICompatible, func(id string) ai.ModelOptions) []api.ActionDesc
func ResolveModelAction(*OpenAICompatible, api.ActionType, id string, func(id string) ai.ModelOptions) api.Action
OpenAICompatible.ListModels func(context.Context, *openai.Client) ([]string, error)
func (*OpenAICompatible) NewModel(id string, ai.ModelOptions) *ai.ModelAction
func (*OpenAICompatible) NewEmbedder(id string, *ai.EmbedderOptions) *ai.EmbedderAction
func (*ModelGenerator) WithParams(openai.ChatCompletionNewParams) *ModelGenerator

// openai: added
OpenAI.Models map[string]ai.ModelOptions
OpenAI.Embedders map[string]ai.EmbedderOptions
func ModelRef(id string, *openai.ChatCompletionNewParams) ai.ModelRef
func NewEmbedderRef(id string, *TextEmbeddingConfig) ai.EmbedderRef
type TextEmbeddingConfig = compat_oai.EmbeddingConfig   // was a struct

// Deprecated, kept
compat_oai.OpenAICompatible.DefineEmbedder, .IsDefinedModel, .IsDefinedEmbedder
compat_oai.ModelGenerator.WithConfig
openai.DefineModel, .DefineEmbedder, .EmbedderRef, .Model, .Embedder

// Deprecated, kept with released behavior until the end of the train
compat_oai.OpenAICompatible.DefineModel
```

## What this does not do

- **No shared generation config**: it would advertise fields providers reject; the conformance sweep keeps the spelled-out fields honest.
- **No sub-plugin changes**: `anthropic`, `dashscope`, `kimi`, and `zai` stay on the transitional path here and migrate up the train.
- **No `Versions` validation against the provider**: the declared list stays a curated allowlist; dynamic models accept any version.
- **No migration of other families**: `modelgarden/llama.go` moves to `NewModel` only because it called the base's `DefineModel`.
- **No `-pro` models**: they need the Responses API; naming one fails at request time, pre-existing and unchanged.

## Live tests share one checklist

New test-only `internal/livetest` walks one checklist against a provider's real API (generation with usage and finish reason, history, system prompt, streaming with the usage opt-in, tools plain and streamed, tool choice, structured output typed/streamed/JSON-mode, reasoning, vision from an in-code image, `extra` passthrough), switched by the catalog's capability claims. It is the only live coverage per provider: the base's old `ModelGenerator` live tests re-proved the same wire behavior through a lower layer and are deleted, and `openai` keeps only its extras (the embedder, the SDK-typed config). Live runs also set the response_format policy: the text form is never sent, `json_object` goes only to a model whose declared output formats include `"json"` (`WithOutputFormats` carries the declaration), and a schema always rides `json_schema`.

## Testing

Go 1.26.3, full `go test ./...` in `go/` passes at this level of the stack with no failures; the `compat_oai` root package and `openai` are 80 cases counting subtests (65 and 15). The openai live suite is green against the real API: 16/16 on `gpt-4o-mini`, reasoning on `gpt-5-nano`, vision on `gpt-4.1-nano`; the checklist has also run green through the anthropic and xai PRs. The curation is pinned three ways: hidden fields advertise as `true`, every override path resolves against the linked SDK's schema, and each rejection names its `ai.With*` option. The four existing sub-plugins pass their current tests unmodified against the new core, including the thinking-config tests that pin the transitional path. Each new test was checked by breaking the code it covers. The `go/samples/compat_oai` openai and custom (OpenRouter) samples are rewritten to the single-flow shape of `go/samples/basic`, the openai one verified against the live API.


adesinah:
@apascal07  I found three issues (I reproduced all three locally) that I think need addressing before merge:

1. `TextEmbeddingConfig` is not fully source-compatible. An unkeyed literal that compiles against the published `v1.11.0` now fails because the alias points to a larger struct with a different field order. We should preserve the original two-field type or treat this as a breaking change.

2. The OpenAI config schema exposes tools, but tools supplied through `ChatCompletionNewParams` are not cleared when no Genkit tools are provided. I confirmed they reach the outgoing request, even though Genkit has no handler for them. Genkit-managed fields should be removed from the advertised config and explicitly overwritten or cleared.

3. RegisterModel cannot customize a curated model as documented. Models such as `gpt-4o` are registered during initialization, so calling `RegisterModel` afterward panics with action "/model/openai/gpt-4o" is already registered. The override mechanism needs to happen before registration or use plugin configuration instead.

The full Go test suite otherwise passes.

apascal07:
All three reproduced and addressed in 86a98002f, 562929669 and 5752d1268.

1. Treated as a documented break rather than reverting the alias, since no field order avoids it: an unkeyed literal needs one value per field. `Dimensions` and `EncodingFormat` keep their released order, types and tags, so keyed literals and the wire format are unchanged.

2. `messages`, `tools`, `tool_choice`, `functions` and `function_call` are stripped from the advertised schema and cleared from any config that carries them. Naming one now fails the request instead of reaching the provider.

3. `RegisterModel` and `RegisterEmbedder` are gone rather than fixed. Plugins take a `Models` map instead, overlaid on what they resolve, which reaches the curated models `Init` registers. Same shape as `googlegenai` and the native `anthropic` plugin. Nothing named `Register*` shipped in `go/v1.11.0`, so none of it is a breaking change.

Details in the updated description. Go 1.26.3, `go test ./...` in `go/`: 43 packages pass.

adesinah (approved):

PR #5982, opened 2026-08-07
fix(py): raise GenkitError for generate model resolution failures

huangjeff5:
## Summary
- When `generate` has no model configured, callers get a clear invalid-argument error instead of a bare exception.
- When a model name can't be resolved, the error says so explicitly and, if the name has no plugin prefix, hints that it should look like `plugin/model`.
- Adds tests for both failure paths.


hilariie (approved):
LGTM. Gemini's comment is a nit

PR #5986, opened 2026-08-07
feat(py): Make ModelRequest generic

huangjeff5:
## Summary
- Plugin authors can annotate `request: ModelRequest[MyConfig]`. `ai.generate(config={'temperature': 0.7})` arrives as a `MyConfig` instance. Bare `ModelRequest` still gets a raw dict. Unknown keys land on `model_extra`; a plugin that sets `extra='forbid'` rejects them.
- Output settings live on the request as `output: { format, schema, constrained, contentType }`. `ai.generate(output_format='json', ...)` is unchanged. Plugins still read `request.output_format` (properties over that object). Construct a request with `output=OutputConfig(...)`; passing `output_format=` into the constructor does not set output.
- A Pydantic config instance is only legal if it is the plugin's schema (or a subclass). `ai.generate(config=OpenAIConfig(...))` on a Gemini-typed plugin is `GenkitError(INVALID_ARGUMENT)`. A mapping still coerces. `generate()` accepts a dict or a `ModelConfig`; a plain `BaseModel` is a `ValidationError` at `generate()`. The error names the public import (`genkit_openai.OpenAIConfig`, not `genkit_openai.typing.OpenAIConfig`).
- `ai.generate(docs=..., config={'temperature': 'high'})` is still that `GenkitError`, not an `AttributeError`. Dumped docs rebuild from dicts the same way messages already do.
- `define_model` rejects `ModelRequest | None` / a raw `dict` at startup so typed construction cannot silently turn off. A bad declared field (`temperature='high'`) is `GenkitError(INVALID_ARGUMENT)`.

```python
async def my_model(request: ModelRequest[MyConfig], ctx):
    request.config.temperature
    if request.output_format == 'json':
        ...
```

### DIFF
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -33,6 +33,7 @@ from typing_extensions import TypeVar
 from genkit._core._channel import Channel, CloseableQueue
 from genkit._core._compat import StrEnum
 from genkit._core._error import GenkitError
+from genkit._core._model import config_type_path, declared_config_type
 from genkit._core._schema import to_json_schema
 from genkit._core._trace._suppress import suppress_telemetry
 from genkit._core._tracing import SpanMetadata, run_in_new_span
@@ -703,13 +704,25 @@ class Action(Generic[InputT, OutputT, ChunkT, InitT]):
         if input is None and self._first_arg_optional:
             return input
         payload: object = input
-        # If input is a BaseModel of a different type (e.g. ModelRequest[dict] vs
-        # ModelRequest[PluginConfig]), Pydantic rejects direct class validation.
-        # Dump to a plain dict so we can re-parse into the action's schema.
+        # A differently-typed ModelRequest with a mapping config is dumped and
+        # re-parsed into the plugin class. A Pydantic config instance of the
+        # wrong class is a caller mistake — dump would silently coerce it.
         if isinstance(input, BaseModel):
             try:
                 return self._input_type.validate_python(input)
             except ValidationError:
+                config = getattr(input, 'config', None)
+                if isinstance(config, BaseModel):
+                    expected = declared_config_type(self._input_class) if self._input_class is not None else None
+                    want = config_type_path(expected) if isinstance(expected, type) else 'the plugin config class'
+                    raise GenkitError(
+                        message=(
+                            f"Invalid input for action '{self.name}': "
+                            f'config must be {want} or a mapping, '
+                            f'got {config_type_path(type(config))}'
+                        ),
+                        status='INVALID_ARGUMENT',
+                    ) from None
                 payload = input.model_dump(mode='python')
 
         try:
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -472,6 +472,11 @@ class Action(Generic[InputT, OutputT, ChunkT, InitT]):
         return self._input_type
 
     @property
+    def input_class(self) -> type | None:
+        """The action's input annotation as a concrete class, when it is one."""
+        return getattr(self, '_input_class', None)
+
+    @property
     def input_schema(self) -> dict[str, object]:
         return self._input_schema
 
@@ -626,10 +631,12 @@ class Action(Generic[InputT, OutputT, ChunkT, InitT]):
             type_adapter = TypeAdapter(arg_types[0])
             self._input_schema: dict[str, object] = type_adapter.json_schema()
             self._input_type: TypeAdapter[InputT] | None = cast(TypeAdapter[InputT], type_adapter)
+            self._input_class: type | None = arg_types[0] if isinstance(arg_types[0], type) else None
             self._metadata[ActionMetadataKey.INPUT_KEY] = self._input_schema
         else:
             self._input_schema = TypeAdapter(object).json_schema()
             self._input_type = None
+            self._input_class = None
             self._metadata[ActionMetadataKey.INPUT_KEY] = self._input_schema
 
         if ActionMetadataKey.RETURN in annotations:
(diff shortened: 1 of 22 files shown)


py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -290,20 +304,35 @@ def gemini_model(request: ModelRequest[GeminiConfig]) -> ModelResponse:
...
+    config: ModelRequestConfigT | None = None
     tools: list[ToolDefinition] | None = None
     tool_choice: ToolChoice | None = Field(default=None)
-    # Flat output fields (no nested OutputConfig)
-    output_format: str | None = None
-    output_schema: dict[str, Any] | None = None
-    output_constrained: bool | None = None
-    output_content_type: str | None = None
+    # Wire-shaped output storage; flat access via the properties below.
+    output: OutputConfig = Field(default_factory=OutputConfig)
```
apascal07: `extra='allow'` makes the old flat constructor form fail silently rather than loudly.

```python
r = ModelRequest(
    messages=[...],
    output_format='json',
    output_schema={'type': 'object'},
    output_constrained=True,
)
r.output_format          # -> None
r.model_dump()           # -> {..., 'output': {}, 'output_format': 'json',
                         #     'output_schema': {...}, 'output_constrained': True}
```

The kwargs land in `__pydantic_extra__`, so the properties read `None` and the stray keys ride onto the wire next to an empty `output`. A plugin or middleware still building requests this way (the supported shape one commit ago) generates unconstrained text with no schema and no error anywhere.

The class docstring documents this, but a docstring is not a guardrail. A `model_validator(mode='before')` that folds `output_format`/`output_schema`/`output_constrained`/`output_content_type` into `output` — or raises `ValueError` naming `output=OutputConfig(...)` — turns a silent wrong answer into a fixable error.

py/packages/genkit/src/genkit/_core/_action.py
```diff
@@ -695,22 +702,25 @@ def _validate_input(self, input: InputT | None) -> InputT | None:
         # signal that "no input" is a legitimate way to invoke this action.
         if input is None and self._first_arg_optional:
             return input
+        payload: object = input
+        # If input is a BaseModel of a different type (e.g. ModelRequest[dict] vs
+        # ModelRequest[PluginConfig]), Pydantic rejects direct class validation.
+        # Dump to a plain dict so we can re-parse into the action's schema.
+        if isinstance(input, BaseModel):
```
apascal07: The dump-and-revalidate fallback does not preserve the error contract when the request carries `docs`.

`ModelRequest._wrap_docs` reads `d.content` / `d.metadata` as attributes, so re-validating a dumped payload hands it dicts:

```python
await ai.generate(
    model=<handler annotated ModelRequest[Cfg]>,
    prompt='hi',
    docs=[Document.from_text('ctx')],
    config={'temperature': 'high'},   # invalid for Cfg
)
# AttributeError: 'dict' object has no attribute 'content'
```

Path: `action_to_generate_request` rejects the bad config on the fast path, falls back to the bare carrier, and `_validate_input` then takes the dump branch. So the comment's "the pre-fast-path error contract, preserved exactly" does not hold — callers get an unhandled `AttributeError` instead of `GenkitError(INVALID_ARGUMENT)`.

Either make `_wrap_docs` accept mappings (`Document(**d)` / `Document(d['content'], d.get('metadata'))`), or dump with `round_trip=True` and re-raise the original `ValidationError` if the second pass throws anything that is not a `ValidationError`.

py/packages/genkit/src/genkit/_core/_model.py
apascal07, line 366: `extra='allow'` turns the old flat constructor form into a silent no-op:

```python
r = ModelRequest(messages=[...], output_format='json', output_schema={'type': 'object'})
r.output_format   # None
r.model_dump()    # {..., 'output': {}, 'output_format': 'json', 'output_schema': {...}}
```

The kwargs land in `__pydantic_extra__`, so the properties read `None` and the stray keys ship on the wire next to an empty `output`. Any plugin or middleware still building requests the way this file supported one commit ago generates unconstrained text with no schema, and nothing raises. The docstring note below records the hazard but cannot enforce it.

py/packages/genkit/src/genkit/_core/_action.py
apascal07, line 710: The dump fallback loses the error contract when the request carries `docs`. `_wrap_docs` reads `d.content` as an attribute, so re-validating a dumped payload hands it dicts:

```python
await ai.generate(
    model=<handler annotated ModelRequest[Cfg]>,
    prompt='hi',
    docs=[Document.from_text('ctx')],
    config={'temperature': 'high'},   # invalid for Cfg
)
# AttributeError: 'dict' object has no attribute 'content'
```

The typed fast path rejects the config, falls back to the bare carrier, and this branch then dumps. Callers get an unhandled `AttributeError` where the comment promises `GenkitError(INVALID_ARGUMENT)`.

The shallow dump gemini-code-assist suggested on this line closes this too, since the `Document` instances survive instead of becoming dicts.

apascal07 (approved):

py/packages/genkit/src/genkit/_core/_action.py
```diff
@@ -695,22 +695,25 @@ def _validate_input(self, input: InputT | None) -> InputT | None:
...
+            try:
+                return self._input_type.validate_python(input)
+            except ValidationError:
+                payload = input.model_dump(mode='python')
```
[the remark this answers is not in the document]
huangjeff5: Decided to remove this logic anyway; this path was meant to support covariant Config classes. Decided that if you pass a config BaseModel, it has to be the one that the plugin requested. If you pass a dict, the framework will still validate/coerce it into the model requested by the plugin.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Ack, but this isn't ai.generate. generate(output_format='json') goes through OutputConfig and is pinned. ModelRequest(output_format=...) is not a constructor arg — the flats are properties. Ignoring those kwargs is intentional; there are no in-repo callers and we don't want a leftover constructor to start raising. The extras on the wire are the extra='allow' cost.

py/packages/genkit/src/genkit/_core/_action.py
huangjeff5: Reproduced — real, and generate() with docs= plus a rejected config. Dump turns docs into dicts; _wrap_docs did d.content and blew up before the GenkitError. Only docs (messages already take a dict). Fix: _wrap_docs accepts a dict the same way Message already does. Not a shallow dump — that only helps this Action path; the next dict docs (Dev UI, model_validate) would still AttributeError.

Pinned:
```
await ai.generate(
    model='conforming',
    prompt='hi',
    docs=[Document.from_text('ctx')],
    config={'temperature': 'high'},
)
# GenkitError(INVALID_ARGUMENT): Invalid input for action 'conforming'
# not AttributeError
```

PR #5989, opened 2026-08-07
feat(py): Use ModelConfigDict instead of ModelConfig to enable autocomplete at veneer layer

huangjeff5:
## What this PR does

Tightens app-developer model configuration while keeping plugin authoring explicit.

> [!WARNING]
> **BREAKING CHANGE:** `from genkit import ModelConfig` and `from genkit.model import ModelConfig` no longer work.
>
> - **App code**: Pass plain dicts (`ModelConfigDict` with IDE autocomplete) or typed config classes.
> - **Plugin authors**: `ModelConfig` is now exported from `genkit.plugin_api`.

---

## Rationale

App developers have two blessed patterns for model config: dict literals (loose/autocompleted) and typed config classes (strict). Importing generic `ModelConfig` at call sites added no extra capability and blurred the app/plugin boundary.


pavelgj (approved):

pavelgj (approved):

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -67,6 +67,23 @@
 ModelConfig = GenerationCommonConfig  # public name for GenerationCommonConfig
 ModelUsage = GenerationUsage  # public name for GenerationUsage
 
+
+class ModelConfigDict(TypedDict, extra_items=Any, total=False):
```
pavelgj, line 74: Drift risk: these keys are a hand-copied subset of `GenerationCommonConfig`/`ModelConfig`. If someone adds or renames a field on the pydantic model, this TypedDict silently goes stale and autocomplete quietly drops the new knob. Could we add a small test that asserts the keys stay in sync (e.g. `set(ModelConfigDict.__annotations__) <= set(GenerationCommonConfig.model_fields)`), or at least a comment on both sides pointing at each other so the next editor updates both? Not a blocker.

huangjeff5: Good idea, added.

PR #5992, opened 2026-08-07
fix(js/plugins/google-genai): Do not persist clientOptions in the …

ifielker:
operation for long running models

This is a breaking change for certain edge cases involving background models with per request overrides only.

Previously the clientOptions were persisted in the operation metadata so it was easier to poll using the same information, however apparently people were just returning the full operation including auth to untrusted third parties without any modification, so there was a potential for leaking apiKeys if they did not remove the metadata from the operation first. This implementation removes that leak potential for security purposes and the operation no longer persists any client options passed in to the original generate request. Unfortunately this breaks an edge case that previously worked: If any overrides were present for longrunning models  (such as apiKeys), previously it just worked, but now those overrides must also be passed in for each checkOperation or cancelOperation. 

Checklist (if applicable):

### DIFF
--- a/js/plugins/google-genai/src/googleai/utils.ts
+++ b/js/plugins/google-genai/src/googleai/utils.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { GenerateRequest, GenkitError, z } from 'genkit';
+import { GenerateRequest, GenkitError, z, type ActionContext } from 'genkit';
 import process from 'process';
 import { extractMedia } from '../common/utils.js';
 import { ClientOptions, ImagenInstance, VeoImage, VeoVideo } from './types.js';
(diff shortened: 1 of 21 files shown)


js/plugins/google-genai/tests/vertexai/veo_test.ts
```diff
@@ -199,6 +199,23 @@ describe('Vertex AI Veo', () => {
         controller.abort();
         sinon.assert.calledOnce(abortSpy);
       });
+
+      it('should apply apiKey override from options.context', async () => {
+        mockFetchResponse({ name: 'operations/ctx-key', done: false });
+
+        const { start } = captureModelRunner(defaultRegionalClientOptions);
+        await await defineModel(modelName, defaultRegionalClientOptions).start(
```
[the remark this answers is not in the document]
ifielker: fixed

js/plugins/google-genai/src/googleai/utils.ts
```diff
@@ -225,3 +225,39 @@ export function removeClientOptionOverrides<
...
+  clientOptions: ClientOptions,
+  context?: ActionContext
+): ClientOptions {
+  if (!context) {
+    return clientOptions;
+  }
+
+  let newOptions = { ...clientOptions };
+
+  if (typeof context.auth?.apiKey === 'string') {
```
pavelgj: Let's use `secrets` instead of `auth`. Auth us reserved for incoming user credentials (ex. from the incoming http call)

```suggestion
  if (typeof context.secrets?.apiKey === 'string') {
```

ifielker: done

pavelgj (approved):

PR #6012, opened 2026-08-11
build(py): bump twine floor to 7.0.0

hilariie:
## Why

Every Python PR is failing at the Build Distributions job. hatchling 1.32.0 now emits Metadata-Version 2.5 and twine 6 rejects it:

```
InvalidDistribution: Invalid distribution metadata: '2.5' is not a valid metadata version
```

Nothing in the repo changed to trigger this. The build backend is declared as `requires = ["hatchling"]` with no upper bound, and hatchling is not in uv.lock, so `uv build` picks up whatever hatchling is current on PyPI at build time. 1.32.0 shipped and every PR started failing.

twine 6.2.0 is the last 6.x release and it still rejects 2.5, so the floor has to move to 7.

## Changes

`py/pyproject.toml`: twine dev dependency floor from >=6.1.0 to >=7.0.0. `py/uv.lock`: regenerated.

The lock diff is bigger than the one-line change suggests, so worth explaining. twine 7 requires packaging>=26.1 and rich>=14.3.3, so packaging (25.0 to 26.3) and rich (14.3.2 to 15.0.0) move with it. The rest is uv re-serializing dependency markers on re-resolution, not a semantic change: the markers it drops are already implied by the parent package's own marker (jinxed only installs on Windows, so its ansicon dep does not need a win32 marker of its own). The whole file is plain `uv lock` output.

rich is a runtime dep of genkit but only a declared one; it is not imported anywhere in the source, and the published wheel still carries rich>=13.0.0, so users are unaffected.

## Verification

Built genkit against hatchling 1.32.0 and confirmed the wheel carries Metadata-Version: 2.5. Ran `twine check` on that wheel and sdist both ways:

| twine   | result                     |
| ------- | -------------------------- |
| 6.2.0   | InvalidDistribution, exit 1 |
| 7.0.0   | PASSED                     |

`./py/bin/build_dists` now runs end-to-end and exits 0 with "Twine check passed".


huangjeff5:
Thanks for fixing this.

One small observation in [`py/bin/build_dists`](https://github.com/genkit-ai/genkit/blob/main/py/bin/build_dists#L53-L57): when `twine check` fails, the error output is currently swallowed in CI logs.

```bash
TWINE_CHECK=$(uv run --directory "${TOP_DIR}"/py twine check "${dist_files[@]}")
echo "$TWINE_CHECK"
```

Because the script runs with `set -e`, the command substitution `TWINE_CHECK=$(...)` immediately aborts bash on a non-zero exit code before `echo "$TWINE_CHECK"` can run. In CI, developers only see `Process completed with exit code 1.` without the actual `twine` error output.

Consider updating `py/bin/build_dists` to execute `twine check` directly:

```bash
if ! uv run --directory "${TOP_DIR}"/py twine check "${dist_files[@]}"; then
  echo "Twine check failed."
  exit 1
fi
```

Otherwise looks good!

huangjeff5 (approved):

PR #6015, opened 2026-08-12
feat(py): normalize model+config for str | ModelRef

huangjeff5:
### Product (this is the layer for #6074)

`generate` / `define_prompt` / agents / `Genkit(model=)` will take one `model=` that is a **name or a `ModelRef`**:

```python
await ai.generate(model='googleai/gemini-2.5-flash', prompt='hi')
await ai.generate(model=flash, prompt='hi')
await ai.generate(model=flash, prompt='hi', config={'temperature': 0.2})

ai = Genkit(model=flash)
await ai.generate(prompt='hi')
```

`flash` carries a wire name plus defaults (`version`, `config`). A string is just the name. Same slot, two shapes. That union — including the constructor — is #6074.

To implement it, every path has to become **one wire name + one config bag** before the plugin runs. This PR is that normalization layer.

### What the layer has to survive

E2e of these lives on #6074. They are listed here because they are what the helpers are for.

```python
# Call-time keys win; None clears a ref default; 0 / '' / [] / False stay.
await ai.generate(model=flash, config={'temperature': 0.2})
await ai.generate(model=flash, config={'temperature': None})
await ai.generate(model=flash, config={'temperature': 0, 'stop_sequences': []})

# Either spelling is the same knob. The ref dumped max_output_tokens=100;
# the call's maxOutputTokens folds onto that slot and wins. No leftover sibling.
await ai.generate(model=flash, config={'maxOutputTokens': 5})
await ai.generate(model=flash, config={'maxOutputTokens': None})   # clears the ref default

# Nested bag replaces whole — thinking.level from the ref is gone.
# The outer key folds (thinkingConfig -> thinking_config); inner keys stay
# as written and are the nested schema's job at the plugin edge.
await ai.generate(model=flash, config={'thinkingConfig': {'thinkingBudget': 256}})

# Switching models does not strip leftover keys (safety_settings still there).
joke = ai.define_prompt(model=flash, config=GeminiConfig(safety_settings=...), prompt='hi')
await joke(model=openai_gpt)

# Explicit name is a different model; constructor knobs do not leak onto it.
ai = Genkit(model=flash)
await ai.generate(model='openai/gpt-4o', prompt='hi')

# '' is omitted (unset MODEL=). 123 is a hard error, not the default.
await ai.generate(model='', prompt='hi')
await ai.generate(model=123, prompt='hi')  # INVALID_ARGUMENT
```

With a bare string `model=` there is no schema, so nothing folds — dict keys pass through as written.

### Principles (this PR)

The pipeline is four steps, each doing one thing:

- **Dump** (`normalize_config`): a config object or dict becomes a plain dict. Pydantic dumps the Python field names; dict keys stay as written; no fold, no merge. `api_key` (`exclude=True`) is copied back so a per-request key still reaches the plugin.
- **Fold** (`fold_config_aliases`): schema aliases rename to the field name (`maxOutputTokens` → `max_output_tokens`). A rename, not validation — unknown keys keep their spelling.
- **Overlay** (`overlay_config`): fold each layer, last layer wins, drop `None`. Both spellings of one knob land in one slot, so a camel override or clear cannot leave a stale snake sibling.
- **Resolve** (`resolve_model_ref` / `resolve_call_model`): stack `ref.version` → dumped `ref.config` → call `config=`, overlay once. Merge is shallow, by key. Only `None` means omit; empty values are explicit. The helper does not know families.

The merged result is a dict. `ModelRequest` rebuilds generic `ModelConfig` today; plugin-typed parse is #5986.

`Genkit(model=)` stays a **string** here. Dev UI `/api/values` unwraps a stored `ModelRef` to the wire name so listing does not 500; that path is ready for #6074.

### What lands here vs later

- **Here:** `normalize_config` / `fold_config_aliases` / `overlay_config` / `resolve_model_arg` / `resolve_model_ref` / `resolve_call_model`. Constructor remains `str`.
- **#6074:** the union on `Genkit(model=)`, `generate(model=)`, `define_prompt`, agents, stream, `generate_operation`.
- **#5986:** `ModelRequest[GeminiConfig]` — plugin-only fields as real attributes.


huangjeff5:
@gemini review

pavelgj (approved):

pavelgj (approved):

PR #6019, opened 2026-08-12
feat(go): add Meta Muse model support

### LINKED ISSUE #5801: feat(go/plugins/compat_oai): add dedicated Meta subplugin
Meta has no compat_oai subplugin. Provider/endpoint to confirm — "Muse" unclear, possibly Meta's Llama API. Template: #5461.

adesinah:
This PR add Meta Muse model support.

Closes: #5801

Checklist (if applicable):

### DIFF
--- a/go/plugins/compat_oai/generate_test.go
+++ b/go/plugins/compat_oai/generate_test.go
@@ -757,6 +757,7 @@ func TestWithParams(t *testing.T) {
 		TopP:        openai.Float(0.8),
 	}
 	params.SetExtraFields(map[string]any{
+		"prompt_cache_retention": "24h",
 		"thinking": map[string]any{
 			"type":           "enabled",
 			"clear_thinking": false,
@@ -771,6 +772,9 @@ func TestWithParams(t *testing.T) {
 	request := marshalParams(t, *g.GetRequest())
 	if got := request["model"]; got != "test-model-2026-01-01" {
 		t.Errorf("model = %v, want the pinned %q", got, "test-model-2026-01-01")
+	}
+	if got := request["prompt_cache_retention"]; got != "24h" {
+		t.Errorf("prompt_cache_retention = %v, want %q", got, "24h")
 	}
 	if got := request["temperature"]; got != 0.3 {
 		t.Errorf("temperature = %v, want 0.3", got)
--- a/go/plugins/compat_oai/generate_test.go
+++ b/go/plugins/compat_oai/generate_test.go
@@ -757,7 +757,6 @@ func TestWithParams(t *testing.T) {
 		TopP:        openai.Float(0.8),
 	}
 	params.SetExtraFields(map[string]any{
-		"prompt_cache_retention": "24h",
 		"thinking": map[string]any{
 			"type":           "enabled",
 			"clear_thinking": false,
@@ -772,9 +771,6 @@ func TestWithParams(t *testing.T) {
 	request := marshalParams(t, *g.GetRequest())
 	if got := request["model"]; got != "test-model-2026-01-01" {
 		t.Errorf("model = %v, want the pinned %q", got, "test-model-2026-01-01")
-	}
-	if got := request["prompt_cache_retention"]; got != "24h" {
-		t.Errorf("prompt_cache_retention = %v, want %q", got, "24h")
 	}
 	if got := request["temperature"]; got != 0.3 {
 		t.Errorf("temperature = %v, want 0.3", got)
--- a/go/plugins/compat_oai/meta/README.md
+++ b/go/plugins/compat_oai/meta/README.md
@@ -1,0 +1,51 @@
+# Meta Model API Plugin
+
+This plugin provides Genkit support for Meta's OpenAI-compatible Model API and
+the Muse Spark 1.1 multimodal reasoning model.
+
+## Usage
+
+Set a Meta Model API key:
+
+```bash
+export MODEL_API_KEY=<your-api-key>
+```
+
+`META_API_KEY` is also accepted. The plugin uses
+`https://api.meta.ai/v1` by default; set `META_BASE_URL` to use another
+compatible endpoint.
+
+```go
+import (
+    "context"
+
+    "github.com/firebase/genkit/go/ai"
+    "github.com/firebase/genkit/go/genkit"
+    "github.com/firebase/genkit/go/plugins/compat_oai/meta"
+)
+
+ctx := context.Background()
+plugin := &meta.Meta{}
+g := genkit.Init(
+    ctx,
+    genkit.WithPlugins(plugin),
+    genkit.WithDefaultModel("meta/"+meta.ModelMuseSpark11),
+)
+
+response, err := genkit.Generate(
+    ctx,
+    g,
+    ai.WithPrompt("Explain mixture-of-experts models."),
+)
+```
+
+Meta Model API is currently in public preview. This plugin uses its
+OpenAI-compatible Chat Completions endpoint.
+
+## Live tests
+
+Live tests are skipped unless `MODEL_API_KEY` or `META_API_KEY` is set:
+
+```bash
+go test -v ./plugins/compat_oai/meta
+```
--- a/go/plugins/compat_oai/meta/README.md
+++ b/go/plugins/compat_oai/meta/README.md
@@ -1,7 +1,7 @@
 # Meta Model API Plugin
 
 This plugin provides Genkit support for Meta's OpenAI-compatible Model API and
-the Muse Spark 1.1 multimodal reasoning model.
+the Muse Spark multimodal reasoning models.
 
 ## Usage
 
@@ -11,9 +11,9 @@ Set a Meta Model API key:
 export MODEL_API_KEY=<your-api-key>
 ```
 
-`META_API_KEY` is also accepted. The plugin uses
-`https://api.meta.ai/v1` by default; set `META_BASE_URL` to use another
-compatible endpoint.
+The plugin uses `https://api.meta.ai/v1` by default. Set `META_BASE_URL` or
+provide `option.WithBaseURL` through `meta.Meta.Opts` to use another compatible
+endpoint.
 
 ```go
 import (
@@ -25,26 +25,21 @@ import (
 )
 
 ctx := context.Background()
-plugin := &meta.Meta{}
-g := genkit.Init(
-    ctx,
-    genkit.WithPlugins(plugin),
-    genkit.WithDefaultModel("meta/"+meta.ModelMuseSpark11),
+g := genkit.Init(ctx,
+    genkit.WithPlugins(&meta.Meta{}),
+    genkit.WithDefaultModel("meta/muse-spark-1.2"),
 )
 
-response, err := genkit.Generate(
-    ctx,
-    g,
+response, err := genkit.Generate(ctx, g,
     ai.WithPrompt("Explain mixture-of-experts models."),
 )
 ```
 
-Meta Model API is currently in public preview. This plugin uses its
-OpenAI-compatible Chat Completions endpoint.
+The plugin uses Meta Model API's OpenAI-compatible Chat Completions endpoint.
 
 ## Live tests
 
-Live tests are skipped unless `MODEL_API_KEY` or `META_API_KEY` is set:
+Live tests are skipped unless `MODEL_API_KEY` is set:
 
 ```bash
 go test -v ./plugins/compat_oai/meta
--- a/go/plugins/compat_oai/meta/meta.go
+++ b/go/plugins/compat_oai/meta/meta.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/core/api"
-	"github.com/firebase/genkit/go/genkit"
 	"github.com/firebase/genkit/go/plugins/compat_oai"
 	"github.com/openai/openai-go/option"
 )
@@ -102,21 +101,11 @@ func (m *Meta) Init(ctx context.Context) []api.Action {
 	actions := m.openAICompatible.Init(ctx)
 
 	for model, modelOpts := range supportedModels {
-		if action, ok := m.DefineModel(model, modelOpts).(api.Action); ok {
+		if action, ok := m.openAICompatible.DefineModel(provider, model, modelOpts).(api.Action); ok {
 			actions = append(actions, action)
 		}
 	}
 	return actions
-}
-
-// Model returns a registered Meta model.
-func (m *Meta) Model(g *genkit.Genkit, id string) ai.Model {
-	return m.openAICompatible.Model(g, api.NewName(provider, id))
-}
-
-// DefineModel registers a Meta model, including models not in the built-in list.
-func (m *Meta) DefineModel(id string, opts ai.ModelOptions) ai.Model {
-	return m.openAICompatible.DefineModel(provider, id, opts)
 }
 
 // ListActions lists models exposed by the configured Meta endpoint.
--- a/go/plugins/compat_oai/meta/meta.go
+++ b/go/plugins/compat_oai/meta/meta.go
@@ -22,43 +22,114 @@ import (
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/core/api"
 	"github.com/firebase/genkit/go/plugins/compat_oai"
+	"github.com/openai/openai-go"
 	"github.com/openai/openai-go/option"
+	"github.com/openai/openai-go/shared"
 )
 
 const (
 	provider       = "meta"
 	defaultBaseURL = "https://api.meta.ai/v1"
-
-	// ModelMuseSpark11 is Meta's Muse Spark 1.1 multimodal reasoning model.
-	ModelMuseSpark11 = "muse-spark-1.1"
 )
 
+// ReasoningEffort controls how much reasoning Muse Spark performs.
+type ReasoningEffort string
+
+const (
+	// ReasoningEffortMinimal uses the least reasoning and lowest latency.
+	ReasoningEffortMinimal ReasoningEffort = "minimal"
+	// ReasoningEffortLow uses a small reasoning budget.
+	ReasoningEffortLow ReasoningEffort = "low"
+	// ReasoningEffortMedium balances reasoning depth and latency.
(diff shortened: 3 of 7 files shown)


adesinah:
@apascal07 this is good for review now.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Model returns a registered Meta model.
+func (m *Meta) Model(g *genkit.Genkit, id string) ai.Model {
+	return m.openAICompatible.Model(g, api.NewName(provider, id))
+}
+
+// DefineModel registers a Meta model, including models not in the built-in list.
+func (m *Meta) DefineModel(id string, opts ai.ModelOptions) ai.Model {
+	return m.openAICompatible.DefineModel(provider, id, opts)
+}
```
apascal07:
```suggestion
```

These are deprecated in existing plugins, new plugins don't get them.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/firebase/genkit/go/plugins/compat_oai"
+	"github.com/openai/openai-go/option"
+)
+
+const (
+	provider       = "meta"
+	defaultBaseURL = "https://api.meta.ai/v1"
+
+	// ModelMuseSpark11 is Meta's Muse Spark 1.1 multimodal reasoning model.
+	ModelMuseSpark11 = "muse-spark-1.1"
```
apascal07: No exported constants for model names, models get deprecated and removed but we're stuck with exported symbols that we can't remove without a breaking change.

go/plugins/compat_oai/meta/meta_live_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/firebase/genkit/go/ai"
+	"github.com/firebase/genkit/go/genkit"
+	"github.com/firebase/genkit/go/plugins/compat_oai/meta"
+)
+
+func TestPluginLive(t *testing.T) {
```
apascal07: We have a live test suite that's consistent across `compat_oai` models, please use it.

go/plugins/compat_oai/meta/README.md
```diff
@@ -0,0 +1,51 @@
...
+plugin := &meta.Meta{}
+g := genkit.Init(
+    ctx,
+    genkit.WithPlugins(plugin),
+    genkit.WithDefaultModel("meta/"+meta.ModelMuseSpark11),
+)
+
+response, err := genkit.Generate(
+    ctx,
+    g,
```
apascal07:
```suggestion
response, err := genkit.Generate(ctx, g,
```

go/plugins/compat_oai/meta/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+    "github.com/firebase/genkit/go/ai"
+    "github.com/firebase/genkit/go/genkit"
+    "github.com/firebase/genkit/go/plugins/compat_oai/meta"
+)
+
+ctx := context.Background()
+plugin := &meta.Meta{}
+g := genkit.Init(
+    ctx,
```
apascal07:
```suggestion
g := genkit.Init(ctx,
```

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+
+	m.openAICompatible.Provider = provider
+	m.openAICompatible.Opts = opts
+	actions := m.openAICompatible.Init(ctx)
+
+	for model, modelOpts := range supportedModels {
+		if action, ok := m.DefineModel(model, modelOpts).(api.Action); ok {
+			actions = append(actions, action)
+		}
+	}
```
apascal07: Why not `compat_oai.NewChatModel[Config]` with custom config? There's little reason to have a custom subplugin that doesn't call for a custom config.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+		option.WithBaseURL(baseURL),
+	}
+	opts = append(opts, m.Opts...)
+
+	m.openAICompatible.Provider = provider
+	m.openAICompatible.Opts = opts
+	actions := m.openAICompatible.Init(ctx)
+
+	for model, modelOpts := range supportedModels {
+		if action, ok := m.DefineModel(model, modelOpts).(api.Action); ok {
```
apascal07: `OpenAICompatible.DefineModel` is deprecated. `compat_oai.NewChatModel[ChatConfig]` with a `ChatConfig` embedding `compat_oai.RequestConfig` is the typed path, and it also brings back `version` pinning, the per-request `apiKey`, and the `extra` passthrough.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+		Label: "Meta Muse Spark 1.1",
+		Supports: &ai.ModelSupports{
+			Multiturn:  true,
+			Tools:      true,
+			SystemRole: true,
+			Media:      true,
+			ToolChoice: true,
+			Output:     []string{"text", "json"},
+		},
+		Versions: []string{ModelMuseSpark11},
```
apascal07: A non-empty `Versions` turns validation on, and this list can only reject. `ai.WithConfig({"version": "muse-spark-1.2"})` fails with `model "meta/muse-spark-1.1" does not support version "muse-spark-1.2", supported versions: [muse-spark-1.1]`.

The other half appears once the catalog is updated: on the untyped path a pin can never be applied, because `generate.go:219` deletes the `version` key before the request is built. Pinning a real snapshot returned no error and put the unpinned model on the wire. Siblings declare no versions for the first reason; the second goes away with `RequestConfig`.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+	if apiKey == "" {
+		apiKey = os.Getenv("META_API_KEY")
+	}
+
+	opts := []option.RequestOption{
+		// Apply this even when apiKey is empty. The OpenAI client otherwise
+		// falls back to OPENAI_API_KEY, which must never be sent to Meta.
+		// A missing Meta key is reported by the service when a request is made,
+		// rather than preventing the plugin from initializing.
+		option.WithAPIKey(apiKey),
```
apascal07, line 158: The comment says OpenAI credentials must never reach Meta, but only the key is overridden. `OPENAI_ORG_ID` and `OPENAI_PROJECT_ID` come from the same `DefaultClientOptions()` and still go out. With a correct Meta key set:

```
Authorization: Bearer <meta key>
Openai-Organization: org-must-not-leak
Openai-Project: proj-must-not-leak
```

`TestPluginInitializesWithoutAPIKey` checks only the key, so the gap is invisible. This affects every plugin in the package, so the guard belongs in `OpenAICompatible.Init` rather than here.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+		if action, ok := m.DefineModel(model, modelOpts).(api.Action); ok {
+			actions = append(actions, action)
+		}
+	}
+	return actions
+}
+
+// Model returns a registered Meta model.
+func (m *Meta) Model(g *genkit.Genkit, id string) ai.Model {
+	return m.openAICompatible.Model(g, api.NewName(provider, id))
```
apascal07: `api.NewName` concatenates unconditionally, so a provider-prefixed ID double-prefixes. `Model(g, "meta/muse-spark-1.1")` returns non-nil, but it is a dynamically resolved action named `meta/meta/muse-spark-1.1` that gets registered permanently, and generating through it puts `"model":"meta/muse-spark-1.1"` on the wire. `compat_oai.ActionName` trims first.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// DefineModel registers a Meta model, including models not in the built-in list.
+func (m *Meta) DefineModel(id string, opts ai.ModelOptions) ai.Model {
+	return m.openAICompatible.DefineModel(provider, id, opts)
+}
+
+// ListActions lists models exposed by the configured Meta endpoint.
+func (m *Meta) ListActions(ctx context.Context) []api.ActionDesc {
```
apascal07, line 188: These describe the same model differently from `Init`. For `muse-spark-1.1`, `Init` gives label `Meta Muse Spark 1.1`, versions `[muse-spark-1.1]`, output `[text json]`, no config schema; list and resolve give label `meta - muse-spark-1.1`, versions `[]`, output `[]`, and the full SDK schema. So `prompt_cache_retention` is accepted on the curated ID and rejected on every other Meta model.

Routing all three through one `modelOptions` built on `compat_oai.ModelOptionsFor` keeps them in agreement. Without a `Models` field there is also no way to correct the curated entry, since `Init` has already registered it.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+var supportedModels = map[string]ai.ModelOptions{
+	ModelMuseSpark11: {
+		Label: "Meta Muse Spark 1.1",
+		Supports: &ai.ModelSupports{
+			Multiturn:  true,
+			Tools:      true,
+			SystemRole: true,
+			Media:      true,
+			ToolChoice: true,
+			Output:     []string{"text", "json"},
```
apascal07: `Constrained` is unset while `Output` declares json, so native structured output is never requested. `ai.WithOutputType` sends `response_format {"type":"json_object"}` plus the schema pasted into the user message. Meta documents `response_format {"type":"json_schema"}`, so this should be `ai.ConstrainedSupportAll`.

`Output` itself is inert here as well: `DefineModel` never chains `WithOutputFormats`, which `NewChatModel` does.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+
+const (
+	provider       = "meta"
+	defaultBaseURL = "https://api.meta.ai/v1"
+
+	// ModelMuseSpark11 is Meta's Muse Spark 1.1 multimodal reasoning model.
+	ModelMuseSpark11 = "muse-spark-1.1"
+)
+
+var supportedModels = map[string]ai.ModelOptions{
```
apascal07, line 106: Only `muse-spark-1.1` is curated, and Meta's model page now lists it as the earlier checkpoint. `muse-spark-1.2` and `muse-spark-1.2-contributor` shipped on 2026-08-05. The README and both tests point at the superseded ID as the default model, and per the `Versions` comment it cannot be pinned from the 1.1 action either. "Currently in public preview" in the README is stale for the same reason.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+		baseURL = defaultBaseURL
+	}
+
+	apiKey := m.APIKey
+	if apiKey == "" {
+		apiKey = os.Getenv("MODEL_API_KEY")
+	}
+	if apiKey == "" {
+		apiKey = os.Getenv("META_API_KEY")
+	}
```
apascal07, line 155: No fail-fast when no key resolves. `option.WithAPIKey("")` still sets the header, so requests go out with a literal `Authorization: Bearer` and the misconfiguration surfaces as a 401 per call inside a flow. All eight siblings panic here with `<provider> plugin initialization failed: apiKey is required`.

The comment argues a missing key is best reported by the service, but that argument applies to every plugin in the package, not to Meta specifically.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+	if baseURL == "" {
+		baseURL = defaultBaseURL
+	}
+
+	apiKey := m.APIKey
+	if apiKey == "" {
+		apiKey = os.Getenv("MODEL_API_KEY")
+	}
+	if apiKey == "" {
+		apiKey = os.Getenv("META_API_KEY")
```
apascal07: `META_API_KEY` is not a name Meta documents; `MODEL_API_KEY` is the only one in their reference. The alias also carries four different precedence stories: this file says `MODEL_API_KEY` then `META_API_KEY`, `meta/README.md` says only that `META_API_KEY` "is also accepted", `meta_live_test.go` skips with "META_API_KEY and MODEL_API_KEY are not set", and the `compat_oai/README.md` hunk lists `MODEL_API_KEY` alone.

go/plugins/compat_oai/meta/meta.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// Meta configures the Meta Model API plugin.
+type Meta struct {
+	// APIKey is the Meta Model API key. If empty, MODEL_API_KEY and then
+	// META_API_KEY are consulted.
+	APIKey string
+	// BaseURL overrides the Meta Model API endpoint. If empty, META_BASE_URL
+	// and then the default endpoint are used.
+	BaseURL string
```
apascal07: No sibling has a `BaseURL` field; they take the endpoint from `<PROVIDER>_BASE_URL` or `option.WithBaseURL` in `Opts`. This adds a fourth mechanism, and it loses to the third: `Opts` is appended after `option.WithBaseURL(baseURL)` on line 98, so an `option.WithBaseURL` there silently beats the field whose doc says it overrides the endpoint. The embedded `OpenAICompatible` also carries its own unused `BaseURL`. Once shipped this is permanent public API.

go/plugins/compat_oai/meta/meta_test.go
```diff
@@ -0,0 +1,223 @@
...
+	g := genkit.Init(
+		ctx,
+		genkit.WithPlugins(plugin),
+		genkit.WithDefaultModel("meta/"+meta.ModelMuseSpark11),
+	)
+
+	if plugin.Name() != "meta" {
+		t.Fatalf("Name() = %q, want %q", plugin.Name(), "meta")
+	}
+	model := plugin.Model(g, meta.ModelMuseSpark11)
```
apascal07: This cannot fail. `plugin.Model` is `genkit.LookupModel`, which resolves dynamically on a miss, and the resolved descriptor advertises the same five capabilities asserted below. `LookupModel("meta/never-registered-9.9")` is non-nil with `media`, `multiturn`, `systemRole`, `tools`, and `toolChoice` all true. Deleting the whole `for model, modelOpts := range supportedModels` loop from `Init` leaves lines 111 to 127 passing. Only `label` and `versions` separate registered from resolved, and neither is checked. Line 211 has the same hole.

go/plugins/compat_oai/meta/meta_test.go
```diff
@@ -0,0 +1,223 @@
...
+		if body.PromptCacheRetention != "24h" {
+			t.Errorf("prompt_cache_retention = %q, want %q", body.PromptCacheRetention, "24h")
+		}
+		if body.ToolChoice != "required" {
+			t.Errorf("tool_choice = %q, want %q", body.ToolChoice, "required")
+		}
+
+		if body.Stream {
+			w.Header().Set("Content-Type", "text/event-stream")
+			for _, event := range []string{
```
apascal07: The fixture sends no usage chunk, but the plugin sets `stream_options.include_usage` on every streaming call, so the streamed response reports 0/0/0 tokens where the non-streaming path in the same test reports 2/3/5. Neither subtest asserts `resp.Usage`, so streamed usage accounting is untested against a stream a real endpoint would not send. The handler also never flushes, so a plugin that buffered the whole stream before firing a callback would pass.

go/plugins/compat_oai/meta/meta_test.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+		if got := streamed.String(); got != "Muse streaming works" {
+			t.Fatalf("streamed text = %q, want %q", got, "Muse streaming works")
+		}
+		if got := resp.Text(); got != streamed.String() {
+			t.Fatalf("final text = %q, want streamed %q", got, streamed.String())
+		}
+	})
+
+	if requests != 2 {
```
apascal07: This breaks running a subtest alone: `-run 'TestPluginRegistersMuseSparkAndGenerates/streaming'` fails with `requests = 1, want 2` while the selected subtest passes.

It also blocks the coverage that is missing. Both subtests force `tool_choice: required`, but the fixture answers with plain content and `finish_reason: stop`, which is not a response Meta can give for that request, so the tool round trip is never exercised. A correct fixture costs two requests per exchange and would trip this counter.

`requests` and `authorization` are also read unsynchronized. Round-trip causality makes that safe today, but the sibling tests guard the same counter with a mutex.

go/plugins/compat_oai/meta/meta_live_test.go
apascal07, line 27: `compat_oai/internal/livetest` is the shared checklist all eight siblings run: 15 subtests covering history, system prompts, tool calling, tool choice, structured output, json mode, reasoning, vision, and the extra config passthrough. This covers 2, and the catalog claims `Tools`, `ToolChoice`, and `Media` with none exercised live.

The streaming subtest passes on any non-empty output. The prompt asks for exactly "Muse streaming works" and nothing checks for it, and `resp.Text() == streamed.String()` holds by construction since the final response is accumulated from the same deltas. A refusal would pass. The complete subtest does check for "paris".

Both subtests force `reasoning_effort: high` on prompts that need no reasoning.

go/plugins/compat_oai/generate_test.go
```diff
@@ -217,10 +217,11 @@ func TestWithMessagesSkipsNilMessagesAndParts(t *testing.T) {
 
 func TestWithConfigPreservesProviderSpecificFields(t *testing.T) {
 	g := newGen().WithConfig(map[string]any{
-		"maxOutputTokens": 123,
-		"model":           "must-not-override",
-		"temperature":     0.3,
-		"topP":            0.8,
+		"maxOutputTokens":        123,
+		"model":                  "must-not-override",
+		"prompt_cache_retention": "24h",
```
apascal07: `TestWithConfigPreservesProviderSpecificFields` was removed in #5976, so this hunk does not apply to main. The assertion is also redundant: `meta_test.go:60` already checks `prompt_cache_retention` on the wire against a server, and `WithConfig`'s map branch is key-agnostic, so the existing `thinking` key already covered pass-through.

apascal07 (changes requested):
No custom config?

PR #6027, opened 2026-08-12
refactor(go): deduplicate plugin helpers and fix the defects the copies hid

apascal07:
googlegenai and `internal/anthropic` each carried the same walker for curating a reflected config schema, and provider naming, catalog overlays, and error classification were spelled several ways across the plugins. #5874 added `plugins/internal/catalog.go` for the overlay half of that.

Collapses the copies and migrates every plugin onto one spelling of each. Doing so is what surfaced the defects in the second half of this description: each was live in one copy of a mechanism and already fixed in another.

The shared code stays internal. Plugins are due to be sharded into their own modules, and that is when their common code needs a permanent home worth committing to; until then `plugins/internal` costs nothing to move.

## Where the shared code lives

`plugins/internal` gets the plugin-facing helpers: `SchemaOverrides` and `ApplySchemaOverrides`, `LookupOverride`, `TrimProvider`, and `ProviderLabel`.

The two schema primitives underneath, `WalkSubschemas` and `CloneSchema`, go to `internal/base` instead, because `ai/config.go` uses them for the null-tolerance and required-stripping work below and cannot import plugin code. `internal/base` is a leaf package with no genkit imports that `ai`, `core`, and five plugin packages already depend on.

**Before:** every plugin that curates a schema reimplements the navigation.

```go
func applyConfigOverrides(schema *jsonschema.Schema, o configOverrides)
func parsePath(path string) []string
func schemaAtPath(schema *jsonschema.Schema, steps []string) *jsonschema.Schema
func hideAtPath(schema *jsonschema.Schema, steps []string) bool
```

**After:** the plugin keeps only its data.

```go
var gccOverrides = internal.SchemaOverrides{
    Descriptions: map[string]string{"temperature": "Sampling temperature.", ...},
    Hidden:       []string{"tools", "responseSchema", ...},
}

schema := base.SchemaAsMap(r.Reflect(config))
internal.ApplySchemaOverrides(schema, overridesFor(config))
```

The shared version works on `map[string]any`, which is what `ModelOptions.ConfigSchema` already is, so a plugin reflecting with a different library converts once at the edge rather than being locked to invopop.

A hidden property is replaced by the permissive `true` schema rather than deleted, because the config schema is also enforced on every request: the field must still reach the plugin's own check, which names the Genkit option to use instead. Deleting it would fail validation as an unknown property.

`LookupOverride` also fixes a real inconsistency. Callers write `googleai/gemini-flash-latest` everywhere else, so a `Models` map keyed that way silently missed.

**Before:**

```go
if override, ok := c.models[id]; ok {
    return internal.OverlayModelOptions(base, override)
}
```

**After:**

```go
if override, ok := internal.LookupOverride(c.models, c.provider, id); ok {
    return base.Overlay(override)
}
```

`Overlay` is hand-written per struct rather than reflective, so a new field is a silent drop. `TestOverlayCoversEveryField` sets each field to a distinct value and fails naming the field Overlay ignored. It also rejects any `bool` field added to these structs, since `false` cannot mean "not specified". `TestCloneSupportsDetachesEveryField` is the same guard for the capability structs, which the constructors now copy.

## Defects the copies hid

**Short-circuited tool calls reported the wrong span subtype.** `recordToolShortCircuit` built its own span metadata with subtype `tool`, while real tool actions emit `tool.v2`. A tool answered by middleware was a different kind of span from the same tool answered normally. The googlecloud telemetry plugin matched the literal `"tool"`, so it now accepts both; that filter had already gone stale for real tool spans.

**The retry and fallback middleware misclassified unclassified errors.** Their local `classifiedStatus` treated `errors.As(err, &e)` as a hit even when it matched a typed-nil `*status.Error`, which `Convert` produces and the generated `AgentOutput.Error` field is whenever nothing failed. Such a chain read as a deliberate `INTERNAL` and was retried on that basis. `status.Classified` reports whether anything in the chain actually carries a status, and `Of` is now two lines on top of it rather than a second copy of the same walk.

**`Label` only defaulted when the whole options pointer was nil.** `NewModelAction("m", &ModelOptions{Versions: v}, fn)` shipped a blank dev UI label. All five constructors now default it whenever it is empty.

**The constructors kept the caller's `*ModelSupports`.** A plugin pointing a model table at one shared capabilities value had every action aliasing it. They now copy, slice fields included.

**A declared config schema was enforced verbatim while an inferred one was widened.** A partial config marshals its unset pointer and slice fields to `null`, so googlegenai rejected `{"temperature": 0.5, "stopSequences": null}` from the dev UI while an inferred schema accepted the equivalent. Enforcement now widens a clone, leaving the advertised schema exactly what the plugin declared.

`format.go` also loses 260 lines to a shared `baseHandler` embedded by all five output handlers, which were five copies of one streaming state machine.

## Not a breaking change

Nothing changes shape for an application. `internal.OverlayModelOptions`, `internal.OverlayEmbedderOptions`, `anthropic.ProviderLabel`, and the `apiModel` parameter on `anthropic.NewModel` (which every caller passed twice) are all under `plugins/internal`, which no external package can import.

The runtime changes are relaxations: a config that used to be rejected for an explicit `null` is now accepted, a label that used to be blank is now filled, and options that used to be aliased are now copied.

The one observable change is the tool short-circuit span subtype, `tool` to `tool.v2`. Custom telemetry matching the old value would stop seeing those spans, which is the point: it was matching a value real tool spans stopped emitting some time ago.

## New API

```go
// ai
func (ModelOptions) Overlay(override ModelOptions) ModelOptions
func (EmbedderOptions) Overlay(override EmbedderOptions) EmbedderOptions

// core/status
status.Classified(err error) (Name, bool)
```

That is the whole diff in exported surface across `ai`, `core`, `core/api`, `core/status`, and `genkit`, verified by diffing `go doc -short -all` against the merge base.

`SchemaOverrides.UnresolvedPaths` (internal) exists because an override path that stops resolving is a silent no-op. Both plugins assert it is empty in a test, so a renamed SDK field fails there instead of quietly dropping the curation.

## What this does not do

**Does not consolidate the two plugin generations.** The older `compat_oai` derivatives and the newer plugins still differ structurally. That is in flight separately, and it is why `zai`, `kimi`, and `dashscope` keep their triplicated test files here: rewriting them now would only create conflicts.

**Does not make the plugins independent.** Four cross-plugin packages remain, three of them pre-existing: `internal/anthropic` (used by `plugins/anthropic` and `vertexai/modelgarden`), `internal/jsonschema` (`compat_oai`, `internal/anthropic`), `internal/uri` (`googlegenai`, `ollama`, `internal/anthropic`), and now `plugins/internal` itself. Deliberate, per above.

**Does not touch the model catalogs.** `gemini-2.5-flash` is retired from the API but still listed in `supportedGeminiModels`, which breaks `TestGoogleAILive` on `main` today. Pruning it belongs with the catalog work.

## Verification

Go 1.26.3, from `go/`: `go test ./...` passes, 42 packages, no failures. `go test -race` on the 32 packages this touches is clean. `gofmt -l .` and `go vet ./...` are both clean.

Live, with real keys: `TestAnthropicLive` passes, 14 subtests, exercising prefixed override keys and partial typed configs through the changed paths. `TestGoogleAILive` fails on `models/gemini-2.5-flash is not found` (404), which reproduces identically on an unmodified worktree at the merge base `feb5f335`.

Each fix above has a test that fails without it. Two were verified by breaking the code deliberately: removing `Versions` from `Overlay` produces `ModelOptions.Versions: Overlay kept [value-1], want the override's [value-2]`, and adding a status without a sentinel produces `status "PROBE_ONLY" has no base sentinel`.

### DIFF
--- a/go/ai/background_model.go
+++ b/go/ai/background_model.go
@@ -203,22 +203,21 @@ func NewBackgroundModelAction[Config any](
 		panic("ai.NewBackgroundModelAction: checkFn is required")
 	}
 
-	if opts == nil {
-		opts = &BackgroundModelOptions{}
+	o := BackgroundModelOptions{}
+	if opts != nil {
+		o = *opts
 	}
-	labelExplicit := opts.Label != ""
+	labelExplicit := o.Label != ""
 	if !labelExplicit {
-		opts.Label = name
+		o.Label = name
 	}
-	if opts.Supports == nil {
-		opts.Supports = &ModelSupports{}
-	}
+	o.Supports = cloneModelSupports(o.Supports)
 
-	configSchema, inputSchema := modelConfigSchemas[Config](opts.ConfigSchema, opts.Versions)
+	configSchema, inputSchema := modelConfigSchemas[Config](o.ConfigSchema, o.Versions)
 
 	// The top-level Metadata wins over the embedded ModelOptions.Metadata on
 	// key conflicts.
-	metadata := modelActionMetadata(api.ActionTypeBackgroundModel, &opts.ModelOptions, configSchema, opts.ModelOptions.Metadata, opts.Metadata)
+	metadata := modelActionMetadata(api.ActionTypeBackgroundModel, &o.ModelOptions, configSchema, o.ModelOptions.Metadata, o.Metadata)
 
 	typedStartFn := func(ctx context.Context, req *ModelRequest) (*ModelOperation, error) {
 		// req.Config was normalized to the exact Config type by
@@ -230,12 +229,12 @@ func NewBackgroundModelAction[Config any](
 		return startFn(ctx, req, cfg)
 	}
 
-	mopts := &opts.ModelOptions
+	mopts := &o.ModelOptions
 
 	// normalizeConfig runs outermost so that the built-in wrappers and the
 	// start function all see the typed, converted config on the request.
 	fn := core.ChainMiddleware(
-		normalizeConfig[Config](name, opts.Versions),
+		normalizeConfig[Config](name, o.Versions),
 		simulateSystemPrompt(mopts, nil),
 		augmentWithContext(mopts, nil),
 		validateSupport(name, mopts),
@@ -255,7 +254,7 @@ func NewBackgroundModelAction[Config any](
 	// defaulted from the name yields to an explicit caller
 	// Metadata["description"]: leaving Description empty lets core's
 	// metadata fallback apply it.
-	description := opts.Label
+	description := o.Label
 	if !labelExplicit {
 		if _, ok := metadata["description"].(string); ok {
 			description = ""
@@ -266,7 +265,7 @@ func NewBackgroundModelAction[Config any](
 		Metadata:    metadata,
 		InputSchema: inputSchema,
 		Check:       checkFn,
-		Cancel:      opts.Cancel,
+		Cancel:      o.Cancel,
 	}, wrappedFn)}
 }
 
--- a/go/ai/config.go
+++ b/go/ai/config.go
@@ -19,11 +19,11 @@ package ai
 import (
 	"context"
 	"errors"
-	"fmt"
 	"maps"
 	"slices"
 
-	"github.com/firebase/genkit/go/core"
+	"github.com/firebase/genkit/go/core/api"
+	"github.com/firebase/genkit/go/core/status"
 	"github.com/firebase/genkit/go/internal/base"
 )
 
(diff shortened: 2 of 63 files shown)


adesinah (changes requested):
Good work on this. I have 2 observations, neither blocking:

1. vertexai/modelgarden/anthropic.go's Init drops a `if a == nil { a = &Anthropic{} } `guard. Pretty sure this is genuinely dead code (reassigning a local pointer receiver doesn't help the caller's nil, and the only construction site always uses &Anthropic{}), but it's not mentioned in the description and isn't really "deduplicate plugin helpers" — might be worth a one-line callout so it doesn't look like an accidental drop.
 
2. The description says "all five constructors" default Label when empty — I only count four in ai/ (Model, BackgroundModel, Embedder, Retriever); NewEvaluatorAction has no Label field. Probably just a miscount, not a code issue.

apascal07:
> Good work on this. I have 2 observations, neither blocking:
> 
> 1. vertexai/modelgarden/anthropic.go's Init drops a `if a == nil { a = &Anthropic{} } `guard. Pretty sure this is genuinely dead code (reassigning a local pointer receiver doesn't help the caller's nil, and the only construction site always uses &Anthropic{}), but it's not mentioned in the description and isn't really "deduplicate plugin helpers" — might be worth a one-line callout so it doesn't look like an accidental drop.
> 2. The description says "all five constructors" default Label when empty — I only count four in ai/ (Model, BackgroundModel, Embedder, Retriever); NewEvaluatorAction has no Label field. Probably just a miscount, not a code issue.

Ack.

adesinah (approved):
Approved with notes on what needs to be addressed in future PRs.

PR #6029, opened 2026-08-12
fix(go/ai): surface abnormal finish reasons instead of failing output parsing

apascal07:
Skips output parsing when a response did not run to a normal completion, so blocked and filtered responses reach the caller with their `FinishReason` and `FinishMessage` for every plugin and every format.

`GenerateWithRequest` ran `ParseMessage` on every response whenever an output format was set. A safety-blocked response has no content, so parsing failed and `Generate` returned `status.ErrInvalidOutput` ("model failed to generate output matching expected schema") instead of the response: the documented `resp.FinishReason == ai.FinishReasonBlocked` check could never fire when `WithOutputType` or `WithOutputSchema` was in play, and a blocked response that does carry text (Veo attaches filter explanations) surfaced as a bogus schema mismatch.

Parsing is now skipped only for finish reasons known to be abnormal (`blocked`, `aborted`, `interrupted`, `other`), at the single call site, which covers the json and enum handlers alike (both reject contentless messages). `unknown`, where plugin mappers land unrecognized provider reasons, keeps the parse path: `ParseMessage` is the only place the output schema is enforced (`Output()` extracts without validating), so skipping an unclassified reason would silently drop validation. Skipped responses still carry their format handler, so `resp.Output(&v)` works on demand when content exists. `length` output still parses and still fails on truncation, matching JS, which validates length-capped output but raises `GenerationBlockedError` before any schema validation. Callers could not have keyed on the old error to detect blocking (its status never identified it), so the visible change is contentless abnormal responses returning instead of erroring.

New `TestGenerateAbnormalFinishSkipsOutputParsing` covers blocked with a contentless message, blocked with a nil message, blocked with enum output, `other` with non-JSON text surviving untouched, and `stop` with non-conforming text still failing with `ErrInvalidOutput`. Go 1.26.3, full `go test ./...` in `go/`: 41 packages pass, plus a `-race` run on `go/ai`.

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -583,11 +583,19 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
 
 			if formatHandler != nil {
 				resp.formatHandler = streamingHandler
-				// This is legacy behavior. New format handlers should implement ParseMessage as a passthrough.
-				resp.Message, err = formatHandler.ParseMessage(resp.Message)
-				if err != nil {
-					logger.FromContext(ctx).Debug("model failed to generate output matching expected schema", "error", err.Error())
-					return nil, status.Errorf(status.ErrInvalidOutput, "model failed to generate output matching expected schema: %w", err)
+				// Only parse output when the model delivered it: a natural
+				// stop, a length-capped response, or a model that reports no
+				// finish reason. An abnormal termination (blocked, aborted,
+				// ...) carries no conforming output, and a schema error here
+				// would mask the FinishReason and FinishMessage the caller
+				// needs to handle it, so those responses pass through as-is.
+				if resp.FinishReason == FinishReasonStop || resp.FinishReason == FinishReasonLength || resp.FinishReason == "" {
+					// This is legacy behavior. New format handlers should implement ParseMessage as a passthrough.
+					resp.Message, err = formatHandler.ParseMessage(resp.Message)
+					if err != nil {
+						logger.FromContext(ctx).Debug("model failed to generate output matching expected schema", "error", err.Error())
+						return nil, status.Errorf(status.ErrInvalidOutput, "model failed to generate output matching expected schema: %w", err)
+					}
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


go/ai/generate.go
```diff
@@ -583,11 +583,19 @@ func GenerateWithRequest(ctx context.Context, r api.Registry, opts *GenerateActi
 
 			if formatHandler != nil {
 				resp.formatHandler = streamingHandler
-				// This is legacy behavior. New format handlers should implement ParseMessage as a passthrough.
-				resp.Message, err = formatHandler.ParseMessage(resp.Message)
-				if err != nil {
-					logger.FromContext(ctx).Debug("model failed to generate output matching expected schema", "error", err.Error())
-					return nil, status.Errorf(status.ErrInvalidOutput, "model failed to generate output matching expected schema: %w", err)
+				// Only parse output when the model delivered it: a natural
+				// stop, a length-capped response, or a model that reports no
```
cabljac: nit: hmm `Unknown` is the `default:` in the finish reason mappers, so an unrecognised lands there and skips ParseMessage. Wondering if it'd be better to invert and skip only the known bad set, so an unrecognised reason keeps the old behaviour instead of silently getting new behaviour. I think this would only show up for resp.Text() though, since Output() does its own extraction

cabljac (approved):
lgtm, one tiny nit/question, not blocking.

go/ai/generate.go
apascal07: Inverted in 795e2654b. What tipped it: `ParseMessage` is the only place the output schema is actually enforced. `Output()` extracts but never validates, and on garbage it returns `nil, nil`, so a skipped parse can silently hand back zero values. Skipping on `unknown` would therefore have dropped validation entirely for responses that may well be complete, which is worse than the failure mode in the other direction: an unmapped provider block reason surfaces as the old schema error until the plugin maps it, which is loud and gets fixed at the right layer. Now only `blocked`, `aborted`, `interrupted`, and `other` pass through unparsed, and there's a subtest pinning that `unknown` still fails eager validation.

PR #6030, opened 2026-08-12
feat(go/plugins/googlegenai): add Express Mode, HTTP options, and `Client` accessors; harden response handling

apascal07:
Closes the highest-impact gaps between the Go plugin and its JS sibling (`@genkit-ai/google-genai`), in two groups: configuration surface the Go plugin lacked entirely (Vertex AI Express Mode, custom endpoints, per-plugin HTTP options, credential injection, access to the underlying SDK client) and response-handling defects (blocked content surfaced as errors or silent empties, streaming dropped grounding metadata and could panic, embeds past the service's per-call limit failed outright).

## Blocked content is a response, not an error

A safety-blocked candidate arrives with a finish reason and no `Content`; the plugin treated that as malformed and returned `no valid candidates were found in the generate response`. A prompt-level block (no candidates, `PromptFeedback.BlockReason` set) returned an empty successful response with no finish reason at all. Both now return a `FinishReasonBlocked` response. For candidate-level blocks that matches JS; prompt-level blocks deliberately go further, where JS throws `FAILED_PRECONDITION: No valid candidates returned.` and drops the reason:

```go
resp, err := genkit.Generate(ctx, g, ...)
if resp.FinishReason == ai.FinishReasonBlocked {
    log.Printf("blocked: %s", resp.FinishMessage)
    // ratings under resp.Custom["candidates"], resp.Custom["promptFeedback"]
}
```

Two adjacent guards keep blocked turns from corrupting sessions: request building skips contentless messages (a blocked model turn replayed via `resp.History()` used to 400 every later turn), and a response with zero candidates and no block reason is now an error instead of a silent empty success.

## Streaming keeps candidate metadata and stops panicking

The stream fold rebuilt the final candidate from finish reason and text alone, so `GroundingMetadata`, `SafetyRatings`, `CitationMetadata`, and `URLContextMetadata` vanished from `Custom["candidates"]`, exactly where the README's Google Search and Maps examples read them. Chunks now merge metadata the way JS aggregates it (latest value wins, citations accumulate). Fixed in the same path: a nil dereference on empty streams, a use-before-error-check on the translated response, metadata-only interstitial chunks aborting the whole generation, and callbacks firing with empty content.

## Vertex AI Express Mode, custom endpoints, and HTTP options

**Before:** `VertexAI` required a project, a location, and Application Default Credentials, and panicked otherwise. `GoogleAI` accepted nothing but an API key.

**After:**

```go
// Express Mode: an API key, nothing else. Falls back to VERTEX_API_KEY,
// GOOGLE_API_KEY, or GOOGLE_GENAI_API_KEY when nothing names a project.
&googlegenai.VertexAI{APIKey: key}

// Custom endpoint: the gateway owns auth.
&googlegenai.VertexAI{BaseURL: "https://my-gateway.example.com"}

// Both plugins: endpoint, headers, transport, and (Vertex) identity.
&googlegenai.GoogleAI{BaseURL: gw, Headers: hdrs, HTTPClient: client}
&googlegenai.VertexAI{Credentials: creds}
```

Mode selection is explicit-config first, and a project or location (configured or in the environment) always outranks an ambient API key, the same precedence the genai SDK applies, so no existing deployment changes auth paths. The env key names follow JS and deliberately exclude `GEMINI_API_KEY`, which names a Gemini Developer API key Vertex rejects (verified live: such a key fails Express Mode requests with `PERMISSION_DENIED`). When an env var selects Express Mode, Init logs which one.

## The SDK client is reachable

`Client()` returns the configured `*genai.Client`, unlocking the Files, Caches, Batches, and Tunings APIs without constructing and authenticating a second client by hand.

## Embeds batch to the service's per-call limits

One `EmbedContent` call carried every input document, which fails past 100 docs on the Gemini API. Requests now split at the real limits: 100 (Gemini API `batchEmbedContents`), 250 (Vertex prediction), and 1 for Vertex Gemini and MaaS embedding models, which take a single input per request (the SDK enforces this client-side for most; `gemini-embedding-001` enforces it server-side). Embeddings come back one per input, in order.

## Rate-limit errors carry the service's retry delay

`RetryDelay(err)` extracts `google.rpc.RetryInfo` from 429s, through wrapped errors, and `wrapAPIError` stamps `retryAfterMs` into the status error's details, the key JS's `ErrorResponseMetadata` uses. Embedder, Imagen, and Veo errors now route through `wrapAPIError` too, so every action's failures are status-classified rather than only generate's.

## Cache TTL survives JSON

`WithCacheTTL` metadata asserted `.(int)`, so any JSON round-trip (dev UI, reflection server) delivered `ttlSeconds` as a `float64` and failed the request. All numeric forms are accepted now; non-positive values, including fractions that truncate to zero and would silently skip cache creation, are rejected loudly.

## API surface

```go
// Added: configuration fields
googlegenai.GoogleAI{BaseURL string, Headers http.Header, HTTPClient *http.Client}
googlegenai.VertexAI{APIKey string, Credentials *auth.Credentials, BaseURL string,
    Headers http.Header, HTTPClient *http.Client}

// Added: methods and helpers
func (*GoogleAI) Client() (*genai.Client, error)
func (*VertexAI) Client() (*genai.Client, error)
func RetryDelay(err error) (time.Duration, bool)
```

## Behavior changes, none breaking at compile time

Everything added is new fields and functions; no signature changed. Runtime deltas, all on paths that previously failed or misbehaved:

- Blocked content: a response with `FinishReasonBlocked` instead of an error (or, for prompt blocks, instead of a silent empty success).
- Zero candidates with no block reason: an error instead of an empty success.
- Streaming callbacks no longer fire for empty-content chunks.
- Embedder, Imagen, and Veo errors are `status.Error`-wrapped; `errors.As` still reaches the underlying `genai.APIError`.
- An embed request with zero documents returns zero embeddings instead of a server 400.
- A supplied `HTTPClient` is used verbatim (documented on the fields): the plugin does not layer its otel transport or, on Vertex, credentials over it.

## What was deliberately left out

- **The `ResponseJsonSchema` migration.** The lossy `genai.Schema` conversion (anyOf collapse, `$ref` recursion overflow) is a separate, larger change.
- **Blocked responses under `ai.WithOutputType`.** Core parses structured output before checking the finish reason, so a blocked response still surfaces as `ErrInvalidOutput` there. The fix belongs in `go/ai` and is queued separately.
- **Concurrent embed batches.** Sequential keeps quota behavior predictable; bounded concurrency is a follow-up if the single-input Vertex models prove too slow.

## Testing

Go 1.26.3. `go test -race` on the plugin: 94 tests pass, including fake-server SSE streams (metadata merge, empty stream, prompt block, mid-stream block), embed batching against a counting server (150 docs split 100+50), Express Mode env precedence with fabricated ADC, and RetryInfo extraction in both proto JSON forms. Full `go test ./...` in `go/`: 41 packages pass, no failures. Express-key rejection verified live against `aiplatform.googleapis.com`.

### DIFF
--- a/go/plugins/googlegenai/embedder.go
+++ b/go/plugins/googlegenai/embedder.go
@@ -5,25 +5,56 @@ package googlegenai
 
 import (
 	"context"
+	"slices"
+	"strings"
 
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/core/api"
 	"google.golang.org/genai"
 )
 
+// Per-request input limits for embedding calls. The Gemini API's
+// batchEmbedContents endpoint accepts up to 100 requests per call; the
+// Vertex AI prediction service accepts up to 250 instances per call.
+const (
+	googleAIEmbedBatchSize = 100
+	vertexAIEmbedBatchSize = 250
+)
+
+// embedBatchSize returns how many inputs a single EmbedContent call may
+// carry for the given backend and model. Vertex AI serves its Gemini
+// embedding models and the open-source MaaS embedding models through
+// endpoints that take one input per request (the SDK enforces this for most
+// of them, and the gemini-embedding-001 prediction service enforces it
+// server side), so those batch one at a time.
+func embedBatchSize(backend genai.Backend, model string) int {
+	if backend != genai.BackendVertexAI {
+		return googleAIEmbedBatchSize
+	}
+	if strings.Contains(model, "gemini") || strings.Contains(model, "maas") {
+		return 1
+	}
+	return vertexAIEmbedBatchSize
+}
+
 // newEmbedder creates an embedder without registering it. The framework
 // validates and deserializes the request's options into
 // [genai.EmbedContentConfig] before the embedder function runs; the config
 // schema is inferred from that type unless the caller overrides it.
+//
+// Requests with more inputs than the service accepts per call are split into
+// sequential batches transparently; the response carries one embedding per
+// input, in input order.
 func newEmbedder(client *genai.Client, id string, embedOpts *ai.EmbedderOptions) *ai.EmbedderAction {
+	backend := client.ClientConfig().Backend
 	provider := googleAIProvider
-	if client.ClientConfig().Backend == genai.BackendVertexAI {
+	if backend == genai.BackendVertexAI {
 		provider = vertexAIProvider
 	}
+	batch := embedBatchSize(backend, id)
 
 	return ai.NewEmbedderAction(api.NewName(provider, id), embedOpts, func(ctx context.Context, req *ai.EmbedRequest, embedConfig genai.EmbedContentConfig) (*ai.EmbedResponse, error) {
-		var content []*genai.Content
-
+		content := make([]*genai.Content, 0, len(req.Input))
 		for _, doc := range req.Input {
 			parts, err := toGeminiParts(doc.Content)
 			if err != nil {
@@ -34,13 +65,15 @@ func newEmbedder(client *genai.Client, id string, embedOpts *ai.EmbedderOptions)
 			})
 		}
 
-		r, err := client.Models.EmbedContent(ctx, id, content, &embedConfig)
-		if err != nil {
-			return nil, err
-		}
-		var res ai.EmbedResponse
-		for _, emb := range r.Embeddings {
-			res.Embeddings = append(res.Embeddings, &ai.Embedding{Embedding: emb.Values})
+		res := ai.EmbedResponse{Embeddings: make([]*ai.Embedding, 0, len(content))}
+		for group := range slices.Chunk(content, batch) {
+			r, err := client.Models.EmbedContent(ctx, id, group, &embedConfig)
+			if err != nil {
+				return nil, wrapAPIError(err)
+			}
+			for _, emb := range r.Embeddings {
+				res.Embeddings = append(res.Embeddings, &ai.Embedding{Embedding: emb.Values})
+			}
 		}
 		return &res, nil
 	})
--- a/go/plugins/googlegenai/embedder.go
+++ b/go/plugins/googlegenai/embedder.go
@@ -22,16 +22,17 @@ const (
 )
 
 // embedBatchSize returns how many inputs a single EmbedContent call may
-// carry for the given backend and model. Vertex AI serves its Gemini
-// embedding models and the open-source MaaS embedding models through
-// endpoints that take one input per request (the SDK enforces this for most
-// of them, and the gemini-embedding-001 prediction service enforces it
-// server side), so those batch one at a time.
+// carry for the given backend and model. On Vertex AI the SDK routes some
+// models to the embedContent API, which takes one content per request, and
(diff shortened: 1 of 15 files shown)


go/plugins/googlegenai/embedder.go
```diff
@@ -5,25 +5,56 @@ package googlegenai
...
+// carry for the given backend and model. Vertex AI serves its Gemini
+// embedding models and the open-source MaaS embedding models through
+// endpoints that take one input per request (the SDK enforces this for most
+// of them, and the gemini-embedding-001 prediction service enforces it
+// server side), so those batch one at a time.
+func embedBatchSize(backend genai.Backend, model string) int {
+	if backend != genai.BackendVertexAI {
+		return googleAIEmbedBatchSize
+	}
+	if strings.Contains(model, "gemini") || strings.Contains(model, "maas") {
```
cabljac: The service takes 250 for this model, not 1: 250 docs go through in one call, 251 gets a 400 saying "supported range is from 1 (inclusive) to 251 (exclusive)". The SDK routes it to `:predict` and skips the one-content guard (`models.go:5714`), and JS sends one `instances` array (`vertexai/embedder.ts:181`). As-is it is 250 sequential calls.

The Vertex text embeddings docs do say one input per request though, which is presumably where the note comes from. Deliberate conservative read, or worth taking the batching? Happy to send the one-liner:

    if (strings.Contains(model, "gemini") && model != "gemini-embedding-001") || strings.Contains(model, "maas") {

go/plugins/googlegenai/gemini.go
```diff
@@ -554,18 +630,36 @@ func translateCandidate(cand *genai.Candidate) (*ai.ModelResponse, error) {
...
-	if len(resp.Candidates) > 0 {
+	switch {
+	case len(resp.Candidates) > 0:
 		r, err = translateCandidate(resp.Candidates[0])
 		if err != nil {
 			return nil, err
 		}
-	} else {
-		r = &ai.ModelResponse{}
+	case promptBlocked(resp):
```
cabljac, line 655: nit: the description says prompt blocks "now return what JS returns", but I think JS throws on zero candidates, prompt blocks included (`googleai/gemini.ts:913-917`, `FAILED_PRECONDITION: No valid candidates returned.`). Candidate-level blocks do match it. This looks like the better behaviour, just not parity.

go/plugins/googlegenai/googlegenai.go
```diff
@@ -113,6 +207,44 @@ func (v *VertexAI) Name() string {
...
+		return "", ""
+	}
+	if os.Getenv("GOOGLE_CLOUD_PROJECT") != "" || vertexLocationFromEnv() != "" {
+		return "", ""
+	}
+	return vertexAPIKeyFromEnv()
+}
+
+// customEndpointOnly reports whether the only configuration present, here or
+// in the environment, is a custom base URL. The SDK treats that as a valid
```
cabljac: nit: I think the SDK supports a GOOGLE_VERTEX_BASE_URL as well, but  isn't checked here?

cabljac (approved):
lgtm

PR #6031, opened 2026-08-12
feat(go/plugins/compat_oai/anthropic): migrate to a typed config and the shared catalog

apascal07:
Migrates the compat `anthropic` plugin onto the typed-config base (#5976), the first sub-plugin PR of that train.

## The plugin and its config, at a glance

```go
type Anthropic struct {
	APIKey string
	Opts   []option.RequestOption
	Models map[string]ai.ModelOptions
}

type ChatConfig struct {
	compat_oai.RequestConfig
	Temperature     *float64        `json:"temperature,omitempty"`
	MaxOutputTokens int             `json:"maxOutputTokens,omitempty"`
	TopP            *float64        `json:"topP,omitempty"`
	StopSequences   []string        `json:"stopSequences,omitempty"`
	Thinking        *ThinkingConfig `json:"thinking,omitempty"`
}

type ThinkingConfig struct {
	Type         string `json:"type,omitempty"`
	BudgetTokens int    `json:"budgetTokens,omitempty"`
}
```

The released plugin took its credential from the environment alone; `APIKey` is additive, field-then-env. Either source derives both wire credentials (the bearer token for chat, `x-api-key` for the native models list); a key supplied only through `Opts` is opaque to the plugin, which is why the field exists, pinned by a precedence test.

The schema enforces the endpoint's documented limits: `temperature` runs 0 to 1 because that is where this endpoint caps it, and `thinking.budgetTokens` carries Anthropic's 1,024 floor (the API rejects smaller budgets, verified live), so an undersized budget fails boundary validation instead of at the endpoint. Two absences are deliberate and pinned by tests: `topP` declares no range (the endpoint documents none), and `thinking.type` no enum (the native set has grown before, so a list would reject values Anthropic accepts).

The catalog moves to the shared `ModelOptionsFor` overlay with a `Models` field, listing pages through the full models list, and the sample pins its config through `anthropic.ModelRef`. The package conformance sweep lands here with its first plugin (canonical camelCase names, a `version` key, no credential key, a description on every property); each later PR adds its own.

## Released surface

`compat_oai/anthropic` shipped in `go/v1.11.0`: `DefineModel` and `Model` are deprecated in place with signatures kept. Runtime tightens the way the base PR's `openai` does: an undeclared config field is rejected instead of riding along as an extra key, and a `temperature` above 1 fails instead of being sent and silently capped by the endpoint.

```go
// Added
type ChatConfig struct { compat_oai.RequestConfig; Temperature *float64; MaxOutputTokens int; TopP *float64; StopSequences []string; Thinking *ThinkingConfig }
type ThinkingConfig struct { Type string; BudgetTokens int }
Anthropic.Models map[string]ai.ModelOptions
func ModelRef(id string, config *ChatConfig) ai.ModelRef

// Deprecated, kept
anthropic.DefineModel, anthropic.Model
```

## The first live run fixed three things

**ANTHROPIC_BASE_URL now names the API origin.** The native SDKs treat the variable as the origin and add the version segment themselves, and environments export it in that form (Claude Code sets `https://api.anthropic.com`). Passed verbatim to the OpenAI SDK, whose paths carry no version, every request went to `/chat/completions` and answered 404. Init restores the `/v1` unless already present; the listing test's server answers only version-prefixed paths.

**Structured outputs are native from the 4.5 generation on.** The endpoint takes `response_format` in its `json_schema` form and rejects every other form with a 400. Models from 4.5 on declare `ConstrainedSupportAll`, and every capability set declares its output formats without `"json"`, so a schema-less JSON request sends no `response_format` at all (mechanism in the base PR).

**The 4.x `-latest` aliases never existed.** The wire answers `claude-haiku-4-5-latest` with model-not-found: from Claude 4 on the alias is the bare name, and from 4.6 on IDs are dateless pinned snapshots. Versions now carry the real aliases, and the catalog gains Fable 5, Opus 5, Sonnet 5, Opus 4.6 through 4.8, and Sonnet 4.6.

## Testing

Go 1.26.3: `go build ./...` and `go test ./plugins/compat_oai/...` pass at this level of the stack; the package is 10 cases counting subtests, the live one skipping without `ANTHROPIC_API_KEY`. The schema test pins the constraint values including the budget floor and both deliberate absences, a regression drives a 512-token budget into boundary validation, and the conformance sweep walks the advertised schema. The live test runs the shared checklist from the base PR's `internal/livetest` harness: Haiku 4.5 carries the whole list, thinking included (the endpoint accepts the knob but never returns the content), and a `thinking` key rides the `extra` passthrough. Against the real API: 15/15 pass, native structured output and vision included.

### DIFF
--- a/go/plugins/compat_oai/README.md
+++ b/go/plugins/compat_oai/README.md
@@ -186,9 +186,10 @@ providers disagree about which ones exist and what they are called: DeepSeek
 dropped the frequency and presence penalties, Z.ai caps `temperature` at 1,
 Kimi's K-series takes neither, and `maxOutputTokens` is `max_tokens` on some
 providers and `max_completion_tokens` on others. Use the same camelCase name
-other plugins use for the same setting.
+other plugins use for the same setting; `conformance_test.go` enforces that
+across the package.
 
-See the `openai` directory for a complete implementation.
+See the `openai` and `anthropic` directories for complete implementations.
 
 ## Running Tests
 
--- a/go/plugins/compat_oai/anthropic/README.md
+++ b/go/plugins/compat_oai/anthropic/README.md
@@ -36,8 +36,7 @@ import (
 
 ctx := context.Background()
 plugin := &anthropic.Anthropic{}
-g := genkit.Init(
-    ctx,
+g := genkit.Init(ctx,
     genkit.WithPlugins(plugin),
     genkit.WithDefaultModel("anthropic/claude-haiku-4-5-20251001"),
 )
--- a/go/plugins/compat_oai/anthropic/README.md
+++ b/go/plugins/compat_oai/anthropic/README.md
@@ -1,59 +1,99 @@
 # Anthropic Plugin
 
-This plugin provides a simple interface for using Anthropic's services.
+This plugin provides Genkit support for Claude models through Anthropic's
+OpenAI-compatible Chat Completions endpoint.
 
-## Prerequisites
+Anthropic positions that endpoint as a compatibility layer for testing and
+comparing Claude rather than a long-term integration surface: it does not
+return thinking content, ignores `response_format`, and hoists system
+messages to the start of the conversation. The full support matrix is at
+https://platform.claude.com/docs/en/cli-sdks-libraries/libraries/openai-sdk.
+For the native Messages API surface (thinking output, prompt caching,
+citations, structured outputs), use the `go/plugins/anthropic` plugin
+instead; this one is for code that must speak the OpenAI shape.
 
-- Go installed on your system
-- An Anthropic API key
+## Setup
 
-## Running Tests
-
-First, set your Anthropic API key as an environment variable:
+Set an Anthropic API key:
 
 ```bash
 export ANTHROPIC_API_KEY=<your-api-key>
 ```
-By default, `baseURL` is set to "https://api.anthropic.com/v1". However, if you
-want to use a custom value, you can set `ANTHROPIC_BASE_URL` environment variable:
 
-```bash
-export ANTHROPIC_BASE_URL=<your-custom-base-url>
+The plugin's `APIKey` field overrides the environment. The plugin uses
+`https://api.anthropic.com/v1` by default; set `ANTHROPIC_BASE_URL` or pass
+`option.WithBaseURL` through the plugin's `Opts` to use another compatible
+endpoint.
+
+```go
+import (
+    "context"
+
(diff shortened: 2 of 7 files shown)


adesinah (changes requested):
A few finding while doing some raw tests:

1. Claude 4.5+ is marked as supporting native constrained output, but Anthropic currently documents response_format as ignored by its OpenAI-compatible endpoint. This also prevents Genkit from adding its fallback schema instructions, so structured output may appear to work without actually being schema-enforced. Could we leave constrained support disabled for this endpoint?Implementation (https://github.com/genkit-ai/genkit/blob/332fdb3c420f4cdb6cefb55b92f010d9799f8258/go/plugins/compat_oai/anthropic/anthropic.go#L130-L138), Genkit fallback behavior (https://github.com/genkit-ai/genkit/blob/332fdb3c420f4cdb6cefb55b92f010d9799f8258/go/ai/generate.go#L417-L436), Anthropic docs (https://platform.claude.com/docs/en/cli-sdks-libraries/libraries/openai-sdk#request-fields).

2. thinking.budgetTokens currently accepts values from 1, while Anthropic requires a minimum of 1,024 and rejects anything lower.
 Could the schema use minimum=1024 so Genkit catches this before sending the request? Implementation (https://github.com/ genkit-ai/genkit/blob/332fdb3c420f4cdb6cefb55b92f010d9799f8258/go/plugins/compat_oai/anthropic/anthropic.go#L68-L76),
Anthropic docs (https://platform.claude.com/docs/en/build-with-claude/extended-thinking#budget-rules-and-tuning).

apascal07:
Checked both against the live endpoint.

1. Keeping constrained support: the compatibility table is stale against the server. Live, `response_format` with `json_schema` returns exactly schema-conforming JSON (`{"capital":"Paris","population":2161000}`) from a prompt that never mentions JSON, the same prompt with no response_format answers in prose, and `json_object` gets 400 "response_format.type: Input should be 'json_schema'". An ignored field cannot selectively 400 by variant, so the endpoint parses and enforces the field; the table's "Ignored" row predates its json_schema support. The 4.5+ entries keep native constrained support, which is also what the live suite's structured output cases exercise (they pass with no injected instructions).

2. Fixed: `thinking.budgetTokens` now declares `minimum=1024` (live, a 512 budget gets 400 "Input should be greater than or equal to 1024"), with a schema assertion and a regression driving an undersized budget into boundary validation before the wire.

adesinah (approved):
Looks good.

PR #6040, opened 2026-08-13
feat(go/plugins/anthropic): add `Opts` to pass SDK request options through

apascal07:
Adds `Opts []option.RequestOption` to the `Anthropic` plugin, opening the anthropic-sdk-go client configuration that had no way in: `option.WithMiddleware`, `option.WithMaxRetries`, `option.WithRequestTimeout`, extra headers, and the SDK's `bedrock` and `vertex` packages, whose routing helpers are request options too. `compat_oai` carries the same field for the same reason: a Stainless SDK's functional options are its entire client-config language, so the plugin passes the language through rather than re-enumerating it as fields.

The options derived from `APIKey` and `BaseURL` are applied first and the SDK applies options in order, so an explicit option in `Opts` wins on conflict, matching `compat_oai`'s precedence.

```go
// Tune the transport.
&anthropic.Anthropic{Opts: []option.RequestOption{
	option.WithMaxRetries(5),
	option.WithRequestTimeout(30 * time.Second),
}}

// Claude on Bedrock: the routing option signs requests, no API key involved.
&anthropic.Anthropic{Opts: []option.RequestOption{bedrock.WithLoadDefaultConfig(ctx)}}
```

## The API-key panic now fires only when nothing authenticates

Init panicked without an API key, which killed exactly the setups `Opts` exists for: the SDK's Bedrock and Vertex options sign requests themselves, and the SDK serves `ANTHROPIC_AUTH_TOKEN` bearer auth from the environment on its own, yet the plugin refused to start before any of that could apply. The panic now fires only when no authentication is configured anywhere: no key in `APIKey` or `ANTHROPIC_API_KEY`, no `ANTHROPIC_AUTH_TOKEN`, and no `Opts`. Options are opaque, so a non-empty `Opts` is trusted to carry authentication. A strict relaxation: every setup that started before starts identically.

## Scope

**No more flat fields.** `Headers`, `HTTPClient`, and the like stay out: `option.WithHeader` and `option.WithHTTPClient` already say it in the SDK's own words. `APIKey` and `BaseURL` stay as the two flat conveniences every plugin names the same way.

**Client-level only.** Nothing threads per-request options; the SDK carries them outside the params type, so that is its own design if ever needed.

## Testing

Go 1.26.3, `go test -race ./plugins/anthropic/`: 22 tests pass, four of them new, covering the auth waiver, `ANTHROPIC_AUTH_TOKEN`-only startup, option application, and field-vs-`Opts` precedence, observed at a recording `httptest` server.

### DIFF
--- a/go/plugins/anthropic/anthropic.go
+++ b/go/plugins/anthropic/anthropic.go
@@ -57,13 +57,27 @@ var dateSuffix = regexp.MustCompile(`-\d{8}$`)
 type Anthropic struct {
 	// APIKey is the key requests are authenticated with. When empty, the
 	// ANTHROPIC_API_KEY environment variable is used, and [Anthropic.Init]
-	// panics if that is empty too.
+	// panics unless authentication arrives another way: an auth token in
+	// ANTHROPIC_AUTH_TOKEN, or a non-empty Opts.
 	APIKey string
 
 	// BaseURL overrides the API endpoint requests are sent to. When empty, the
 	// ANTHROPIC_BASE_URL environment variable is used, and failing that the
 	// SDK's own default.
 	BaseURL string
+
+	// Opts are anthropic-sdk-go request options applied to every request the
+	// client sends. They open the SDK's full client configuration to the
+	// plugin: [option.WithMiddleware], [option.WithMaxRetries],
+	// [option.WithRequestTimeout], extra headers, and the SDK's bedrock and
+	// vertex packages, whose routing helpers are request options too.
+	//
+	// They are applied after the options derived from APIKey and BaseURL, and
+	// the SDK applies options in order, so an option here wins over those
+	// fields on conflict. Options are opaque, so a non-empty Opts is trusted
+	// to carry authentication when no API key is configured; that is what
+	// makes key-less setups such as Bedrock or Vertex routing work.
+	Opts []option.RequestOption
 
 	// Models overrides what the plugin knows about a Claude model, keyed by
 	// model ID, bare or provider-prefixed. Every Claude model already works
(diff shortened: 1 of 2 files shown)


go/plugins/anthropic/anthropic.go
```diff
@@ -57,14 +57,28 @@ var dateSuffix = regexp.MustCompile(`-\d{8}$`)
...
 	BaseURL string
 
+	// Opts are anthropic-sdk-go request options applied to every request the
+	// client sends. They open the SDK's full client configuration to the
+	// plugin: [option.WithMiddleware], [option.WithMaxRetries],
+	// [option.WithRequestTimeout], extra headers, and the SDK's bedrock and
+	// vertex packages, whose routing helpers are request options too.
+	//
+	// They are applied after the options derived from APIKey and BaseURL, and
+	// the SDK applies options in order, so an option here wins over those
```
cabljac, line 76: nit: might be wrong here, but I think this "wins over those fields" holds when an Opt sets the same setting, which is what `TestInitOptsWinOverFields` covers, but not across auth schemes:

`WithAPIKey` writes `X-Api-Key` and `WithAuthToken` writes `authorization`:
https://github.com/anthropics/anthropic-sdk-go/blob/v1.23.0/option/requestoption.go#L272-L285

so neither clears the other and both go out - with ANTHROPIC_API_KEY exported, the Anthropic key rides along as X-Api-Key to whatever host the Opts route to.

i'm not sure the plugin is able to guard it, so probably just doc change like

Maybe worth a sentence "key-less means ANTHROPIC_API_KEY unset in the environment too"

cabljac (approved):
lgtm, one nit

PR #6041, opened 2026-08-13
docs(go/plugins/anthropic): add a README

apascal07:
Adds a README for the native Anthropic plugin, which had none: setup, the dynamic model catalog with the `Models` overlay, the SDK-typed `MessageNewParams` config with its managed-field rejections and the server-side tools carve-out, and links to Anthropic's current documentation. Follows the structure of the `compat_oai` provider READMEs from the typed-config train (#5976).


cabljac (approved):
lgtm!

PR #6046, opened 2026-08-13
feat(py/plugins/amazon-bedrock): add rerank helper (Cohere, Amazon)

hilariie:
## Why

Slice 7 of the Bedrock Python port (#5820) adds reranking. Stacked on #6011. Reranking ships the same way Go ships it, as a plugin helper and not as a first-class ActionKind.RERANKER action as suggested by the issue.

## Changes

`Bedrock.rerank(model_id, query=..., documents=..., options=...)` calls Cohere Rerank over the existing InvokeModel transport. No new boto3 client, and no IAM beyond `bedrock:InvokeModel`.

Both rerank families are supported, and the body is built from the model ID because they are mutually incompatible: `cohere.rerank-*` requires `api_version` and rejects a body without it, `amazon.rerank-*` rejects the key outright as extraneous. Everything else about the two is identical, including the response shape. Go always sends `api_version`, so it cannot call `amazon.rerank-v1:0` at all (filed upstream as genkit-ai/aws-bedrock-go-plugin#190). That is the second divergence from `rerank.go`. The first: a malformed result entry is an error, where Go's `json.Unmarshal` zero-fills it, so a result missing `index` silently scores the first document. 

`resolve()` now returns None for rerank IDs under `ActionKind.MODEL`. They previously resolved as Converse chat models, which Cohere's model card marks unsupported, so the action could only fail when called. This flips an assertion added in slice 3.

### DIFF
--- a/py/packages/genkit-amazon-bedrock/README.md
+++ b/py/packages/genkit-amazon-bedrock/README.md
@@ -3,10 +3,11 @@
 Amazon Bedrock plugin for Genkit Python. Provides text generation with
 Bedrock-hosted models (Anthropic Claude, Amazon Nova, Meta Llama, Mistral,
 Cohere, and others) through the Bedrock Converse and ConverseStream APIs, and
-embeddings and image generation through InvokeModel.
+embeddings, image generation, and reranking through InvokeModel.
 
-> Status: in progress. Text generation (streaming and non-streaming),
-> embedders, and image generation are available. Reranking is still to come.
+> Status: text generation (streaming and non-streaming), embedders, image
+> generation, and reranking are available. The work left is sample app and
+> docsite coverage, not plugin surfaces.
 
 ## Installation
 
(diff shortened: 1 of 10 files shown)


cabljac (approved):
Looks good, approving. Cohere body matches Go's exactly, the Amazon branch correctly drops `api_version`, index bounds-checking handles the bool-is-an-int trap, and service order is pinned by a test.

You called the reranker-as-a-helper question right and #5820 is the document that's wrong. That enum member at `_core/_action.py:128` is the only occurrence of "rerank" in `py/packages/genkit/src/`: no `define_reranker`, no `ai.rerank()`, schema types denylisted at `py/scripts/schema_to_typing.py:17-28`. Registering would have worked mechanically but left callers on `resolve_action` plus `arun` with hand-rolled types. I'll fix the issue.

Only cost worth noting: a registered action would get reflection listing and tracing spans for free. Worth revisiting if `ai.rerank()` ever lands upstream.

PR #6050, opened 2026-08-13
feat(py/plugins/amazon-bedrock): scaffold the plugin package

### LINKED ISSUE #5820: [plugins] port Amazon Bedrock to Python

hilariie:
## Why

Closes #5820. Genkit Python has no Amazon Bedrock plugin; this ports the mature Go plugin (genkit-ai/aws-bedrock-go-plugin). This PR started as the base of a stacked-PR train, but the train has since been collapsed, so this PR now carries the entire plugin.

Each slice was reviewed separately, and those PR descriptions hold the detailed rationale and screenshots:

- #5699 model capability registry
- #5876 Converse non-streaming generate path
- #5970 ConverseStream streaming path
- #6010 embedders (Titan, Cohere v3, Nova)
- #6011 image generation (Nova Canvas, Stability)
- #6046 rerank helper (Cohere, Amazon)
- #6048 sample app and README
- #6084 Claude 4.7, 4.8 and 5 model registrations
- #6087 debug logging and CHANGELOG

On the dependency choice, plain boto3 was selected instead of aioboto3, because aiobotocore pins botocore to a narrow patch range that would dictate boto3 versions for every installer. A thread bridge keeps the event loop free, and all SDK calls sit behind a transport seam so the official async SDK can slot in once it matures.

## Changes

New `genkit-amazon-bedrock` package under `py/packages`, wired into the uv workspace:

- `Bedrock` plugin resolving any Bedrock model ID lazily, with a capability registry for known families and safe defaults for unknown ones.
- Text generation over Converse and ConverseStream: multi-turn chat, system prompts, tool calling, image/document input, reasoning content including Claude extended thinking, and prompt caching via `cache_point_part()`.
- `BedrockConfig` with camelCase aliases; unknown keys are rejected, and `additionalModelRequestFields` passes model-specific options through verbatim.
- Embedders for the Titan text, Titan multimodal, Cohere v3 and Nova-2 families.
- Image generation for Nova Canvas, Titan Image and Stability, configured with `BedrockImageConfig`.
- Rerank helper for the Cohere and Amazon rerank models.
- AWS wiring: region via the SDK resolution chain (no silent default), retry/timeout/pool knobs that defer to ambient AWS config, AWS errors mapped to Genkit statuses, and a whole-call `total_timeout` that also bounds streaming.
- Sample app under `py/samples/amazon-bedrock-sample` with flows covering each surface, plus package and sample READMEs.


py/packages/genkit-amazon-bedrock/tests/plugin_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+def test_bedrock_name_prefixes_model_id() -> None:
+    assert bedrock_name('anthropic.claude-sonnet-4-5-20250929-v1:0') == (
+        'bedrock/anthropic.claude-sonnet-4-5-20250929-v1:0'
+    )
+
+
+def test_constructor_defaults() -> None:
+    plugin = Bedrock()
+    assert plugin.region == 'us-east-1'
```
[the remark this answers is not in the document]
hilariie: resolved in [e2b0d5f](https://github.com/genkit-ai/genkit/pull/6050/commits/e2b0d5f2ac15dbd5b95776a5018a0a891ce4c872)

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/plugin.py
```diff
@@ -0,0 +1,121 @@
...
+                back to ``us-east-1`` (Go plugin parity).
+            max_retries: Retry limit for Bedrock API calls.
+            request_timeout: Per-call timeout in seconds.
+            session: Optional pre-configured ``boto3.session.Session`` for custom
+                credentials or advanced SDK wiring.
+            models: Bedrock models to register. Models not listed can still be
+                resolved dynamically by namespaced name.
+            embedders: Embedding model IDs to register (Titan, Cohere, Nova).
+        """
+        self.region = region or DEFAULT_REGION
```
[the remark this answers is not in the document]
hilariie: Resolved in [e2b0d5f](https://github.com/genkit-ai/genkit/pull/6050/commits/e2b0d5f2ac15dbd5b95776a5018a0a891ce4c872)

py/samples/amazon-bedrock-sample/src/main.py
```diff
@@ -0,0 +1,774 @@
...
+
+    Reasoning text streams delta by delta; the signature arrives in its own
+    delta that streams nothing, and is attached to the final reasoning part so
+    it can be replayed on the next turn.
+    """
+    stream_response = ai.generate_stream(
+        model=CLAUDE,
+        prompt=f'What is 17 * 23? Think it through, then state the answer. Mention {data.topic} once.',
+        config={
+            'maxTokens': 4096,
```
cabljac: Follow-through from my #5876 comment (https://github.com/genkit-ai/genkit/pull/5876#discussion_r3784183740): dropping the duplicate `maxTokens` key was the right call, but these four sample sites (:373, :603, :622, :623) still use it. With `extra='allow'` the key validates fine and is silently discarded, so from what I can see these flows now run with no token cap at all. This one matters most since it enables extended thinking with `budget_tokens: 1024` while the intended 4096 cap never gets sent. The other flows (:404, :430, :446) already use `maxOutputTokens`.

Can you switch these to `maxOutputTokens` and run one of them to check the cap lands in `inferenceConfig` on the trace? Longer term I'd lean towards `extra='forbid'` on `BedrockConfig` rather than aliasing `maxTokens` back in: an alias reintroduces the dual-key surface we removed in #5876 and only covers one spelling, while rejecting unknown keys catches the whole class of silent typos. Passthrough already has `additionalModelRequestFields` as its home. Worth checking nothing in core injects extra config keys first.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/transport.py
```diff
@@ -0,0 +1,365 @@
...
+        except asyncio.TimeoutError as e:
+            raise GenkitError(
+                message=f'bedrock converse failed: call exceeded the {self._total_timeout}s total timeout',
+                status='DEADLINE_EXCEEDED',
+            ) from e
+
+    def _converse_sync(self, kwargs: dict[str, Any]) -> dict[str, Any]:  # noqa: ANN401
+        return self.client().converse(**kwargs)
+
+    async def converse_stream(self, **kwargs: Any) -> AsyncGenerator[dict[str, Any], None]:  # noqa: ANN401
```
cabljac, line 201: Related to my timeout-config comment on #5876 (https://github.com/genkit-ai/genkit/pull/5876#discussion_r3784183735): `converse()` wraps its worker in `asyncio.wait_for(call, self._total_timeout)` at :188-196, but `converse_stream()` applies no deadline at all: not on the initial call, not per event, not overall. The docstrings (plugin.py:123-125, config.py:32-34) pitch `total_timeout` as the whole-call deadline covering the slow-dribble case the read timeout can't see, and streaming is exactly where that case lives. A stream trickling one event per hour runs forever, and `Bedrock(total_timeout=60)` gives no deadline on any streamed generate.

Suggest computing a monotonic deadline up front and bounding each event wait with the remaining budget. If we'd rather not touch the pump, the fallback is documenting `total_timeout` as non-streaming only, but that undercuts its own stated rationale.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/converters.py
```diff
@@ -0,0 +1,739 @@
...
+        raise GenkitError(message='bedrock: model request is missing', status='INVALID_ARGUMENT')
+    config = normalize_config(request.config)
+    messages, system = convert_messages(request.messages)
+
+    tools = request.tools or []
+    tool_choice = _requested_tool_choice(request, config) if tools else ''
+    # "none" means omit toolConfig entirely — Bedrock has no none mode.
+    send_tool_config = bool(tools) and tool_choice != 'none'
+    # Bedrock rejects a trailing assistant message only when a toolConfig is sent.
+    if send_tool_config and messages and messages[-1]['role'] == 'assistant':
```
cabljac, line 528: nit: When tools are configured and a toolConfig is actually sent (`tool_choice='none'` bypasses this, per the test at converters_test.py:543), a conversation ending in a model-role message gets that whole message dropped from the outbound request. The Bedrock constraint is real (ValidationException on a trailing assistant message with `toolConfig` present) and the behaviour is tested at :520, but the mitigation is silent: a caller doing assistant prefill with tools registered has the prefill discarded and the model re-answers the previous user turn, with nothing in logs or docs to say why. Silently rewriting caller-supplied history feels worse than the error it avoids.

I'd lean towards raising INVALID_ARGUMENT naming the constraint, so the caller decides. If there's a reason to keep the drop, a warning log plus a README note would at least make it discoverable. Not a blocker for this PR though, happy to leave it to a follow-up. If we go that way, let's open an issue to track it so it doesn't get lost.

py/samples/amazon-bedrock-sample/src/main.py
hilariie: Switched all four to `maxOutputTokens` in `163c77ac`.

Took `extra='forbid'` over the alias too. Nothing injects extra config keys (`version` is a declared field on `GenerationCommonConfig`) in core. `normalize_config` turns the pydantic error into INVALID_ARGUMENT listing the unknown keys and pointing at `additionalModelRequestFields`.

Screenshot showing `maxOutputTokens` now working:

Screenshot 2026-08-18 at 16 09 58

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/transport.py
hilariie: Fixed in `2b876a9c`. `converse_stream()` takes a monotonic deadline up front and bounds both the initial call and every event wait with what's left, raising DEADLINE_EXCEEDED like `converse()` does. The stream still gets closed on the way out, which is also what frees the parked worker thread.

Tests cover the deadline firing before the first event and mid-stream, plus `total_timeout=None` streaming freely.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/converters.py
hilariie: Raising INVALID_ARGUMENT now in `952c0b33`. Names the constraint and the two ways out: end on a user/tool message, or `tool_choice='none'`. Test at converters_test.py:542 updated to expect the error.

cabljac (approved):

PR #6054, opened 2026-08-13
feat(go): stream logs to the Dev UI, overhaul logging, classify provider errors

apascal07:
Adds trace-correlated log streaming to the Dev UI and rebuilds framework logging around it: a context-first API, one level taxonomy across core, genkit, ai, and the plugins, and a dev-mode exporter that ships every record at debug and above to the telemetry server, attached to the span that emitted it. The ingestion side already exists (from #4981); this makes Go the first runtime to feed it by default, matching the JS exporter's wire format. A companion Dev UI change, genkit-ai/genkit-ui#2041, renders each record's attributes inline in that panel.

image

## Logs reach the Dev UI attached to their span

```go
logger.Info(ctx, "looking up weather", "city", city)
```

In dev, a slog handler converts each record to OTLP-JSON, stamps the trace and span IDs from ctx, and a background worker batches and POSTs them. It never blocks: bounded queue, drop-and-count on overflow, one warning if the server is unreachable. Export turns on whenever a telemetry server URL is configured; `GENKIT_OTEL_ENABLE_LOGS=false` turns it off. JS reads the same variable but exports only on an explicit `true`, which the CLI never sets.

Console level and export are independent: the terminal shows info and above (`GENKIT_LOG_LEVEL`, `logger.SetLevel`), while the Dev UI always receives debug and above, so an interactive app keeps a quiet terminal while the full narrative lands in the trace viewer.

## Context-first logging API

**Before:** the helper resolved a logger nothing ever put in a context, and dropped ctx, so no handler could correlate a record with the active span:

```go
logger.FromContext(ctx).Debug("Action.Run", "name", a.Name())
```

**After:** package-level functions take ctx and hand it to the handler, and `FromContext` returns a logger bound to its context, so the old shape correlates too:

```go
logger.Debug(ctx, "span started", "name", name, "type", "action")
ctx = logger.WithContext(ctx, logger.FromContext(ctx).With("reqId", id))
```

```go
// Added
logger.Debug(ctx, msg, args...)   logger.Info(ctx, msg, args...)
logger.Warn(ctx, msg, args...)    logger.Error(ctx, msg, args...)
logger.WithContext(ctx, l)        logger.AddHandler(h)
logger.SetDefaultHandler(h)       logger.HasCustomDefault()
tracing.EnableLogExport(url)      tracing.LogExportHandler()
```

`AddHandler` tees a handler onto the default logger and survives `SetLevel`, which previously clobbered the default unconditionally. `SetLevel` and `GENKIT_LOG_LEVEL` now leave an application-installed handler in place and warn. `SetDefaultHandler` swaps the base while keeping sinks, so googlecloud's Cloud Logging handler no longer disconnects Dev UI streaming.

`AddHandler` substitutes Genkit's console when it would otherwise capture the stdlib default handler, which is not a real sink: it writes through the `log` package, whose output `slog.SetDefault` redirects back, so teeing it deadlocks on `log`'s mutex.

## One taxonomy across the framework

Debug carries the per-request narrative: span start and finish, the resolved generate request, each model turn with finish reason and token counts, tool batches, and middleware hooks. Middleware gets no spans of its own, so each hook is bracketed by debug records carrying the middleware name, duration, and whether it short-circuited the chain. Info is rare lifecycle. Warn is misconfiguration and skipped work, and a model finishing abnormally now warns instead of passing silently. Error is reserved for errors not also returned to the caller, so nothing is double-reported. This diverges from JS deliberately, which has no framework `info` and is silent out of the box. Hot paths build debug arguments only when a handler accepts them, so production pays one `Enabled` check per span.

## Provider errors carry a status, so retries mean something

`middleware.Retry` reissues any error nothing in its chain classified, whatever status set it is configured with, so a request that could never succeed was retried through the whole backoff schedule:

```go
return nil, fmt.Errorf("at least one message is required in generate request") // retried 3x
```

The larger half was the SDK boundary. Only googlegenai wrapped what its SDK returns for an HTTP response, so across the Anthropic and OpenAI-compatible families every provider 400, 401, 403, and 404 was retried. Both SDKs expose `StatusCode` on their error type, so each family gets a `WrapAPIError` mirroring googlegenai's, applied at all 13 boundaries where an SDK error escapes. Transport failures stay unclassified, since a dial timeout is worth retrying. Three sites also dropped a status the server had already sent, including two googlegenai cache paths that flattened theirs with `%v`.

The remaining 59 sites are request-shaping failures deterministic in their input, now `InvalidArgument`, `InvalidSchema`, or `Unimplemented`. Failures a retry can fix keep a retryable status on purpose: an empty stream is `Unavailable`, a truncated one `Internal`.

Classifying these surfaced a bug they had masked. The reflection error envelope carries a code the dev UI validates against its `Status` schema, canonical numbering where `INVALID_ARGUMENT` is 3, and `toReflectionError` filled it with the HTTP status, so every classified failure went out outside that enum and the UI had no status to render. Reflection v2, JS, and the cancellation path beside it all send canonical codes. That field also fed `WriteHeader`, where a canonical code would panic, so the transport status is now derived from it.

## A step can own its context, and text means text

```go
out, err := genkit.RunWithContext(ctx, "fetch", func(ctx context.Context) (string, error) { ... })
```

`Run` hands its callback no context, so work the callback starts nests beside the enclosing flow instead of under the step. `RunWithContext` gives the step its own; `Run` keeps its signature for pure work. `MediaParts`, on `Message` and `ModelResponse`, returns every media part with its content type, where `Media` returns only the first and leaves the type behind.

`Message.Text` and `ModelResponseChunk.Text` now concatenate text parts only. A data part carries a blob the plugins send as bytes, so counting one as text spliced base64 into prose. This matches JS, where text and data are separate fields, and the `accumulate` and `splitTextParts` helpers, which already filtered this way.

## Reported version tracks the module

`internal.Version` was a hand-bumped constant sitting at 1.4.0 with 1.11 released. It now resolves from the binary's embedded build info, or `dev` when Genkit is the main module, and runtime files, `x-genkit-version` headers, and googlecloud `sourceVersion` follow.

## Init no longer hangs when reflection startup fails

The V1 reflection launcher and Init's startup select both received from the same one-shot error channel, so a startup failure was a race: if the launcher won, it swallowed the error and Init blocked forever. A supervisor goroutine now owns the channel and hands Init a dedicated startup verdict.

## What this does not do

- **No CLI changes, and the Logs panel stays experiment-gated**: enable `otel_logs` at `/lab`. Un-gating it and having the CLI set `GENKIT_OTEL_ENABLE_LOGS` for JS are tools-repo follow-ups.
- **No googlecloud rework**: it keeps its Cloud Logging pipeline; only the handler installation goes through `logger.SetDefaultHandler`.
- **No exporter flush on exit**: records inside the 100ms batch window are lost when a short-lived process exits, and draining needs a shutdown hook the framework lacks.
- **Records logged outside any span** are exported and stored but not shown, since the UI only lists logs per span.

### DIFF
--- a/go/core/logger/logger.go
+++ b/go/core/logger/logger.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 
 	"github.com/firebase/genkit/go/internal/base"
+	otrace "go.opentelemetry.io/otel/trace"
 )
 
 var (
@@ -64,17 +65,29 @@ func isStdlibDefaultHandler(h slog.Handler) bool {
 }
 
 // SetLevel sets the minimum level of Genkit's console log handler and installs
-// it as the process-wide default logger, replacing the current one. Handlers
-// previously installed with [AddHandler] are preserved.
+// it as the process-wide default logger. Handlers previously installed with
+// [AddHandler] are preserved.
 //
-// Applications that manage their own [slog] handler should set their handler's
-// level directly instead of calling SetLevel.
+// A default handler the application installed itself (via [slog.SetDefault]
+// or [SetDefaultHandler]) is not Genkit's to manage: SetLevel records the
+// level for the managed console handler but leaves the application's handler
+// in place and warns, exactly as GENKIT_LOG_LEVEL does, since that handler's
+// own level is what governs output. Set the application handler's level
+// directly instead.
 func SetLevel(l slog.Level) {
 	mu.Lock()
-	defer mu.Unlock()
 	level.Set(l)
-	ensureConsole()
-	install(console)
+	custom := hasCustomDefaultLocked()
+	if !custom {
+		ensureConsole()
+		install(console)
+	}
+	mu.Unlock()
+	if custom {
+		// Logged after unlocking: the warning flows through the application's
+		// handler, which must be free to call back into this package.
+		slog.Warn("logger.SetLevel: the application installed its own default log handler; set its level directly", "level", l)
+	}
 }
 
 // ensureConsole creates the managed console handler if it does not exist yet.
@@ -127,11 +140,32 @@ func SetDefaultHandler(h slog.Handler) {
 func HasCustomDefault() bool {
 	mu.Lock()
 	defer mu.Unlock()
-	h := slog.Default().Handler()
-	if t, ok := h.(*teeHandler); ok {
-		h = t.handlers[0]
-	}
+	return hasCustomDefaultLocked()
+}
+
+// hasCustomDefaultLocked is [HasCustomDefault] for callers already holding mu.
+func hasCustomDefaultLocked() bool {
+	h := baseHandler(slog.Default().Handler())
 	return !isStdlibDefaultHandler(h) && h != console
+}
+
+// baseHandler strips the composition layers this package wraps around a base
+// handler: the tee that install builds, and the context binding [FromContext]
+// applies (in case an application handed a FromContext logger to
+// slog.SetDefault). What remains is the handler whose identity decides
+// whether the base is the stdlib default, the managed console, or
+// application-owned.
+func baseHandler(h slog.Handler) slog.Handler {
+	for {
+		switch v := h.(type) {
+		case *teeHandler:
+			h = v.handlers[0]
+		case *ctxHandler:
+			h = v.inner
+		default:
+			return h
+		}
+	}
 }
 
 // currentBase returns the handler that console output should continue to flow
@@ -142,10 +176,7 @@ func HasCustomDefault() bool {
 // carried over so the substitution does not change verbosity. Callers must
 // hold mu.
 func currentBase() slog.Handler {
-	h := slog.Default().Handler()
-	if t, ok := h.(*teeHandler); ok {
-		h = t.handlers[0]
-	}
+	h := baseHandler(slog.Default().Handler())
 	if isStdlibDefaultHandler(h) {
 		if console == nil {
 			seedLevelFromStdlib(h)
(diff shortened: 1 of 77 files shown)


go/core/logger/logger.go
```diff
@@ -58,3 +194,38 @@ func FromContext(ctx context.Context) *slog.Logger {
 	}
```
huangjeff5, line 244: `FromContext(ctx).Info(...)` still drops the span. `FromContext` only looks up the logger; stdlib `.Info` then calls `Log(context.Background(), ...)`, so `traceId` / `spanId` stay empty. The telemetry server stores the record; the UI only lists logs by span, so it never appears.

The sweep left the sites you'd actually look up: `go/ai/exp/agent.go` (state-transform failure, heartbeat refresh, snapshot finalize — two of those still use "err") and every MCP sample.

Migrating call sites one by one invites the next leftover. Can `FromContext` return a logger whose handler already closes over `ctx`, so the old `.Info` / `.Error` methods correlate?

go/core/logger/logger.go
```diff
@@ -21,33 +21,169 @@ import (
...
+	return t.Elem().PkgPath() == "log/slog" && t.Elem().Name() == "defaultHandler"
+}
+
+// SetLevel sets the minimum level of Genkit's console log handler and installs
+// it as the process-wide default logger, replacing the current one. Handlers
+// previously installed with [AddHandler] are preserved.
+//
+// Applications that manage their own [slog] handler should set their handler's
+// level directly instead of calling SetLevel.
+func SetLevel(l slog.Level) {
```
huangjeff5, line 77: `SetLevel` always `install(console)`. Sinks from `AddHandler` survive; the `SetDefaultHandler` base does not.

`HasCustomDefault` exists so `GENKIT_LOG_LEVEL` won't clobber a custom base, and `configureLogging` actually checks it. `SetLevel` does not. After `googlecloud.EnableGoogleCloudTelemetry` + `Init`, `logger.SetLevel(slog.LevelDebug)` silently disconnects Cloud Logging. That's the call `doc.go` teaches.

Can `SetLevel` do what `configureLogging` already does — if the current base isn't the managed console, warn and leave it alone (or only flip `level` when it is the console)?

huangjeff5 (approved):

PR #6079, opened 2026-08-17
feat(go/plugins/compat_oai/openrouter): add an OpenRouter plugin

apascal07:
Stacked on #6071, which lands first.

Adds an OpenRouter plugin to the `compat_oai` family, and the three response fields a gateway answers with that the shared converter dropped.

```go
g := genkit.Init(ctx,
    genkit.WithPlugins(&openrouter.OpenRouter{}),
    genkit.WithDefaultModel("openrouter/anthropic/claude-sonnet-4.5"),
)
```

A model ID keeps its upstream vendor's prefix, so an action name carries two slashes: the first is this plugin's provider prefix and the rest is the ID OpenRouter serves. Variant suffixes work the same way, including `:free`, `:nitro`, and `:floor`.

## The catalog is resolved, not curated

OpenRouter is not a model vendor. It fronts hundreds of models from dozens of vendors and adds more weekly, so the plugin curates no `supportedModels` map, registers nothing at `Init`, and returns no descriptors from `ListActions`. That last one is a size decision rather than a stylistic one: a descriptor carries full request and response schemas, so returning that catalog would put megabytes on every reflection poll. Models stay usable by name and only the browsable Dev UI list is absent.

Every resolved model is described permissively, because the two ways to be wrong are not symmetric. A capability declared too narrow is refused by Genkit before the request leaves, which blocks a model that would have worked; one declared too wide reaches OpenRouter and comes back with the real reason. Native constrained output is the exception, left unclaimed so structured output falls back to schema instructions in the prompt, which every model handles and which returns the same typed result.

`Models` narrows a model whose real capabilities are known, and a zero-value field keeps what the plugin resolves:

```go
&openrouter.OpenRouter{Models: map[string]ai.ModelOptions{
    "mistralai/mistral-7b-instruct": {Supports: &ai.ModelSupports{
        Multiturn: true, Tools: true, SystemRole: true,
    }},
}}
```

## The gateway controls are the config

A typed `ChatConfig` carries the fields that are the reason to route through a gateway at all, rather than leaving them to `extra`:

```go
genkit.Generate(ctx, g,
    ai.WithModel(openrouter.ModelRef("openai/gpt-5", &openrouter.ChatConfig{
        Provider: &openrouter.ProviderRouting{
            Sort:           openrouter.ProviderSortPrice,
            DataCollection: openrouter.DataCollectionDeny,
        },
        Models:    []string{"anthropic/claude-sonnet-4.5"},
        Reasoning: &openrouter.ReasoningConfig{Effort: openrouter.ReasoningEffortHigh},
    })),
    ai.WithPrompt("Work through this step by step."),
)
```

`provider` picks which upstreams may serve the request, `models` is a fallback chain tried in order, and `reasoning` sets effort or an exact token budget. `plugins` stays `[]map[string]any` on purpose: the roster and each plugin's options change on OpenRouter's schedule, so typing them would date the config faster than the gateway ships.

`maxOutputTokens` reaches OpenRouter as `max_tokens`, which reasoning models spend on thinking before emitting anything visible, so a budget comfortable for a plain model can return an empty answer with a `length` finish reason.

## Three response fields the shared converter dropped

These land in `generate.go`, so all eight subplugins get them.

**`usage.cost`** reaches `Usage.Custom["cost"]`. Presence decides rather than value: a `:free` model is priced at an explicit `0`, which is an answer and is kept, while a provider that does not price requests has no field and gains no key. OpenRouter used to gate this behind `usage: {include: true}`, which is now deprecated and does nothing.

**A provider failing mid-generation** arrives as a 200. The status and headers are committed by the time the upstream dies, so the failure cannot arrive as an error status. Where the error object lands depends on the transport, and so does what is left to recover.

On a non-streaming response it rides on the choice, as `finish_reason: "error"` plus an error object beside the text produced first. The response is otherwise well-formed, so before this it read as a complete answer with an `unknown` finish reason and no message. It now maps to `other`, the error's message becomes `FinishMessage`, and the object rides whole on `Raw["error"]`, carrying the code and the typed `error_type` naming the class of failure. The partial text stays: that 200 is the only delivery, and erroring would discard it.

On a stream the object arrives at the top level of a chunk instead. The SDK checks for that field before it unmarshals and ends the stream when it finds one, so the chunk never reaches the loop and no response is assembled. That stays an error rather than becoming a partial response. What the model generated first has already reached the stream callback, so the failure costs the aggregate rather than the output, and retry and fallback middleware key on the error rather than the finish reason: a truncated answer returned as a success would reach neither. The error's own code now classifies it, so a rate limit and a request the provider will refuse again are no longer alike reissued until the attempts run out.

A choice that finished cleanly keeps an empty `FinishMessage` even when an error-shaped extra rides beside it, since a finish message on a `stop` reads as the reason generation ended.

**`reasoning`**, the field OpenRouter normalizes every vendor's thinking onto, is read beside DeepSeek and Kimi's `reasoning_content` with the same precedence on both the streaming and non-streaming paths.

## `DefaultModelOptions` returns a fresh `Supports`

It returned `Supports: &Multimodal`, a pointer to the package-level var. Go hands out that pointer to every caller, so `opts.Supports.Media = false` wrote through to the global and silently narrowed every dynamically resolved model in the process, racing with concurrent resolution. `ai.NewModelAction` clones `Supports` at construction, which reads the global at resolution time, so the poison spread to everything resolved afterward rather than staying with the one caller. Each call now returns its own copy.

## API reference

```go
// Plugin
type OpenRouter struct{ APIKey, SiteURL, AppName string; Opts []option.RequestOption; Models map[string]ai.ModelOptions }
func (o *OpenRouter) Name() string
func (o *OpenRouter) Init(ctx) []api.Action
func (o *OpenRouter) ListActions(ctx) []api.ActionDesc
func (o *OpenRouter) ResolveAction(atype api.ActionType, id string) api.Action

// Config
func ModelRef(id string, config *ChatConfig) ai.ModelRef
type ChatConfig struct{ ...; Provider *ProviderRouting; Models []string; Reasoning *ReasoningConfig;
                        Plugins []map[string]any; Transforms []string; SessionID string;
                        ServiceTier ServiceTier; Metadata map[string]string;
                        TopK, MinP, TopA, RepetitionPenalty float64 }
func (c ChatConfig) ApplyToChatCompletion(params *openai.ChatCompletionNewParams)
type ProviderRouting struct{ Order, Only, Ignore, Quantizations []string; Sort ProviderSort;
                             MaxPrice *MaxPrice; DataCollection DataCollection;
                             ZDR, RequireParameters, AllowFallbacks *bool }
type MaxPrice struct{ Prompt, Completion, Request, Image *float64 }
type ReasoningConfig struct{ Effort ReasoningEffort; MaxTokens int; Exclude *bool }

// Closed sets, per the family rule that an enum-tagged field is a named type
type ReasoningEffort string  // None, Minimal, Low, Medium, High, XHigh, Max
type ProviderSort string     // Price, Throughput, Latency
type DataCollection string   // Allow, Deny
type ServiceTier string      // Auto, Default, Fast, Flex, Priority, Scale
```

## What this deliberately leaves out

- **No model catalog and no Dev UI list.** Both scale with OpenRouter's catalog rather than with what the application uses. `Models` covers correcting a specific model.
- **Three request fields.** `n` bills for choices Genkit never reads, and `route` and `usage` are deprecated by OpenRouter and do nothing. Everything else undeclared still reaches the wire through `extra`.
- **Chat completions only.** OpenRouter's image and audio generation endpoints are separate surfaces.
- **`url_citation` annotations.** OpenRouter's web plugin and OpenAI's `web_search_options` both return them and both drop them today. Surfacing them needs a shape decision (source part, part metadata, or `Raw`) that applies to the whole family, so it belongs in its own change.
- **`native_finish_reason`.** It carries the upstream vendor's own reason beside the normalized one, with no `ai.ModelResponse` field to land in. `Raw["error"]` covers the failure case that motivated reading it.

### DIFF
--- a/go/plugins/compat_oai/generate.go
+++ b/go/plugins/compat_oai/generate.go
@@ -506,8 +506,12 @@ func (g *ModelGenerator) generateStream(ctx context.Context, req *ai.ModelReques
 	// The accumulator knows nothing of the citations xAI answers a live search
 	// with either, so they are carried out of the chunk that has them.
 	var citations any
-	// The error object a gateway attaches to a failing choice is raw JSON the
-	// accumulator drops with the rest, so it too is carried out by hand.
+	// An error object a provider attaches to a failing choice is raw JSON the
+	// accumulator drops with the rest, so it too is carried out by hand. That
+	// is the shape of a failure a gateway reports on the response as a whole; a
+	// failure part-way through a stream is reported at the top level of a chunk
+	// instead, which ends the stream rather than reaching this loop. See
+	// [wrapStreamError].
 	var failure map[string]any
 
 	for stream.Next() {
@@ -570,10 +574,14 @@ func (g *ModelGenerator) generateStream(ctx context.Context, req *ai.ModelReques
 		}
 	}
 
-	// NewStreaming defers its HTTP error to the stream, so a 4xx on the
-	// request that opened it surfaces here rather than at the call.
+	// NewStreaming defers its HTTP error to the stream, so a 4xx on the request
+	// that opened it surfaces here rather than at the call, as does a gateway's
+	// mid-stream failure. Whatever was generated before the failure has already
+	// reached the callback, and the aggregate response is dropped: a caller and
+	// the middleware around it are told the generation failed, rather than
+	// handed a short answer that reads as a complete one.
 	if err := stream.Err(); err != nil {
-		return nil, fmt.Errorf("stream error: %w", WrapAPIError(err))
+		return nil, wrapStreamError(err)
 	}
 
 	if usageSeen {
@@ -614,6 +622,37 @@ func (g *ModelGenerator) generateStream(ctx context.Context, req *ai.ModelReques
 	}
 	resp.Request = req
 	return resp, nil
+}
+
+// wrapStreamError classifies the error a stream ended with, so that
+// status-aware middleware can tell a rate limit from a request the provider
+// will refuse again.
+//
+// A gateway whose upstream fails part-way through a generation reports it as an
+// error object at the top level of a chunk. The SDK looks for that field before
+// it unmarshals and ends the stream when it finds one, so the chunk never
+// reaches the caller and the object survives only as raw JSON inside the error
+// message. Reading the code back out of it is the one way to recover the status
+// the failure carries.
+//
+// An error that classifies to nothing is left unclassified rather than marked
+// Unknown, since the retry middleware reissues an unclassified error and gives
+// up on an Unknown one, and a failure this cannot read is not a reason to stop
+// trying.
+func wrapStreamError(err error) error {
+	err = WrapAPIError(err)
+	if _, classified := status.Classified(err); classified {
+		return fmt.Errorf("stream error: %w", err)
+	}
+	message := err.Error()
+	if brace := strings.IndexByte(message, '{'); brace >= 0 {
+		if code, ok := extractErrorObject(message[brace:])["code"].(float64); ok {
+			if name := status.FromHTTPCode(int(code)); name != status.Unknown {
+				return status.Errorf(status.Base(name), "stream error: %w", err)
+			}
+		}
+	}
+	return fmt.Errorf("stream error: %w", err)
 }
 
 // extractTokenCount reads a token count a provider reports as a usage field the
@@ -818,9 +857,9 @@ func convertChatCompletionToModelResponse(completion *openai.ChatCompletion) (*a
 	if citations := extractJSONValue(completion.JSON.ExtraFields["citations"].Raw()); citations != nil {
 		custom["citations"] = citations
 	}
-	// The error object rides whole beside the finish message it fed: retrying
-	// around a failed upstream takes its name and the failure's code, which
-	// have no [ai.ModelResponse] field but belong to the response's metadata.
+	// The error object rides whole beside the finish message it fed: the code
+	// and the typed error_type saying which class of failure it was have no
+	// [ai.ModelResponse] field but belong to the response's metadata.
 	if failure != nil {
 		custom["error"] = failure
 	}
--- a/go/plugins/compat_oai/generate.go
+++ b/go/plugins/compat_oai/generate.go
@@ -506,6 +506,9 @@ func (g *ModelGenerator) generateStream(ctx context.Context, req *ai.ModelReques
 	// The accumulator knows nothing of the citations xAI answers a live search
 	// with either, so they are carried out of the chunk that has them.
 	var citations any
+	// The error object a gateway attaches to a failing choice is raw JSON the
+	// accumulator drops with the rest, so it too is carried out by hand.
+	var failure map[string]any
 
 	for stream.Next() {
 		chunk := stream.Current()
@@ -524,13 +527,19 @@ func (g *ModelGenerator) generateStream(ctx context.Context, req *ai.ModelReques
 			continue
 		}
 
+		if value := extractErrorObject(chunk.Choices[0].JSON.ExtraFields["error"].Raw()); value != nil {
+			failure = value
+		}
+
 		// Create chunk for callback
 		modelChunk := &ai.ModelResponseChunk{}
 
-		// Some OpenAI-compatible reasoning models, including Kimi, return
-		// reasoning in the non-standard reasoning_content field.
-		if reasoningDelta := extractReasoningContent(
+		// Some OpenAI-compatible reasoning models return reasoning in a
+		// non-standard field: reasoning_content for Kimi and DeepSeek,
+		// reasoning for OpenRouter.
+		if reasoningDelta := firstReasoning(
 			chunk.Choices[0].Delta.JSON.ExtraFields["reasoning_content"].Raw(),
+			chunk.Choices[0].Delta.JSON.ExtraFields["reasoning"].Raw(),
 		); reasoningDelta != "" {
 			reasoning.WriteString(reasoningDelta)
 			modelChunk.Content = append(modelChunk.Content, ai.NewReasoningPart(reasoningDelta, nil))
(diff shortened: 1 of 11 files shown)


go/plugins/compat_oai/generate.go
```diff
@@ -524,13 +527,19 @@ func (g *ModelGenerator) generateStream(ctx context.Context, req *ai.ModelReques
 			continue
 		}
 
+		if value := extractErrorObject(chunk.Choices[0].JSON.ExtraFields["error"].Raw()); value != nil {
```
adesinah, line 534: There seems to be an issue with how we handle/campture error in mid-stream errors situation. Going by the doc: (https://openrouter.ai/docs/api_reference/errors-and-debugging#mid-stream-errors) with error at the top level of the SSE chunk, but this code reads choices[0].error.

The pinned openai-go stream decoder (https://github.com/openai/openai-go/blob/v1.8.2/packages/ssestream/ssestream.go#L150-L170) intercepts a top-level error before unmarshalling the chunk. The loop therefore never sees it, and stream.Err() returns:
https://github.com/genkit-ai/genkit/blob/c2ac1f5efd68385600e8b4fd8000f5106f570c46/go/plugins/compat_oai/generate.go#L573-L577 losing the final partial response.

I reproduced this using OpenRouter’s documented event shape: the callback received the partial text, but Generate returned a nil response and a stream error. Could we add a regression test with the top-level shape and preserve the partial response here?

I was able to reproduce that using this code snippet:
```go

// TestGenerateStreamReportsTopLevelProviderFailure uses OpenRouter's
// documented mid-stream error shape. Unlike a non-streaming failure, the
// error object is at the top level of the terminal SSE chunk, alongside the
// choices array. The partial response must remain available to the caller.
func TestGenerateStreamReportsTopLevelProviderFailure(t *testing.T) {
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "text/event-stream")
		for _, event := range []string{
			`{"id":"1","object":"chat.completion.chunk","created":1,"model":"openai/gpt-5","choices":[{"index":0,"delta":{"role":"assistant","content":"partial out"},"finish_reason":null}]}`,
			`{"id":"1","object":"chat.completion.chunk","created":1,"model":"openai/gpt-5","provider":"Together","error":{"code":502,"message":"Provider disconnected","metadata":{"error_type":"provider_unavailable"}},"choices":[{"index":0,"delta":{"content":""},"finish_reason":"error","native_finish_reason":"provider_error"}]}`,
		} {
			_, _ = io.WriteString(w, "data: "+event+"\n\n")
		}
		_, _ = io.WriteString(w, "data: [DONE]\n\n")
	}))
	t.Cleanup(srv.Close)

	client := openai.NewClient(option.WithBaseURL(srv.URL), option.WithAPIKey("stub"))
	messages := []*ai.Message{ai.NewUserTextMessage("hi")}
	resp, err := NewModelGenerator(&client, "openai/gpt-5").
		WithMessages(messages).
		Generate(context.Background(), &ai.ModelRequest{Messages: messages},
			func(context.Context, *ai.ModelResponseChunk) error { return nil })
	if err != nil {
		t.Fatalf("Generate() error = %v, want the partial response returned", err)
	}
	if resp == nil {
		t.Fatal("Generate() response = nil, want the partial response returned")
	}

	if resp.FinishReason != ai.FinishReasonOther {
		t.Errorf("FinishReason = %q, want %q", resp.FinishReason, ai.FinishReasonOther)
	}
	if want := "Provider disconnected"; resp.FinishMessage != want {
		t.Errorf("FinishMessage = %q, want %q", resp.FinishMessage, want)
	}
	if want := "partial out"; resp.Text() != want {
		t.Errorf("Text() = %q, want %q", resp.Text(), want)
	}
	custom, _ := resp.Custom.(map[string]any)
	failure, _ := custom["error"].(map[string]any)
	if got, _ := failure["code"].(float64); got != 502 {
		t.Errorf("custom[error][code] = %v, want 502", failure["code"])
	}
}

```

adesinah (changes requested):

go/plugins/compat_oai/generate.go
apascal07: Confirmed on both counts, thank you.

The docs are explicit that a mid-stream failure puts the error object at the top level of the chunk, and they document no choice-level `error` for streaming at all. `openai-go` checks that field before it unmarshals (`packages/ssestream/ssestream.go:168-172`), so the chunk never reaches the loop and the stream ends instead. Your shape reproduces exactly as you describe:

```
streamed to callback = "partial out"
resp == nil: true
err = stream error: received error while streaming: {"code":502,"message":"Provider disconnected","metadata":{"error_type":"provider_unavailable"}}
```

The choice-level read stays, because it is the documented shape for a non-streaming response and this converter is shared by eight providers. Its comment now says that, rather than implying it covers a stream.

I have kept the failure as an error rather than returning the partial response. `Retry` and `Fallback` both key on `err != nil` (`plugins/middleware/retry.go:143`, `plugins/middleware/fallback.go:89`), not on the finish reason. A partial returned with `FinishReasonOther` would stop a dead upstream from being retried or rerouted, which is the case a gateway most needs rerouted. The generated text is not lost either way: the stream callback has already received it, so only the aggregate goes.

That also keeps the plugin with the rest of Genkit. Every Go plugin drops the aggregate on a stream failure (`googlegenai/gemini.go:211`, `internal/anthropic/anthropic.go:293`, `ollama/ollama.go:719`), and JS does the same (`compat-oai/src/model.ts:718`, `anthropic/src/models.ts:186`). Nothing in either returns in-progress content beside an error.

The two transports differ for a real reason. Non-streaming, the 200 is the only delivery, so keeping the partial needs a nil error. Streaming, the partial is already delivered, so the error costs only the aggregate and buys back retryability.

What your comment did change:

- The stream error is now classified from the code inside the error object, so a 429 reaches middleware as `RESOURCE_EXHAUSTED` and a 502 as `INTERNAL`. It was unclassified before, which the retry middleware reissues regardless of what it says, so a request the provider would refuse again burned every attempt. A code that maps to nothing stays unclassified rather than becoming `UNKNOWN`, which retry would instead give up on.
- `TestGenerateStreamTopLevelFailureEndsStream` pins your shape: the callback receives the partial, the response is nil, the gateway's own message survives the wrapping, and each code classifies as expected.
- `TestErrorStatusClassifiedLive` checks the classification against the real gateway on both transports, with two refusals rather than one so the codes have to be told apart: a rejected key arrives as `UNAUTHENTICATED`, a model no allowed provider serves as `NOT_FOUND`. A mid-stream failure is not in there, since it needs an upstream to die mid-response and no request can provoke that.
- The error metadata in the tests is now `error_type`, which is what the docs document for a provider failure. `provider_name` appears only on moderation errors.

The full live suite is green against OpenRouter: 15 suite cases, cost accounting, gateway controls, and the refusal statuses above.

adesinah (approved):

PR #6090, opened 2026-08-18
fix(go/ai): stop masking abnormal finish reasons with schema errors

apascal07:
`Generate` skips output parsing when a response finished `blocked`, `aborted`, `interrupted`, or `other`, so the caller reads the finish reason rather than a parse failure. The typed helpers did not: `GenerateData`, `GenerateDataStream`, `DataPrompt.Execute`, and `DataPrompt.ExecuteStream` parsed unconditionally, so a safety-blocked response came back as `data did not match expected schema: (root): Invalid type. Expected: object, given: null` while `resp.FinishReason` said `blocked`. `DataPrompt` also never had the no-text escape the generate helpers have, so a turn holding only tool requests, which is what `WithReturnToolRequests` asks for, failed the same way.

All four now share one guard. Skipping the parse is only half of it: handing back a nil value with a nil error reads as success, and the `DefineDataPrompt` example in `genkit.go` does exactly that, printing an empty capital for a refusal it never sees. A refusal is now an error.

```go
// Added
ai.ErrGenerationBlocked = status.ErrFailedPrecondition.Subtype("generation blocked")
```

The typed helpers return it when `FinishReason` is `blocked`, carrying the provider's `FinishMessage`, with the response alongside it. FAILED_PRECONDITION is the status JS `GenerationResponseError` defaults to, so a refusal reports the same way in both; Python does the same in #6100.

Everything else that yields nothing to extract keeps a nil output and a nil error, since none of it is a failure: interrupts, tool requests, no text at all, and the `aborted` and `other` finishes. For those a string `Out` now leaves the text on `resp.Text()` rather than returning it as the answer. `Generate` is unchanged and still hands a blocked response back as a value. Matching JS `assertValid` there would touch every existing caller and belongs in its own PR.

**Streaming chunks are provisional.** They parse as they arrive, before any finish reason exists, so a generation that streams `{"title":"partial` and then blocks has already yielded a populated chunk. The terminal value settles it, and for a refusal that is now an error rather than a zeroed `Output`.

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -328,6 +328,10 @@ func LookupModel(r api.Registry, name string) Model {
 // FinishReasonUnknown is not in the set on purpose: plugins map unrecognized
 // provider reasons to it, so treating it as abnormal would silently drop
 // output validation for responses the model may well have completed.
+//
+// FinishReasonBlocked is in the set, because a refusal must skip parsing like
+// any other abnormal finish, but the typed helpers test for it first and
+// report [ErrGenerationBlocked] instead of reaching here.
 func (fr FinishReason) isAbnormal() bool {
 	switch fr {
 	case FinishReasonBlocked, FinishReasonAborted, FinishReasonInterrupted, FinishReasonOther:
@@ -335,6 +339,17 @@ func (fr FinishReason) isAbnormal() bool {
 	default:
 		return false
 	}
+}
+
+// blockedError reports a refusal as [ErrGenerationBlocked], carrying the
+// provider's explanation when there is one. Only the typed helpers call it:
+// they promise a value that a refusal cannot produce, whereas [Generate]
+// hands the response back and lets the caller read FinishReason.
+func blockedError(resp *ModelResponse) error {
+	if resp.FinishMessage == "" {
+		return status.Errorf(ErrGenerationBlocked, "generation blocked")
+	}
+	return status.Errorf(ErrGenerationBlocked, "generation blocked: %s", resp.FinishMessage)
 }
 
 // GenerateWithRequest is the central generation implementation for ai.Generate(), prompt.Execute(), and the GenerateAction direct call.
@@ -1038,12 +1053,14 @@ func GenerateText(ctx context.Context, r api.Registry, opts ...GenerateOption) (
 	return res.Text(), nil
 }
 
-// GenerateData runs a generate request and returns strongly-typed output.
-// If the response doesn't contain text output (e.g., contains tool requests
-// or interrupts instead), or generation ended abnormally (blocked, aborted,
-// interrupted, other), the output will be nil and no error is returned. Check
-// resp.FinishReason, resp.Interrupts() or resp.ToolRequests() to handle these
-// cases.
+// A refusal is an error: when the response finished blocked, the output is nil
+// and the error is [ErrGenerationBlocked], carrying the provider's explanation.
+// The response is still returned alongside it.
+//
+// Every other finish that yields nothing to extract is not an error. If the
+// response carries no text (tool requests or interrupts instead), or ended
+// aborted, interrupted, or other, the output is nil and the error is nil.
+// Check resp.FinishReason, resp.Interrupts(), and resp.ToolRequests().
 //
 // The output format is JSON with a schema inferred from Out; an explicit
 // [WithOutputSchema] or [WithOutputSchemaName] overrides the schema while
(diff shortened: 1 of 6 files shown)


huangjeff5 (commented):
Skipping ParseMessage so we don't mask FinishReason is the right cut. Two leftovers this PR doesn't touch:

1. GenerateText still `return res.Text(), nil` (go/ai/generate.go:1005). DataPrompt[In,string] now withholds text on an abnormal finish; jokesFlow in go/samples/basic will still serve "Response was blocked for safety reasons." as the joke. Can GenerateText consult isAbnormal the same way?

2. GenerateDataStream still parses chunks first (chunk.Output at go/ai/generate.go:1138). A model that streams leftover JSON then finishes blocked will yield a typed chunk, then Done zeros it — or chunk.Output errors and FinishReason never arrives. The new tests return a finished response in one shot, so this path is unexercised.

go/ai/generate.go
```diff
@@ -1018,10 +1030,12 @@ func GenerateData[Out any](ctx context.Context, r api.Registry, opts ...Generate
...
-	// If there's no text content to parse (e.g., the response contains tool
-	// requests or interrupts), return nil output. The caller should check
-	// resp.Interrupts() or resp.ToolRequests() to handle these cases.
-	if resp.Text() == "" {
+	// Two responses have no conforming output to extract: one that ended
+	// abnormally, whose FinishReason is the news the caller needs, and one with
+	// no text at all, which is what a turn holding tool requests, interrupts, or
+	// media looks like. Both hand the response back unparsed rather than report a
+	// schema error naming the wrong cause. See [FinishReason.isAbnormal].
+	if resp.FinishReason.isAbnormal() || resp.Text() == "" {
```
huangjeff5, line 1038: Skipping ParseMessage so we don't mask FinishReason with `expected object, given null` is the right cut. Handing `data=nil, err=nil` back on `blocked` is the leftover — the documented caller in go/genkit/genkit.go (DefineDataPrompt example) only checks `err`, then prints `output.Capital`. A refusal becomes an empty capital. Retry.wrapModel also treats err==nil as success.

Is err=nil on blocked intended? Python #6100 / JS assertValid() throw GenerationBlockedError on the same event. Can the typed helpers error on a refusal and keep isAbnormal only for the parse skip? Interrupted + tool requests can stay nil + no error.

go/ai/generate.go
```diff
@@ -1150,10 +1167,12 @@ func GenerateDataStream[Out any](ctx context.Context, r api.Registry, opts ...Ge
...
-		// If there's no text content to parse (e.g., the response contains tool
-		// requests or interrupts), return zero-value output. The caller should check
-		// resp.Interrupts() or resp.ToolRequests() to handle these cases.
-		if resp.Text() == "" {
+		// Two responses have no conforming output to extract: one that ended
+		// abnormally, whose FinishReason is the news the caller needs, and one with
+		// no text at all, which is what a turn holding tool requests, interrupts, or
+		// media looks like. Both hand the response back unparsed rather than report a
+		// schema error naming the wrong cause. See [FinishReason.isAbnormal].
+		if resp.FinishReason.isAbnormal() || resp.Text() == "" {
```
huangjeff5, line 1175: This final guard is right. Chunks still parse first (chunk.Output above, unchanged in this diff). A model that streams leftover JSON and then finishes blocked will yield a typed Report chunk, then Done zeros it — or chunk.Output errors and FinishReason never arrives. The tests return a finished response in one shot, so this path is unexercised. Can we hold the parse until the finish reason is known?

go/ai/generate.go
apascal07: Not intended, and now changed. `GenerateData`, `GenerateDataStream`, `DataPrompt.Execute`, and `DataPrompt.ExecuteStream` return `ai.ErrGenerationBlocked` on a blocked finish, carrying the provider's finish message, with the response still returned alongside it. The status is FAILED_PRECONDITION, which is what `GenerationResponseError` defaults to in JS.

`isAbnormal` keeps `blocked` for the parse skip in `GenerateWithRequest`; the helpers test for it first. Interrupts, tool requests, and the remaining abnormal reasons keep nil output and nil error, as you suggested.

`Generate` is unchanged, so it still hands a blocked response back as a value. Matching `assertValid` there would change every existing caller and belongs in its own PR.

go/ai/generate.go
apascal07: Test added; that path was unexercised.

Holding the parse is not possible for a stream, since the finish reason only exists once the stream ends. Chunks can only ever be provisional.

The partial-JSON half of your reading reproduces: `{"title":"partial` yields a populated `Report` chunk, then the terminal value zeroes it. The other half does not. Non-JSON chunk text parses to a zero value rather than erroring, so the finish reason still reaches the caller.

What changed instead is the terminal value. A refused generation now ends in `ErrGenerationBlocked` rather than a zeroed `Output`, and the godoc states that the Done value is the authoritative one.

pavelgj (approved):

PR #6096, opened 2026-08-18
docs(py/plugins/amazon-bedrock): point the status note at the docsite page

hilariie:
Points status note at the docsite page. Expected to land after docsite lands so the url isn't broken.


cabljac (approved):
lgtm

PR #6098, opened 2026-08-18
fix(py): classify provider HTTP errors so retry skips 400s

huangjeff5:
A provider HTTP 400 is a bad request; a 503 is an outage. Retry should skip the first and wait out the second.

```python
ai.generate(model='googleai/gemini-2.5-flash', prompt='hi')
```

Gemini used to map only a handful of codes and treat everything else — including a 503 — as `INTERNAL`. OpenAI and Ollama leaked the raw SDK exception, so Retry retried a 400. After this, plugins go through shared `from_http_code` / `wrap_http_error`:

| Provider HTTP | Status | Retry |
|---|---|---|
| 400 | `INVALID_ARGUMENT` | skip |
| 429 + `Retry-After: 60` | `RESOURCE_EXHAUSTED` | wait 60s, then retry |
| 503 | `UNAVAILABLE` | retry |
| no status (`None`, `'nope'`, `0`, Ollama mid-stream `-1`) | unclassified (raw error) | retry |

The callable / reflection wire shows the provider text (`original_message`), not the SDK `repr`. Anthropic 529 stays `UNAVAILABLE` (overloaded). Anthropic 404 is `NOT_FOUND`, so Fallback can try the next model.

Veo `start()` and `check()` polling are wrapped the same way, so a 503 on the poll stays retryable.

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
@@ -42,7 +42,7 @@ from google import genai
 from google.auth import default as google_auth_default
 from google.auth.exceptions import DefaultCredentialsError
 from google.genai import types as genai_types
-from google.genai.errors import ClientError
+from google.genai.errors import APIError
 from pydantic import BaseModel, ConfigDict, Field, ValidationError, WithJsonSchema
 
 from genkit import (
@@ -64,7 +64,7 @@ from genkit.model import Candidate, FinishReason, get_basic_usage_stats
 from genkit.plugin_api import (
     ActionRunContext,
     ModelConfig,
-    StatusName,
+    wrap_http_error,
 )
 
 
@@ -1507,27 +1507,11 @@ class GeminiModel:
                 contents=cast(genai_types.ContentListUnion, request_contents),
                 config=request_cfg,
             )
-        except ClientError as e:
-            status: StatusName = 'INTERNAL'
-            if e.code == 400:
-                status = 'INVALID_ARGUMENT'
-            elif e.code == 401:
-                status = 'UNAUTHENTICATED'
-            elif e.code == 403:
-                status = 'PERMISSION_DENIED'
-            elif e.code == 404:
-                status = 'NOT_FOUND'
-            elif e.code == 429:
-                status = 'RESOURCE_EXHAUSTED'
-
-            raise GenkitError(
-                status=status,
-                message=e.message or 'Unknown error',
-                cause=e,
-            ) from e
+        except APIError as e:
+            raise wrap_http_error(e, status_code=e.code, message=e.message or str(e)) from e
         except Exception as e:
-            # Catch any other exceptions and provide a clear error message
-            # This helps debug issues like authentication errors that might not be ClientError
+            # Auth and other SDK failures are not APIError — still fail the
+            # generate so the caller is not left with a partial reply.
             import logging
 
             logger = logging.getLogger(__name__)
@@ -1607,24 +1591,8 @@ class GeminiModel:
                 contents=cast(genai_types.ContentListUnion, request_contents),
                 config=request_cfg,
             )
-        except ClientError as e:
-            status: StatusName = 'INTERNAL'
-            if e.code == 400:
-                status = 'INVALID_ARGUMENT'
-            elif e.code == 401:
-                status = 'UNAUTHENTICATED'
-            elif e.code == 403:
-                status = 'PERMISSION_DENIED'
-            elif e.code == 404:
-                status = 'NOT_FOUND'
-            elif e.code == 429:
-                status = 'RESOURCE_EXHAUSTED'
-
-            raise GenkitError(
-                status=status,
-                message=e.message or 'Unknown error',
-                cause=e,
-            ) from e
+        except APIError as e:
+            raise wrap_http_error(e, status_code=e.code, message=e.message or str(e)) from e
 
         accumulated_content: list[Part] = []
         finish_reason = FinishReason.UNKNOWN
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
@@ -89,9 +89,18 @@ def _to_finish_reason(fr: Any) -> FinishReason:  # noqa: ANN401
         'LANGUAGE',
         'MALICIOUS',
         'IMAGE_SAFETY',
+        'IMAGE_PROHIBITED_CONTENT',
+        'IMAGE_RECITATION',
     ):
         return FinishReason.BLOCKED
-    if fr_name in ('OTHER', 'MALFORMED_FUNCTION_CALL', 'MISSING_THOUGHT_SIGNATURE'):
+    if fr_name in (
+        'OTHER',
+        'MALFORMED_FUNCTION_CALL',
+        'MISSING_THOUGHT_SIGNATURE',
+        'UNEXPECTED_TOOL_CALL',
+        'NO_IMAGE',
+        'IMAGE_OTHER',
+    ):
         return FinishReason.OTHER
     return FinishReason.UNKNOWN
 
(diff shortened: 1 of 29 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1607,24 +1591,8 @@ async def _streaming_generate(
...
-                status = 'NOT_FOUND'
-            elif e.code == 429:
-                status = 'RESOURCE_EXHAUSTED'
-
-            raise GenkitError(
-                status=status,
-                message=e.message or 'Unknown error',
-                cause=e,
-            ) from e
+        except APIError as e:
```
apascal07, line 1637: This `try` does not cover the streaming request on the default path. In google-genai 1.63.0, `generate_content_stream` returns an un-started async generator whenever AFC is enabled, which is the default for a request with no tools. The `await` completes without any HTTP call, and the request is issued inside `async for response_chunk in generator` below, outside this handler.

Confirmed against the installed SDK: a fake client whose generator raises `APIError(503, ...)` on first iteration leaks the raw `APIError` rather than `GenkitError(UNAVAILABLE)`. The same applies to a 400, which then reaches Retry as a non-`GenkitError` and is retried, which is the case this PR removes for `_generate`. Wrapping the `async for` loop in the same handler would close both this and mid-stream errors on the eager branches.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
```diff
@@ -35,6 +38,19 @@
...
+
+
+def reraise_openai_error(error: Exception) -> NoReturn:
+    """Re-raise an OpenAI HTTP or request-shaping error as a classified GenkitError.
+
+    Transport failures pass through so retry can still try again.
+    """
+    if isinstance(error, APIStatusError):
+        raise wrap_http_error(error, status_code=error.status_code) from error
+    if isinstance(error, ValueError):
```
apascal07, line 55: `pydantic.ValidationError` and `json.JSONDecodeError` are both `ValueError` subclasses, so this arm catches more than request shaping. If OpenAI returns a payload that genkit's own `ModelResponse`/`Part` models reject, the `ValidationError` becomes `INVALID_ARGUMENT`, which is not in Retry's default status list. Before this PR that error reached Retry raw and was retried. Consider matching on the specific request-shaping errors, or letting `ValidationError` fall through to `INTERNAL`.

py/packages/genkit-ollama/src/genkit_ollama/models.py
```diff
@@ -249,6 +250,21 @@ async def generate(
             streaming=self.is_streaming_request(ctx=ctx),
         )
 
+        try:
+            return await self._generate_classified(request=request, ctx=ctx, client=client, content=content)
+        except ollama_api.ResponseError as e:
+            raise wrap_http_error(e, status_code=getattr(e, 'status_code', None)) from e
+        except ValueError as e:
```
apascal07, line 261: Same concern as the OpenAI helper: `pydantic.ValidationError` subclasses `ValueError`, so a response shape genkit rejects in `_build_multimodal_chat_response` now becomes `INVALID_ARGUMENT` and is no longer retried.

Also, `raise ValueError(f'Unresolved API type: ...')` at line 293 is a plugin configuration bug, not a bad caller request, and it now reports as `INVALID_ARGUMENT`.

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -429,11 +430,13 @@ async def generate(self, request: ModelRequest, ctx: ActionRunContext) -> ModelR
         """
         request.config = self.normalize_config(request.config)
```
apascal07: `normalize_config` raises `ValueError('Expected request.config to be a dict or OpenAIConfig, ...')` (line 459), and that is the one `ValueError` here that really is a bad request. It stays above the `try`, so it escapes unclassified and Retry burns `max_retries` attempts on a request that can never succeed. Moving this line inside the `try` covers it.

py/packages/genkit/src/genkit/_core/_error.py
```diff
@@ -100,6 +113,95 @@ def http_status_code(status: StatusName) -> int:
...
+        try:
+            resolved = int(code)  # type: ignore[arg-type]
+        except (TypeError, ValueError):
+            return None
+    if 100 <= resolved <= 599:
+        return resolved
+    return None
+
+
+def from_http_code(code: int) -> StatusName:
```
apascal07, line 139: 408 Request Timeout has no entry and is under 500, so it lands on `UNKNOWN`, which is not in `_DEFAULT_RETRY_STATUSES`. Gemini previously mapped every unlisted `ClientError` code to `INTERNAL` and retried it, so a transient upstream timeout that used to recover on attempt two now fails outright. Worth an explicit `408: 'DEADLINE_EXCEEDED'` (and possibly 425) in the overlay.

py/packages/genkit/src/genkit/_core/_error.py
```diff
@@ -100,6 +113,95 @@ def http_status_code(status: StatusName) -> int:
...
+        return None
+    resolved: int
+    if isinstance(code, int):
+        resolved = code
+    else:
+        try:
+            resolved = int(code)  # type: ignore[arg-type]
+        except (TypeError, ValueError):
+            return None
+    if 100 <= resolved <= 599:
```
apascal07, line 134: The 100-599 gate lets success and redirect codes through. `from_http_code(200)` returns `'OK'`, so `wrap_http_error(err, status_code=200)` builds a `GenkitError` with `http_code` 200 and `StatusCodes['OK'].value == 0`, and the FastAPI/Flask/Django callable handlers serialize a failure as status `OK`. Narrowing this to `>= 400`, or refusing to classify a success code in `from_http_code`, keeps the helper from producing an error that claims success.

apascal07 (approved):

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
huangjeff5: Took. The HTTP call is the first iteration, not the await that created the generator. The classify now wraps the async for as well — first-chunk 400/503 and a mid-stream 503 all become GenkitError.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
huangjeff5: Took the over-catch, not the blanket ValidationError → INTERNAL.

JSONDecodeError and "Unable to determine content part" are INTERNAL so Retry can try again. Request-shaping ValueError stays INVALID_ARGUMENT. So does OpenAIConfig ValidationError — after moving normalize_config into the try, a bad config dict is a request we should not retry.

py/packages/genkit-ollama/src/genkit_ollama/models.py
huangjeff5: Took. Response ValidationError is INTERNAL. Unresolved API type is INTERNAL too — that's a plugin misconfig, not a bad caller prompt. Request-shaping ValueErrors (bad URI, unknown role, unsupported tool schema) stay INVALID_ARGUMENT.

py/packages/genkit-openai/src/genkit_openai/models/model.py
huangjeff5: Took. normalize_config is inside the try, so a config type we cannot send is INVALID_ARGUMENT and Retry skips it.

py/packages/genkit/src/genkit/_core/_error.py
huangjeff5: Took 408 → DEADLINE_EXCEEDED so a transient timeout is retried. Skipped 425 — none of the providers we wrap actually return it.

py/packages/genkit/src/genkit/_core/_error.py
huangjeff5: Took the bug, not the from_http_code change.

wrap_http_error now leaves 2xx/3xx unclassified (same as a missing status) so a failure cannot serialize as OK. from_http_code(200) still returns OK — that's the reverse map, not an error classifier.

PR #6099, opened 2026-08-18
feat(py): stream debug logs to the Dev UI

huangjeff5:
Under `genkit start`, `ai.generate()` should leave a breadcrumb next to the span in the Dev UI log panel — without dumping the response blob onto the shared TTY.

```python
ai.generate(model='googleai/gemini-2.5-flash', prompt='hi')
```

Debug used to serialize a (redacted) `ModelResponse`. That was either too big for the log panel or missing entirely when `GENKIT_LOG=info`. After this, generate emits named records — `generate request resolved`, `calling model`, `model responded`, `middleware hook started` / `finished`, `executing tool requests` — and a background OTLP sink POSTs them to `/api/otlp`.

`ExportTee` splits the two sinks. The console still honours `GENKIT_LOG`. The Dev UI sink is debug and above, so a quiet terminal can still fill the trace viewer's log panel when `GENKIT_TELEMETRY_SERVER` is set (`GENKIT_ENV=dev`).

Collector I/O stays off the generate thread (300s batch POST; 2s shutdown flush) so a wedged Dev UI does not stall Ctrl-C.

### DIFF
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -91,6 +91,46 @@ logger = get_logger(__name__)
 HookParamsT = TypeVar('HookParamsT')
 HookResultT = TypeVar('HookResultT')
 
+# A termination known to be abnormal carries no conforming output, so a schema
+# error here would mask the finish reason the caller needs to handle it.
+ABNORMAL_FINISH_REASONS = frozenset({
+    FinishReason.BLOCKED,
+    FinishReason.ABORTED,
+    FinishReason.INTERRUPTED,
+    FinishReason.OTHER,
+})
+
+
+def log_output_parse(
+    *,
+    model: str | None,
+    finish_reason: FinishReason | None,
+    finish_message: str | None,
+    formatter: Formatter[Any, Any] | None,
+    message: Message | None,
+) -> None:
+    """Warn on an abnormal finish; debug when the formatter cannot parse."""
+    if formatter is None:
+        return
+    if finish_reason in ABNORMAL_FINISH_REASONS:
+        logger.warning(
+            'model finished abnormally, skipping output parsing',
+            model=model,
+            finishReason=finish_reason,
+            finishMessage=finish_message,
+        )
+        return
+    if message is None or not is_debug_enabled(logger):
+        return
+    try:
+        formatter.parse_message(message)
+    except Exception as e:
+        logger.debug(
+            'model output does not match the expected schema',
+            model=model,
+            error=e,
+        )
+
 
 def middleware_name(mw: MiddlewareDef) -> str:
     """Class name is what shows up on hook log records."""
@@ -107,10 +147,11 @@ def hook_finished(
     extra: dict[str, object],
 ) -> dict[str, object]:
     """Attributes for the ``middleware hook finished`` record."""
+    ms = max(0, round((time.monotonic() - start) * 1000))
     out: dict[str, object] = {
         'middleware': name,
         'hook': hook,
-        'duration_ms': int((time.monotonic() - start) * 1000),
+        'duration': f'{ms}ms',
         **extra,
     }
     if not next_called:
@@ -600,9 +641,8 @@ async def generate_with_request(
             'max_turns': raw_request.max_turns,
             'streaming': on_chunk is not None,
         }
-        if raw_request.output is not None:
-            resolved['format'] = raw_request.output.format
-            resolved['constrained'] = raw_request.output.constrained
+        resolved['format'] = raw_request.output.format if raw_request.output else None
+        resolved['constrained'] = raw_request.output.constrained if raw_request.output else None
         if middleware:
             resolved['middleware'] = [middleware_name(m) for m in middleware]
         logger.debug('generate request resolved', **resolved)
@@ -660,7 +700,10 @@ class ChunkAccumulator:
         """Send one framework-wrapped chunk through the current stream chain."""
         if ctx.on_chunk is None:
             return
-        ctx.on_chunk(self.make(role=role, chunk=chunk))
+        try:
+            ctx.on_chunk(self.make(role=role, chunk=chunk))
+        except Exception as e:
+            logger.debug('tool stream callback failed, dropping chunk', error=e)
 
     @contextlib.contextmanager
     def intercept_model_stream(
@@ -955,6 +998,14 @@ async def _generate_action_turn(
                 responded['input_tokens'] = response.usage.input_tokens
                 responded['output_tokens'] = response.usage.output_tokens
             logger.debug('model responded', **responded)
+
+        log_output_parse(
+            model=turn_options.model,
+            finish_reason=response.finish_reason,
+            finish_message=response.finish_message,
+            formatter=formatter,
+            message=generated_msg,
+        )
 
         response.assert_valid()
 
@@ -1428,6 +1479,7 @@ async def resolve_tool_requests(
             intr = _interrupt_from_tool_exc(e)
             if intr is None:
                 raise
+            logger.debug('tool triggered an interrupt', tool=trp.tool_request.name)
             return (None, _interrupt_request_part(trp, intr))
 
     outs = await asyncio.gather(*[_resolve_one_tool(tool, trp) for _, tool, trp in work])
@@ -1760,6 +1812,13 @@ async def _run_restart_through_middleware(
     except Exception as e:
         intr = _interrupt_from_tool_exc(e)
         if intr is not None:
+            # run_tool_after_restart already logged when the tool body interrupted.
+            # wrap_tool can raise Interrupt itself; that's the only leftover case.
+            if not isinstance(e, GenkitError):
+                logger.debug(
+                    'restarted tool triggered an interrupt',
+                    tool=restart_trp.tool_request.name,
+                )
             # Re-interrupting during restart is a hard error — same as the legacy
             # run_tool_after_restart path, which raises FAILED_PRECONDITION when
             # the inner tool throws an Interrupt during restart. Surface the
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -19,11 +19,11 @@
 import asyncio
 import contextlib
 import copy
-import re
 import secrets
+import time
 from collections.abc import Awaitable, Callable, Generator, Sequence
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, TypeVar, cast
 
 from pydantic import BaseModel
 from typing_extensions import Never
@@ -83,6 +83,88 @@ from genkit._core._typing import (
 DEFAULT_MAX_TURNS = 5
 
 logger = get_logger(__name__)
+
+HookParamsT = TypeVar('HookParamsT')
+HookResultT = TypeVar('HookResultT')
+
+
+def middleware_name(mw: MiddlewareDef) -> str:
+    """Class name is what shows up on hook log records."""
+    return type(mw).__name__
+
+
+def hook_finished(
+    *,
+    name: str,
+    hook: str,
+    start: float,
+    next_called: bool,
+    error: str | None,
+    extra: dict[str, object],
+) -> dict[str, object]:
+    """Attributes for the ``middleware hook finished`` record."""
+    out: dict[str, object] = {
+        'middleware': name,
+        'hook': hook,
+        'duration_ms': int((time.monotonic() - start) * 1000),
+        **extra,
+    }
+    if not next_called:
+        out['short_circuited'] = True
+    if error is not None:
+        out['error'] = error
+    return out
+
+
+async def run_logged_hook(
+    *,
+    mw: MiddlewareDef,
+    hook: str,
+    params: HookParamsT,
+    ctx: GenerateMiddlewareContext,
+    wrap: Callable[
+        [
+            HookParamsT,
+            GenerateMiddlewareContext,
+            Callable[[HookParamsT, GenerateMiddlewareContext], Awaitable[HookResultT]],
+        ],
+        Awaitable[HookResultT],
+    ],
+    inner: Callable[[HookParamsT, GenerateMiddlewareContext], Awaitable[HookResultT]],
+    extra: dict[str, object] | None = None,
+) -> HookResultT:
+    """Run one middleware hook, with started/finished records when debug is on."""
+    if not is_debug_enabled(logger):
+        return await wrap(params, ctx, inner)
+    attrs = extra or {}
+    name = middleware_name(mw)
+    logger.debug('middleware hook started', middleware=name, hook=hook, **attrs)
+    start = time.monotonic()
+    next_called = False
+
+    async def tracked(tp: HookParamsT, tc: GenerateMiddlewareContext) -> HookResultT:
+        nonlocal next_called
+        next_called = True
+        return await inner(tp, tc)
+
+    err: str | None = None
+    try:
+        return await wrap(params, ctx, tracked)
+    except Exception as e:
+        err = str(e)
+        raise
+    finally:
+        logger.debug(
+            'middleware hook finished',
+            **hook_finished(
+                name=name,
+                hook=hook,
+                start=start,
+                next_called=next_called,
+                error=err,
+                extra=attrs,
+            ),
+        )
 
 
 class ScopedGenkitView:
@@ -220,7 +302,15 @@ async def dispatch_tool(
             _m: MiddlewareDef = _mw,
             _i: Callable[[ToolHookParams, GenerateMiddlewareContext], Awaitable[MultipartToolResponse]] = _inner,
         ) -> MultipartToolResponse:
-            return await _m.wrap_tool(p, c, _i)
+            return await run_logged_hook(
+                mw=_m,
+                hook='tool',
+                params=p,
+                ctx=c,
+                wrap=_m.wrap_tool,
+                inner=_i,
+                extra={'tool': p.tool.name},
+            )
 
         runner = run_next
     return await runner(params, ctx)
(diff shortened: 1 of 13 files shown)


py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,353 @@
...
+            # The caller already returned; this wait lives on the worker.
+            with httpx.Client(timeout=EXPORT_TIMEOUT_S) as client:
+                response = client.post(url, content=body, headers=LOG_HEADERS)
+            if response.status_code != 200:
+                self.note_transport_failure(error=RuntimeError(f'HTTP {response.status_code}'))
+                return
+            self.last_result_ok = True
+            self.warned_unreachable = False
+        except (httpx.RequestError, OSError) as error:
+            self.note_transport_failure(error=error)
```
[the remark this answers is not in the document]
huangjeff5: Good call on reusing the client across worker batches to keep connection pools alive—updated in 803193b0a.

huangjeff5:
@gemini review

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -86,6 +86,88 @@
...
+        nonlocal next_called
+        next_called = True
+        return await inner(tp, tc)
+
+    err: str | None = None
+    try:
+        return await wrap(params, ctx, tracked)
+    except Exception as e:
+        err = str(e)
+        raise
```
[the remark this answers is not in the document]
huangjeff5: Good catch on capturing task cancellation—updated to BaseException with class name fallback in 6f8e9a1c0.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,353 @@
...
+        record['droppedAttributesCount'] = dropped
+    span = trace_api.get_current_span()
+    ctx = span.get_span_context()
+    if ctx is not None and ctx.is_valid:
+        record['traceId'] = format(ctx.trace_id, '032x')
+        record['spanId'] = format(ctx.span_id, '016x')
+    # Fail on this thread if the record cannot be JSON — a broken payload
+    # must not look like a quiet collector miss after generate() returns.
+    json.dumps(record)
+    return record
```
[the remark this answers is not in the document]
huangjeff5: Removed the redundant synchronous json.dumps check in 6f8e9a1c0.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,353 @@
...
+    if isinstance(value, int):
+        return {'intValue': value}
+    if isinstance(value, str):
+        return {'stringValue': _clip_attr(value)}
+    if isinstance(value, float):
+        return {'stringValue': _clip_attr(format(value, 'g'))}
+    try:
+        return {'stringValue': _clip_attr(json.dumps(value))}
+    except (TypeError, ValueError):
+        return {'stringValue': _clip_attr(str(value))}
```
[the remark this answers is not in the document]
huangjeff5: Attributes are already bounded by MAX_ATTRIBUTES = 32 and MAX_ATTR_CHARS = 500, so keeping this simple without extra recursive tree traversal.

huangjeff5:
@gemini review

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,350 @@
...
+    'token',
+    'access_token',
+})
+
+SEVERITY: dict[int, tuple[int, str]] = {
+    logging.DEBUG: (5, 'DEBUG'),
+    logging.INFO: (9, 'INFO'),
+    logging.WARNING: (13, 'WARN'),
+    logging.ERROR: (17, 'ERROR'),
+    logging.CRITICAL: (17, 'ERROR'),
```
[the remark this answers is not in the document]
huangjeff5: Updated logging.CRITICAL mapping to (21, 'FATAL') to align with the OTLP log data model in 2831b0c26.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,350 @@
...
+
+def _otlp_value(*, value: object) -> dict[str, object]:
+    if isinstance(value, bool):
+        return {'boolValue': value}
+    if isinstance(value, int):
+        return {'intValue': value}
+    if isinstance(value, str):
+        return {'stringValue': _clip_attr(value)}
+    if isinstance(value, float):
+        return {'stringValue': _clip_attr(format(value, 'g'))}
```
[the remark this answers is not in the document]
huangjeff5: Switched float attributes to native doubleValue in 2831b0c26.

huangjeff5:
@gemini review

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,350 @@
...
+    except Full:
+        pass
+    try:
+        queue.get_nowait()
+    except Empty:
+        pass
+    try:
+        queue.put_nowait(None)
+    except Full:
+        pass
```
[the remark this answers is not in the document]
huangjeff5: Great catch on the task counter leak—added queue.task_done() on the discarded slot and added unit test coverage for queue.join() in 8d65c9ba7.

huangjeff5:
@gemini review

py/packages/genkit/src/genkit/_core/_logger.py
```diff
@@ -114,9 +118,92 @@ def configure_logging(*, shared_tty: bool | None = None) -> None:
...
+            return emit_log
+        if name not in _EMIT_LEVELS or not callable(target):
+            return target
+        emit_level = _EMIT_LEVELS[name]
+
+        def emit_then(event: str, *args: object, **kw: object) -> object:
+            self._emit(emit_level, event, kw)
+            return target(event, *args, **kw)
+
+        return emit_then
```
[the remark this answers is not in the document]
huangjeff5: Good call on caching the dynamically bound methods on the instance—updated in f92b3f557.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,351 @@
...
+        try:
+            # The caller already returned; this wait lives on the worker.
+            response = client.post(url, content=body, headers=LOG_HEADERS)
+            if response.status_code != 200:
+                self.note_transport_failure(error=RuntimeError(f'HTTP {response.status_code}'))
+                return
+            self.last_result_ok = True
+            self.warned_unreachable = False
+        except (httpx.RequestError, OSError) as error:
+            self.note_transport_failure(error=error)
```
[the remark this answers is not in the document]
huangjeff5: Wrapped serialization inside try and caught general Exception to protect the worker thread in f92b3f557.

huangjeff5:
@gemini review

py/packages/genkit/src/genkit/_core/_logger.py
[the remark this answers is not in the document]
huangjeff5: Update on setattr caching: because Structlog reconfigures and wraps logger levels dynamically at runtime, caching the closure via setattr permanently captures the initial level handler (e.g. dropping debug events even if log level is later changed). Reverted in 9722fd9f3 to preserve dynamic level filtering.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,351 @@
...
+    if isinstance(value, int):
+        return {'intValue': value}
+    if isinstance(value, float):
+        return {'doubleValue': value}
+    if isinstance(value, str):
+        return {'stringValue': _clip_attr(value)}
+    try:
+        return {'stringValue': _clip_attr(json.dumps(value))}
+    except (TypeError, ValueError):
+        return {'stringValue': _clip_attr(str(value))}
```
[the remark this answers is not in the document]
huangjeff5: Added explicit handling for bytes, bytearray, and memoryview to summarize as <N bytes> in f4ed779f0.

huangjeff5:
@gemini review

py/packages/genkit/src/genkit/_core/_logger.py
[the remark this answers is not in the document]
huangjeff5: Genkit's logging is strictly structured around Structlog's key-value event model (e.g. logger.info('event', key=val)). Where positional args are used for printf-style interpolation, Structlog's processor pipeline handles formatting the event string directly. We prefer not to introduce a bespoke positional_args attribute into the OTLP payload schema.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,353 @@
...
+    return encoded, dropped
+
+
+def _otlp_value(*, value: object) -> dict[str, object]:
+    if isinstance(value, bool):
+        return {'boolValue': value}
+    if isinstance(value, int):
+        return {'intValue': value}
+    if isinstance(value, float):
+        return {'doubleValue': value}
```
apascal07, line 208: `fromOtlpValue` in `genkit-tools/telemetry-server/src/utils/otlp.ts` decodes only `stringValue`, `intValue`, `boolValue` and `arrayValue`. Anything else returns `undefined`, and `toLogRecordData` then skips the attribute entirely. So a float attribute (`logger.debug('scored', confidence=0.87)`) is accepted by the collector and silently missing from the log panel, with no warning on either side.

`test_build_log_record_wire_shape` asserts the `doubleValue` shape, so the test pins the encoding the Dev UI cannot read.

py/packages/genkit/src/genkit/_core/_logger.py
```diff
@@ -114,9 +118,92 @@ def configure_logging(*, shared_tty: bool | None = None) -> None:
...
+            def emit_log(level: int, event: str, *args: object, **kw: object) -> object:
+                self._emit(level, event, kw)
+                return target(level, event, *args, **kw)
+
+            return emit_log
+        if name not in _EMIT_LEVELS or not callable(target):
+            return target
+        emit_level = _EMIT_LEVELS[name]
+
+        def emit_then(event: str, *args: object, **kw: object) -> object:
```
apascal07, line 197: `emit_then` accepts `*args` but `_emit(emit_level, event, kw)` drops them, so printf-style records reach the sink uninterpolated. Same in the `name == 'log'` branch above.

Existing call sites use that form: `_aio.py:1092` `logger.error('Startup failed: %s: %s', type(e).__name__, e)` and `_reflection_v2.py:498` `logger.debug('Action failed: %s: %s', type(exc).__name__, exc, exc_info=True)`. On a startup failure the console shows the real message and the Dev UI panel shows the literal `Startup failed: %s: %s`, which is the one place someone would go looking for it.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -86,6 +86,129 @@
...
+    model: str | None,
+    finish_reason: FinishReason | None,
+    finish_message: str | None,
+    formatter: Formatter[Any, Any] | None,
+    message: Message | None,
+) -> None:
+    """Warn on an abnormal finish; debug when the formatter cannot parse."""
+    if formatter is None:
+        return
+    if finish_reason in ABNORMAL_FINISH_REASONS:
```
apascal07, line 116: The comment says a schema error here "would mask the finish reason", and the record says "skipping output parsing", but nothing is actually skipped: `response.assert_valid_schema()` still runs unconditionally at line 961 when there are no tool requests.

A safety-blocked response on a request with `output=GenerateActionOutputConfig(format='json')` logs this warning and then raises the schema validation error anyway, so the caller sees a parse failure rather than `FinishReason.BLOCKED`. Either gate the assertion on the same set, or reword both the comment and the message to match what the code does.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -86,6 +86,129 @@
...
+HookParamsT = TypeVar('HookParamsT')
+HookResultT = TypeVar('HookResultT')
+
+# A termination known to be abnormal carries no conforming output, so a schema
+# error here would mask the finish reason the caller needs to handle it.
+ABNORMAL_FINISH_REASONS = frozenset({
+    FinishReason.BLOCKED,
+    FinishReason.ABORTED,
+    FinishReason.INTERRUPTED,
+    FinishReason.OTHER,
```
apascal07: `OTHER` is the providers' catch-all for unmapped stop reasons, not a signal that anything went wrong. `genkit_anthropic/models.py:304,308` map `compaction` and `pause_turn` (normal continuations) to `OTHER`, and `genkit_amazon_bedrock/converters.py:711` defaults every unrecognised or missing stop reason to `OTHER`.

Since `formatter` is non-None for any request carrying an output format or schema, a structured-output Anthropic call that ends in `pause_turn` now logs `model finished abnormally` at WARNING on every turn, at the default `GENKIT_LOG=info`. BLOCKED, ABORTED and INTERRUPTED are unambiguous; OTHER is not.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,353 @@
...
+def _severity(*, level: int) -> tuple[int, str]:
+    for threshold in (logging.CRITICAL, logging.ERROR, logging.WARNING, logging.INFO, logging.DEBUG):
+        if level >= threshold:
+            return SEVERITY[threshold]
+    return SEVERITY[logging.DEBUG]
+
+
+def _should_redact(*, key: str) -> bool:
+    lowered = key.lower()
+    return lowered in REDACT_KEYS or lowered.endswith('_key') or lowered.endswith('_token')
```
apascal07: This matches lowercase and snake_case only. `apiKey`, `authToken`, `accessToken` and `apikey` all pass through: none end in `_key`/`_token`, and `'apikey'` is not in `REDACT_KEYS`.

camelCase attribute names are used in this codebase, including in this PR (`finishReason=`, `finishMessage=` in `log_output_parse`). A plugin logging `logger.debug('calling provider', apiKey=key)` POSTs the key in cleartext to `/api/otlp`, where the telemetry server persists it. Normalizing the key (strip `_`, lowercase) and substring-matching key/token/secret/password would close it.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -610,7 +698,10 @@ def stream_chunk(
         """Send one framework-wrapped chunk through the current stream chain."""
         if ctx.on_chunk is None:
             return
-        ctx.on_chunk(self.make(role=role, chunk=chunk))
+        try:
+            ctx.on_chunk(self.make(role=role, chunk=chunk))
+        except Exception as e:
+            logger.debug('tool stream callback failed, dropping chunk', error=e)
```
apascal07: This swallows every callback exception, but only for tool-response chunks. Model tokens go through `intercept_model_stream.handler`, which calls `downstream(...)` unguarded.

So the same user callback bug (say `chunk.tex`) fails the generate call loudly on a model chunk and silently drops the chunk on a tool chunk, leaving only a DEBUG line. Worth guarding both paths or neither.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
```diff
@@ -0,0 +1,353 @@
...
+            self.dropped += 1
+            if not self.warned_overflow:
+                self.warned_overflow = True
+                _diag(
+                    level=logging.WARNING,
+                    event='log export queue is full; dropping records from the Dev UI log view',
+                )
+
+    def run_worker(self) -> None:
+        with httpx.Client(timeout=EXPORT_TIMEOUT_S) as client:
```
apascal07: Only `post_batch`'s body is guarded. `httpx.Client(timeout=...)` construction and the context-manager exit are not, so if either raises the worker dies with `self.stopped` still `False`.

`log_export_is_enabled()` then returns `True` for the rest of the process, which makes `is_debug_enabled()` return `True` everywhere: every debug branch in `_generate.py` keeps building records and `formatter.parse_message` keeps running for output that is never exported. `TraceServerExporter.run_worker` guards against exactly this ("The worker has to survive a surprise exception").

apascal07 (approved):

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
huangjeff5: Took. fromOtlpValue now reads doubleValue, so a float attribute shows up in the log panel instead of being dropped.

py/packages/genkit/src/genkit/_core/_logger.py
huangjeff5: Took. The sink interpolates printf-style args before export, so a startup failure in the panel is the same line as the console, not the literal %s.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Took. assert_valid_schema is gated on the same abnormal set as the skip log, so a blocked JSON request keeps finish_reason=blocked instead of raising a parse error.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Took. OTHER is no longer in that set. A structured-output turn that ends in pause_turn / compaction does not log "model finished abnormally" at info.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
huangjeff5: Took. Keys are compacted (strip _, lowercase) and matched against key/token/secret/password, so apiKey / authToken / apikey redact before the POST to /api/otlp.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Took — neither path swallows. A sinking on_chunk fails generate on a tool chunk the same way it does on a model chunk.

py/packages/genkit/src/genkit/_core/_trace/_log_exporter.py
huangjeff5: Took. The whole worker is guarded. If Client() or context-exit raises, stopped flips so log_export_is_enabled() goes false and generate stops building records that will never export.

PR #6100, opened 2026-08-18
fix(py): return generate() after the model has already replied

huangjeff5:
generate() used to throw when the model refused or wrote something that was not the schema you asked for. That discarded the reply. After this, you get the ModelResponse back.
```python
    response = await ai.generate(prompt='give me a recipe', output_schema=Recipe)

    if response.output is not None:
        return response.output  # a Recipe

    # The model already replied. finish_reason / error say why it is not a Recipe.
    response.text
    response.finish_reason   # failed | blocked | aborted | ...
    response.finish_message   # only on failed or aborted
```

- Matching JSON is still stop, and .output is the Recipe.
- A string that is not a Recipe (not json, wrong shape) comes back as finish_reason=failed with error.status=INVALID_ARGUMENT. The string stays on .text.
- A safety refusal stays finish_reason=blocked and has no .error. The refusal text stays on .text.
- Hitting max tool turns is failed / ABORTED. A tool name that is not registered is failed / NOT_FOUND.
- A model abort stays aborted and sets error.status=ABORTED.

generate() still throws when there was no turn: empty messages, an unknown output_format or model, or an output_schema that is not a JSON Schema.


py/packages/genkit/src/genkit/_ai/_formats/_array.py
```diff
@@ -92,10 +92,19 @@ def handle(self, schema: dict[str, object] | None) -> Formatter[Any, Any]:
...
             """Parses a complete message into a list of items."""
+            start = msg.text.find('[')
+            if start == -1:
+                return None
             result = extract_json_array_from_text(msg.text, 0)
-            return result.items
+            if result.items:
+                return result.items
+            # a closed [] is a real empty list; a dangling '[' is leftover
+            close = msg.text.find(']', start)
```
apascal07: `extract_json_array_from_text` only collects `{...}` objects, so an array of scalars yields `items == []`. This fallback then reads the text between the first `[` and the first `]`, finds it non-empty, and returns `None`.

With `output_format='array'` and `output_schema=list[str]`, a model reply of `["a", "b"]` now comes back `finish_reason=failed` / `INVALID_ARGUMENT` with `.output` None, because `assert_valid_schema` reads `None` as a parse miss. Before this change the same reply parsed to `[]`.

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
```diff
@@ -756,18 +756,17 @@ async def finalize_detach(
...
         turn_err = self.session_runner.last_turn_error
+        result = result_holder[0] if result_holder else None
+        finish_reason = (
+            result.finish_reason if result and result.finish_reason else self.session_runner.last_turn_finish_reason
+        )
+        if fn_err is not None:
+            finish_reason = AgentFinishReason.FAILED
         is_failed = (
             fn_err is not None
             or turn_err is not None
```
apascal07: `last_turn_error` is now set from `turn_result.error`, so it is non-None for finish reasons that are not failures. `assert_valid()` always stamps `.error` on an aborted response, so an aborted turn reaches here with `turn_err` set and `finish_reason=aborted`.

The two paths then disagree on the same turn: `emit_turn_end` computes `is_failed` from `{FAILED, BLOCKED}` only, so attached writes `status=completed` and drops the error, while detached writes `status=failed` and keeps it. The attached snapshot loses the error the response now guarantees.

py/packages/genkit/src/genkit/_ai/_agents/_client.py
```diff
@@ -1357,9 +1355,21 @@ def _update_from_output(self, *, raw: AgentOutput, message_count_before: int) ->
...
-            self._resume_snapshot_id = raw.snapshot_id
-
-        if raw.finish_reason in (AgentFinishReason.FAILED, AgentFinishReason.ABORTED):
+            # A failed/blocked row is inspectable, not a place the next
+            # send() can resume from. Keep the last completed id.
+            if raw.finish_reason not in {
+                AgentFinishReason.FAILED,
+                AgentFinishReason.BLOCKED,
+                AgentFinishReason.ABORTED,
+                AgentFinishReason.DETACHED,
```
apascal07: `detach()` still appends the user message optimistically (line 1199), but `DETACHED` is in this set and absent from the rollback tuple below. The local transcript keeps the message while `_resume_snapshot_id` stays on the previous completed snapshot.

The next `send()` builds `AgentInit(snapshot_id=_resume_snapshot_id)`, so the server resumes from before the detached prompt and the model never sees it, while `self.messages` still shows it and the new turn appends on top. Either roll the optimistic append back for `DETACHED` or keep advancing the resume id.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -509,14 +544,97 @@ def model_post_init(self, __context: object) -> None:
...
+        # is the output.
+        if parsed is None and (self.text or '').strip() != 'null':
+            preview = (self.text or '')[:200]
+            self._mark_failed(
+                message=f'Model output was not valid JSON for the requested schema: {preview}',
+                status='INVALID_ARGUMENT',
+            )
+            return
+
+        if self._string_is_output and isinstance(parsed, str):
```
apascal07: `_string_is_output` returns here before `parse_schema` runs, and `output` does the same at line 675, so an enum reply is never checked against its own schema.

`output_format='enum'` with a three-value schema and a model answering `MAYBE` gives `finish_reason=stop` and `.output == 'MAYBE'`. `py/samples/output-formats/src/main.py` returns that straight out of a function typed `-> Sentiment`. Enum is the shape most likely to come back off-list, so it is the one case the new contract does not cover.

py/samples/fastapi-bugbot/src/main.py
```diff
@@ -79,20 +79,23 @@ class DiffInput(BaseModel):
 async def analyze_security(input: CodeInput) -> Analysis:
     """Analyze code for security vulnerabilities."""
     response = await security_prompt(input=input)
+    assert response.output is not None
```
apascal07: `assert` is stripped under `python -O`, so `return response.output` hands back None from a function typed `-> Analysis` and FastAPI serializes null. Without `-O` the caller gets a bare `AssertionError` rather than `finish_reason` / `error`, which is what this change exists to expose. Same pattern at 90, 98, 117 and in `py/samples/output-formats/src/main.py`.

apascal07 (approved):

py/packages/genkit/src/genkit/_ai/_formats/_array.py
huangjeff5: Took. A finished array reply is parsed as one JSON array, so ["a", "b"] is the output instead of a leftover.

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
huangjeff5: Took. Detached finalize uses the same failed rule as emit_turn_end (FAILED / an escaped fn error). An aborted turn is not failed. The error stays on the snapshot when one is present.

py/packages/genkit/src/genkit/_ai/_agents/_client.py
huangjeff5: Took — keep advancing the resume id. DETACHED points send() at the pending snapshot and keeps the prompt. Failed / blocked / aborted stay inspectable and do not become the next resume id.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Took. An enum string still goes through parse_schema. MAYBE against a three-value schema is finish_reason=error / .output None, with MAYBE left on .text.

py/samples/fastapi-bugbot/src/main.py
huangjeff5: Took. The samples skip a leftover with `if result.output` instead of assert, so python -O cannot hand None out of a flow typed → Analysis.

py/packages/genkit/src/genkit/_ai/_agents/_client.py
huangjeff5: BLOCKED is a completed refusal, not a dead end. The prompt stays and send() resumes from that snapshot — same as STOP / DETACHED. Only FAILED / ABORTED stay inspectable and keep the last completed resume id.

PR #6105, opened 2026-08-18
feat(py): Anthropic.claude_model and OpenAI.gpt_model typed ref constructors

huangjeff5:
## Summary
- `Anthropic.claude_model` namespaces the id and binds `AnthropicConfig`. Unknown ids are allowed; every Anthropic generate model takes that schema.
- `OpenAI.gpt_model` is chat-only. Image, TTS, STT, and embedding ids validate different request shapes, so binding `OpenAIConfig` would let chat-only keys like `frequency_penalty` ride into the wrong endpoint. Those ids raise `INVALID_ARGUMENT` and keep working as plain strings (`openai_model(...)`).
- Foreign prefixes (`vertexai/…`, `azure/…`) stay in the name. This constructor does not remap a Vertex or Azure paste onto Anthropic/OpenAI — that would hit the wrong API.
- Chat actions opt into `ModelRequest[OpenAIConfig]` / `ModelRequest[AnthropicConfig]`, so the **string** generate path validates plugin config today (`temperature: 3` is `INVALID_ARGUMENT`). `OpenAIConfig` accepts camelCase aliases so Dev UI / reflection knobs (`frequencyPenalty`, `maxOutputTokens`) land. `maxOutputTokens` is accepted; it is not mapped to `max_tokens`.
- Unlisted chat ids that ask for JSON (no schema) get `json_object`. The provider rejects the request if that model cannot do JSON mode.
- `KnownClaude` / `KnownGpt` are catalog Literals so quote autocomplete and the supported-model dict are the same set.
- `openai_model()` stays the string name helper.
- refs can be minted but all the end-user surfaces (generate, agent, prompt, Genkit() constructor) still want a string, until #6074 lands.

### DIFF
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
@@ -40,21 +40,13 @@ logger = structlog.get_logger(__name__)
 
 ANTHROPIC_PLUGIN_NAME = 'anthropic'
 
-# Prefixes people paste from action keys, Dev UI traces, or another plugin's
-# samples. Stripping them first means this constructor decides the namespace.
-_STRIP_PREFIXES = (
-    'background-model/',
-    'model/',
-    'models/',
-    'embedders/',
-    'googleai/',
-    'vertexai/',
-    f'{ANTHROPIC_PLUGIN_NAME}/',
-)
+# Only this plugin's namespace. A vertexai/ paste is a different name —
+# remapping it here would send the request to the Anthropic API.
+_STRIP_PREFIXES = (f'{ANTHROPIC_PLUGIN_NAME}/',)
 
 
 def _strip_ref_prefixes(name: str) -> str:
-    """Reduce a pasted name to the bare model id."""
+    """Peel this plugin's prefix so it is not stamped twice."""
     local = name
     changed = True
     while changed:
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
@@ -21,8 +21,8 @@ from typing import Any, Literal, cast
 import structlog
 from anthropic import AsyncAnthropic
 
-from genkit import ModelRequest, ModelResponse
-from genkit.model import model_action_metadata
+from genkit import GenkitError, ModelRequest, ModelResponse
+from genkit.model import ModelRef, model_action_metadata, model_ref
 from genkit.plugin_api import (
     Action,
     ActionKind,
@@ -33,7 +33,7 @@ from genkit.plugin_api import (
     to_json_schema,
 )
 from genkit_anthropic.config import AnthropicConfig
-from genkit_anthropic.model_info import SUPPORTED_ANTHROPIC_MODELS, get_model_info
+from genkit_anthropic.model_info import SUPPORTED_ANTHROPIC_MODELS, KnownClaude, get_model_info
 from genkit_anthropic.models import AnthropicModel
 
 logger = structlog.get_logger(__name__)
@@ -60,6 +60,21 @@ class Anthropic(Plugin):
     """
 
     name = ANTHROPIC_PLUGIN_NAME
+
+    @classmethod
+    def claude_model(
+        cls, name: KnownClaude | str, *, config: AnthropicConfig | None = None
+    ) -> ModelRef[AnthropicConfig]:
+        """Typed ref for a Claude model, e.g. ``Anthropic.claude_model('claude-sonnet-4-5')``.
+
+        Unknown ids are allowed so new Claude releases work before this
+        plugin learns their names; every Anthropic generate model takes
+        AnthropicConfig.
+        """
+        local = str(name).removeprefix(f'{ANTHROPIC_PLUGIN_NAME}/')
+        if not local:
+            raise GenkitError(status='INVALID_ARGUMENT', message='Anthropic.claude_model: model name is required.')
+        return model_ref(local, config_schema=AnthropicConfig, namespace=ANTHROPIC_PLUGIN_NAME, config=config)
 
     def __init__(
         self,
@@ -129,7 +144,7 @@ class Anthropic(Plugin):
 
         model_info = get_model_info(clean_name)
 
-        async def _generate(request: ModelRequest, ctx: ActionRunContext) -> ModelResponse:
+        async def _generate(request: ModelRequest[AnthropicConfig], ctx: ActionRunContext) -> ModelResponse:
             model = AnthropicModel(
                 model_name=clean_name,
                 client=self._runtime_client(),
--- a/py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
+++ b/py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
@@ -40,6 +40,31 @@ logger = structlog.get_logger(__name__)
 
 ANTHROPIC_PLUGIN_NAME = 'anthropic'
 
+# Prefixes people paste from action keys, Dev UI traces, or another plugin's
+# samples. Stripping them first means this constructor decides the namespace.
+_STRIP_PREFIXES = (
+    'background-model/',
+    'model/',
+    'models/',
+    'embedders/',
+    'googleai/',
+    'vertexai/',
+    f'{ANTHROPIC_PLUGIN_NAME}/',
+)
+
+
+def _strip_ref_prefixes(name: str) -> str:
+    """Reduce a pasted name to the bare model id."""
+    local = name
+    changed = True
+    while changed:
+        changed = False
+        for prefix in _STRIP_PREFIXES:
+            if local.startswith(prefix):
+                local = local[len(prefix) :]
+                changed = True
+    return local
+
 
 def anthropic_name(name: str) -> str:
     """Get Anthropic model name.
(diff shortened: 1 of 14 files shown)


py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -48,6 +48,36 @@
 
 logger = structlog.get_logger(__name__)
 
+# Genkit common fields that are not chat.completions.create() kwargs.
+# version becomes the wire model id; stop_sequences becomes stop.
+_GENKIT_ONLY = frozenset({'api_key', 'top_k', 'version', 'max_output_tokens', 'stop_sequences'})
```
apascal07: Peeling `max_output_tokens` without mapping it to `max_tokens` silently disables Genkit's portable output cap on this plugin.

```python
action = OpenAI(api_key='k')._create_model_action('openai/gpt-4o')
validated = action._validate_input({
    'messages': [{'role': 'user', 'content': [{'text': 'hi'}]}],
    'config': {'maxOutputTokens': 256, 'temperature': 0.3},
})
validated.config.max_output_tokens   # 256.0  -- bound fine
await OpenAIModel('gpt-4o', client)._get_openai_request_config(validated)
# body keys: ['messages', 'model', 'temperature']   -- no max_tokens
```

`normalize_config` already does exactly this mapping (`max_tokens=int(config.max_output_tokens)`) for plain `ModelConfig` input, so only the new typed `ModelRequest[OpenAIConfig]` path loses it — and that is now the path every registered chat action takes.

`maxOutputTokens` is the cross-plugin knob users reach for; someone who set it to bound cost gets uncapped responses with no warning. Suggest mapping it to `max_tokens` when `max_tokens`/`max_completion_tokens` are unset (same shape as the existing `stop_sequences` -> `stop` fallback just below), and keeping the docstring note for the explicit-`max_tokens` case.

py/packages/genkit-openai/src/genkit_openai/models/handler.py
```diff
@@ -49,7 +50,7 @@ def __init__(self, model: OpenAIModel, source: PluginSource = PluginSource.OPENA
         self._source = source
 
     @staticmethod
-    def _get_supported_models(source: PluginSource) -> dict[str, ModelInfo]:
+    def _get_supported_models(source: PluginSource) -> Mapping[str, ModelInfo]:
```
apascal07: Adjacent to this signature change: `_validate_version` (line 111) still has the unlisted-id hole that `_get_response_format` was hardened against in this same PR.

```python
h = OpenAIModelHandler(OpenAIModel('my-custom-ft', client))
req = ModelRequest(messages=[...], config=OpenAIConfig(model='my-custom-ft'))
await h.generate(req, ctx)
# KeyError: 'my-custom-ft'
```

`supported_models[self._model.name]` indexes directly. `_create_model_action` builds the handler without going through `get_model_handler`'s membership check, and this PR promotes `config.model`/`config.version` to first-class typed knobs, so fine-tunes and dated snapshots now reach this line and crash with a bare `KeyError` rather than a `GenkitError`.

```suggestion
        model_info = supported_models.get(self._model.name)
        if model_info is None:
            return
```
(applied at line 111, replacing the direct index; the `model_info.versions` check below then stays as-is)

py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
```diff
@@ -61,6 +78,21 @@ class Anthropic(Plugin):
...
+    def claude_model(
+        cls, name: KnownClaude | str, *, config: AnthropicConfig | None = None
+    ) -> ModelRef[AnthropicConfig]:
+        """Typed ref for a Claude model, e.g. ``Anthropic.claude_model('claude-sonnet-4-5')``.
+
+        Unknown ids are allowed so new Claude releases work before this
+        plugin learns their names; every Anthropic generate model takes
+        AnthropicConfig.
+        """
+        local = _strip_ref_prefixes(str(name))
```
apascal07: `str(name)` with no type guard mints a real-looking ref from a non-string:

```python
Anthropic.claude_model(None).name   # 'anthropic/None'
Anthropic.claude_model(123).name    # 'anthropic/123'
```

`_model_refs.family_model_ref` in genkit-google-genai guards this deliberately — its comment calls out `str(None) is 'None'` — and `model_refs_test.py` there pins the behaviour. The two constructors added here skip it, so a ref built from an unset variable becomes valid-looking and fails much later as a confusing API 404.

```suggestion
        if not isinstance(name, str):
            raise GenkitError(
                status='INVALID_ARGUMENT', message='Anthropic.claude_model: model name must be a string.'
            )
        local = _strip_ref_prefixes(name)
```

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
```diff
@@ -200,6 +218,34 @@ class OpenAI(Plugin):
...
 
+    @classmethod
+    def gpt_model(cls, name: KnownGpt | str, *, config: OpenAIConfig | None = None) -> ModelRef[OpenAIConfig]:
+        """Typed ref for an OpenAI chat model, e.g. ``OpenAI.gpt_model('gpt-4o')``.
+
+        Chat only: image, TTS, STT, and embedding ids validate different
+        request shapes, so binding OpenAIConfig to them would let chat-only
+        keys like frequency_penalty ride into the wrong endpoint.
+        """
+        local = _strip_ref_prefixes(str(name))
```
apascal07: Same missing type guard as `Anthropic.claude_model`:

```python
OpenAI.gpt_model(None).name   # 'openai/None'
OpenAI.gpt_model(123).name    # 'openai/123'
```

`_classify_model('None')` returns `CHAT`, so the wrong-family gate below does not catch it either and the bad ref reaches the API as a 404.

```suggestion
        if not isinstance(name, str):
            raise GenkitError(
                status='INVALID_ARGUMENT', message='OpenAI.gpt_model: model name must be a string.'
            )
        local = _strip_ref_prefixes(name)
```

py/packages/genkit-openai/src/genkit_openai/models/model.py
apascal07, line 53: Peeling `max_output_tokens` without mapping it to `max_tokens` makes Genkit's portable output cap a no-op on this plugin:

```python
action = OpenAI(api_key='k')._create_model_action('openai/gpt-4o')
validated = action._validate_input({
    'messages': [{'role': 'user', 'content': [{'text': 'hi'}]}],
    'config': {'maxOutputTokens': 256, 'temperature': 0.3},
})
validated.config.max_output_tokens   # 256.0
await OpenAIModel('gpt-4o', client)._get_openai_request_config(validated)
# body keys: ['messages', 'model', 'temperature']
```

`normalize_config` maps it onto `max_tokens` for plain `ModelConfig` input, so only the typed `ModelRequest[OpenAIConfig]` path drops it, and that is now the path every registered chat action takes. Someone who set `maxOutputTokens` to bound cost gets uncapped responses with nothing raised.

py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
apascal07, line 91: `str(name)` with no type guard mints a valid-looking ref from a non-string:

```python
Anthropic.claude_model(None).name   # 'anthropic/None'
Anthropic.claude_model(123).name    # 'anthropic/123'
```

A ref built from an unset variable then fails as an API 404 instead of at construction. `family_model_ref` in genkit-google-genai rejects this case explicitly (its comment calls out that `str(None)` is `'None'`), so the constructors added across this stack disagree on it.

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
apascal07, line 229: Same missing guard as `Anthropic.claude_model`:

```python
OpenAI.gpt_model(None).name   # 'openai/None'
OpenAI.gpt_model(123).name    # 'openai/123'
```

`_classify_model('None')` returns `CHAT`, so the wrong-family gate below does not catch it either.

apascal07 (approved):

py/packages/genkit-openai/src/genkit_openai/models/model.py
huangjeff5: Leaving this unmapped.

`config={'maxOutputTokens': 256}` binds on `OpenAIConfig.max_output_tokens`, then `_openai_create_kwargs` peels it. The cap on this plugin is `max_tokens` / `max_completion_tokens`.

Copying 256 onto `max_tokens` works on gpt-4o and 400s on o-series ("Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead."). Those two knobs also mean different things — `max_completion_tokens` includes hidden reasoning tokens.

JS `toOpenAIRequestBody` already leaves `maxOutputTokens` unused. Go `WithConfig` deletes it. Same starting dict, two OpenAI fields — we should not pick one.

py/packages/genkit-anthropic/src/genkit_anthropic/plugin.py
huangjeff5: Agreed. `isinstance(name, str)` before strip, same reason as `family_model_ref`: `str(None)` is `'None'` and unknown ids are allowed, so it would mint `anthropic/None`. Covered in `test_claude_model_rejects_non_string_name`.

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
huangjeff5: Agreed. Same guard. `_classify_model('None')` is CHAT, so the family gate would not catch it. Covered in `test_gpt_model_rejects_non_string_name`.

huangjeff5:
@gemini review

PR #6109, opened 2026-08-19
fix(go/plugins/compat_oai): stop forwarding OpenAI credentials to other providers

apascal07:
`openai.NewClient` prepends `DefaultClientOptions()`, which reads `OPENAI_API_KEY`, `OPENAI_ORG_ID`, and `OPENAI_PROJECT_ID` from the environment. Every `compat_oai` plugin builds its client through it, so all of them forwarded OpenAI's identity to whatever endpoint they pointed at: with the two IDs set, a request to DeepSeek or xAI or any other sibling carried `Openai-Organization` and `Openai-Project` next to the correct provider key. `compat_oai/anthropic` applied `option.WithAPIKey` only when the key was non-empty, so with `ANTHROPIC_API_KEY` unset it sent `Authorization: Bearer $OPENAI_API_KEY` to Anthropic as well. The other siblings escaped that one only by panicking at Init on a missing key.

`OpenAICompatible.Init` now clears all three before applying anything the plugin composes. Each is cleared twice, because the option that sets one writes both a field on the request config and a header, and an empty value leaves the header behind rather than removing it:

```go
option.WithAPIKey(""), option.WithHeaderDel("Authorization"),
option.WithOrganization(""), option.WithHeaderDel("OpenAI-Organization"),
option.WithProject(""), option.WithHeaderDel("OpenAI-Project"),
```

The scrub is folded into `o.Opts` rather than passed to `NewClient` alone. There are two construction sites: `clientForKey` clones `o.Opts` for the per-request key override, so a scrub living only in `Init`'s client would leave that path leaking.

It is a floor, not a ban. Landing before the plugin's own options means a plugin that wants one of the three sets it and wins, which is how `openai` keeps its behavior: it reads `OPENAI_ORG_ID` and `OPENAI_PROJECT_ID` itself now, as it already did `OPENAI_API_KEY`. Nothing changes for it on the wire.

`vertexai/modelgarden` builds on the same base and was sending both headers to Vertex AI. Its oauth2 transport overwrites `Authorization`, so the key never leaked there.

**`OPENAI_BASE_URL` is left alone.** The SDK inherits it too, but every plugin sets a base URL that wins over it. Closing it at the base would make `OpenAICompatible.BaseURL` required, since `option.RequestOption` takes an openai-go-internal type and the base cannot see a base URL passed through `Opts`. That breaks third-party plugins and belongs in its own PR.

### DIFF
--- a/go/plugins/compat_oai/compat_oai_test.go
+++ b/go/plugins/compat_oai/compat_oai_test.go
@@ -53,6 +53,17 @@ func headerRecordingServer(t *testing.T) (url string, lastHeader func() http.Hea
 	}
 }
 
+// assertHeaderAbsent fails unless the header carries no values at all.
+// [http.Header.Get] cannot tell an absent header from one sent with an empty
+// value, and clearing the identity without a WithHeaderDel leaves exactly
+// that on the wire, so asserting on Get would pass either way.
+func assertHeaderAbsent(t *testing.T, header http.Header, name string) {
+	t.Helper()
+	if values := header.Values(name); len(values) != 0 {
+		t.Errorf("%s = %q, want the header dropped rather than blanked", name, values)
+	}
+}
+
 // generateOnce runs one request through an initialized plugin.
 func generateOnce(t *testing.T, o *OpenAICompatible) {
 	t.Helper()
@@ -102,15 +113,13 @@ func TestInitDropsInheritedOpenAIIdentity(t *testing.T) {
 			generateOnce(t, o)
 
 			got := lastHeader()
-			if auth := got.Get("Authorization"); auth != tt.wantAuth {
+			if tt.wantAuth == "" {
+				assertHeaderAbsent(t, got, "Authorization")
+			} else if auth := got.Get("Authorization"); auth != tt.wantAuth {
 				t.Errorf("Authorization = %q, want %q", auth, tt.wantAuth)
 			}
-			if org := got.Get("OpenAI-Organization"); org != "" {
-				t.Errorf("OpenAI-Organization = %q, want the inherited organization dropped", org)
-			}
-			if project := got.Get("OpenAI-Project"); project != "" {
-				t.Errorf("OpenAI-Project = %q, want the inherited project dropped", project)
-			}
+			assertHeaderAbsent(t, got, "OpenAI-Organization")
+			assertHeaderAbsent(t, got, "OpenAI-Project")
 		})
 	}
 }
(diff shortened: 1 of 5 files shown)


go/plugins/compat_oai/compat_oai_test.go
```diff
@@ -0,0 +1,182 @@
...
+			url, lastHeader := headerRecordingServer(t)
+			o := &OpenAICompatible{Provider: "testprovider", APIKey: tt.apiKey, BaseURL: url}
+			o.Init(context.Background())
+			generateOnce(t, o)
+
+			got := lastHeader()
+			if auth := got.Get("Authorization"); auth != tt.wantAuth {
+				t.Errorf("Authorization = %q, want %q", auth, tt.wantAuth)
+			}
+			if org := got.Get("OpenAI-Organization"); org != "" {
```
huangjeff5: There's a subtle testing blind spot here with `http.Header.Get()`:

In Go, `got.Get("OpenAI-Organization")` returns `""` both when the header is completely absent **and** when the header is present with an empty value (`"OpenAI-Organization: "`).

If you temporarily remove `WithHeaderDel` from `clearInheritedOpenAIIdentity()`, the empty header is still sent over HTTP (`got.Values("OpenAI-Organization")` is `[]string{""}`), but this test still passes green because `Get()` returns `""`.

Could we update the assertions in these tests to check that the header is actually absent (e.g. `len(got.Values("OpenAI-Organization")) == 0` or checking `got["OpenAI-Organization"]`)? That way the test directly verifies `WithHeaderDel` is doing its job.

---
**Minor note on README:**
The example snippet above the new paragraph only sets `Provider` and calls `Init()`. Since `OPENAI_API_KEY` from the environment is now scrubbed by default, that example won't send any authorization. Updating the example to explicitly pass `APIKey` or `WithAPIKey` would align it with the new behavior.

huangjeff5 (approved):
The scrub works. I ran the anthropic-shaped Init (empty APIKey field, WithAPIKey in Opts, OPENAI_* in the env) and it sent the provider key with no org/project. One test gap on the thing the comment is about.

go/plugins/compat_oai/compat_oai_test.go
apascal07: Switched the assertions to `Values`, so an absent header and a blanked one no longer read the same. Dropping any of the three `WithHeaderDel` calls now fails: the organization and project arrive as `[""]`, the authorization as `["Bearer"]`.

PR #6110, opened 2026-08-19
fix(go/core): prefer completion over cancellation in bidi `Send`

apascal07:
`BidiConnection.run` closes `doneCh` and then releases the connection context, so a context that reads as done is usually the action finishing rather than an abort. `Output` and `Receive` both account for that by re-checking `doneCh` inside their context arm. `Send` did not: it tested completion and cancellation in two separate selects, so a teardown landing between them returned `context.Canceled`, and its blocking select carried the same gap.

Callers tolerate `ErrActionCompleted` and go to `Output` for the outcome, but take a cancellation as the outcome. `exp.Agent.Run` is where it surfaced: an agent whose init is rejected fails before reading any input, so the caller saw `CANCELLED: context canceled` instead of the `FAILED_PRECONDITION` sitting on `Output`. That flaked `TestAgent_ResumeFromSessionID_PendingTipRejected`.

Both context arms now route through one helper that prefers completion. The window is a few instructions wide, so the test drives it under contention rather than forcing it, and reports 18 to 24 cancellations per run without the fix against none with it.

### DIFF
--- a/go/core/bidi_action.go
+++ b/go/core/bidi_action.go
@@ -561,7 +561,7 @@ func (c *BidiConnection[In, Out, Stream]) Send(input In) (err error) {
 	}
 	select {
 	case <-c.ctx.Done():
-		return c.ctxErr()
+		return c.completedOrCtxErr()
 	default:
 	}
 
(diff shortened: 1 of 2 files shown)


go/core/bidi_action.go
```diff
@@ -561,20 +561,37 @@ func (c *BidiConnection[In, Out, Stream]) Send(input In) (err error) {
...
+// checks above, or racing the blocking select, must still surface as
+// [ErrActionCompleted] so the caller goes to Output for the real result
+// instead of reporting a cancellation that never happened.
+func (c *BidiConnection[In, Out, Stream]) completedOrCtxErr() error {
+	select {
+	case <-c.doneCh:
+		return status.Errorf(ErrActionCompleted, "action has completed")
+	default:
+	}
+	return c.ctxErr()
```
huangjeff5, line 592: This default arm is the live-abort path, and nothing in the PR would fail if it disappeared.

`TestSendPrefersCompletionOverTeardownCancellation` only counts `context.Canceled` on teardown and always calls `Output`. `TestBidiConnectionCancel` never calls `Send`. Forcing this helper to always return `ErrActionCompleted` left those tests green; `Send` then reported "action has completed" after `Cancel` on a session still blocked on `ctx.Done()`.

Callers like `Agent.Run` treat any non-`ErrActionCompleted` send error as the outcome. Could we add `Cancel` then `Send` on a still-running session and assert `errors.Is(err, context.Canceled) && !errors.Is(err, ErrActionCompleted)`?

huangjeff5 (changes requested):
The completion-preference arm is covered. The live-abort arm of `completedOrCtxErr` is not: nothing in this PR would fail if the helper always returned `ErrActionCompleted`.

huangjeff5 (approved):

go/core/bidi_action.go
apascal07: Added, covering both of the helper's call sites: a send issued after `Cancel`, and one already blocked in the select when it lands. Forcing the helper to always return `ErrActionCompleted` now fails both.

PR #6119, opened 2026-08-19
feat(go/plugins/middleware): add background sub-agent delegation (`Agents.Async`)

apascal07:
Second PR of the train, on #6118 (`AgentHandle`, `waitForSnapshot`). Adds background delegation to the `Agents` middleware: with `Async` set, every delegation tool takes a `background` flag that launches the sub-agent through its detach support and returns a task ID at once, and three shared tools let the orchestrator check, join, and stop what it launched. `Agents.Async` is the only new exported symbol.

## One control per thing the orchestrator can do with a task

```go
ai.WithUse(&middlewarex.Agents{
    Agents: []aix.AgentRef{{Name: "researcher"}},
    Async:  true,
})
```

```
delegate_to_researcher    {task, background: true}
                          -> {response, taskId: "researcher:<snapshotId>", status: "pending"}
check_background_tasks    {taskIds}
                          -> {tasks: [{taskId, agent, status, response?, artifacts?, error?}]}
wait_for_background_tasks {taskIds, timeoutSeconds?, waitFor?: "all" | "first"}
                          -> the same, plus timedOut
abort_background_tasks    {taskIds}
                          -> the same, each task reported where the stop left it
```

The async delegation input is a distinct Go type, since tool schemas are inferred statically from the struct; `ToolPrefix` namespaces the shared tools as it does the delegation tools. A launch spends a `MaxDelegations` slot, and a launch the runtime rejects before any work ran refunds it, so the hinted synchronous retry is not refused by a cap the refusal consumed. History is never forwarded to a background launch: detach requires a server-managed sub-agent, and server-managed init rejects seeded state.

## The sub-agent's snapshot is the whole record

The middleware keeps no task registry. The task ID rides in the tool result, so the orchestrator's history is the registry, and a re-instantiated orchestrator collects with nothing but the IDs it finds there. Reports key on the row: `pending` is status only; `completed` carries the last model message and the artifacts, merged under the deterministic `<agent>_<snapshotId[:8]>` namespace so a re-check after a restart overwrites rather than duplicates; a completed row whose finish reason carries no answer (`failed`, `interrupted`) reports as `failed` with the reason in `error`, because a model told `completed` moves on without reading `error`; `failed`, `aborted`, and `expired` carry an explanatory error. A completed run with no model text says so and names its artifact count, so silence is not read as failure. NOT_FOUND means "delegate again", FAILED_PRECONDITION and INVALID_ARGUMENT are reported verbatim as dead ends, and any other read error is "check again later". Settled reports are cached for the rest of the generate call.

## Wait follows tasks instead of polling them

The wait tool dispatches each task's `waitForSnapshot` companion action once, concurrently, so a trace carries one span per task per call and a task is reported the moment it settles. An elapsed `timeoutSeconds` returns the current statuses with `timedOut`; zero waits until settled, negative reports immediately. `waitFor: "first"` returns as soon as any listed task settles (`unknown` counts, since an unresolvable handle is an answer to act on), and a won race is not a timeout. Abort flips the pending row, then reads and reports the settled row through the check path, because the abort action's bare status would answer `completed` for a task that had already finished and strand its response.

## Middleware build errors keep their own status

Both wrap sites in `ai` classified every `Middleware.New` failure as INVALID_ARGUMENT, so an UNAVAILABLE from a network-backed `New` or a cancelled context reached the caller as a caller mistake. A classified error now keeps its status; only unclassified errors default to INVALID_ARGUMENT.

- **No interrupt resume for background tasks.** Same report-as-text policy as synchronous delegation.
- **Task handles are not access-scoped**, mirroring the sub-agent's own unscoped companion actions; the config doc tells multi-tenant deployments to treat snapshot IDs as capability-like secrets.
- **No operator cap on the wait.** `timeoutSeconds` is the model's to choose, bounded in practice by the calling context.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -504,7 +504,7 @@ ai.WithUse(&middlewarex.Agents{
 })
 ```
 
-The orchestrator then launches with `{"task": "...", "background": true}`, posts an update while the sub-agent runs, and collects the result later. `wait_for_background_tasks` takes an optional `timeoutSeconds`, so a slow task becomes an interim answer instead of a blocked turn. An abort is safe to call on any task: one that had already finished is left alone and reports its result, so the orchestrator never loses an answer by giving up on it.
+The orchestrator then launches with `{"task": "...", "background": true}`, posts an update while the sub-agent runs, and collects the result later. `wait_for_background_tasks` takes an optional `timeoutSeconds`, so a slow task becomes an interim answer instead of a blocked turn, and `waitFor: "first"` turns the join into a race: the tool returns as soon as any listed task settles while the rest keep running, so the orchestrator can act on the first answer. An abort is safe to call on any task: one that had already finished is left alone and reports its result, so the orchestrator never loses an answer by giving up on it.
 
 Sub-agents are named by `aix.AgentRef`, either captured from an agent value with `agent.Ref()` or written by hand (`aix.AgentRef{Name: "researcher"}`). The middleware composes with the `Artifacts` middleware: give a sub-agent `&middlewarex.Artifacts{}` so it can save output, set `ArtifactStrategy: middlewarex.ArtifactStrategySession` to merge those artifacts into the orchestrator's session instead of inlining them in the tool result, and add `&middlewarex.Artifacts{Readonly: true}` on the orchestrator so it can review them before answering.
 
--- a/go/README.md
+++ b/go/README.md
@@ -446,7 +446,9 @@ Detach requires a store that implements `SnapshotSubscriber` (both bundled local
 > [!WARNING]
 > This API is in preview and may experience breaking changes in minor releases.
 
-The experimental `Agents` middleware (in `plugins/middleware/exp`) lets one agent delegate to others. It injects one `delegate_to_<name>` tool per sub-agent and a `<sub-agents>` listing into the orchestrator's system prompt, then runs the chosen sub-agent and returns its result when the model calls the tool. Each sub-agent's `aix.WithDescription` (captured by `agent.Ref()`) tells the orchestrator when to reach for it:
+The experimental `Agents` middleware (in `plugins/middleware/exp`) lets one agent delegate to others. It injects one `delegate_to_<name>` tool per sub-agent and a `<sub-agents>` listing into the orchestrator's system prompt, then runs the chosen sub-agent and returns its result when the model calls the tool. Each sub-agent's `aix.WithDescription` (captured by `agent.Ref()`) tells the orchestrator when to reach for it.
+
+Set `Async: true` to let the orchestrator keep working while a sub-agent runs: each delegation tool gains a `background` flag that returns a task ID instead of waiting, and two shared tools, `check_background_tasks` and `wait_for_background_tasks`, collect the results. Background delegation needs a sub-agent whose session store supports it (see `AgentMetadata.Abortable`).
 
 ```go
 import (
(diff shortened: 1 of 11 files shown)


go/plugins/middleware/exp/agents_async.go
```diff
@@ -0,0 +1,565 @@
...
+		}, fmt.Sprintf("%s_%s", ref.Name, shortSnapshotID(snapshotID)))
+		switch snap.FinishReason {
+		case aix.AgentFinishReasonFailed, aix.AgentFinishReasonInterrupted:
+			// The row committed, so the status is completed, but the agent
+			// declared a reason that carries no answer (it can do so without
+			// erroring, which is why the two disagree). The fold produced
+			// explanatory text, not a response: report it as the error it is,
+			// or the report would claim success in one field and failure in
+			// the next.
+			report.Error = folded.Response
```
pavelgj, line 655: Want to sanity-check the reporting contract here: when a snapshot committed as `completed` but the finish reason is `failed`/`interrupted`, we report `Status: "completed"` and route the explanation into `Error`. The comment explains the persistence-vs-declared-reason split well, but the model on the other end only sees the JSON, and `status: "completed"` reads like success. There's a real chance it treats the task as done and never looks at `error`.

Could we make the reported `Status` reflect the outcome the reader should act on, e.g. surface `failed`/`interrupted` (or a dedicated value) in `Status` when the fold produced an error rather than a response? Then the status and body agree for the consumer, not just internally. Fine to be told it's intentional if the collection contract is meant to key off `error` presence.

go/README.md
```diff
@@ -404,7 +404,9 @@ Detach requires a store that implements `SnapshotSubscriber` (both bundled local
...
+The experimental `Agents` middleware (in `plugins/middleware/exp`) lets one agent delegate to others. It injects one `delegate_to_<name>` tool per sub-agent and a `<sub-agents>` listing into the orchestrator's system prompt, then runs the chosen sub-agent and returns its result when the model calls the tool. Each sub-agent's `aix.WithDescription` (captured by `agent.Ref()`) tells the orchestrator when to reach for it.
+
+Set `Async: true` to let the orchestrator keep working while a sub-agent runs: each delegation tool gains a `background` flag that returns a task ID instead of waiting, and two shared tools, `check_background_tasks` and `wait_for_background_tasks`, collect the results. Background delegation needs a sub-agent whose session store supports it (see `AgentMetadata.Abortable`).
```
pavelgj: Nice that `Async` gets a prose paragraph. Two small follow-ups:

1. The synchronous `Agents` section just above has a `go` snippet; a tiny `Async: true` example here would make it symmetric and copy-pasteable (the `commander.go` sample is great, but README skimmers won't see it).
2. This adds a user-facing capability (the `background` flag plus two tools), so it probably warrants a docsite update in the separate docs repo. Worth filing a follow-up if there isn't one already.

go/plugins/middleware/exp/agents_async.go
apascal07: Changed. `Status` now reports `failed` when the fold produced an explanation instead of a response, so `completed` always carries a `Response`.

go/README.md
apascal07: Added, and moved the `Async` prose below the synchronous example so each mode sits next to its own snippet.

Will update docsite before release.

pavelgj (approved):
Read the delegation accounting and report path end to end. Cap accounting is sound (seq vs delegations kept separate, only the sync-fallback refusal refunds), report-by-status is correct (a committed-but-answerless row reports failed, abort reads before it stops), and the concurrency is safe (per-slot pointers, mutex-guarded cache; wg.Go needs 1.25 and go.mod is 1.25.0). Good first real consumer of the AgentHandle surface. Approving, to merge after #6118.

PR #6122, opened 2026-08-19
fix(go/core): name operation companions for the operation they perform

apascal07:
Fixes the keys a background action's lifecycle companions register under. Go named the check and cancel companions after the action itself, producing `/check-operation/{model}` and `/cancel-operation/{model}`. JS and Python both name them for the operation they perform, `{name}/check` and `{name}/cancel`, and the Dev UI builds that form, so its background-task panel got NOT_FOUND on every poll and every cancel against a Go background model.

Both companions now carry the suffix. `googlegenai` resolves the model its check companion belongs to, because dynamic resolution now hands the plugin `veo-3.0-generate-001/check`, a name `ClassifyModel` does not recognize.

`Operation.Action` still carries the start key, so anything round-tripping an operation through `LookupBackgroundAction` is unaffected. Only a hand-built check key changes, and those were already broken against the other two runtimes.

### DIFF
--- a/go/plugins/googlegenai/actions.go
+++ b/go/plugins/googlegenai/actions.go
@@ -6,6 +6,7 @@ package googlegenai
 import (
 	"context"
 	"slices"
+	"strings"
 
 	"github.com/firebase/genkit/go/core/api"
 	"google.golang.org/genai"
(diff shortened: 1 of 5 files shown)


go/plugins/googlegenai/actions.go
```diff
@@ -79,10 +80,16 @@ func resolveAction(client *genai.Client, c catalog, atype api.ActionType, id str
 	// and check actions, so the same value resolves either key. The registry
 	// registers what we return and then looks up the key it was asked for.
 	case api.ActionTypeBackgroundModel, api.ActionTypeCheckOperation:
-		if mt != ModelTypeVeo {
+		// The check companion is named for the model plus the operation it
+		// performs, so resolve the model the companion belongs to.
+		modelID := id
+		if atype == api.ActionTypeCheckOperation {
+			modelID = strings.TrimSuffix(id, "/check")
```
huangjeff5, line 88: This sentence a few lines up is now wrong: "the same value resolves either key." Start is the bare model, check is `{model}/check`. The new comment under it already says that. I'd drop the old one so the next background-model author doesn't skip the strip.

huangjeff5 (approved):
LGTM. Keys match what Dev UI builds, Operation.Action is still the start key, Veo still has no cancel companion.

Issue #4343, opened 2026-01-30
[Python] Implement with_model in model.py


yesudeep:
## Description

The with_model method needs implementation.

## Location
- File: `py/packages/genkit/src/genkit/blocks/model.py`
- Lines: 199, 210
- Current code: `# TODO: implement`

## Acceptance Criteria
- [ ] Implement the with_model method
- [ ] Add tests for the implementation

PR #6130, opened 2026-08-20
feat(py): box background operation before wrap_model

huangjeff5:
## Summary
- `generate()` on a background model returns a `ModelResponse` whose `operation` is the start handle and whose `message` is `None`. It does not poll until done.
- `generate_operation()` is that same start and returns the `Operation`. If `generate()` came back without a handle, this is the missing-handle error.
- `start()` is boxed into a `ModelResponse` **before** `wrap_model`, so a hook that reads `.message` is safe. Boxing before the chain is deliberate.
- After the hook, a background start takes the same path as chat: pin `response.request` to the pre-`wrap_model` request, debug-log, `assert_valid()`, persist. `message is None` is the exit, not a kind-keyed early return.
- A hook (Fallback, cache, a drop) may replace the boxed start. `generate()` keeps that answer. It does not demand an `operation` from every post-middleware response.
- A `start()` that is not an `Operation` (or a boxed handle) raises `FAILED_PRECONDITION: Background model 'X' did not return an operation`. A dict is not coerced into a ticket.
- `define_model` that returns an `operation` raises. Use `define_background_model`. That combo is not a product.
- Resume against a background model is rejected before `start()` is billed. A start cannot finish a tool interrupt.
- `ModelResponse.__eq__` compares `message`, `finish_reason`, and `[operation.id](http://operation.id/)`. Timing on the handle is not the job.
- `.messages` is a `@property`, so attaching `request` after a first read still shows history.

### DIFF
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -47,7 +47,7 @@ from genkit._core._action import (
     ActionKind,
     ActionRunContext,
 )
-from genkit._core._background import _ensure_operation
+from genkit._core._background import _ensure_operation, missing_operation_error
 from genkit._core._error import GenkitError
 from genkit._core._logger import get_logger, is_debug_enabled
 from genkit._core._middleware import (
@@ -657,10 +657,7 @@ def box_background_start(*, raw: object, request: ModelRequest, name: str) -> Mo
     """
     if isinstance(raw, ModelResponse):
         if raw.operation is None:
-            raise GenkitError(
-                status='FAILED_PRECONDITION',
-                message=f"Background model '{name}' did not return an operation",
-            )
+            raise missing_operation_error(name=name)
         if raw.latency_ms is None:
             raw.latency_ms = _latency_ms_from_operation(raw.operation)
         return raw
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -793,9 +793,6 @@ async def _generate_action_turn(
         # through persist and the tool loop.
         if model.kind == ActionKind.BACKGROUND_MODEL:
             if model_response.operation is None:
-                # wrap_model may read .message. The job id lives on
-                # .operation; a rebuild that only copies message /
-                # finish_reason drops it.
                 raise GenkitError(
                     status='FAILED_PRECONDITION',
                     message=(
@@ -915,7 +912,13 @@ async def _generate_action_turn(
         iteration=current_turn,
         message_index=chunks.message_index,
     )
-    return await dispatch_generate(generate_params, run_ctx, run_one_iteration)
+    response = await dispatch_generate(generate_params, run_ctx, run_one_iteration)
+    if model.kind == ActionKind.BACKGROUND_MODEL and response.operation is None:
+        raise GenkitError(
+            status='FAILED_PRECONDITION',
+            message=('wrap_generate returned no operation for a background model; pass through response.operation'),
+        )
+    return response
 
 
 def apply_format(
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -648,6 +648,43 @@ def _latency_ms_from_operation(operation: Operation) -> float | None:
     return None
 
 
+def box_background_start(*, raw: object, request: ModelRequest, name: str) -> ModelResponse:
+    """Turn a background start into the ModelResponse wrap_model reads.
+
+    start() is a poll handle. Already-boxed responses pass through; a chat
+    turn with no handle is a model-author mistake.
+    """
+    if isinstance(raw, ModelResponse):
+        if raw.operation is None:
+            raise GenkitError(
+                status='FAILED_PRECONDITION',
+                message=f"Background model '{name}' did not return an operation",
+            )
+        if raw.latency_ms is None:
+            raw.latency_ms = _latency_ms_from_operation(raw.operation)
+        return raw
+    if not isinstance(raw, Operation):
+        raise GenkitError(
+            status='FAILED_PRECONDITION',
+            message=f"Background model '{name}' did not return an operation",
+        )
+    return ModelResponse(
+        operation=raw,
+        request=request,
+        latency_ms=_latency_ms_from_operation(raw),
+    )
+
+
+def reject_foreground_operation(*, raw: object, name: str) -> ModelResponse:
+    """A chat model returns a message, not a job handle."""
+    if isinstance(raw, Operation) or (isinstance(raw, ModelResponse) and raw.operation is not None):
+        raise GenkitError(
+            status='FAILED_PRECONDITION',
+            message=f"Model '{name}' is a define_model and returned an operation; use define_background_model",
+        )
+    return cast(ModelResponse, raw)
+
+
 def _persist_threaded_conversation(response: ModelResponse, messages: list[Message]) -> ModelResponse:
     """Persist the threaded conversation onto the response's request.
 
@@ -795,37 +832,9 @@ async def _generate_action_turn(
                     abort_signal=c.abort_signal,
                 )
             ).response
-            # wrap_model reads .message. A background start is a poll handle,
-            # so box it before the hook. A chat model that returns a handle
-            # is registered on the wrong kind.
             if model.kind == ActionKind.BACKGROUND_MODEL:
-                if isinstance(raw, ModelResponse):
-                    if raw.operation is None:
-                        raise GenkitError(
-                            status='FAILED_PRECONDITION',
-                            message=f"Background model '{model.name}' did not return an operation",
-                        )
-                    if raw.latency_ms is None:
-                        raw.latency_ms = _latency_ms_from_operation(raw.operation)
-                    return raw
-                if not isinstance(raw, Operation):
-                    raise GenkitError(
-                        status='FAILED_PRECONDITION',
-                        message=f"Background model '{model.name}' did not return an operation",
-                    )
-                return ModelResponse(
-                    operation=raw,
-                    request=params.request,
-                    latency_ms=_latency_ms_from_operation(raw),
-                )
-            if isinstance(raw, Operation) or (isinstance(raw, ModelResponse) and raw.operation is not None):
-                raise GenkitError(
-                    status='FAILED_PRECONDITION',
-                    message=(
-                        f"Model '{model.name}' is a define_model and returned an operation; use define_background_model"
-                    ),
-                )
-            return raw
+                return box_background_start(raw=raw, request=params.request, name=model.name)
+            return reject_foreground_operation(raw=raw, name=model.name)
 
         with chunks.intercept_model_stream(ctx, role=Role.MODEL):
             model_response = await dispatch_model(
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -72,6 +72,7 @@ from genkit._core._typing import (
     FinishReason,
     MiddlewareRef,
     MultipartToolResponse,
+    Operation,
     Part,
     Role,
     TextPart,
@@ -766,7 +767,7 @@ async def _generate_action_turn(
             request = _augment_with_context(request)
 
         async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> ModelResponse:
-            return (
+            raw = (
                 await model.run(
                     input=params.request,
                     context=c.custom_context,
@@ -774,6 +775,11 @@ async def _generate_action_turn(
                     abort_signal=c.abort_signal,
                 )
             ).response
+            # start() returns a poll handle. wrap_model is documented as
+            # seeing a ModelResponse, so wrap before the hook runs.
+            if isinstance(raw, Operation):
+                return ModelResponse(operation=raw, request=params.request)
+            return raw
 
         with chunks.intercept_model_stream(ctx, role=Role.MODEL):
             model_response = await dispatch_model(
@@ -781,6 +787,11 @@ async def _generate_action_turn(
                 ctx,
                 next_fn,
             )
+
+        if model_response.operation is not None:
+            if model_response.request is None:
+                model_response.request = request
+            return model_response
 
         def message_parser(msg: Message) -> Any:  # noqa: ANN401
             if formatter is None:
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -675,14 +675,16 @@ def box_background_start(*, raw: object, request: ModelRequest, name: str) -> Mo
     )
 
 
-def reject_foreground_operation(*, raw: object, name: str) -> ModelResponse:
+def assert_foreground_model_has_no_operation(*, raw: object, name: str) -> None:
     """A chat model returns a message, not a job handle."""
     if isinstance(raw, Operation) or (isinstance(raw, ModelResponse) and raw.operation is not None):
         raise GenkitError(
             status='FAILED_PRECONDITION',
-            message=f"Model '{name}' is a define_model and returned an operation; use define_background_model",
+            message=(
+                f"Model '{name}' is a regular model that returns a response immediately. "
+                'Use define_background_model for background models that return operations.'
+            ),
         )
-    return cast(ModelResponse, raw)
 
 
 def _persist_threaded_conversation(response: ModelResponse, messages: list[Message]) -> ModelResponse:
@@ -834,7 +836,8 @@ async def _generate_action_turn(
             ).response
             if model.kind == ActionKind.BACKGROUND_MODEL:
                 return box_background_start(raw=raw, request=params.request, name=model.name)
-            return reject_foreground_operation(raw=raw, name=model.name)
+            assert_foreground_model_has_no_operation(raw=raw, name=model.name)
+            return cast(ModelResponse, raw)
 
         with chunks.intercept_model_stream(ctx, role=Role.MODEL):
             model_response = await dispatch_model(
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -637,6 +637,17 @@ class ChunkAccumulator:
             ctx.replace_on_chunk(previous)
 
 
+def _latency_ms_from_operation(operation: Operation) -> float | None:
+    """Copy start timing off the handle if the wrapper stamped it."""
+    meta = operation.metadata
+    if not isinstance(meta, dict):
+        return None
+    raw_ms = meta.get('latencyMs')
+    if isinstance(raw_ms, int | float):
+        return float(raw_ms)
+    return None
+
+
 def _persist_threaded_conversation(response: ModelResponse, messages: list[Message]) -> ModelResponse:
     """Persist the threaded conversation onto the response's request.
 
@@ -663,6 +674,15 @@ async def _generate_action_turn(
     raise_if_aborted(run_ctx.abort_signal)
 
     model, tools, format_def = await resolve_parameters(registry, raw_request)
+
+    if model.kind == ActionKind.BACKGROUND_MODEL and raw_request.resume is not None:
+        raise GenkitError(
+            status='FAILED_PRECONDITION',
+            message=(
+                f"Cannot resume background model '{model.name}'; "
+                'a background start cannot satisfy an interrupted tool turn'
+            ),
+        )
 
     raw_request, formatter = apply_format(raw_request, format_def)
 
@@ -775,10 +795,36 @@ async def _generate_action_turn(
                     abort_signal=c.abort_signal,
                 )
             ).response
-            # start() returns a poll handle. wrap_model is documented as
-            # seeing a ModelResponse, so wrap before the hook runs.
-            if isinstance(raw, Operation):
-                return ModelResponse(operation=raw, request=params.request)
+            # wrap_model reads .message. A background start is a poll handle,
+            # so box it before the hook. A chat model that returns a handle
+            # is registered on the wrong kind.
+            if model.kind == ActionKind.BACKGROUND_MODEL:
+                if isinstance(raw, ModelResponse):
+                    if raw.operation is None:
+                        raise GenkitError(
+                            status='FAILED_PRECONDITION',
+                            message=f"Background model '{model.name}' did not return an operation",
+                        )
+                    if raw.latency_ms is None:
+                        raw.latency_ms = _latency_ms_from_operation(raw.operation)
+                    return raw
+                if not isinstance(raw, Operation):
+                    raise GenkitError(
+                        status='FAILED_PRECONDITION',
+                        message=f"Background model '{model.name}' did not return an operation",
+                    )
+                return ModelResponse(
+                    operation=raw,
+                    request=params.request,
+                    latency_ms=_latency_ms_from_operation(raw),
+                )
+            if isinstance(raw, Operation) or (isinstance(raw, ModelResponse) and raw.operation is not None):
+                raise GenkitError(
+                    status='FAILED_PRECONDITION',
+                    message=(
+                        f"Model '{model.name}' is a define_model and returned an operation; use define_background_model"
+                    ),
+                )
             return raw
 
         with chunks.intercept_model_stream(ctx, role=Role.MODEL):
@@ -787,21 +833,6 @@ async def _generate_action_turn(
                 ctx,
                 next_fn,
             )
-
-        # A background start is a poll handle, not a conversation turn.
-        # define_model LRO replies that also carry a message still go
-        # through persist and the tool loop.
-        if model.kind == ActionKind.BACKGROUND_MODEL:
-            if model_response.operation is None:
-                raise GenkitError(
-                    status='FAILED_PRECONDITION',
-                    message=(
-                        'wrap_model returned no operation for a background model; pass through response.operation'
-                    ),
-                )
-            if model_response.request is None:
-                model_response.request = request
-            return model_response
 
         def message_parser(msg: Message) -> Any:  # noqa: ANN401
             if formatter is None:
@@ -912,13 +943,7 @@ async def _generate_action_turn(
         iteration=current_turn,
         message_index=chunks.message_index,
     )
-    response = await dispatch_generate(generate_params, run_ctx, run_one_iteration)
-    if model.kind == ActionKind.BACKGROUND_MODEL and response.operation is None:
-        raise GenkitError(
-            status='FAILED_PRECONDITION',
-            message=('wrap_generate returned no operation for a background model; pass through response.operation'),
-        )
-    return response
+    return await dispatch_generate(generate_params, run_ctx, run_one_iteration)
 
 
 def apply_format(
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -47,6 +47,7 @@ from genkit._core._action import (
     ActionKind,
     ActionRunContext,
 )
+from genkit._core._background import _ensure_operation
 from genkit._core._error import GenkitError
 from genkit._core._logger import get_logger, is_debug_enabled
 from genkit._core._middleware import (
@@ -663,15 +664,11 @@ def box_background_start(*, raw: object, request: ModelRequest, name: str) -> Mo
         if raw.latency_ms is None:
             raw.latency_ms = _latency_ms_from_operation(raw.operation)
         return raw
-    if not isinstance(raw, Operation):
-        raise GenkitError(
-            status='FAILED_PRECONDITION',
-            message=f"Background model '{name}' did not return an operation",
-        )
+    op = _ensure_operation(response=raw, name=name)
     return ModelResponse(
-        operation=raw,
+        operation=op,
         request=request,
-        latency_ms=_latency_ms_from_operation(raw),
+        latency_ms=_latency_ms_from_operation(op),
     )
 
 
--- a/py/packages/genkit/src/genkit/_ai/_generate.py
+++ b/py/packages/genkit/src/genkit/_ai/_generate.py
@@ -47,7 +47,7 @@ from genkit._core._action import (
     ActionKind,
     ActionRunContext,
 )
-from genkit._core._background import _ensure_operation, missing_operation_error
+from genkit._core._background import _ensure_operation, missing_operation_error, stamp_operation_action
 from genkit._core._error import GenkitError
 from genkit._core._logger import get_logger, is_debug_enabled
 from genkit._core._middleware import (
@@ -638,39 +638,24 @@ class ChunkAccumulator:
             ctx.replace_on_chunk(previous)
 
 
-def _latency_ms_from_operation(operation: Operation) -> float | None:
-    """Copy start timing off the handle if the wrapper stamped it."""
-    meta = operation.metadata
-    if not isinstance(meta, dict):
-        return None
-    raw_ms = meta.get('latencyMs')
-    if isinstance(raw_ms, int | float):
-        return float(raw_ms)
-    return None
+def box_background_start(
+    *,
+    raw: object,
+    request: ModelRequest,
+    name: str,
+    latency_ms: float | None = None,
+) -> ModelResponse:
+    """Turn a start() Operation into the ModelResponse wrap_model reads.
 
-
-def box_background_start(*, raw: object, request: ModelRequest, name: str) -> ModelResponse:
-    """Turn a background start into the ModelResponse wrap_model reads.
-
-    start() is a poll handle. Already-boxed responses pass through; a chat
-    turn with no handle is a model-author mistake.
+    Timing comes from Action.run, not from the ticket.
     """
-    if isinstance(raw, ModelResponse):
-        if raw.operation is None:
-            raise missing_operation_error(name=name)
-        if raw.latency_ms is None:
-            raw.latency_ms = _latency_ms_from_operation(raw.operation)
-        return raw
     op = _ensure_operation(response=raw, name=name)
-    return ModelResponse(
-        operation=op,
-        request=request,
-        latency_ms=_latency_ms_from_operation(op),
-    )
+    stamp_operation_action(operation=op, name=name)
+    return ModelResponse(operation=op, request=request, latency_ms=latency_ms)
 
 
-def assert_foreground_model_has_no_operation(*, raw: object, name: str) -> None:
-    """A chat model returns a message, not a job handle."""
+def require_model_response(*, raw: object, name: str) -> ModelResponse:
+    """A chat model returns a ModelResponse, not a dict or a job handle."""
     if isinstance(raw, Operation) or (isinstance(raw, ModelResponse) and raw.operation is not None):
         raise GenkitError(
             status='FAILED_PRECONDITION',
@@ -679,6 +664,25 @@ def assert_foreground_model_has_no_operation(*, raw: object, name: str) -> None:
                 'Use define_background_model for background models that return operations.'
             ),
         )
+    if not isinstance(raw, ModelResponse):
+        raise GenkitError(
+            status='FAILED_PRECONDITION',
+            message=f"Model '{name}' did not return a ModelResponse.",
+        )
+    return raw
+
+
+@dataclass
+class BoxedStart:
+    """The ModelResponse start() produced, before wrap_model / wrap_generate."""
+
+    response: ModelResponse | None = None
+
+
+def assert_hook_kept_operation(*, boxed: ModelResponse | None, after_hooks: ModelResponse, name: str) -> None:
+    """A hook that called start() and then dropped the ticket orphans the job."""
+    if boxed is not None and boxed.operation is not None and after_hooks.operation is None:
+        raise missing_operation_error(name=name)
 
 
 def _persist_threaded_conversation(response: ModelResponse, messages: list[Message]) -> ModelResponse:
@@ -707,6 +711,7 @@ async def _generate_action_turn(
     raise_if_aborted(run_ctx.abort_signal)
 
     model, tools, format_def = await resolve_parameters(registry, raw_request)
+    boxed_start = BoxedStart()
 
     if model.kind == ActionKind.BACKGROUND_MODEL and raw_request.resume is not None:
         raise GenkitError(
@@ -820,18 +825,22 @@ async def _generate_action_turn(
             request = _augment_with_context(request)
 
         async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> ModelResponse:
-            raw = (
-                await model.run(
-                    input=params.request,
-                    context=c.custom_context,
-                    on_chunk=c.on_chunk,
-                    abort_signal=c.abort_signal,
-                )
-            ).response
+            result = await model.run(
+                input=params.request,
+                context=c.custom_context,
+                on_chunk=c.on_chunk,
+                abort_signal=c.abort_signal,
+            )
+            raw = result.response
             if model.kind == ActionKind.BACKGROUND_MODEL:
-                return box_background_start(raw=raw, request=params.request, name=model.name)
-            assert_foreground_model_has_no_operation(raw=raw, name=model.name)
-            return cast(ModelResponse, raw)
+                boxed_start.response = box_background_start(
+                    raw=raw,
+                    request=params.request,
+                    name=model.name,
+                    latency_ms=result.latency_ms,
+                )
+                return boxed_start.response
+            return require_model_response(raw=raw, name=model.name)
 
         with chunks.intercept_model_stream(ctx, role=Role.MODEL):
             model_response = await dispatch_model(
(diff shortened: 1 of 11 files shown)


py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -765,14 +766,19 @@ async def run_one_iteration(
...
                     context=c.custom_context,
                     on_chunk=c.on_chunk,
                     abort_signal=c.abort_signal,
                 )
             ).response
+            # start() returns a poll handle. wrap_model is documented as
+            # seeing a ModelResponse, so wrap before the hook runs.
+            if isinstance(raw, Operation):
+                return ModelResponse(operation=raw, request=params.request)
+            return raw
```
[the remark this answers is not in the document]
huangjeff5: next_fn is boxing model.run(...).response, which is already an Operation or a ModelResponse. A start fn that returns a dict is a broken action, not a second wire format. Guessing Operation vs ModelResponse from 'id' in raw would also mis-read some dumps. Persisted dict handles are #6131 (check/cancel), not this generate() path. Leaving this as-is.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -233,6 +233,16 @@ def define_background_model(
...
 
+    # generate_operation looks at this flag. A background model is a
+    # poll-handle model, so the flag is set on registration.
+    supports = model_options.get('supports')
+    if not isinstance(supports, dict):
+        supports = {}
+    else:
+        supports = dict(supports)
+    supports['longRunning'] = True
+    model_options['supports'] = supports
```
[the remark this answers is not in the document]
huangjeff5: supports here is read after info.model_dump(...), so it is already a dict. We do not take a live Supports object in this block, so the isinstance(supports, dict) branch does not drop flags. No change.

apascal07 (commented):
Max-effort review of the background-operation boxing. 15 findings, each reproduced by running the code against this branch.

Three sit outside the diff and could not be anchored inline:

**`_generate.py:824` — a background start never reaches the `generate response` debug log.** `_loggable_response` is only called below the early return at 804. That is the call most worth logging, because the result arrives minutes later through a separate `check_operation`, and nothing ties the returned operation id back to the request that started it.

**`_generate.py:742` — the resumed-tool chunk is streamed for a response that has no message.** With `resume_respond` against a background model, `generate_stream` yields a `TOOL` chunk while the awaited response carries only an operation. A consumer accumulating chunks into history gets a turn that appears in no `ModelResponse.messages` and, because the early return skips `_persist_threaded_conversation`, is recorded nowhere else.

**`_aio.py:1756` — three competing "no operation" errors.** `generate_operation` calls `generate()`, so for any `BACKGROUND_MODEL` one of the two new raises fires first and this check never runs (the new test asserts exactly that). It stays live only for `define_model` LRO, so the same mistake reports different wording depending on how the plugin registered the model. JS has one such error (`js/ai/src/generate.ts:574`); Python now has three.

Two test cleanups not worth their own threads: `register_bg_model` (line 47) is `async def` but awaits nothing, so the five `await`s and the `BackgroundAction` import exist for a value no caller reads; and the two `assert 'did not return an operation' not in ...` checks (lines 167, 190) are tautological, because the enclosing `pytest.raises(match=...)` already proved which error fired.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -782,6 +788,21 @@ async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> Mode
...
+            if model_response.operation is None:
+                raise GenkitError(
+                    status='FAILED_PRECONDITION',
+                    message=(
+                        'wrap_model returned no operation for a background model; pass through response.operation'
+                    ),
+                )
+            if model_response.request is None:
+                model_response.request = request
+            return model_response
```
apascal07: **The early return skips `_persist_threaded_conversation`, so injected context becomes durable history.**

Every other exit from `run_one_iteration` (830, 853, 882) funnels through that helper. Its docstring states the policy this path bypasses: the model-facing request "carries per-call extras (docs/format injection, middleware edits) we don't want in saved history."

Reproduced: `ai.generate(model=<background>, prompt='render a cat', docs=[Document(...'SECRET-CONTEXT-DOC')])` returns `response.request.messages` containing the injected `Use the following information to complete your task: - [0]: SECRET-CONTEXT-DOC` block. The identical call against a `define_model` model returns the clean prompt.

This is not cosmetic: `persist_turn_messages` (`_ai/_agents/_runtime.py:1102`) reads `response.request.messages` and writes them to the session store via `set_messages(clean)`. An agent on a background model persists injected RAG and format text into durable history, then re-sends it next turn.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -517,9 +517,16 @@ def assert_valid_schema(self) -> None:
...
+        """Compare responses by message, finish_reason, and operation.
+
+        Two start handles with different job ids are not the same response.
+        """
         if isinstance(other, ModelResponse):
-            return self.message == other.message and self.finish_reason == other.finish_reason
+            return (
+                self.message == other.message
+                and self.finish_reason == other.finish_reason
+                and self.operation == other.operation
```
apascal07: **Comparing `operation` makes equality wall-clock dependent, and it can now raise.**

`wrapped_start` stamps `op.metadata['latencyMs']` on every start (`_core/_background.py:262`), and `Operation.__eq__` is pydantic field-wise. Two starts of the *same* job compare unequal:

```
op1.metadata = {'latencyMs': 0.166}
op2.metadata = {'latencyMs': 0.0022}
r1 == r2  ->  False
```

The docstring promises that only differing job ids make responses differ. The new test passes only because it builds bare `Operation(id='unique-a')` objects with `metadata=None`, never a real response.

`Operation.output` is also typed `Any`, so `==` can raise instead of returning a bool. With an operation carrying a numpy array: `ValueError: The truth value of an array with more than one element is ambiguous`.

Worth noting: deleting this `__eq__` override entirely leaves 1257 tests green (genkit + openai + middleware), because pydantic's default equality already compares `operation`. If the override stays, compare `operation.id` rather than the whole object, which is what the docstring actually describes.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -782,6 +788,21 @@ async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> Mode
                 next_fn,
             )
 
+        # A background start is a poll handle, not a conversation turn.
+        # define_model LRO replies that also carry a message still go
+        # through persist and the tool loop.
+        if model.kind == ActionKind.BACKGROUND_MODEL:
```
apascal07: **`resume_respond` against a background model is accepted, billed, and then discarded.**

`_resolve_resume_options` runs before this gate, so the resume path completes and only then hits the early return.

Reproduced with `ai.generate(model=<background>, messages=[user, interrupted-tool-request], resume_respond=[ToolResponsePart(...)])`:

```
start() saw roles: [USER, MODEL, TOOL]
billed start() calls = 1
response.message = None | operation = bg-op-123
```

The synthesized TOOL message is folded into the request handed to `start()`, which a video model ignores. A fresh render job is charged, the interrupt is consumed, and because the early return skips `_persist_threaded_conversation` nothing records that it was resolved.

A background start can never satisfy an interrupt resume. Reject it up front rather than running halfway and charging for it.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -782,6 +788,21 @@ async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> Mode
...
+            if model_response.operation is None:
```
apascal07: **This demands an operation from every response leaving the chain, not just from responses that came from `start()`.**

That forbids a whole class of legitimate middleware. The shipped `Fallback` is the concrete case:

`ai.generate(model=<background>, use=[Fallback(models=['alt-model'])])` where the start raises `UNAVAILABLE: veo capacity exhausted`. Fallback resolves the alternate, gets a good `ModelResponse`, and returns it. This check then raises `FAILED_PRECONDITION: wrap_model returned no operation...`. The alternate's answer is thrown away and the `UNAVAILABLE` cause is gone from the final message.

The same hard failure hits a cache middleware serving an already-rendered video, and a replay/mock middleware in tests. JS has no equivalent constraint.

Related: `_fallback.py:92` returns `result.response` un-boxed, and `_resolve_fallback_model` resolves only `ActionKind.MODEL`, so Fallback can never chain to another background model either.

Consider firing only when `next_fn` actually produced an operation that a hook then dropped.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -766,14 +767,19 @@ async def run_one_iteration(
...
                 await model.run(
                     input=params.request,
                     context=c.custom_context,
                     on_chunk=c.on_chunk,
                     abort_signal=c.abort_signal,
                 )
             ).response
+            # start() returns a poll handle. wrap_model is documented as
+            # seeing a ModelResponse, so wrap before the hook runs.
+            if isinstance(raw, Operation):
```
apascal07: **A start action that returns anything other than an `Operation` escapes as a bare `AttributeError`.**

This is narrower than the codebase's own `_ensure_operation` (`_core/_background.py:154`), which handles `Operation`, `dict`, and raises a typed error otherwise. It exists because a start action's `.response` can arrive as a dict.

Reproduced with a plugin-style action registered directly via `registry.register_action(kind=BACKGROUND_MODEL, ...)` — the shape `genkit_google_genai/google.py:397` uses, which has no `wrapped_start` coercion — whose fn returns `None`:

```
AttributeError: 'NoneType' object has no attribute 'operation'
```

from line 795. No status, no `GenkitError`, nothing actionable. A dict return behaves the same. So `ai.generate()` and `BackgroundAction.start()` now diverge on identical input.

Reusing `_ensure_operation` here would fix it. The deeper cause is `_registry.py:779`, which casts a BACKGROUND_MODEL start action to `Action[..., ModelResponse, ...]` — that false cast is what forces an isinstance sniff at this layer at all.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -782,6 +788,21 @@ async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> Mode
...
 
+        # A background start is a poll handle, not a conversation turn.
+        # define_model LRO replies that also carry a message still go
+        # through persist and the tool loop.
+        if model.kind == ActionKind.BACKGROUND_MODEL:
+            if model_response.operation is None:
+                raise GenkitError(
+                    status='FAILED_PRECONDITION',
+                    message=(
+                        'wrap_model returned no operation for a background model; pass through response.operation'
```
apascal07: **This blames `wrap_model` even when no middleware is installed.**

The branch also fires when the background action itself returned a response without an operation.

Reproduced: register a `BACKGROUND_MODEL` action whose fn returns `ModelResponse(message=...)`, then call `ai.generate(model=..., prompt='x')` with `use` unset. The user gets:

```
FAILED_PRECONDITION: wrap_model returned no operation for a background model; pass through response.operation
```

and goes looking for a middleware chain that is empty. The plugin's start fn is the culprit.

This inverts the goal stated in the PR description. The older model-blaming wording (`_aio.py:1759`, `Model 'X' did not return an operation`) is the accurate one for this case, and the new tests assert it is absent. Distinguishing "the action returned a non-Operation" from "middleware ran and dropped it" would let each message name the real offender.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -891,7 +912,13 @@ def message_parser(msg: Message) -> Any:  # noqa: ANN401
         iteration=current_turn,
         message_index=chunks.message_index,
     )
-    return await dispatch_generate(generate_params, run_ctx, run_one_iteration)
+    response = await dispatch_generate(generate_params, run_ctx, run_one_iteration)
+    if model.kind == ActionKind.BACKGROUND_MODEL and response.operation is None:
```
apascal07: **Both `model.kind` gates read a model that middleware may have replaced.**

`model` is bound once by `resolve_parameters` (line 665) and never re-derived from `turn_options.model`, even though `turn_tools` is re-resolved per turn and the comment at 758-759 claims `params.options` edits "aren't lost".

Reproduced with a `wrap_generate` that sets `params.options.model`:

```
bg   -> reroute to 'plain':  still ran the background model (operation returned, text '')
plain -> reroute to 'bg':    still ran the plain model (text 'PLAIN', operation None)
```

So model-routing middleware is silently a no-op. That is pre-existing, but these gates now depend on it: a reroute away from a background model would raise a spurious FAILED_PRECONDITION here, and a reroute toward one would take the non-background path and fall into `response.request = request` on an operation-only response.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -782,6 +788,21 @@ async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> Mode
...
+        # through persist and the tool loop.
+        if model.kind == ActionKind.BACKGROUND_MODEL:
+            if model_response.operation is None:
+                raise GenkitError(
+                    status='FAILED_PRECONDITION',
+                    message=(
+                        'wrap_model returned no operation for a background model; pass through response.operation'
+                    ),
+                )
+            if model_response.request is None:
```
apascal07: **This conditional repair diverges from the normal path, and it cannot fix an already-primed `cached_property`.**

Two problems.

(a) `next_fn` stamps `request=params.request` (the post-`wrap_model` request), so this guard almost never fires and the middleware's request wins. Line 813 on the normal path is unconditional and pins the pre-middleware `request`. Reproduced with a `wrap_model` that rewrites `params.request.messages` to `['REWRITTEN']`: the background response reflects the rewrite, the normal model's does not. A redaction or logging middleware's in-place edits therefore surface to the user for background models only.

(b) `messages`, `text`, and `output` are `cached_property`. A `wrap_model` that rebuilds the response without `request` and reads `resp.messages` before returning latches the cache. Assigning `request` afterwards does not invalidate it. Reproduced: `response.request.messages` has 1 entry while `response.messages == []` and `response.text == ''`.

So the line looks like it repairs the drop but does not. Pick one provenance rule and apply it unconditionally.

py/packages/genkit/tests/genkit/ai/background_generate_test.py
```diff
@@ -0,0 +1,241 @@
...
+        name='lro-model',
+        fn=model_fn,
+        info=ModelInfo(supports=Supports(long_running=True, tools=True)),
+    )
+
+    response = await ai.generate(model='lro-model', prompt='x', tools=['ping'])
+
+    assert tool_ran == 1
+    assert response.text == 'done'
+    assert response.operation is None
```
apascal07: **This assertion locks in a silent operation loss that orphans a started job.**

A `define_model` LRO returning an operation alongside a tool request loses the handle when the tool loop runs another turn: turn 2's response replaces turn 1's.

The user-visible consequence:

```
await ai.generate_operation(model='lro-model', prompt='x', tools=['ping'])
-> FAILED_PRECONDITION: Model 'lro-model' did not return an operation.
```

The model did return one. Job `lro-1` keeps running with no handle to poll or cancel.

The gate at `_generate.py:794` deliberately excludes this path, so the one case where an operation is genuinely dropped is the one left unguarded, and this test ratifies it. Carrying the last non-`None` operation forward would fix it. At minimum the assertion deserves a comment saying the drop is deliberate.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -766,14 +767,19 @@ async def run_one_iteration(
...
                     input=params.request,
                     context=c.custom_context,
                     on_chunk=c.on_chunk,
                     abort_signal=c.abort_signal,
                 )
             ).response
+            # start() returns a poll handle. wrap_model is documented as
+            # seeing a ModelResponse, so wrap before the hook runs.
+            if isinstance(raw, Operation):
+                return ModelResponse(operation=raw, request=params.request)
```
apascal07: **The boxed response never gets `latency_ms`, although the framework already measured it.**

`_record_latency` (`_core/_action.py:52`) is `if hasattr(output, 'latency_ms')`. `Operation` is `extra='forbid'` with no such field, so the measurement is taken and then dropped.

```
background: response.latency_ms = None   (operation.metadata = {'latencyMs': 0.0019})
normal:     response.latency_ms = 0.0152
```

Telemetry and the Dev UI see a null model latency for every background start. Forwarding `raw.metadata['latencyMs']` into `ModelResponse.latency_ms` at this line fixes it.

apascal07: **An operation that comes back already failed is boxed into an apparent success.**

`start()` returning `Operation(id='op-failed', done=True, error=Error(message='quota exceeded'))` yields:

```
text = ''   message = None   finish_reason = None   (no exception)
```

A caller reading `.text` or `.output` sees empty success. The failure is reachable only through `.operation.error`.

JS boxes the same way (`js/ai/src/generate/action.ts:485`), so this is a shared gap rather than a new divergence. Worth noting though that Python has no backstop at all here, because `ModelResponse.assert_valid` is still a `pass` stub (TODO #4343), whereas JS raises in `GenerateResponse` when there is neither a message nor an operation.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -782,6 +788,21 @@ async def next_fn(params: ModelHookParams, c: GenerateMiddlewareContext) -> Mode
                 next_fn,
             )
 
+        # A background start is a poll handle, not a conversation turn.
```
apascal07: **This block is not load-bearing. It buys one word in one error message.**

I removed lines 791-804 and ran the suite: 1033 tests in `packages/genkit/tests` pass, and 7 of the 8 new tests in `background_generate_test.py` pass. The only failure is a wording assertion:

```
Expected regex: 'wrap_model returned no operation'
Actual message: 'FAILED_PRECONDITION: wrap_generate returned no operation ...'
```

The outer check at 916 already catches the missing-operation case, and the rest of the pipeline is inert for an operation-only response (`assert_valid` is a stub, and `_message_parser`/`_schema_type` are unused when `message is None`). So 14 lines of `kind`-keyed special case sit permanently in the shared generate loop, and the inner raise duplicates the outer one.

There is a cost beyond the duplication: when #4343 lands and `assert_valid` becomes real, the background path silently bypasses it. JS puts this rule in `GenerateResponse` (`js/ai/src/generate/response.ts:110`, `if (!this.message && !this.operation) throw`), which is the layer that generalizes.

Also worth recording in the PR description: JS boxes *after* the middleware chain (`js/ai/src/generate/action.ts:483`), so a `wrap_model` ported from JS sees an `Operation` there and a `ModelResponse` here. Boxing before the chain is the better choice, but it is a deliberate parity break.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Gone. message is None case hits _persist_threaded_conversation like every other exit. Injected RAG is not on response.request.messages.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Agreed, a start cannot finish a tool interrupt. We reject up front now. Start() shouldn't be called.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: That wording is gone. If the start fn returns no handle, the error names the model. We didn't add a second 'middleware dropped it' string, generate() lets a hook replace the response; generate_operation is the peel.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: I removed the kind gates that made this behavior worse. Good point that model re-routing is silently a no-op. I'll track that as a separate issue.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Update to forward at the box. Start timing lands on response.latency_ms now.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Deleted. Persist / log / assert_valid  now run. We also dropped the outer duplicate (one path, no kind-keyed return). Boxing before wrap_model is deliberate; the hook is documented as seeing a ModelResponse.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: JS boxes the same way (finish reason is left empty, use `.operation` to get the actual operation details. This is a shared gap / product call rather than a divergence.

py/packages/genkit/tests/genkit/ai/background_generate_test.py
huangjeff5: Interesting. I looked at JS behavior and it seems to also drop the operation in the hybrid case. 

Unless there is a good reason, I don't think define_model() + Operation should be a supported use case. If a user needs an operation, that's a long-running job and should be registered as a background model. 

So that's what we do now here. define_model + any operation raises in next_fn on turn 1 with: `Model '{name} is a regular model... Use define_background_model for background models that return operations.`

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Updated to check operation.id rather than the whole object.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Makes sense. Rather than coercing dict to Operation here, we enforce that the start fn has to return an Operation and we assert that through ensure_operation now. Now both paths return GenkitError.

We didn't change the resolve_model cast but I agree with you we should do that in an a followup PR.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: We dropped the gate in `generate`. If start raises and Fallback returns flash, generate() keeps that answer. 

In `generate_operation`, we still raise if there's no Operation returned.

huangjeff5:
_generate.py:824 — a background start never reaches the generate response debug log. _loggable_response is only called below the early return at 804. That is the call most worth logging, because the result arrives minutes later through a separate check_operation, and nothing ties the returned operation id back to the request that started it.

--> Background starts now hit the same generate response debug log.

_generate.py:742 — the resumed-tool chunk is streamed for a response that has no message. With resume_respond against a background model, generate_stream yields a TOOL chunk while the awaited response carries only an operation. A consumer accumulating chunks into history gets a turn that appears in no ModelResponse.messages and, because the early return skips _persist_threaded_conversation, is recorded nowhere else.

--> resume never reaches start() or the TOOL chunk. Rejected as a failed precondition.

_aio.py:1756 — three competing "no operation" errors. generate_operation calls generate(), so for any BACKGROUND_MODEL one of the two new raises fires first and this check never runs (the new test asserts exactly that). It stays live only for define_model LRO, so the same mistake reports different wording depending on how the plugin registered the model. JS has one such error (js/ai/src/generate.ts:574); Python now has three.

--> There is 1 shared helper now `missing_operation_error` that raises and contains the unified message.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: (a) The early-return if request is None is gone. Background starts now take the same path as a normal generate: we always set response.request to the request built before wrap_model. A hook that rewrites params.request.messages does not show up on the user's response. You get 'a cat', not 'REWRITTEN'.

(b) .messages is a normal @property now, not @cached_property. A hook that builds a ModelResponse without request and reads .messages first no longer sticks [] after we attach request. .text / .output are still cached (parse cost). wrap_model messages are params.request.messages; wrap_generate messages are params.options.messages (persist uses the latter).

py/packages/genkit/src/genkit/_core/_model.py
apascal07: The id comparison fixes the latency flake, but it now ignores the fields that change across a poll:

```
ModelResponse(operation=Operation(id='job1', done=False))
  == ModelResponse(operation=Operation(id='job1', done=True, output={...}))   -> True
```

Done vs errored on the same id is also True. A caller polling with `while resp == prev: resp = await ai.check_operation(...)` never exits.

Comparing `operation.id` and `done`, or dropping the override (which left the suite green), both avoid it.

py/packages/genkit/src/genkit/_ai/_generate.py
apascal07: The Fallback case works now, confirmed against the real middleware: start raises `UNAVAILABLE`, `generate()` returns the alternate's text.

Removing the gate outright also removed the drop case, and `test_generate_operation_fails_when_wrap_model_drops_operation` locks that in. After `DropsOperation`, `generate()` returns success with `operation=None` and `text=''` while `start()` has already billed a job. The id is gone, so the job cannot be polled or cancelled, and the caller's next step fails untyped:

```
ai.check_operation(response.operation)
-> AttributeError: 'NoneType' object has no attribute 'action'
```

That is what the narrow gate was for: fire only when `next_fn` produced an operation that a hook then dropped. It leaves Fallback alone and still catches this.

`_fallback.py` is unchanged, so the second half stands. `_resolve_fallback_model` resolves only `ActionKind.MODEL` (`_fallback.py:58`), and it is called outside the try at `:84`, so `Fallback(models=['some-bg-model'])` raises `NOT_FOUND` and masks the primary's `UNAVAILABLE`.

py/packages/genkit/src/genkit/_ai/_generate.py
apascal07: The start path is fixed. A start fn returning `None` or a dict now raises `FAILED_PRECONDITION` instead of `AttributeError`, and `ai.generate()` and `BackgroundAction.start()` agree on that input.

`_ensure_operation` is shared with `BackgroundAction.check/cancel`, though, so the strictness reaches past the start path. A check action registered directly (the `Action(kind=ActionKind.CHECK_OPERATION, fn=_check)` shape at `genkit_google_genai/google.py:407`) whose fn returns the parsed REST payload fails at the first poll:

```
ai.check_operation(op)
-> FAILED_PRECONDITION: Background model 'raw-bg/check' did not return an operation.
```

Same for `ai.cancel_operation`. JS coerces at this boundary: `js/core/src/action.ts:615` parses the response against the action's output schema, and `background-action.ts:61` declares that schema as `OperationSchema`. Restoring the dict branch fails no tests, so neither direction is covered.

Keeping the strictness at the start-boxing site and leaving the shared helper coercive gets both.

py/packages/genkit/src/genkit/_ai/_generate.py
apascal07: Naming the model is right for the generate path. The shared factory hardcodes "Background model", though, and two of its three call sites are not background models:

- `_background.py:135,152` pass `check_action.name` / `cancel_action.name`, so a check failure reads `Background model 'raw-bg/check' did not return an operation.` That is a CHECK_OPERATION action.
- `_aio.py:1758` reports `Background model 'plain-lro' did not return an operation.` for a `define_model` model, since `_model_supports_long_running` still admits MODEL kind. That also drops parity with `js/ai/src/generate.ts:574`, which still reads `Model '${...name}' did not return an operation.`

Passing the actor in, or keeping the model-neutral wording at the `_aio.py` site, covers both.

py/packages/genkit/src/genkit/_ai/_generate.py
apascal07: (a) Confirmed. `response.request` is the pre-`wrap_model` request for both kinds now, including when a hook returns a different response object.

(b) `messages` no longer latches, but the parse-cost argument for keeping the others cached does not extend to `interrupts`. `_generate.py:915` copies with `deep=False`, then `:918` reassigns `.message`, so any `wrap_model` that reads `resp.interrupts` leaves the caller with `finish_reason=INTERRUPTED`, the interrupt metadata still on the message, and `len(response.interrupts) == 0`:

```
no hook  : finish=interrupted  interrupts=1
with hook: finish=interrupted  interrupts=0
```

An app gated on `if response.interrupts:` skips the approval step and never learns why. `output` is the same shape: a hook whose body is `_ = resp.output` downgrades `response.output` from the typed model to a raw dict, because `_schema_type` is attached at `:858`, after the chain.

One footgun from the swap itself: `cached_property` is a non-data descriptor and `property` is a data descriptor, so `resp.model_copy(update={'messages': trimmed})` is silently ignored. The value lands in `__dict__` and `__pydantic_fields_set__`, then the descriptor shadows it on read, so the copy reports the full untrimmed history with no error. A `__setattr__` that pops the `messages` cache when `request` is assigned keeps the cache, avoids that, and still passes `test_model_response_messages_sees_request_set_after_first_read`.

py/packages/genkit/tests/genkit/ai/background_generate_test.py
apascal07: Fair call on `define_model` + Operation not being a product. Three things still say otherwise, so today the ban reads as a bug to anyone who hits it:

- `_aio.py:151 _model_supports_long_running` still returns True for a MODEL action declaring `supports.longRunning`, so `generate_operation` accepts the model and the loop then tells the caller it is a regular model.
- `docs/model-spec.md:31` still documents `longRunning` under `Action Type: model`.
- JS still implements it. `js/ai/src/generate.ts:574` gates on `supports.longRunning` and is kind-agnostic.

The guard also ignores `done`, so a MODEL action returning a complete usable response alongside a finished operation is rejected. That case has no test.

Rejecting at `define_model` registration time instead names the registration site, lets the MODEL branch of `_model_supports_long_running` be deleted rather than left as a trap, and costs nothing per call.

py/packages/genkit/src/genkit/_ai/_generate.py
apascal07: Works for `define_background_model`. It does not reach the shipping shape.

The forward reads `operation.metadata['latencyMs']`, which only `wrapped_start` writes, and Veo builds its start action directly (`genkit_google_genai/google.py:397`, no `wrapped_start`). That model still reports `latency_ms=None` with `op.metadata=None`. `test_generate_forwards_start_latency` covers only the `define_background_model` fake, so the gap is invisible.

`:661-662` is also dead. `Action.run` runs `_record_latency` (`_core/_action.py:790`), which stamps `latency_ms` on any output that has the attribute, so `raw.latency_ms` is never None by the time boxing sees it (0 of 5 arrivals across streaming, non-streaming, and an explicit `latency_ms=None` return). The line also writes through to the caller's object.

`_record_latency` is where the measurement already exists, and it is correct for every registration shape.

apascal07: Agreed on the product call.

For the record: the background path now runs `assert_valid()` (`_generate.py:863`), so #4343 closes this in Python with no further plumbing. JS returns at `action.ts:497` before its own `assertValid`, so Python ends up ahead here rather than behind.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -636,6 +638,49 @@ def handler(chunk: ModelResponseChunk[Any]) -> None:
...
+    return None
+
+
+def box_background_start(*, raw: object, request: ModelRequest, name: str) -> ModelResponse:
+    """Turn a background start into the ModelResponse wrap_model reads.
+
+    start() is a poll handle. Already-boxed responses pass through; a chat
+    turn with no handle is a model-author mistake.
+    """
+    if isinstance(raw, ModelResponse):
```
apascal07: An already-boxed response passes through with `message` intact, and the only background exit from the loop is `if generated_msg is None` at `:866`. So a start that returns a message keeps running the tool loop.

With `ModelResponse(operation=Operation(id='job-1'), message=<toolRequest ping>)`, the shape `test_background_start_already_boxed_model_response` legalizes:

```
ai.generate(model='boxed-bg', prompt='a cat', tools=['ping'])
-> GenerationResponseError: Exceeded maximum tool call iterations (5)
   start() billed 6 times
```

Six real render jobs, and the caller gets an exception instead of a handle. JS returns unconditionally right after boxing (`js/ai/src/generate/action.ts:493`), before tool scanning and before `assertValid`.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -636,6 +638,49 @@ def handler(chunk: ModelResponseChunk[Any]) -> None:
...
+        if raw.operation is None:
+            raise missing_operation_error(name=name)
+        if raw.latency_ms is None:
+            raw.latency_ms = _latency_ms_from_operation(raw.operation)
+        return raw
+    op = _ensure_operation(response=raw, name=name)
+    return ModelResponse(
```
apascal07: `name` is in hand here, but `operation.action` is never set, so a handle from a start fn that does not stamp it itself cannot be polled. With this PR's own `_register_raw_background` fixture:

```
generate()                          -> Operation(action=None, id='boxed-1')
ai.check_operation(resp.operation)  -> ValueError: Provided operation is missing original request information
```

`test_background_start_boxes_raw_action_operation` asserts only the id, so it passes on a handle that is dead on arrival. `wrapped_start` (`_background.py:272`) and `google.py:386` both set `op.action` by hand, which is the duplication this layer could absorb.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -766,14 +820,18 @@ async def run_one_iteration(
...
                 await model.run(
                     input=params.request,
                     context=c.custom_context,
                     on_chunk=c.on_chunk,
                     abort_signal=c.abort_signal,
                 )
             ).response
+            if model.kind == ActionKind.BACKGROUND_MODEL:
+                return box_background_start(raw=raw, request=params.request, name=model.name)
+            assert_foreground_model_has_no_operation(raw=raw, name=model.name)
```
apascal07: The guard inspects only `Operation` and `ModelResponse`, and `cast` is a runtime no-op, so a plain model returning anything else still escapes untyped. A model fn returning `{'operation': {'id': 'op-1', 'done': False}}` (the wire shape) or `None` surfaces at `:854` as:

```
AttributeError: 'dict' object has no attribute 'request'
```

No status, no model name, no hint. That is the same defect class the line above now routes through `_ensure_operation` for background models.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -119,7 +120,7 @@ async def start(
             An Operation with an ID to track the job.
         """
         result = await self.start_action.run(input)
-        return _ensure_operation(result.response)
+        return _ensure_operation(response=result.response, name=self.start_action.name)
```
apascal07, line 130: `box_background_start` accepts a start fn returning a boxed `ModelResponse` and this path rejects it, so one registered action behaves differently by entry point:

```
ai.generate()                 -> OK, operation='boxed-1'
BackgroundAction.start(req)   -> FAILED_PRECONDITION: Background model 'boxed-bg' did not return an operation.
```

The message is false: it did return one. `py/samples/google-genai-media/src/main.py:128` calls `action.start(...)` directly, so it is the failing path that ships.

py/packages/genkit/tests/genkit/ai/background_generate_test.py
```diff
@@ -0,0 +1,375 @@
...
+        next_fn: Callable[[ModelHookParams, GenerateMiddlewareContext], Awaitable[ModelResponse]],
+    ) -> ModelResponse:
+        resp = await next_fn(params, ctx)
+        return ModelResponse(message=resp.message, finish_reason=resp.finish_reason)
+
+
+@pytest.mark.asyncio
+async def test_generate_operation_fails_when_wrap_model_drops_operation(ai: Genkit) -> None:
+    """generate() stays quiet; generate_operation is the one missing-handle error."""
+    register_bg_model(ai)
```
apascal07, line 156: `start()` has already billed a job at this point and the handle is gone, so this asserts an orphaned job. The caller's next step fails untyped:

```
ai.check_operation(response.operation)
-> AttributeError: 'NoneType' object has no attribute 'action'
```

If the drop is deliberate the assertion needs a comment saying so. See the `wrap_model returned no operation` thread for the narrower gate that keeps Fallback working and still catches this.

py/packages/genkit/tests/genkit/ai/generate_operation_test.py
```diff
@@ -292,5 +236,4 @@ async def model_fn(request: ModelRequest, ctx: ActionRunContext) -> ModelRespons
     )
 
     assert captured_request is not None
-    # Verify config was passed
     assert captured_request.config is not None
```
apascal07: This cannot fail. `action_to_generate_request` sets `config=options.config if options.config is not None else {}`, so `captured_request.config` is never None.

The forwarding block it is meant to cover is unguarded as a result. Replacing 13 of the 17 kwargs `generate_operation` passes to `self.generate()` with `None` (system, messages, tools, tool_choice, max_turns, context, output_schema, output_format, output_content_type, output_instructions, output_constrained, return_tool_requests, docs) leaves 1064 passed and 2 failed, and both failures are unrelated tests about dropped operations. `system=None` alone is fully green at 1066 passed.

apascal07 (commented):
Two items with no line to anchor to.

`_ai/_agents/_runtime.py:1060` gates persistence on `if response.message:`, so an agent on a background model returns the caller's own prompt as `out.text`, with `finish_reason=None` and the started job unreachable. `AgentResult` has no `operation` field, and the id appears nowhere in the result or the stream. The job is billed and then lost.

`define_background_model` and `generate_operation` appear in no doc and no sample. The one Veo sample drives `BackgroundAction.start/check` directly, which is the API this PR is trying to make unnecessary.

huangjeff5:
Two items from the review body:

**Agent persist.** Not a product — `AgentResult` has no place for a handle in any SDK, and agents don't sit on background models. Same persist-on-message shape in JS. Out of scope here.

**Docs / Veo sample.** Fair. Docs and the Veo sample still sit on `BackgroundAction`. Out of scope for this PR — I'll do that in a follow-up.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Updated — `==` is the poll snapshot (`id`, `done`, `error`, `output`). Timing on the handle is still ignored.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Accepted. `generate()` now raises if `start()` produced a handle and a hook dropped it. Fallback still works — `start()` never boxed, so the gate stays quiet.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: We're consistent: start / check / cancel must return an `Operation`. A dict is a plugin bug on every door — we raise, we don't coerce.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Softened to `'{name}' did not return an operation.` The name is the action that failed. No kind in the sentence.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: All of those views are live now. A hook that reads them before `message` / schema is attached no longer freezes the empty value.

py/packages/genkit/tests/genkit/ai/background_generate_test.py
huangjeff5: `define_model` raises if you set `longRunning`. `generate_operation` only takes a background model.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Timing comes from `Action.run` now. The boxed response gets `result.latency_ms` for every registration shape.

huangjeff5: Agreed — no change.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Updated — if the boxed start has a handle, we return it. No tool loop.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: The box now stamps `operation.action`. A handle from `generate()` can be polled.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: Updated — a chat model must return a `ModelResponse`. A dict/`None` is a plugin bug — we raise.

py/packages/genkit/src/genkit/_core/_background.py
huangjeff5: Updated — `start()` returns an `Operation`. A `ModelResponse` is a plugin bug on both doors. `generate()` is what boxes.

py/packages/genkit/tests/genkit/ai/background_generate_test.py
huangjeff5: `generate()` now raises when a hook drops a billed ticket. Same as `generate_operation`.

py/packages/genkit/tests/genkit/ai/generate_operation_test.py
huangjeff5: Tightened the test to assert the forwarded `config` and system prompt.

apascal07 (approved):

PR #6131, opened 2026-08-20
feat(py): require a live Operation on check and cancel

huangjeff5:
## Summary
- `check_operation` / `cancel_operation` take a live `Operation`. A persist dump is `Operation.model_validate(dumped)` first. Extra keys (`latencyMs`) are ignored so that validate does not 500.
- Published check/cancel input schema stays one `Operation` (`extra='ignore'`). Not `Operation | dict`.
- A raw dump, a boxed `ModelResponse`, or a `str` is `INVALID_ARGUMENT` (not `AttributeError`). Messages: `got a dump; pass Operation.model_validate(...)`, `got ModelResponse; pass response.operation`, `got str, expected Operation`.
- The same gate runs on `BackgroundAction.check` / `BackgroundAction.cancel`, so a dump cannot `AttributeError` on `.action` before `UNIMPLEMENTED`.
- Missing cancel fn is `UNIMPLEMENTED` on both the veneer and a real `BackgroundAction.cancel`. Unknown action is `INVALID_ARGUMENT`.

### DIFF
--- a/py/packages/genkit/src/genkit/_core/_background.py
+++ b/py/packages/genkit/src/genkit/_core/_background.py
@@ -18,7 +18,6 @@
 
 from __future__ import annotations
 
-import time
 from collections.abc import Awaitable, Callable, Mapping
 from typing import Any, Generic, TypeVar
 
@@ -263,16 +262,10 @@ def define_background_model(
     output_schema_meta = to_json_schema(ModelResponse)
     model_meta['outputSchema'] = output_schema_meta
 
-    # Wrap the start function to add the action key and timing
     async def wrapped_start(request: ModelRequest, ctx: ActionRunContext) -> Operation:
-        start_time = time.perf_counter()
         op = await start(request, ctx)
-        # Set action key in format: /{action_type}/{name}
+        # The handle needs this key so check/cancel can find the job later.
         op.action = action_key
-        latency_ms = (time.perf_counter() - start_time) * 1000
-        if op.metadata is None:
-            op.metadata = {}
-        op.metadata['latencyMs'] = latency_ms
         return op
 
     # Wrap the check function (no ctx parameter)
--- a/py/packages/genkit/src/genkit/_core/_background.py
+++ b/py/packages/genkit/src/genkit/_core/_background.py
@@ -135,6 +135,10 @@ class BackgroundAction(Generic[OutputT]):
 
         Returns:
             Updated Operation with current status.
+
+        Raises:
+            GenkitError: INVALID_ARGUMENT if ``operation`` is not a live
+                ``Operation`` (e.g. a dump or a ``ModelResponse``).
         """
         operation = require_operation(value=operation)
         result = await self.check_action.run(operation)
@@ -150,9 +154,13 @@ class BackgroundAction(Generic[OutputT]):
             Updated Operation reflecting cancellation attempt.
 
         Raises:
-            GenkitError: If this action does not implement cancel.
+            GenkitError: UNIMPLEMENTED if this action does not implement
+                cancel, INVALID_ARGUMENT if ``operation`` is not a live
+                ``Operation``.
         """
         operation = require_operation(value=operation)
+        # Raising here is deliberate: returning the operation unchanged would
+        # make "this model can't cancel" indistinguishable from "cancelled".
         if self.cancel_action is None:
             raise GenkitError(
                 status='UNIMPLEMENTED',
@@ -387,7 +395,7 @@ def require_operation(*, value: object) -> Operation:
 async def resolve_operation_action(
     registry: Registry,
     operation: Operation,
-) -> tuple[Operation, BackgroundAction]:
+) -> BackgroundAction[ModelResponse]:
     """Turn a poll handle into the background action that owns it."""
     operation = require_operation(value=operation)
     if not operation.action:
@@ -396,13 +404,21 @@ async def resolve_operation_action(
             message='Provided operation is missing original request information',
         )
 
-    background_action = await lookup_background_action(registry, operation.action)
+    try:
+        background_action = await lookup_background_action(registry, operation.action)
+    except ValueError as e:
+        # operation.action is caller data (often reloaded from storage), so a
+        # mangled key is the caller's bad argument, not an internal failure.
+        raise GenkitError(
+            status='INVALID_ARGUMENT',
+            message=f'Failed to resolve background action from original request: {operation.action}',
+        ) from e
     if background_action is None:
         raise GenkitError(
             status='INVALID_ARGUMENT',
             message=f'Failed to resolve background action from original request: {operation.action}',
         )
-    return operation, background_action
+    return background_action
 
 
 async def check_operation(
@@ -422,8 +438,8 @@ async def check_operation(
         GenkitError: If the handle is missing action, or the action is
             not found.
     """
-    resolved, background_action = await resolve_operation_action(registry, operation)
-    return await background_action.check(resolved)
+    background_action = await resolve_operation_action(registry, operation)
+    return await background_action.check(operation)
 
 
 async def cancel_operation(
@@ -441,12 +457,8 @@ async def cancel_operation(
 
     Raises:
         GenkitError: If the handle is missing action, the action is not
-            found, or cancel is not implemented.
+            found, or cancel is not implemented (UNIMPLEMENTED, raised by
+            ``BackgroundAction.cancel``).
     """
-    resolved, background_action = await resolve_operation_action(registry, operation)
-    if not background_action.supports_cancel:
-        raise GenkitError(
-            status='UNIMPLEMENTED',
-            message=f'Background action {resolved.action} does not support cancellation.',
-        )
-    return await background_action.cancel(resolved)
+    background_action = await resolve_operation_action(registry, operation)
+    return await background_action.cancel(operation)
--- a/py/packages/genkit/src/genkit/_core/_background.py
+++ b/py/packages/genkit/src/genkit/_core/_background.py
@@ -19,10 +19,8 @@
 from __future__ import annotations
 
 import time
-from collections.abc import Awaitable, Callable
+from collections.abc import Awaitable, Callable, Mapping
 from typing import Any, Generic, TypeVar
-
-from pydantic import BaseModel
 
 from genkit._core._action import Action, ActionKind, ActionRunContext
 from genkit._core._error import GenkitError
@@ -138,23 +136,28 @@ class BackgroundAction(Generic[OutputT]):
         Returns:
             Updated Operation with current status.
         """
+        operation = require_operation(value=operation)
         result = await self.check_action.run(operation)
         return _ensure_operation(response=result.response, name=self.check_action.name)
 
     async def cancel(self, operation: Operation) -> Operation:
         """Cancel a background operation.
 
-        If cancellation is not supported, returns the operation unchanged.
-
         Args:
             operation: The operation to cancel.
 
         Returns:
             Updated Operation reflecting cancellation attempt.
+
+        Raises:
+            GenkitError: If this action does not implement cancel.
         """
+        operation = require_operation(value=operation)
         if self.cancel_action is None:
-            # Return operation unchanged if cancel not supported
-            return operation
+            raise GenkitError(
+                status='UNIMPLEMENTED',
+                message=f'Background action {operation.action} does not support cancellation.',
+            )
         result = await self.cancel_action.run(operation)
         return _ensure_operation(response=result.response, name=self.cancel_action.name)
 
@@ -172,24 +175,6 @@ def _ensure_operation(*, response: object, name: str) -> Operation:
     if isinstance(response, Operation):
         return response
     raise missing_operation_error(name=name)
-
-
-class DefineBackgroundModelOptions(BaseModel):
-    """Options for defining a background model.
-
-    Attributes:
-        name: Unique name for this background model.
-        label: Human-readable label (defaults to name).
-        versions: Known version names for this model.
-        supports: Model capability information.
-        config_schema: Custom options schema for this model.
-    """
-
-    name: str
-    label: str | None = None
-    versions: list[str] | None = None
-    supports: dict[str, Any] | None = None
-    config_schema: type | dict[str, Any] | None = None
 
 
 def define_background_model(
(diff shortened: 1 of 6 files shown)


py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -365,29 +385,113 @@ async def lookup_background_action(
...
+    generate() ModelResponse is not a handle — pass response.operation.
+    Extra keys from a dump are ignored so a persist/reload does not 500.
+    """
+    if isinstance(value, ModelResponse):
+        raise GenkitError(
+            status='INVALID_ARGUMENT',
+            message='got ModelResponse; pass response.operation',
+        )
+    if isinstance(value, Operation):
+        return Operation.model_validate(value.model_dump())
```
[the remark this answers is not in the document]
huangjeff5: The round-trip is intentional: a live handle is copied and leftover keys are stripped the same way as a persist dump, so check/cancel see one shape. This is once per poll on a small object, so we are leaving the copy in.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -365,29 +385,113 @@ async def lookup_background_action(
...
+        status='INVALID_ARGUMENT',
+        message=f'got {type(value).__name__}, expected Operation | Mapping',
+    )
+
+
+def is_operation_field(key: str) -> bool:
+    fields = Operation.model_fields
+    if key in fields:
+        return True
+    return any(field.alias == key for field in fields.values())
```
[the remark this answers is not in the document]
huangjeff5: Operation only has a handful of fields, so a scan per check_operation is not worth a module-level set (and the set would have to stay in sync with the generated model). Leaving the explicit field/alias check.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -379,29 +364,89 @@ async def lookup_background_action(
...
+) -> tuple[Operation, BackgroundAction]:
+    """Turn a poll handle into the background action that owns it."""
+    operation = require_operation(value=operation)
+    if not operation.action:
+        raise GenkitError(
+            status='INVALID_ARGUMENT',
+            message='Provided operation is missing original request information',
+        )
+
+    background_action = await lookup_background_action(registry, operation.action)
```
apascal07: `operation.action` is caller data, and a malformed key escapes as a bare `ValueError` rather than the `GenkitError` this function documents.

`parse_action_key` inside `resolve_action_by_key` raises on a bad format or an unknown kind (`_registry.py:656-658`). Against a real registry:

- `action='missing'` gives `ValueError: Invalid action key format`
- `action='/not-a-kind/foo'` gives `ValueError: Invalid action kind`
- `action='/background-model'` gives `ValueError: Invalid action key format`

Same on both `check_operation` and `cancel_operation`. The docstrings below promise `GenkitError`, and the PR body promises "Unknown action is `INVALID_ARGUMENT`". Over the reflection layer a `GenkitError` maps through `_STATUS_CODE_MAP` to a 400; a bare `ValueError` becomes a 500.

The gap sits exactly between the two guards that were converted: the `not operation.action` check above and the `background_action is None` check below.

py/packages/genkit/src/genkit/_core/_typing.py
```diff
@@ -559,7 +559,7 @@ class MultipartToolResponse(GenkitModel):
 class Operation(GenkitModel):
     """Model for operation data."""
 
-    model_config: ClassVar[ConfigDict] = ConfigDict(alias_generator=to_camel, extra='forbid', populate_by_name=True)
+    model_config: ClassVar[ConfigDict] = ConfigDict(alias_generator=to_camel, extra='ignore', populate_by_name=True)
```
apascal07: `extra='ignore'` converts a malformed payload from a hard failure into a silent field drop, and `done` is the field every poll loop reads.

`Operation.model_validate({'id': 'x', 'Done': True, 'doneX': True, 'out_put': {'a': 1}})` now returns `Operation(action=None, id='x', done=None, output=None, error=None, metadata=None)`. Three keys dropped, no error. `extra='forbid'` raised.

The check action validates wire and persisted input through `TypeAdapter(Operation).validate_python` (`_action.py:729`), so a peer sending `Done` yields `done=None`, and a caller polling `while not operation.done` (`py/samples/google-genai-media/src/main.py:110`) spins to its own 180s timeout instead of failing with the bad key named.

If the goal is to accept a persisted dump, having the veneer take a `Mapping` and call `Operation.model_validate` itself keeps the loud failure for a genuinely malformed handle.

apascal07: This also drops `additionalProperties: false` from the published input schema of every check-operation and cancel-operation action.

`TypeAdapter(Operation).json_schema()` now emits no `additionalProperties` key at all. That adapter is what `_action.py:632-636` publishes as `ActionMetadataKey.INPUT_KEY` for `wrapped_check` and `wrapped_cancel`, so the Dev UI and any reflection consumer are told the input is open.

`genkit-tools/genkit-schema.json` still declares `Operation` with `"additionalProperties": false`, and it is the cross-language source of truth. Anything validating a Python-published action schema against the repo schema now sees two contracts for one type. Worth a line in the PR body either way.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
```diff
@@ -78,8 +81,189 @@ async def test_genkit_check_operation_not_found() -> None:
     op = Operation(id='123', done=False, action='missing')
     ai.registry.resolve_action_by_key = AsyncMock(return_value=None)  # type: ignore[assignment]
```
apascal07: This mock removes the only call that would surface the bug the test's own input triggers.

`parse_action_key('missing')` lives inside `resolve_action_by_key` (`_registry.py:660`) and raises `ValueError: Invalid action key format` before any `None` return. Stubbing the method skips it, so the test asserts `GenkitError` on a path that raises `ValueError` unmocked.

A case with a real registry and a syntactically invalid action key would pin the gap in `resolve_operation_action`.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -138,23 +136,28 @@ async def check(self, operation: Operation) -> Operation:
...
             operation: The operation to cancel.
 
         Returns:
             Updated Operation reflecting cancellation attempt.
+
+        Raises:
+            GenkitError: If this action does not implement cancel.
         """
+        operation = require_operation(value=operation)
         if self.cancel_action is None:
```
apascal07, line 163: JS keeps `BackgroundAction.cancel` lenient. `js/core/src/background-action.ts:166-173` returns the operation unchanged when `cancelAction` is unset, and the throw lives one layer up in `js/ai/src/cancel-operation.ts:41-46`.

This PR converges the veneer with JS and diverges the wrapper in the same commit. Two smaller mismatches alongside it: the JS cancel action body throws `UNAVAILABLE` (`background-action.ts:361-364`) where this uses `UNIMPLEMENTED`, and JS has no argument-type gate.

If the wrapper divergence is deliberate, a comment saying so keeps the next parity pass from reverting it.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -379,29 +364,89 @@ async def lookup_background_action(
...
+
+    Returns:
+        Updated Operation reflecting the cancel attempt.
+
+    Raises:
+        GenkitError: If the handle is missing action, the action is not
+            found, or cancel is not implemented.
+    """
+    resolved, background_action = await resolve_operation_action(registry, operation)
+    if not background_action.supports_cancel:
```
apascal07: `supports_cancel` is `self.cancel_action is not None`, so this check and the `if self.cancel_action is None` inside `BackgroundAction.cancel` test the same thing. Once this passes, that branch is unreachable from the veneer.

The message is written out twice, so an edit to one drifts from the other. Keeping the raise in one place, either here or in the wrapper alone, avoids that.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -379,29 +364,89 @@ async def lookup_background_action(
...
+            status='INVALID_ARGUMENT',
+            message='got a dump; pass Operation.model_validate(...)',
+        )
+    raise GenkitError(
+        status='INVALID_ARGUMENT',
+        message=f'got {type(value).__name__}, expected Operation',
+    )
+
+
+async def resolve_operation_action(
```
apascal07, line 388: `resolved` is always the object passed in. `require_operation` returns its argument unchanged on the success path, so the rebind on the next line is a no-op and `resolved is operation` holds at both call sites.

The tuple return and the two-name unpack read as though a dump gets coerced somewhere along the way. Returning the `BackgroundAction` alone says what actually happens.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
```diff
@@ -78,8 +81,189 @@ async def test_genkit_check_operation_not_found() -> None:
...
+async def test_cancel_operation_accepts_dumped_operation() -> None:
+    """Cancel accepts the same persisted Operation dump as check."""
+    ai = Genkit()
+    dumped = {
+        'id': '123',
+        'done': False,
+        'action': '/background-model/test_action',
+        'latencyMs': 42,
+    }
+    mock_background_action = MagicMock()
```
apascal07: `MagicMock().supports_cancel` is an auto-created child mock, so it is truthy and this test clears the `supports_cancel` gate for the wrong reason. Invert the gate in `cancel_operation` and the test still passes.

`test_cancel_operation_without_cancel_is_unimplemented` sets the attribute explicitly for the False case; setting it here too makes the assertion real.

py/scripts/schema_to_typing.py
```diff
@@ -133,6 +133,24 @@ def _models_allowing_extra(schema: dict) -> set[str]:
     return result
 
 
+# Models that silently drop unknown fields (e.g. leftover keys on saved handles)
+# so that Operation.model_validate(saved) does not fail on extra properties.
+_MODELS_IGNORING_EXTRA: frozenset[str] = frozenset({'Operation'})
```
apascal07: `_models_allowing_extra` right above derives its set from `additionalProperties` in the schema. This hardcodes one model name in Python instead, and the schema it contradicts still says `Operation` is closed.

Every object def in `genkit-schema.json` declares `additionalProperties` explicitly (89 false, 2 true), so the schema already has a place to express this. As it stands the Python runtime and the cross-language schema disagree about one type, JS and Go codegen see nothing, and the intent survives only as long as this constant does.

JS's `z.object` already strips unknown keys at runtime, so opening `Operation` in the schema would not break JS. If it should stay closed, the leniency belongs in the veneer instead.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
```diff
@@ -78,8 +81,189 @@ async def test_genkit_check_operation_not_found() -> None:
...
+    assert exc_info.value.status == 'INVALID_ARGUMENT'
+
+
+@pytest.mark.asyncio
+async def test_check_operation_accepts_dumped_operation() -> None:
+    """A persisted dump still polls, even with leftover keys like latencyMs."""
+    ai = Genkit()
+    dumped = {
+        'id': '123',
+        'done': False,
```
apascal07: Nothing in genkit produces a top-level `latencyMs` on an Operation dump. `wrapped_start` writes it nested at `op.metadata['latencyMs']` (`_background.py:267`), `_record_latency` only stamps models that declare a `latency_ms` field (`_action.py:51`) and `Operation` does not, and JS spreads it onto `GenerateResponseData` rather than the operation (`js/ai/src/model.ts:173`).

`Operation(...).model_dump(by_alias=True)` emits `action, id, done, output` and nothing else. So these fixtures pin a payload shape that cannot occur, and the round trip that can, `Operation.model_validate(op.model_dump(by_alias=True))`, has no test.

py/scripts/schema_to_typing.py
```diff
@@ -133,6 +133,24 @@ def _models_allowing_extra(schema: dict) -> set[str]:
...
+
+
+def _models_ignoring_extra(_schema: dict | None = None) -> set[str]:
```
apascal07: `_schema` is never read, and the body returns a copy of a module constant. It is called once, at line 371, with an argument that is discarded.

The signature mirrors `_models_allowing_extra(schema)`, which does read its argument, so the pair suggests both are schema-driven. Using `_MODELS_IGNORING_EXTRA` directly at the call site says what happens.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
```diff
@@ -78,8 +81,189 @@ async def test_genkit_check_operation_not_found() -> None:
...
-    with pytest.raises(ValueError, match='Failed to resolve background action from original request: missing'):
+    with pytest.raises(
+        GenkitError, match='Failed to resolve background action from original request: missing'
+    ) as exc_info:
         await ai.check_operation(op)
+    assert exc_info.value.status == 'INVALID_ARGUMENT'
+
+
+@pytest.mark.asyncio
+async def test_check_operation_accepts_dumped_operation() -> None:
```
apascal07: This passes `Operation.model_validate(dumped)`, not a dump. What it covers is that `model_validate` tolerates the extra key.

`test_check_operation_dump_is_invalid_argument` thirty lines below asserts that `check_operation` rejects a dump, so the two names read as a contradiction. `accepts_validated_dump` would separate them. Same for `test_cancel_operation_accepts_dumped_operation`.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -138,23 +136,28 @@ async def check(self, operation: Operation) -> Operation:
         Returns:
             Updated Operation with current status.
         """
+        operation = require_operation(value=operation)
```
apascal07, line 142: This raises `GenkitError(INVALID_ARGUMENT)` for a dict, a `ModelResponse`, or any non-Operation, which `test_background_action_check_rejects_non_operation` pins, but the docstring above still ends at `Returns:`.

`cancel` below gained a `Raises:` section for the UNIMPLEMENTED case and omits this one. Both are public contract now.

py/packages/genkit/src/genkit/_core/_background.py
```diff
@@ -379,29 +364,89 @@ async def lookup_background_action(
...
+    raise GenkitError(
+        status='INVALID_ARGUMENT',
+        message=f'got {type(value).__name__}, expected Operation',
+    )
+
+
+async def resolve_operation_action(
+    registry: Registry,
+    operation: Operation,
+) -> tuple[Operation, BackgroundAction]:
```
apascal07: nit: `BackgroundAction` is generic and `lookup_background_action` returns `BackgroundAction[ModelResponse] | None`. Leaving it unparameterized here erases `OutputT` for both callers.

apascal07 (approved):

apascal07 (approved):

py/packages/genkit/src/genkit/_core/_background.py
huangjeff5: Good catch — this was a real gap. `resolve_operation_action` now wraps the lookup and re-raises `parse_action_key`'s `ValueError` as `GenkitError(INVALID_ARGUMENT)`, so a mangled reloaded handle gets a 400 over reflection instead of a 500. Also unmocked the test that was hiding this (see below).

py/packages/genkit/src/genkit/_core/_typing.py
huangjeff5: Agreed, the silent `done` drop is worse than what ignore was buying. Reverted `Operation` to `extra='forbid'`. The motivating payload turned out not to exist anyway — genkit nests `latencyMs` under `metadata` on the handle, so `model_dump(by_alias=True)` → `model_validate` round-trips clean under forbid. The top-level `latencyMs` only appears on the `ModelResponse` dump, which this PR already rejects with a pointer to `response.operation`.

huangjeff5: Moot after reverting to forbid — the published input schema for check/cancel matches genkit-schema.json again.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
huangjeff5: Yep — replaced with two unmocked tests against a real registry: `action='missing'` (malformed key, pins the ValueError→GenkitError fix) and `action='/background-model/nope'` (well-formed but unregistered). Both assert INVALID_ARGUMENT.

py/packages/genkit/src/genkit/_core/_background.py
huangjeff5: The wrapper divergence is deliberate — returning the operation unchanged makes "this model can't cancel" indistinguishable from "cancelled", so the wrapper raises UNIMPLEMENTED itself. Added a comment on the raise so a parity pass doesn't revert it. On the UNAVAILABLE mismatch: that JS branch is inside `if (config.cancel)` and guarded by `if (!config.cancel)`, so it's unreachable narrowing code — nothing to match.

py/packages/genkit/src/genkit/_core/_background.py
huangjeff5: Consolidated — the raise now lives only in `BackgroundAction.cancel` and `cancel_operation` just calls it, so the message exists in one place.

py/packages/genkit/src/genkit/_core/_background.py
huangjeff5: Right, the tuple was theater. `resolve_operation_action` returns the `BackgroundAction` alone now.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
huangjeff5: The test now defines a real no-cancel background model and asserts UNIMPLEMENTED end-to-end through `ai.cancel_operation`, so there's no mock attribute to get wrong (and the veneer no longer reads `supports_cancel` at all after the consolidation).

py/scripts/schema_to_typing.py
huangjeff5: Gone with the forbid revert — all the ignore machinery in the codegen script is removed and the schema stays the single source of truth.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
huangjeff5: Confirmed — nothing produces a top-level `latencyMs` on the handle. Replaced the fixtures with the real round trip, including one test that starts an operation via `generate_operation`, asserts the dump has no stray keys (latency is nested in metadata), and reloads it through `model_validate` → `check_operation`.

py/scripts/schema_to_typing.py
huangjeff5: Removed along with the rest of the ignore machinery.

py/packages/genkit/tests/genkit/ai/genkit_api_test.py
huangjeff5: Renamed both to `*_round_trips_persisted_dump`, which is what they actually cover.

py/packages/genkit/src/genkit/_core/_background.py
huangjeff5: Added, and expanded `cancel`'s to cover both INVALID_ARGUMENT and UNIMPLEMENTED.

py/packages/genkit/src/genkit/_core/_background.py
huangjeff5: Fixed — returns `BackgroundAction[ModelResponse]`.

PR #6134, opened 2026-08-20
fix(py): redact auth and secrets on action spans

jeffdh5:
## Summary
- Action spans dump run context onto `genkit:metadata:context` so the Dev UI Context panel can render it. Top-level `auth` and `secrets` become `<redacted>` before that dump.
- Nested keys named `auth` are left alone. The live `ctx.context` the action sees is unchanged.
- Does not add `check_operation(..., context=)` or teach Veo to read a poll-time key.

## Test plan
- [x] `context={'auth': {token}, 'secrets': {apiKey}, 'safeField': ..., 'nestedField': {auth: ...}}` → span has `<redacted>` / `<redacted>` / `hello` / `{auth: this stays}`
- [x] Existing unserializable-context test still passes (`auth` renamed to `session` so it is not the redaction path)

### DIFF
--- a/py/packages/genkit/src/genkit/_core/_action.py
+++ b/py/packages/genkit/src/genkit/_core/_action.py
@@ -59,6 +59,21 @@ def _record_latency(output: object, start_time: float) -> object:
     return output
 
 
+REDACTED_CONTEXT_KEYS = frozenset({'auth', 'secrets'})
+
+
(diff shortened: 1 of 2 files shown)


jeffdh5:
Closing — opened without being asked. Patch stays local / emailed.

PR #6141, opened 2026-08-22
fix(go/plugins/googlegenai): stop resending and re-creating the context cache

### LINKED ISSUE #6137: [Go] Context caching sends the cached prefix inline as well, doubling prompt cost
**Describe the bug**

In the Go `googlegenai` plugin, context caching puts the cached span into the `CachedContent` resource but **also leaves it in the request `contents`**, so the cached prefix is sent twice — once by reference and once inline. The JS plugin removes it; the Go plugin does not.

`handleCache` collects `request.Messages[0..endIndex]` into the cache and never mutates `request.Messages`:

https://github.com/genkit-ai/genkit/blob/main/go/plugins/googlegenai/cache.go

Then the generate path builds `contents` from **every** non-system message, including the ones just cached:

- v1.10.0 — `go/plugins/googlegenai/gemini.go:193-208`
- v1.12.0 — `toGeminiContents`, `go/plugins/googlegenai/gemini.go:327-352`

```go
var contents []*genai.Content
for _, m := range input.Messages {
    if m.Role == ai.RoleSystem {
        continue
    }
    // ... no check against the cached span
    contents = append(contents, &genai.Content{Parts: parts, Role: toGeminiRole(m.Role)})
}
```

Compare the JS implementation, which slices the cached span off the request (`js/plugins/vertexai/src/context-caching/utils.ts`):

```ts
const cachedContent: CachedContent = {
  model: modelVersion,
  contents: chatRequest.history.slice(0, endOfCachedContents + 1),
};
chatRequest.history = chatRequest.history.slice(endOfCachedContents + 1);
```

**Impact**

The feature's purpose is to stop re-sending a large stable prefix. As written in Go, the prefix is billed as cached tokens *and* as full-price inline input, so enabling context caching increases cost instead of reducing it, and the model sees the prefix twice. That makes the Go feature a cost regression rather than an optimisation.

**To Reproduce**

Any two-message Go request with a `cache` marker on the first message:

```go
opts.Messages[0].Metadata = map[string]any{
    "cache": map[string]any{"ttlSeconds": 3600},
}
resp, _ := genkit.GenerateWithRequest(ctx, g, opts, nil, nil)
// resp.Usage.CachedContentTokens > 0, but Usage.InputTokens still
// includes the full prefix inline.
```

**Expected behavior**

After the cached span is created or looked up, the generate request should carry only the messages *after* the cache boundary, matching the JS behaviour.

**Runtime environment**
- `github.com/firebase/genkit/go` v1.10.0, also read in v1.12.0
- `google.golang.org/genai` v1.51.0
- Backend: Vertex AI

**Additional context**

Two related points in the same area, both worked around downstream:

1. A named cache that has expired returns `cache lookup error` and is not recreated — there is an existing `// TODO: if cache expired or not found, create a fresh one` in `handleCache`.
2. `CachedContent` set through `ai.WithConfig` is rejected (`gemini.go` "cached content must be set using Genkit feature: ai.WithCacheTTL()"), so the metadata marker is the only route and there is no way for a caller to work around the duplication.

a2105z:
Fixes #6137

Repairs the explicit context-caching path, which billed a caller more for turning it on than for leaving it off.

## Bugs fixed

**The cached prefix was sent inline as well.** `handleCache` stored `messages[0..endIndex]` on the `CachedContent` resource and left `request.Messages` alone, so `toGeminiContents` sent them again: billed as cached tokens and as full-price input, and seen twice by the model. `handleCache` now returns the messages that still have to go on the wire, so nothing downstream needs to know a cache exists. JS slices the same span off `chatRequest.history`.

**The cached span was encoded differently from the inline one.** It was built backwards, wrote `ai.RoleTool` through as role `"tool"`, which the Content API rejects, and kept messages whose parts converted to nothing, which it also rejects. Both spans now share one converter, so they cannot drift apart again.

**The content hash could not tell two caches apart.** Only `Text` and `InlineData` reached the digest, with no separator, so every `gs://` document hashed to the digest of no bytes at all and `[ab]` collided with `[a][b]`. That hash is the only check that a named cache holds the request's own content. Contents now hash by their JSON encoding, keyed by model.

**Every request created a new cache.** Nothing looked for a resource already holding the same content, and the name reaches the caller only through response metadata. `handleCache` now scans display names first, bounded at 200. A cache that no longer resolves or no longer matches is rebuilt, rather than returned as an error the caller cannot act on.

**The markers cancelled each other, and a name alone did nothing.** `WithCacheTTL` and `WithCacheName` each replaced the whole `cache` metadata map, and `findCacheMarker` ignored a name unless a ttl accompanied it, so caching was silently off and the prefix went inline at full price.

```go
msg := ai.NewUserTextMessage(doc).WithCacheTTL(3600).WithCacheName(known)
// cache metadata: {"ttlSeconds": 3600, "name": known}
```

## Behavior worth noting

Gemini 2.5 and newer cache repeated prefixes implicitly, on by default and with no storage charge, while `WithCacheTTL` bills a resource for its lifetime. `Usage.CachedContentTokens` reports a hit either way, so the explicit path earns its cost only when the hit has to be guaranteed. Marking the last message leaves nothing to send inline, so the request carries a minimal turn rather than failing.

### DIFF
--- a/go/ai/request_helpers.go
+++ b/go/ai/request_helpers.go
@@ -17,8 +17,6 @@
 package ai
 
 import (
-	"maps"
-
 	"github.com/firebase/genkit/go/core"
 )
 
@@ -108,41 +106,30 @@ func NewTextMessage(role Role, text string) *Message {
 
 // WithCacheTTL adds cache TTL configuration for the desired message
 func (m *Message) WithCacheTTL(ttlSeconds int) *Message {
-	metadata := make(map[string]any)
-
-	if m.Metadata != nil {
-		metadata = m.Metadata
-	}
-
-	cache := map[string]any{
-		"cache": map[string]any{
-			"ttlSeconds": ttlSeconds,
-		},
-	}
-	maps.Copy(metadata, cache)
-
-	return &Message{
-		Content:  m.Content,
-		Role:     m.Role,
-		Metadata: metadata,
-	}
+	return m.withCacheMetadata("ttlSeconds", ttlSeconds)
 }
 
 // WithCacheName adds cache name to use in the generate request
 func (m *Message) WithCacheName(n string) *Message {
-	metadata := make(map[string]any)
+	return m.withCacheMetadata("name", n)
+}
 
+// withCacheMetadata sets one key of the message's "cache" metadata and leaves
+// the rest of that map alone, so a message can carry both the ttl that builds
+// a cache and the name of the one an earlier turn already built. Replacing the
+// whole map instead would make the two setters cancel each other out.
+func (m *Message) withCacheMetadata(key string, value any) *Message {
+	metadata := make(map[string]any)
 	if m.Metadata != nil {
 		metadata = m.Metadata
 	}
 
-	cache := map[string]any{
-		"cache": map[string]any{
-			"name": n,
-		},
+	cache, _ := metadata["cache"].(map[string]any)
+	if cache == nil {
+		cache = make(map[string]any)
 	}
-
-	maps.Copy(metadata, cache)
+	cache[key] = value
+	metadata["cache"] = cache
 
 	return &Message{
 		Content:  m.Content,
--- a/go/plugins/googlegenai/README.md
+++ b/go/plugins/googlegenai/README.md
@@ -301,27 +301,80 @@ resp, err := genkit.Generate(ctx, g,
 
 ### Context Caching
 
-Gemini 2.5 and newer models automatically cache common content prefixes. In Genkit Go, you can mark content for caching using `WithCacheTTL` or `WithCacheName`.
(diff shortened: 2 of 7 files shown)


a2105z:
Hi @genkit-ai/genkit-googlers @pavelgj @apascal07 @huangjeff5 @ifielker @MichaelDoyle @yesudeep @ssbushi @shrutip90 @cabljac @adesinah @xcapaldi — this is the fix for #6137.

Context caching was writing the prefix into `CachedContent` and still sending it inline. `toGeminiContents` now skips the cached span, same as JS.

Could someone take a look when you have a moment? Happy to adjust anything. Thank you!

apascal07 (approved):

PR #6142, opened 2026-08-22
feat(py): add Message.media as first-or-null

### LINKED ISSUE #6097: feat(py): add Message.media as first-or-null
## What

Add `Message.media` on the Python `Message` type so callers can read the image on a message without scanning `content`.

The accessor is the **first media part, or `None`** — same shape as JS (`message.media.url`).

```python
msg.media          # Media | None
msg.media.url      # works when present
```

`AgentResponse.media` is already this (`first_media_of`). `Document.media` stays a list (JS `Document.media` is a list too).

`ModelResponse.media` is currently a list. This PR should make it first-or-null as well so `response.media.url` matches JS `GenerateResponse.media`. That is a behavior change for anyone already iterating `response.media`; call it out in the PR.

## Why first-or-null

JS `Message.media` / `GenerateResponse.media` is first-or-null. Go `Media()` is the first URL. A JS porter’s first line is `const { media } = await ai.generate(); media.url`.

If we need every part later, add `media_parts` (or keep scanning `content`). Do not make `.media` a list.

## Out of scope

This is **not** on the HTTP-classify / Dev UI logs / `generate()` throw-on-blocked train. Land it as a small PR off `main`.

## Test plan

- Lone `MediaPart` → `message.media` is that `Media`, `message.text == ''`
- Two media parts → first only
- Text-only / data-only → `None`
- `ModelResponse.media` is that first part when a message is present; `None` when `message` is None

a2105z:
## Summary
- Add `Message.media` on the Python `Message` type so callers can read the image on a message without scanning `content`. The accessor is the first media part, or `None` — same shape as JS (`message.media.url`).

  ```python
  msg.media          # Media | None
  msg.media.url      # works when present
  ```

- `AgentResponse.media` is already this (`first_media_of`). `Document.media` stays a list (JS `Document.media` is a list too).
- **Behavior change:** `ModelResponse.media` was a `list[Media]` (empty list when `message` is `None`). It is now first-or-null (`Media | None`) so `response.media.url` matches JS `GenerateResponse.media`. Anyone iterating `response.media` needs to scan `content` (or a future `media_parts`) instead.

Fixes #6097

## Test plan
- [x] Lone `MediaPart` → `message.media` is that `Media`, `message.text == ''`
- [x] Two media parts → first only
- [x] Text-only / data-only → `None`
- [x] `ModelResponse.media` is that first part when a message is present; `None` when `message` is `None`
- [x] `Document.media` remains a list
- [ ] CLA check

### DIFF
--- a/py/packages/genkit-amazon-bedrock/README.md
+++ b/py/packages/genkit-amazon-bedrock/README.md
@@ -299,7 +299,7 @@ response = await ai.generate(
     model='bedrock/stability.sd3-5-large-v1:0',
     prompt='A tabby cat asleep on a sunlit windowsill, watercolour.',
 )
-image = response.media[0].url  # data:image/png;base64,...
+image = response.media.url  # data:image/png;base64,...
 ```
 
 Declaring is optional here too. An undeclared ID in one of the two families
--- a/py/packages/genkit/src/genkit/_core/_model.py
+++ b/py/packages/genkit/src/genkit/_core/_model.py
@@ -200,7 +200,7 @@ class ModelRef(Generic[ModelRefConfigT]):
 
 
 class Message(MessageData):
-    """Message wrapper with utility properties for text and tool requests."""
+    """Message wrapper with utility properties for text, media, and tool requests."""
 
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


a2105z:
@jeffdh5 this is done.

`Message.media` and `ModelResponse.media` are first-or-null (`Media | None`), so `msg.media.url` / `response.media.url` work when a media part is present. `Document.media` is still a list.

Covered the cases from #6097: lone `MediaPart`, two parts (first only), text-only / data-only → `None`, and `ModelResponse.media` is `None` when `message` is `None`.

Issue #6002, opened 2026-08-10
[JS] OpenAI Responses API support in @genkit-ai/compat-oai (tracking)


cabljac:
Tracking issue for Responses API support in `@genkit-ai/compat-oai`. Follows the direction discussed in #5236 (see RFC discussion there; #5345 tracks the Go side).

## Approach

Per-model routing inside the existing `openAI()` plugin - one namespace, no sibling plugin:

- **Responses-only models** (`gpt-5-pro`, `gpt-5-codex`, `o1-pro`, `o3-pro`, ...) register on `/v1/responses`. They are unreachable via compat-oai today, so this is purely additive.
- **Dual-transport models** opt in explicitly: `openAI.responsesModel('gpt-5')`. Existing model references are unchanged - same wire protocol, same `raw` shape, same config.
- **Default behavior is stateless**: `store: false` pinned, encrypted reasoning items round-tripped via part metadata (`include: ['reasoning.encrypted_content']`), with `previousResponseId` + `store` exposed as opt-in config for server-side threading.
- **Built-in tools** (`web_search_preview`, `file_search`, `code_interpreter`) go through the existing `config.tools` passthrough - no new wrapper fields. Citations surface raw as `metadata.annotations` on text parts, matching the google-genai Interactions handling.

Deferred, out of scope here: flipping the default transport for `gpt-5*`/`o3`/`o4-mini` (breaking, needs its own proposal), and deep-research / computer-use models (long-running, need background-model actions; background mode also requires `store: true`).

## Example

```ts
// After #6003 (PR #6006): Responses-only models work by name
await ai.generate({ model: openAI.model('gpt-5-pro'), prompt });

// After #6005: opt a dual-transport model onto Responses explicitly,
// with built-in tools via the existing config.tools passthrough
await ai.generate({
  model: openAI.responsesModel('gpt-5'),
  prompt: 'What changed in the news today?',
  config: { tools: [{ type: 'web_search_preview' }] },
});
```

## Work items

One PR per item, in order:

- [ ] #6003 transport boundary + non-streaming Responses runner
- [ ] #6004 streaming, tool calling, encrypted reasoning round-trip
- [ ] #6005 `responsesModel()` opt-in surface, discovery, docs

cabljac:
## Design notes: Responses API in compat-oai

Context for #6006 and the follow-up slices. Expands on the direction discussed in #5236.

### Why per-model routing inside `openAI()` (not a sibling plugin)

[...]

PR #6160, opened 2026-08-25
feat(js/plugins/compat-oai): add responsesModel() opt-in and dual-transport dispatch

### LINKED ISSUE #6005: [JS] compat-oai: responsesModel() opt-in, discovery, docs
Part of #6002, follows #6004. Public surface + discovery.

Opt-in surface:
- `openAI.responsesModel(name)`: typed ref with Responses config schema, pins `config: { transport: 'responses' }`; overridden `withConfig` re-injects the pin (core `withConfig` replaces config wholesale); test `.withConfig().withVersion()` chains.
- Responses config schema: `transport` literal, `previousResponseId`, `store`, reasoning-summary options.
- Citations passthrough as `metadata.annotations` (set only when present).

Discovery:
- Route resolver, `listActions`, and `openAI.model()` fallthroughs through the Responses-only predicate; register remaining sync-appropriate Responses-only models (`o1-pro`, `o3-pro`, `gpt-5-codex`, `gpt-5.1-codex-max`).

Docs:
- Opt-in usage, built-in tools via `config.tools`, stateless default + opting into server-side threading, dotprompt frontmatter (`config: { transport: 'responses' }`).

cabljac:
Part of #6002, closes #6005. Slice 3 of the Responses API stack, on top of #6148.

- `openAI.responsesModel(name)`: typed ref pinned to `config: { transport: 'responses' }`; `withConfig`/`withVersion` re-wrapped so the pin survives chaining.
- Dual-transport dispatch: chat-registered OpenAI models route a request carrying `config: { transport: 'responses' }` to the Responses runner, so the opt-in also works from dotprompt frontmatter. Providers that do not opt in keep the INVALID_ARGUMENT rejection.
- Responses config schema declares `previousResponseId`, `reasoningEffort`, `reasoningSummary`, mapped to `previous_response_id` and `reasoning: { effort, summary }` on the wire; declared fields compose over a raw passthrough `reasoning` object.
- README: automatic routing for Responses-only models, the opt-in, built-in tools via `config.tools`, stateless default and opting into server-side threading.

Review hardening applied on top: chat-only config keys (`frequencyPenalty` etc.) are dropped from the Responses wire; dispatch-capable registrations default `supports.constrained` to `'all'` so an output schema reaches the Responses wire natively instead of being simulated into the prompt, while middleware re-applies the simulation to Chat Completions requests on models that never declared constrained support, keeping the plain chat path byte-identical to before; near-miss transport values (e.g. `'Responses'`) throw INVALID_ARGUMENT instead of silently serving Chat Completions; declared `previousResponseId` wins over a raw `previous_response_id` passthrough, matching the reasoning fields.

Known limitations, noted deliberately:

- `responsesModel()`'s config type inherits `.passthrough()` from the schema, so unknown keys (typos included) compile and pass to the wire. Cost of keeping the passthrough escape hatch; consistent with `model()`.
- The registered action for a dual-transport model advertises the Chat Completions config schema, so schema-driven tooling (Dev UI config forms) does not offer `previousResponseId`/`reasoningEffort`/`reasoningSummary` on it. Runtime honors them regardless; inherent to one action serving two transports.
- Dispatch-capable actions advertise `supports.constrained: 'all'` even when their Chat Completions API would reject `response_format: json_schema`; runtime behavior is correct on both transports (see above), the metadata over-claim is display-only.

Live smoke checklist before un-drafting:

- `config.tools: [{ type: 'web_search' }]` passthrough (type absent from the pinned SDK union).
- `previousResponseId` under the default `store: false` (unverified whether OpenAI accepts the combination).
- A non-reasoning model over the dispatch (`responsesModel('gpt-4o')`): confirms the reasoning-gated encrypted-content include.
- A text-only reasoning turn replayed stateless with `reasoningSummary` on (multi-turn, no tools): the assistant message replays without its `msg_` id next to an id-carrying reasoning item; one shape passed live already, ecosystem reports 400s on close variants.

### DIFF
--- a/js/plugins/compat-oai/src/model.ts
+++ b/js/plugins/compat-oai/src/model.ts
@@ -34,7 +34,13 @@ import {
   z,
 } from 'genkit';
 import { parsePartialJson } from 'genkit/extract';
-import type { ModelAction, ModelInfo, ToolDefinition } from 'genkit/model';
+import type {
+  ModelAction,
+  ModelInfo,
+  ModelMiddleware,
+  ToolDefinition,
+} from 'genkit/model';
+import { simulateConstrainedGeneration } from 'genkit/model';
 import { model } from 'genkit/plugin';
 import OpenAI from 'openai';
 import type {
@@ -737,22 +743,34 @@ export function defineCompatOpenAIModel<
           : chatRunner(request, options)
     : chatRunner;
 
-  // A dispatch-capable action must not be wrapped in
-  // simulateConstrainedGeneration: that middleware strips output.schema into
-  // the prompt before the transport decision runs, and both wire formats here
-  // carry the schema natively.
+  // Registering constrained keeps core's simulateConstrainedGeneration off,
+  // because it would strip output.schema into the prompt before the transport
+  // dispatch runs. Models that never declared constrained still need that
+  // simulation on their Chat Completions requests, so the middleware below
+  // re-applies it for exactly those.
+  const ownConstrained = modelRef?.info?.supports?.constrained;
   const supports = params.responsesTransport
     ? {
         ...modelRef?.info?.supports,
-        constrained: modelRef?.info?.supports?.constrained ?? ('all' as const),
+        constrained: ownConstrained ?? ('all' as const),
       }
     : modelRef?.info?.supports;
+  const use: ModelMiddleware[] | undefined =
+    params.responsesTransport && !ownConstrained
+      ? [
+          (req, next) =>
+            req.config?.transport === 'responses'
+              ? next(req)
+              : simulateConstrainedGeneration()(req, next),
+        ]
+      : undefined;
 
   return model(
     {
       name: actionName,
       ...modelRef?.info,
       ...(supports ? { supports } : {}),
+      ...(use ? { use } : {}),
       configSchema: modelRef?.configSchema,
     },
     runner
--- a/js/plugins/compat-oai/src/model.ts
+++ b/js/plugins/compat-oai/src/model.ts
@@ -728,10 +728,22 @@ export function defineCompatOpenAIModel<
           : chatRunner(request, options)
     : chatRunner;
 
+  // A dispatch-capable action must not be wrapped in
+  // simulateConstrainedGeneration: that middleware strips output.schema into
+  // the prompt before the transport decision runs, and both wire formats here
+  // carry the schema natively.
+  const supports = params.responsesTransport
+    ? {
+        ...modelRef?.info?.supports,
+        constrained: modelRef?.info?.supports?.constrained ?? ('all' as const),
+      }
+    : modelRef?.info?.supports;
+
   return model(
     {
       name: actionName,
       ...modelRef?.info,
+      ...(supports ? { supports } : {}),
       configSchema: modelRef?.configSchema,
     },
     runner
--- a/js/plugins/compat-oai/src/model.ts
+++ b/js/plugins/compat-oai/src/model.ts
@@ -537,6 +537,15 @@ export function toOpenAIRequestBody(
       message: `The 'responses' transport is not supported by ${modelVersion ?? modelName}; it is served over the Chat Completions API.`,
     });
   }
+  // Near-miss values ('Responses', a YAML typo) would otherwise be silently
+  // dropped and served over Chat Completions - the one silent cell in the
+  // transport matrix.
+  if (transport !== undefined && transport !== 'chat_completions') {
+    throw new GenkitError({
+      status: 'INVALID_ARGUMENT',
+      message: `Unknown transport '${transport}'.`,
+    });
+  }
 
   const tools: ChatCompletionTool[] = request.tools?.map(toOpenAITool) ?? [];
   if (toolsFromConfig) {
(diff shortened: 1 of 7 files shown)


js/plugins/compat-oai/src/model.ts
```diff
@@ -701,19 +720,60 @@ export function defineCompatOpenAIModel<
...
+    : modelRef?.info?.supports;
+  const use: ModelMiddleware[] | undefined =
+    params.responsesTransport && !ownConstrained
+      ? [
+          (req, next) =>
+            req.config?.transport === 'responses'
+              ? next(req)
+              : simulateConstrainedGeneration()(req, next),
+        ]
+      : undefined;
```
adesinah, line 766: Worth pulling into Known Limitations: core runs `use` middleware **before** `augmentWithContext()`, but its own simulator runs **after** it. So a request carrying both context docs and an output schema now gets its injected text in a different order than before this PR.

You spotted this in 1aacfe0f7's commit message, but right now it only lives in a test comment. It's also wider than it looks — only `MULTIMODAL_MODEL_INFO` declares `constrained`, so this path covers ~19 of the 30 curated chat refs (all of GPT-5, GPT-4 non-multimodal, GPT-4-vision, GPT-3.5), not just edge cases.

Nothing to fix in code — core always prepends `use` — just worth saying out loud in the description.

js/plugins/compat-oai/src/model.ts
```diff
@@ -701,19 +720,60 @@ export function defineCompatOpenAIModel<
...
+          ? responsesRunner(request, options)
+          : chatRunner(request, options)
+    : chatRunner;
+
+  // Registering constrained keeps core's simulateConstrainedGeneration off,
+  // because it would strip output.schema into the prompt before the transport
+  // dispatch runs. Models that never declared constrained still need that
+  // simulation on their Chat Completions requests, so the middleware below
+  // re-applies it for exactly those.
+  const ownConstrained = modelRef?.info?.supports?.constrained;
```
adesinah, line 751: Latent, not actionable today: `ownConstrained ?? 'all'` treats *any* declared value as "no compensation needed".

If a dispatch-capable model ever declared `constrained: 'no-tools'`, core's simulator would fire for tool-carrying requests on **both** transports — simulating a schema into the prompt on the Responses path, where it's supported natively.

Unreachable right now (only `openai` sets `responsesTransport`, and no chat info declares `'no-tools'`), so just a note for whoever extends this.

js/plugins/compat-oai/src/openai/index.ts
```diff
@@ -243,6 +244,7 @@ export function openAIPlugin(options?: OpenAIPluginOptions): GenkitPluginV2 {
             client,
             pluginOptions,
             modelRef,
+            responsesTransport: true,
```
adesinah, line 247: `listActions` above builds its metadata straight from `modelRef.info`, so it never picks up the `constrained: 'all'` override that `defineCompatOpenAIModel` applies here. Net effect: `gpt-5` *lists* as no-constrained but *registers* as `'all'`.

The description calls the over-claim display-only, which is fair — just flagging that the two display surfaces disagree with each other.

js/plugins/compat-oai/src/openai/responses.ts
```diff
@@ -97,6 +97,11 @@ export function isNonStreamingResponsesModelName(name?: string): boolean {
 /** OpenAI Responses API custom configuration schema. */
 export const OpenAIResponsesConfigSchema = ResponsesCommonConfigSchema.extend({
   store: z.boolean().optional(),
+  previousResponseId: z.string().optional(),
+  reasoningEffort: z
+    .enum(['none', 'minimal', 'low', 'medium', 'high', 'xhigh'])
+    .optional(),
```
adesinah, line 103: The pinned SDK types `ReasoningEffort` as `'low' | 'medium' | 'high' | null`, so `none`, `minimal` and `xhigh` sit outside its union and pass straight through to the wire.

Same risk class as the `config.tools: [{ type: 'web_search' }]` passthrough that's already on the live-smoke checklist — and since this landed in the last commit on the branch, it's had the least live exercise. Probably worth its own line on that checklist.

js/plugins/compat-oai/tests/openai_responses_test.ts
```diff
@@ -1697,4 +2003,16 @@ describe('chat completions transport handling', () => {
       })
     ).rejects.toThrow(expect.objectContaining({ status: 'INVALID_ARGUMENT' }));
   });
+
+  test('rejects a near-miss transport value instead of serving it silently', async () => {
```
adesinah, line 2007: This drives `openAIModelRunner` directly, so it skips the action + config-schema layer. The scenario in the description is dotprompt frontmatter, which *does* go through that layer, and `transport` is a `z.enum` there.

An end-to-end case would pin down which layer actually rejects `'Responses'` — otherwise a later change to config validation could relocate the behavior without failing anything.

adesinah (commented):
Went through this against the ticket. The dispatch design holds up well — I checked the ref pinning against core's `modelRef` and it's correct for the right reasons (arrow functions so the unbound method extraction is safe, `withVersion` spreads the original options so the pin survives, `withConfig` replaces wholesale hence the re-injection). `transport` is also stripped on both wire builders, so nothing leaks to either API. Typecheck clean, 199 tests pass.

Five notes inline — mostly documentation rather than code changes.

PR #6167, opened 2026-08-25
fix(py/plugins/openai): omit response_format for gpt-image-1

### LINKED ISSUE #6155: fix(py/plugins/openai): gpt-image-1 rejects response_format sent by default
The Python image model always includes a `response_format` value in image generation requests and defaults it to `b64_json`. The `gpt-image-1` model rejects that parameter, so requests fail. The JS plugin strips the parameter for this model and also ships a richer `gpt-image-1` config schema covering `background`, `moderation`, `output_compression`, and `output_format`.

mikemikimike:
## Motivation and Context

Resolves #6155.

The Python OpenAI image model always sent `response_format`, defaulting to `b64_json`. OpenAI's `gpt-image-1` endpoint rejects that parameter, so image generation requests failed. Its supported configuration options also differ from DALL-E, so invalid values such as `quality: 'hd'` should be rejected by configuration validation before reaching the API.

## Description

- Use the effective image model name, including a `version` override.
- Omit `response_format` for `gpt-image-1`, while preserving the existing DALL-E default.
- Advertise a GPT Image 1-specific configuration schema with the API constraints for `size`, `quality`, `background`, `moderation`, `output_compression`, and `output_format`.
- Add focused regression coverage for the request behavior and model schema.

## Compatibility

Only GPT Image request construction and metadata change. DALL-E request payloads retain their previous defaults and explicit format behavior. GPT Image 1 now exposes model-specific validation constraints and no longer advertises the unsupported `response_format` option.

## Validation

- `python -m pytest packages/genkit-openai/tests/image_model_test.py --no-cov -q` — 22 passed.
- `ruff format --check packages/genkit-openai/src/genkit_openai/models/image.py packages/genkit-openai/tests/image_model_test.py` — passed.
- `ruff check packages/genkit-openai/src/genkit_openai/models/image.py packages/genkit-openai/tests/image_model_test.py` — passed.
- `uv build --package genkit-openai` — passed.
- `git diff --check` — passed.
- `ty check` was attempted, but the available Ty version reports two pre-existing `FinishReason.STOP` errors in unchanged lines of `image.py` and does not recognize one repository configuration rule.
- `uv sync --frozen` on Windows was blocked because `dotpromptz-handlebars==0.1.8` has no Windows wheel. Focused tests used a task-local virtual environment and an untracked import shim for that unavailable dependency; no shim or generated files are part of this PR.
- Equivalent Ubuntu Docker validation was attempted, but PyPI dependency installation hit repeated TLS EOFs before pytest could run.


py/packages/genkit-openai/src/genkit_openai/models/image.py
```diff
@@ -91,12 +91,19 @@ def _to_image_generate_params(
...
         'prompt': prompt,
-        'response_format': config.pop('response_format', 'b64_json'),
     }
 
+    # GPT Image 1 rejects the response_format parameter; its API always
+    # returns base64 image data. DALL-E models retain the existing default.
+    if effective_model != 'gpt-image-1':
+        params['response_format'] = config.pop('response_format', 'b64_json')
+    else:
+        config.pop('response_format', None)
```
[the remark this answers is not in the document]
mikemikimike: Confirmed. The current comparison is case-sensitive, so values such as `GPT-Image-1` would incorrectly retain `response_format`. This is a valid follow-up; the implementation should use `effective_model.lower() == "gpt-image-1"`, with a regression test for mixed-case model names.

hilariie: This matches `gpt-image-1` exactly, but the plugin routes on a substring. `_classify_model` uses `'gpt-image' in name`, so `gpt-image-1-mini` and `gpt-image-1.5` also reach this function and still get `response_format` sent, which is the failure #6155 describes. Both are in the `ImageModel` literal in the `openai` SDK we depend on, and the dependency is unpinned in `pyproject.toml`. The JS builder uses `name.includes('gpt-image-1')` for the same reason.

Could we match on the substring instead? Reading it in the positive form is also a bit clearer.

```suggestion
    if 'gpt-image' in effective_model:
        config.pop('response_format', None)
    else:
        params['response_format'] = config.pop('response_format', 'b64_json')
```

py/packages/genkit-openai/tests/image_model_test.py
```diff
@@ -84,6 +84,19 @@ def test_basic_params(self) -> None:
...
+        request = ModelRequest(
+            messages=[
+                Message(role=Role.USER, content=[Part(root=TextPart(text='a cat'))]),
+            ],
+            config={'response_format': 'url'},
+        )
+
+        got = _to_image_generate_params('gpt-image-1', request)
+
+        assert 'response_format' not in got
```
hilariie, line 98: Could we add a case that goes through the `version` override? That is the path this PR restructured, and today `_to_image_generate_params('gpt-image-1', config={'version': 'gpt-image-1-mini'})` still returns `response_format='b64_json'`. This test only covers the bare model name, so it would not catch that.

hilariie (commented):
Thanks for picking this up. The regression test looks good and fails against the pre-fix code. The test comment below is about widening the match and the issue scope, not the approach.

The `response_format` fix looks right, but #6155 has a second half that this PR does not cover, which is the richer `gpt-image-1` config schema (`background`, `moderation`, `output_compression`, `output_format`, plus the `size` and `quality` enums, which differ from DALL-E). Could we include that here so the issue closes complete? Reference `SUPPORTED_IMAGE_MODELS` in `js/plugins/compat-oai/src/openai/index.ts` for this.

py/packages/genkit-openai/src/genkit_openai/models/image.py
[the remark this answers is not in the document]
mikemikimike: System.Collections.Hashtable[3854691471]

mikemikimike: System.Collections.Hashtable[3861179449]

py/packages/genkit-openai/tests/image_model_test.py
mikemikimike: System.Collections.Hashtable[3861192492]

hilariie (commented):
Thanks, the `version` override test and the substring match both look right.

The part of #6155 I was after is the schema itself. `gpt-image-1` needs `size` limited to `1024x1024|1536x1024|1024x1536|auto` and `quality` to `low|medium|high`, both of which differ from DALL-E, plus `background`, `moderation`, `output_compression`, and `output_format`. Today `quality: 'hd'` on `gpt-image-1` passes straight through and 400s at the API instead of failing config validation. `SUPPORTED_IMAGE_MODELS` in `js/plugins/compat-oai/src/openai/dalle.ts` has the shape.

Worth noting Python image models are registered with no `config_schema` at all right now, so this is adding one rather than extending one.

Happy to differ this to a different slice, but this PR body says `Resolves #6155`, so if you would rather keep this PR to the `response_format` fix, could you change that to `Part of #6155` so the issue stays open for the schema? Otherwise merging this closes it with half the work still to do.

FYI, your last three replies came through as `System.Collections.Hashtable[<comment id>]`, so whatever you wrote did not make it to GitHub.

py/packages/genkit-openai/src/genkit_openai/models/image.py
```diff
@@ -56,6 +75,7 @@
     ),
     'gpt-image-1': ModelInfo(
         label='OpenAI - GPT Image 1',
+        config_schema=_GPT_IMAGE_1_CONFIG_SCHEMA,
```
hilariie, line 78: nothing reads this. `_create_multimodal_action` dumps it into `metadata.model.configSchema`, and the registered `openai/gpt-image-1` action ends up with `customOptions: None`. Could we surface it as `customOptions` there and in `_multimodal_action_metadata`, which hardcodes `ModelConfig`?

mikemikimike:
Implemented in c9105fe. Multimodal metadata now passes each registered model's config_schema through model_action_metadata, so gpt-image-1 exposes its schema as model.customOptions for both init() and list_actions(), without leaking configSchema. Added regressions for both paths; the genkit-openai test package passes locally (173 tests).

hilariie:
fix CI and I'll approve

mikemikimike:
Fixed in 60c33dac2736db1fff5a02e24d4187bd277eac88. The failing Ty check came from nested assertions over dynamically typed action metadata; the tests now narrow the model metadata to dict[str, Any] before indexing it. Verified locally with Ty 0.0.17, Ruff, and 28 focused OpenAI tests. The new fork workflow is currently awaiting upstream approval and has not started any jobs yet.

hilariie (approved):

PR #6168, opened 2026-08-25
fix(py/openai): add Whisper translation support

### LINKED ISSUE #6154: fix(py/plugins/openai): whisper lacks the translate flag (JS parity)
The JS whisper model supports a `translate` flag that switches the request over to the OpenAI Translations API. The Python plugin has no translation path at all.

mikemikimike:
## Summary

Fixes #6154.

The Python OpenAI Whisper model now honors `config={"translate": true}` by sending the request to the OpenAI Audio Translations API. The default and explicit false paths continue to use the Transcriptions API.

## Implementation and compatibility

- Route `translate: true` to `client.audio.translations.create` only for the registered `whisper-1` model.
- Reject `translate: true` for `gpt-4o-transcribe` and `gpt-4o-mini-transcribe` with `GenkitError(status="INVALID_ARGUMENT")` before making an API call.
- Keep the existing transcription request construction and `stream=False` behavior unchanged for the default path.
- Remove transcription-only `language` and `timestamp_granularities` parameters from translation requests.
- Accept the OpenAI `Translation` response type in the shared text-response converter.
- Add regression coverage for translation rejection on both non-Whisper models, plus explicit `translate: false` transcription behavior.

This is backward compatible: callers that omit `translate`, set it to `false`, or use `whisper-1` retain the existing behavior.

## Validation

Run in a Linux `ghcr.io/astral-sh/uv:python3.11-bookworm` container with the repository's frozen `uv` workspace:

- `python -m pytest --no-cov packages/genkit-openai/tests -q` — 166 passed.
- `ruff format --check --preview packages/genkit-openai/src/genkit_openai/models/audio.py packages/genkit-openai/tests/audio_model_test.py` — passed after normalizing the Windows checkout's CRLF line endings for the Linux formatter.
- `ruff check --preview packages/genkit-openai/src/genkit_openai/models/audio.py packages/genkit-openai/tests/audio_model_test.py` — passed.
- `uv lock --check` — passed.
- `uv build --package genkit-openai` — passed.
- `python -m compileall -q packages/genkit-openai/src packages/genkit-openai/tests` — passed.
- `git diff --check` — passed using the Windows Git checkout.

Pyright was run against `packages/genkit-openai/src` with the package environment and reported 13 unrelated diagnostics in existing files/lines; the new translation guard introduced no diagnostic. The full repository suite and Python-version matrix remain for CI; no live OpenAI credentials or external services are required for the focused tests.

### DIFF
--- a/py/packages/genkit-openai/src/genkit_openai/models/audio.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/audio.py
@@ -375,6 +375,8 @@ class OpenAISTTModel:
         params = _to_stt_params(self._model_name, request)
         translate = extract_config_dict(request).get('translate', False)
         if translate:
+            params.pop('language', None)
+            params.pop('timestamp_granularities', None)
             result = await self._client.audio.translations.create(**params)
         else:
             result = await self._client.audio.transcriptions.create(
--- a/py/packages/genkit-openai/src/genkit_openai/models/audio.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/audio.py
@@ -33,6 +33,7 @@ from openai._legacy_response import HttpxBinaryResponseContent
 from openai.types.audio import Transcription, Translation
 
 from genkit import (
+    GenkitError,
     Media,
     MediaPart,
     Message,
@@ -372,8 +373,17 @@ class OpenAISTTModel:
         Returns:
             A ModelResponse containing the transcribed text.
         """
-        params = _to_stt_params(self._model_name, request)
         translate = extract_config_dict(request).get('translate', False)
+        if translate and self._model_name != 'whisper-1':
+            raise GenkitError(
+                status='INVALID_ARGUMENT',
+                message=(
+                    "OpenAI audio translations only support 'whisper-1'; "
+                    f"model '{self._model_name}' cannot use translate=True."
+                ),
+            )
+
+        params = _to_stt_params(self._model_name, request)
         if translate:
             params.pop('language', None)
             params.pop('timestamp_granularities', None)
--- a/py/packages/genkit-openai/src/genkit_openai/models/audio.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/audio.py
@@ -30,7 +30,7 @@ from typing import Any
 
 from openai import AsyncOpenAI
 from openai._legacy_response import HttpxBinaryResponseContent
-from openai.types.audio import Transcription
+from openai.types.audio import Transcription, Translation
 
 from genkit import (
     Media,
@@ -271,7 +271,7 @@ def _to_stt_params(
     return {k: v for k, v in params.items() if v is not None}
 
 
-def _to_stt_response(result: Transcription | str) -> ModelResponse:
+def _to_stt_response(result: Transcription | Translation | str) -> ModelResponse:
     """Convert an OpenAI transcription result to a Genkit ModelResponse.
 
     Handles the full union of types returned by transcriptions.create().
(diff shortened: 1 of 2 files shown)


py/packages/genkit-openai/src/genkit_openai/models/audio.py
```diff
@@ -373,10 +373,14 @@ async def generate(self, request: ModelRequest, ctx: ActionRunContext) -> ModelR
             A ModelResponse containing the transcribed text.
         """
         params = _to_stt_params(self._model_name, request)
-        result = await self._client.audio.transcriptions.create(
-            **params,
-            stream=False,
-        )
+        translate = extract_config_dict(request).get('translate', False)
+        if translate:
+            result = await self._client.audio.translations.create(**params)
```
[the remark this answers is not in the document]
mikemikimike: Addressed in the current head `9afd284e`. When `translate` is enabled, the implementation now removes `language` and `timestamp_granularities` before calling `audio.translations.create`; the regression test asserts both parameters are absent and the transcription endpoint is not called.

py/packages/genkit-openai/src/genkit_openai/models/audio.py
```diff
@@ -373,10 +373,16 @@ async def generate(self, request: ModelRequest, ctx: ActionRunContext) -> ModelR
             A ModelResponse containing the transcribed text.
         """
         params = _to_stt_params(self._model_name, request)
-        result = await self._client.audio.transcriptions.create(
-            **params,
-            stream=False,
-        )
+        translate = extract_config_dict(request).get('translate', False)
```
hilariie, line 376: `OpenAISTTModel` also backs `gpt-4o-transcribe` and `gpt-4o-mini-transcribe`, so `translate: true` on either one sends `model='gpt-4o-transcribe'` to `audio.translations.create`, which OpenAI only supports for `whisper-1`. JS scopes this flag to `defineOpenAIWhisperModel`, which is registered for whisper only, so as written this is wider than the JS behavior it is matching.

So we should gate the translate branch on the whisper model name and raise `GenkitError(status='INVALID_ARGUMENT')`. `openai_plugin.py` already uses that pattern to keep model ids off the wrong endpoint.

py/packages/genkit-openai/tests/audio_model_test.py
```diff
@@ -305,6 +305,53 @@ async def test_generate_calls_speech_create(self) -> None:
 class TestOpenAISTTModel:
     """Tests for the OpenAISTTModel class."""
 
+    @pytest.mark.asyncio
+    async def test_generate_calls_translation_create_when_translate_is_enabled(self) -> None:
```
hilariie, line 310: Can you add the explicit `translate: false` case? The description says the explicit false path still uses Transcriptions, but the only coverage is the pre-existing test with no config at all. The JS suite covers unset and explicit false separately.

py/packages/genkit-openai/src/genkit_openai/models/audio.py
mikemikimike: Addressed in commit `c66f5d79`. `OpenAISTTModel` now gates `translate=True` on the registered model name: only `whisper-1` uses `audio.translations.create`; `gpt-4o-transcribe` and `gpt-4o-mini-transcribe` raise `GenkitError(status="INVALID_ARGUMENT")` before any endpoint call. The parametrized regression test covers both non-Whisper models.

py/packages/genkit-openai/tests/audio_model_test.py
mikemikimike: Added explicit `translate=False` coverage alongside the existing unset-config case. Both paths assert that `audio.transcriptions.create` is called and the translations endpoint is not used.

hilariie (approved):

PR #6172, opened 2026-08-26
feat(py): implement multipart tool response

huangjeff5:
A tool can return a structured JSON value, or a value plus media. The handler's return annotation defines the `outputSchema` advertised to the model. All tool invocations (`await tool()`, Dev UI `Action.run()`, and `generate` tool response messages) return a `MultipartToolResponse` envelope: structured `output` plus optional `content` (media parts). Dynamic MCP tool selectors (`mcp:tool/echo`, `mcp:tool/*`) resolve and bind `/tool.v2/<name>` on the ephemeral `generate` child only.

---

### 1. Schema Binding & Envelope Contract

The return annotation is the schema advertised to the model. Tool execution always returns a `MultipartToolResponse` envelope:

```python
# 1. Primitive return — model schema is string, caller receives envelope
@ai.tool
async def weather(city: str) -> str:
    return f'Sunny in {city}'

out = await weather('Austin')
assert isinstance(out, MultipartToolResponse)
assert out.output == 'Sunny in Austin'
assert out.content is None
# Model outputSchema: {type: 'string'}
# generate role=tool message: {output: 'Sunny in Austin', content: None}
```

```python
# 2. Pydantic return — model schema is ShotOut, caller receives envelope
@ai.tool
async def shot() -> ShotOut:
    return ShotOut(ok=True, label='lab')

out = await shot()
assert out.output == {'ok': True, 'label': 'lab'}
assert out.content is None
```

```python
# 3. Multipart return — return value plus media parts
@ai.tool
async def screenshot() -> MultipartToolResponse[ShotOut]:
    return response(ShotOut(ok=True, label='lab'), parts=[png])

out = await screenshot()
assert out.output == {'ok': True, 'label': 'lab'}
assert out.content == [png]
```

- **Extra Media**: `-> str` returning `response('Sunny', parts=[png])` keeps `{type: 'string'}` as `outputSchema` while attaching the PNG to `content`.
- **Dynamic Schema**: Bare `-> MultipartToolResponse` (unsubscripted) tells the model `outputSchema is None`.
- **Part Metadata**: Metadata passed via `response(..., metadata=...)` lands on the `Part` envelope, not inside `ToolResponse`.

---

### 2. Dynamic Tool Resolution (`mcp:tool/*`)

Callers pass selector patterns directly to `generate`:

```python
await ai.generate(prompt='go', tools=['mcp:tool/echo'])
await ai.generate(prompt='go', tools=['mcp:tool/*'])
```

- **Child-Scoped Binding**: Selectors resolve and register `/tool.v2/<name>` on the generate child action runner only. The root app catalog remains clean.
- **DAP Wire Format**: DAP authors write `{tool: [...]}`. `{tool.v2: [...]}` is not matched by `mcp:tool/echo`.
- **Resolution Boundary**: `mcp:tool.v2/echo` is not a valid `tools=` argument. `resolve_tool('mcp:tool/echo')` fails (expansion during generate is the only bind path).
- **Preflight Validation**: Unknown DAP providers, missing tools, and duplicate tool name collisions (e.g. `tools=['mcp:tool/*', 'toolA']`) raise `INVALID_ARGUMENT` before calling the model.

---

### 3. Decisions & Breaking Changes

- **Universal Envelope Shape**: All tool execution paths return `MultipartToolResponse(output=..., content=...)`.
- **Schema & Media Decoupling**: Return type annotation dictates `outputSchema`. Attached media (`content: Sequence[Part]`) does not alter the advertised schema.
- **Ephemeral Registry Bindings**: MCP tool expansion binds to `/tool.v2/<name>` exclusively on the invocation child.
- **Breaking: Tool Call Return**: `await tool(...)` now returns `MultipartToolResponse` instead of the bare return value.
- **Breaking: Tool Action Kind**: Registry kind is updated from `tool` to `tool.v2` (`/tool.v2/<name>`).


py/packages/genkit/src/genkit/_ai/_tools.py
```diff
@@ -20,16 +20,330 @@
...
+        try:
+            return to_json_schema(output_schema)
+        except Exception as e:
+            raise GenkitError(
+                status='INVALID_ARGUMENT',
+                message=(
+                    f'Tool {tool_name!r} output_schema is not a JSON Schema or Pydantic model, got {output_schema!r}.'
+                ),
+                cause=e,
+            ) from e
```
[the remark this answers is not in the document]
huangjeff5: We only support 3.10+. The check is isinstance(output_schema, type) and issubclass(...), so issubclass is only called when the value is actually a class.

On 3.10, dict[str, int] is a type, but issubclass returns False (no TypeError). On 3.11+, generic aliases are not type, so issubclass is never called. 3.13+ would TypeError if we called issubclass, but the and short-circuits.

output_schema=dict[str, int] already raises the same INVALID_ARGUMENT as output_schema=str. No try/except needed.

py/packages/genkit/src/genkit/_ai/_tools.py
```diff
@@ -401,7 +734,19 @@ async def tool_fn_wrapper(*args: Any) -> Any:  # noqa: ANN401 - arity dispatch;
...
+    # A bare return is the inner output (and the model schema). response() is
+    # already the action result — don't advertise that wrapper as outputSchema
+    # unless output_schema= says otherwise.
+    if output_schema is not None:
+        declared_output_schema = override_output_schema(output_schema, tool_name=tool_name)
+    elif return_annotation_is_envelope(func):
+        declared_output_schema = None
+    else:
+        declared_output_schema = action.output_schema
+    action.metadata[DECLARED_OUTPUT_SCHEMA_KEY] = declared_output_schema
```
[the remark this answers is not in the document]
huangjeff5: action.metadata is a read-only property.

py/samples/multipart-tools/src/main.py
```diff
@@ -0,0 +1,64 @@
...
+    """Take a screenshot. ``response()`` is the action result, PNG included."""
+    return response({'ok': True, 'label': label}, parts=[_PNG], metadata={'src': 'lab-cam'})
+
+
+@ai.flow()
+async def ask_about_the_lab() -> str:
+    """Ask the model to check the weather and snap the lab."""
+    res = await ai.generate(
+        prompt='What is the weather in Austin, and take a screenshot of the lab camera.',
+        tools=['weather', 'screenshot'],
```
huangjeff5: Use inline tools.

py/packages/genkit/src/genkit/__init__.py
```diff
@@ -46,9 +46,12 @@ async def my_flow(prompt: str) -> str:
 )
 from genkit._ai._tools import (
     Interrupt,
+    MultipartToolResponse,
+    PartLike,
```
huangjeff5: Remove PartLike.

huangjeff5: Leaving PartLike for a follow-up. This commit is leftover product-pin tests only.

py/packages/genkit-google-genai/src/genkit_google_genai/models/utils.py
```diff
@@ -72,6 +72,35 @@
...
+        return genai.types.FunctionResponsePart.from_bytes(
+            data=part.inline_data.data,
+            mime_type=part.inline_data.mime_type or 'application/octet-stream',
+        )
+    if part.file_data and part.file_data.file_uri:
+        return genai.types.FunctionResponsePart.from_uri(
+            file_uri=part.file_data.file_uri,
+            mime_type=part.file_data.mime_type,
+        )
+    return None
```
pavelgj, line 93: nit / nice-to-fix (non-blocking): this only maps inline_data/file_data and returns None for everything else, so non-media parts passed via response(parts=[...]) (text/data/reasoning/resource) get silently dropped here. If media-only is intentional, a one-line comment or debug log would save someone a confusing debugging session later.

pavelgj (approved):
LGTM. Nice envelope design, and the sample makes the bare-return vs response() distinction clear. I checked the two bot-flagged items (issubclass on generic aliases, action.metadata possibly None) and neither is real in this diff.

py/packages/genkit-google-genai/src/genkit_google_genai/models/utils.py
huangjeff5: Media-only is intentional: Gemini's wire schema for `FunctionResponse.parts` only accepts media attachments (`inline_data` / `file_data`), while structured tool output lives in `FunctionResponse.response` (`output`). Dart does the same `.where((p) => p.isMedia)` filtering, and Go hard-errors with `ErrInvalidArgument` on non-media parts.

Added the docstring explaining the wire format constraint and a `logger.debug` in `_function_response_part` so dropped non-media parts are visible when debugging.

PR #6173, opened 2026-08-26
feat(go/ai): return partial responses from the generate loop

apascal07:
Returns the generate loop's completed work alongside its classified errors: Go's multiple return values carry both, where JS attaches the response to a thrown `GenerationResponseError`. Base of a three-PR train, on the shared schema from #6202.

## A failed generate returns a conversation you can send again

Once the request has resolved, an error comes back with a non-nil partial `ModelResponse` instead of discarding everything the loop did:

```go
    ai.WithPrompt("Plan the trip."),
    ai.WithTools(searchFlights, bookHotel),
    ai.WithMaxTurns(3),
)
if err != nil && resp != nil {
    resp.History()     // The completed rounds. Send them back to retry the failed step.
    resp.FinishReason  // Failed if something broke, aborted if the caller stopped it.
    resp.FinishMessage // The cause, e.g. "exceeded maximum tool call iterations (3)".
    resp.Error         // The same cause, classified: {status, message, details}.
}
```

`History()` ends at a turn seam: the completed `[model with tool requests, tool with every response]` rounds, and nothing from the turn that failed. That is a seam rather than a high-water mark because no provider accepts a conversation ending in a tool request nothing answered. So a failed tool discards the whole round including the model message that opened it, `WithMaxTurns` discards the round it refused to run, a model call that dies mid-stream discards its message, and `Message` is nil on all three. What is left is a conversation the caller can re-send without repeating the tool calls that already succeeded.

- **One rule for the finish reason**: `FinishReasonFailed` when something broke (a model call, a tool), `FinishReasonAborted` when the caller stopped the loop (a cancelled context, an expired deadline, or a limit it set such as `WithMaxTurns`, whose `ErrMaxTurnsExceeded` is an ABORTED subtype). The cause is the `FinishMessage` either way.
- **A tool that stopped because the call's context ended is a caller stop**, so it carries CANCELLED rather than `ErrToolFailed`, whose INTERNAL status would otherwise win `status.Classified`. The check keys on the call's context, so a tool that returns `context.Canceled` on its own terms is still a tool failure.
- **A failed tool reports as soon as its error arrives**, while its siblings finish detached (`TestGenerateNoGoroutineLeak` pins that they are not leaked). Their outcomes go with the discarded round.
- **A completed response that post-processing rejected** (`ErrInvalidOutput`) is not a loop stop: it keeps the model's own message and finish reason, since the raw output is usually what the caller needs to see.
- **A resume whose restarted tool interrupts again** keeps `FinishReason` interrupted under its FAILED_PRECONDITION error. An interrupt tip is the one non-seam shape a partial carries, because it is answered with `WithResume` rather than re-sent.
- **Errors before the request resolves** (unknown model or tool, invalid options) still return a nil response.

`Error` is a `RuntimeError` on the shared schema, the same shape a session snapshot carries, and it is what makes a partial useful once it has travelled: `FinishMessage` is a string, and nobody reading a response out of a trace or a persisted turn should be matching one. Both stream helpers yield the partial beside their error too, with the value still `Done` and still carrying `Response`, so a consumer that streamed a tool loop reads the same conversation the non-streaming call returns.

The contract survives `WrapGenerate` hooks that drop the response on the way up: the loop records each failing turn's partial and restores it, or synthesizes one from the conversation entering the turn for errors raised outside one.

## The action boundary stops discarding the value

```go
// Before: any error erased what the function produced.
r, err := a.runWithTelemetry(ctx, input, cb, a.fn, nil)
if err != nil {
    return base.Zero[Out](), err
}
return r.Result, nil
```

Three places threw the partial away before anyone could read it. `Action.Run` zeroed its output whenever err was non-nil, `RunJSON` marshaled nothing, and `RunInNewSpan` set `genkit:output` only on the success branch. All three now carry whatever the function returned, so the partial reaches an in-process caller, the JSON surface, and a trace, which is what puts a failed generate's conversation in front of the Dev UI. The value is the function's own, so an action that returns nothing on error still yields the zero one and still stamps no output; the JSON and span paths guard on that explicitly rather than writing a null.

Two consequences beyond generate. An output that failed the action's own schema validation now comes back with its error instead of being erased, which is the same call the loop makes for `ErrInvalidOutput`. And any flow whose function returns a value alongside an error now hands that value to its caller.

## Interrupt replays restore the full tool response

A `pendingOutput` replay used to rebuild a resolved sibling's tool response from its output alone, dropping a multipart response's `Content` and its `Metadata`. JS drops them too. The loop now stashes both beside `pendingOutput` (`pendingContent`, `pendingMetadata`) and restores them on replay, tolerating the generic JSON a persistence round-trip leaves behind. `pendingOutput` itself is unchanged, for cross-SDK parity; the two new keys are Go-first and candidates for JS adoption.

## Resumed tool messages are request-ordered

`handleResumeOption` resolved directives concurrently and emitted tool responses in completion order, while first-run tool messages are deliberately request-ordered. Both paths now order responses by their requests' positions in the model message (`TestResumedToolMessageOrder` pins it). JS is completion-ordered on both paths, an artifact of pushing inside `Promise.all`, so there is no deterministic JS ordering to mirror; models match responses by ref, so the wire is unaffected.

## Not a breaking change

No new exported symbols and no signature changes. `ai.FinishReasonFailed` and `ai.ModelResponse.Error`, which the loop sets, arrive with the schema in #6202. At runtime the change is a strict relaxation: paths that returned a zero value beside an error now return whatever was produced, and a caller that checks err first never reads it. The pattern that could notice is branching on the value without checking the error, and it is worth naming that the action boundary widens this beyond generate: any action, flow, or tool whose function returns a value alongside an error now hands that value to its caller. Relatedly, `GenerateText` returns the partial's text alongside the error on the post-processing path, where it returned `""`.

- **No salvage of the failing turn's output.** The seam rule is uniform, so a single-turn call that dies mid-stream hands back no text on the response. The chunks reached the streaming callback and the trace; keeping them on a response the caller might re-send would produce a conversation providers reject, and the response would have to be two shapes instead of one.
- **The reflection API's error response has no result slot**: on failure the server writes `{"error": {message, code, details}}`, a shape the dev UI validates against, so a "Run" there shows the partial in the trace view and not in its result pane. Widening that envelope is a dev UI protocol change, not a Go one.
- **Agents are unchanged**: a failed turn keeps the last good session snapshot, as the conformance spec pins. Persisting the partial is a separate, spec-visible decision, taken in #6196.

### DIFF
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -1492,6 +1492,19 @@ func clone[T any](obj *T) *T {
 	return &newObj
 }
 
+// toolFailureError classifies a tool's error for the loop. A tool that failed
+// on its own terms is [ErrToolFailed], which the loop reports as a failed
+// generation. A tool that stopped because the call's context ended is not a
+// tool failure at all: the cancellation is returned with its own status, so
+// the partial response carries [FinishReasonAborted] rather than blaming the
+// tool for a stop the caller asked for.
+func toolFailureError(ctx context.Context, name string, cause error) error {
+	if ctx.Err() != nil {
+		return status.Errorf(status.ErrCancelled, "tool %q stopped: %w", name, cause)
+	}
+	return status.Errorf(ErrToolFailed, "tool %q failed: %w", name, cause)
+}
+
 // toolRunnerFunc runs a tool through the WrapTool hook chain and returns the
 // raw [MultipartToolResponse]. Returned by [buildToolRunner].
 type toolRunnerFunc = func(ctx context.Context, tool Tool, req *ToolRequest) (*MultipartToolResponse, error)
@@ -1622,7 +1635,7 @@ func handleToolRequests(ctx context.Context, r api.Registry, req *ModelRequest, 
 					return
 				}
 
-				resultChan <- result[*MultipartToolResponse]{index: idx, err: status.Errorf(ErrToolFailed, "tool %q failed: %w", toolReq.Name, err)}
+				resultChan <- result[*MultipartToolResponse]{index: idx, err: toolFailureError(ctx, toolReq.Name, err)}
 				return
 			}
 
@@ -2241,7 +2254,7 @@ func handleResumedToolRequest(ctx context.Context, r api.Registry, genOpts *Gene
 						}, nil
 					}
 
-					return nil, status.Errorf(ErrToolFailed, "tool %q failed: %w", restartPart.ToolRequest.Name, err)
+					return nil, toolFailureError(ctx, restartPart.ToolRequest.Name, err)
 				}
 
 				newToolReq := clone(p)
--- a/go/ai/generate.go
+++ b/go/ai/generate.go
@@ -372,6 +372,16 @@ func responseError(cause error) *status.Error {
 	return e
 }
 
+// callerStopped are the statuses that say the caller ended the loop rather
+// than something breaking inside it: it cancelled the context, its deadline
+// expired, or it set a limit the loop reached ([ErrMaxTurnsExceeded] is an
+// ABORTED subtype). They report [FinishReasonAborted].
+var callerStopped = map[status.Name]bool{
+	status.Cancelled:        true,
+	status.DeadlineExceeded: true,
+	status.Aborted:          true,
+}
+
 // failurePartial builds the partial [ModelResponse] that accompanies the
 // error when the generate loop stops before it produced a final response.
 //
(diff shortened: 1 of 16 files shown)


go/ai/generate.go
```diff
@@ -1478,6 +1690,31 @@ func handleToolRequests(ctx context.Context, r api.Registry, req *ModelRequest,
...
+		// The partial message is assembled on an isolated copy of the model
+		// message, since the stragglers keep writing into revisedMsg after
+		// this returns. Each received index is safe to take: its goroutine's
+		// revision (pendingOutput, interrupt metadata) was published by the
+		// channel send, and no other goroutine writes that element.
+		partialMsg := clone(resp.Message)
+		for _, idx := range receivedIndexes {
+			partialMsg.Content[idx] = revisedMsg.Content[idx]
+		}
+		return nil, partialMsg, toolErr
```
pavelgj: This error path does careful work to assemble `partialMsg` (the non-blocking `drain:` loop, the isolated `clone(resp.Message)`, the per-index copy), but the only caller discards it: on `err != nil` it returns `failurePartial(resp, req, err)` built from `resp`/`req` and never reads the returned `*Message`. So this block, plus its "isolated copy / published by the channel send" reasoning, produces a value nobody consumes. The channel is also fully buffered (`len(toolRequests)`), so the drain isn't needed to avoid a goroutine leak.

Given the seam rule intentionally drops the whole failing round, could this just be `return nil, nil, toolErr` with the doc comment trimmed to match? If it's kept deliberately for a future consumer, a one-line note saying so would help.

pavelgj (approved):

PR #6196, opened 2026-08-26
feat(go/ai/exp): make a failed agent turn a resume point

apascal07:
Makes a failed agent turn a resume point instead of a dead end. Base of a two-PR train on top of #6173's partial response, which supplies the conversation: an hour of tool calls no longer dies with the model call that blinked.

## A failed turn writes a `failed` snapshot, and resume accepts it

```go
out, _ := agent.RunText(ctx, "Book the full itinerary.")
// out.FinishReason == failed, out.Error carries the classified cause.

snap, _ := agent.GetSnapshot(ctx, out.SnapshotID)
snap.Status         // aix.SnapshotStatusFailed, no longer a dead end
snap.Error          // the same classified failure
snap.State.Messages // what the turn committed: the tool rounds it completed
```

The prompt loop folds #6173's partial into the session, and because that partial ends at a turn seam, what it commits is a conversation the model can be called on again. `resumeSessionFrom` now rejects `aborted` and `pending` only; #6213 takes `aborted` too.

- **Which failures commit**: any the generate call produced a partial for, which is every failure past the model call. A turn that fails before it (an invalid input, a render failure) rolls back, because the only thing in the session by then is the input it just rejected. Either way the newest snapshot is the latest resumable state, which is the rule `AgentOutput.SnapshotID` reports on a failure.
- **Custom agents opt in** by returning a `TurnResult` alongside the turn's error, mirroring generate's `(resp, err)`; a bare error keeps the discard-the-turn contract the spec pins.
- **Client-managed agents** get the same resume point inline: `lastGoodState` advances on every committed turn, failed or not, so the failed output's `State` is what the turn committed.
- **A restarted tool that interrupts again** commits the completed row a first-run interrupt commits, not a failed one. `ai.Generate` reports the second interrupt with a FAILED_PRECONDITION because that call's caller asked for a completed generation; the agent's caller did not, and the tip handed back is the same interrupt `Resume` answers. Calling it failed pointed clients at the empty-input re-attempt, which cannot work there: that tip ends on a model message holding an unanswered tool request, which is not a turn seam.

## The turn-end snapshot write outlives the turn's context

`snapshotTurnEnd` passed the turn's own context to `SaveSnapshot`. A turn that ends because the invocation's context was cancelled is exactly the turn whose snapshot a client needs, and that write ran on the context that had just died. Persistence is best-effort, so the store error was logged and the row silently did not exist. The write now runs on `context.WithoutCancel(ctx)`, still traced and logged under `ctx`.

Every in-memory store ignores its context, so no test could see this. `ctxHonoringStore` wraps one and rejects a write on a dead context the way a store doing real I/O does.

## Re-attempting is an input with no payload

```go
retried, _ := agent.Run(ctx, &aix.AgentInput{}, aix.WithSessionID[any](out.SessionID))
```

The turn runs again on the committed messages, so the tool calls that already succeeded are not repeated. No new field carries this: the "message or resume is required" check relaxes to reject an empty input only when the session has no messages to continue. A new message works on that snapshot too, and rewinding past the failure is a resume from the previous snapshot ID.

**Whether a failure is worth another attempt is the client's decision.** The runtime writes the status on the row and never classifies it. A `RESOURCE_EXHAUSTED` wants a wait, an `INVALID_ARGUMENT` wants a different message, and a `FAILED_PRECONDITION` from a tool guard may want neither; only the caller knows which, and it already has the status.

## Conformance rides the shared spec behind a `requires` gate

Six cases land in `tests/specs/agent.yaml` behind the `requires: [resumable-failures]` gate this PR adds to the spec format, and only the Go harness declares that capability; the JS and Python harnesses learn to read the field and skip. Because every harness skips what it does not recognize, a misspelled capability would skip in all of them and be reported by none, so the spec carries its own top-level `capabilities` registry and all three harnesses fail a `requires` entry absent from it. `docs/agents-conformance-testing.md` gains the format entry, the capability, and the two fixtures it needs. Go runs 45/45; JS 39 passed, 6 skipped; Python 53 passed, 6 skipped.

## Behavior shifts inside exp

The `SessionRunner.Run` callback contract widens: a non-nil `TurnResult` beside an error commits the turn. The package is exp and may change in minor releases, but four shifts are worth naming for anyone already on it.

- Resume from a `failed` snapshot succeeds where it raised FAILED_PRECONDITION.
- `AgentInput{}` runs a turn where it was always INVALID_ARGUMENT.
- A custom agent that already returned a non-nil `TurnResult` together with an error, previously ignored, now commits its failed turns.
- A restarted tool that interrupts again resolves the invocation as interrupted where it failed it.

- **No mid-round resume.** A failed tool round is re-run whole, siblings that succeeded included, because the round cannot be handed back half-answered. Keeping their outputs would need the failed request answered too, which is what `WithSoftFailure` is for.
- **No retryable-status set in the framework.** A canonical list looks tempting here, but it would put the runtime in the business of judging a failure it did not raise. `middleware.Retry` keeps its own list for re-invoking a model call; the agent layer records and hands off.
- **No error redaction seam.** A `failed` row persists the full error text, and `WithStateTransform` shapes only `State`. `convertKeepText` and the failed detached row both predate this PR, so what changes is how often a raw message is stored, not what is stored; the hook is a follow-up.
- **Crash durability**: only a graceful failure commits, so a killed process mid-turn still loses the turn (a detached one surfaces as `expired`). Mid-turn checkpointing is a separate feature.
- **Telemetry counts the failed turn as failed** even when the next attempt succeeds; a recovered-failure marker is a follow-up shared with the soft-failure work in #6205.
- **JS and Python adoption**: the spec cases and harness gating are in place; the runtimes implement later.

### DIFF
--- a/docs/agents-conformance-testing.md
+++ b/docs/agents-conformance-testing.md
@@ -26,10 +26,20 @@ tests:
   - name: <string>              # Human-readable test name
     description: <string>       # Optional description
     agent: <string>             # Name of the harness-provided agent
+    requires: [<capability>]    # Optional capability gate; see below
     steps:                      # Ordered sequence of operations
       - type: send | getSnapshotData | abort | waitUntilCompleted
         ...                     # Fields depend on step type
 ```
+
+`requires` lists the capabilities a test depends on. Each harness declares
+the capabilities its runtime implements and **skips** (not fails) any test
+naming one it does not, so the shared spec can carry cases for features an
+SDK has not adopted yet. Known capabilities:
+
+| Capability | Meaning |
+|------------|---------|
+| `resumable-failures` | A failed turn commits what the generate call left at its last turn seam and persists it as a `failed` snapshot carrying the error; resume accepts that snapshot, and an input with no payload of its own re-attempts the turn. Gates model `error` entries, the empty input, a `failed` snapshot as a resume target, and the `promptAgentWithToolsAndStore` fixture. Implemented by: Go. |
 
 ### Step Types
 
@@ -42,8 +52,8 @@ Sends inputs to the agent via its bidirectional streaming interface (e.g.
 |-------|------|-------------|
 | `type` | `"send"` | Required. |
 | `init` | `AgentInit` | Initialization payload. May contain `snapshotId`, `state`, or be empty `{}`. |
-| `inputs` | `AgentInput[]` | Ordered list of inputs to send. Each may contain `messages`, `resume` (with `respond` and/or `restart`), and/or `detach`. |
-| `modelResponses` | `GenerateResponseData[]` | Pre-programmed responses for the programmable model, one per `generate` call made by the agent. |
+| `inputs` | `AgentInput[]` | Ordered list of inputs to send. Each may contain `messages`, `resume` (with `respond` and/or `restart`), and/or `detach`. An input with none of those continues the conversation already in the session (requires `resumable-failures`). |
+| `modelResponses` | `GenerateResponseData[]` | Pre-programmed turns for the programmable model, one per `generate` call made by the agent. An entry whose `error` is set (`{ status, message }`) fails that call with a classified error rather than returning (requires `resumable-failures`). |
 | `streamChunks` | `GenerateResponseChunkData[][]` | Optional. Pre-programmed streaming chunks, indexed by model call. Each inner array is emitted as a stream before the corresponding `modelResponses` entry. |
 | `expectChunks` | `AgentStreamChunk[]` | **Strict ordered** list of expected stream chunks. |
 | `expectOutput` | Object | Expected fields on the `AgentOutput`. See [Output Assertions](#output-assertions). |
@@ -185,6 +195,7 @@ via the `modelResponses` / `streamChunks` fields in `send` steps.
 | `promptAgentWithTools` | A prompt agent with `testTool` registered. Client-managed state. |
 | `promptAgentWithInterrupt` | A prompt agent with `interruptTool` registered and a server-managed store (for snapshot-based resume). |
 | `promptAgentWithRestartTool` | A prompt agent with `restartTool` registered and a server-managed store. Used for `resume.restart` tests. |
+| `promptAgentWithToolsAndStore` | A prompt agent with `testTool` and `flakyTool` registered and a server-managed store. Used for `resumable-failures` tests; only required by harnesses declaring that capability. |
 
 #### Custom agents (hardcoded behavior)
 
@@ -209,6 +220,7 @@ is not needed for tests targeting these agents.
 | `testTool` | A simple tool | `{}` (empty) | `"tool called"` (string) |
 | `interruptTool` | An interrupt tool | `{ query: string }` | `{ answer: string }` |
 | `restartTool` | A tool that requires confirmation; throws `ToolInterruptError` on first call, succeeds when `resumed` metadata is provided | `{ action: string }` | `{ result: string }` |
+| `flakyTool` | Fails its first call in each test with an UNAVAILABLE error (`flaky tool failed`), succeeds after with `"tool recovered"`. Requires `resumable-failures`. | `{}` (empty) | `"tool recovered"` (string) |
 
 ### Programmable Model
 
--- a/docs/agents-conformance-testing.md
+++ b/docs/agents-conformance-testing.md
@@ -22,6 +22,7 @@ The pattern mirrors `tests/specs/generate.yaml` for the Generate API.
 ### Top-Level Structure
 
 ```yaml
+capabilities: [<capability>]    # Every name `requires` may use; see below
 tests:
   - name: <string>              # Human-readable test name
     description: <string>       # Optional description
@@ -35,7 +36,13 @@ tests:
 `requires` lists the capabilities a test depends on. Each harness declares
 the capabilities its runtime implements and **skips** (not fails) any test
 naming one it does not, so the shared spec can carry cases for features an
-SDK has not adopted yet. Known capabilities:
+SDK has not adopted yet.
+
+Because every harness skips what it does not recognize, a misspelled name
+would otherwise skip the test in all of them and be reported by none. The
+top-level `capabilities` list is the spec's own registry of valid names, and
+a `requires` entry absent from it **fails** in every harness. Adding a
+capability means adding it there and to the table below. Known capabilities:
 
 | Capability | Meaning |
 |------------|---------|
--- a/go/ai/exp/agent.go
+++ b/go/ai/exp/agent.go
@@ -122,22 +122,30 @@ type SessionRunner[State any] struct {
 	// to the fn goroutine (Run and its synchronous onEndTurn callback) until
 	// fn completes, after which the terminal paths read it with a
 	// happens-before edge through the fnDone channel, so no lock is needed.
-	// The same confinement applies to lastTurnFailed and lastGoodState; the
-	// terminal paths that read them (handleFnDone and the detach-failure
-	// paths) all wait on fnDone first.
+	// The same confinement applies to lastTurnErr, lastTurnCommitted, and
+	// lastGoodState; the terminal paths that read them (handleFnDone and the
+	// detach-failure paths) all wait on fnDone first.
 	lastTurnFinishReason AgentFinishReason
 
-	// lastTurnFailed reports whether the most recent turn ended in error.
-	// Set by endTurn each turn.
-	lastTurnFailed bool
+	// lastTurnErr is the error the most recent turn ended with, or nil when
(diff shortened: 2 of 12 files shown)


go/ai/exp/agent.go
```diff
@@ -1146,23 +1188,23 @@ func newAgentRuntime[State any](
 // [Responder] before each Send returns, and fn returned before this
 // runs, so there is no in-flight router work to wait out.
 //
-// The snapshot is skipped when the turn failed (the live state holds the
-// turn's partial mutations) and when detach has landed (snapshotTurnEnd
-// observes the suspension under snapMu; the pending row already captures
-// the invocation and a single finalize rewrite records the cumulative
-// state once the queued inputs drain).
+// The snapshot is skipped when the turn failed without committing (the live
+// state holds mutations nothing can be resumed from) and when detach has
```
huangjeff5, line 1192: `finalizePendingSnapshot` captures `finalState := *rt.session.State()` — the live session state — and writes it under `SnapshotStatusFailed` whenever `fnErr != nil`. But per this PR's own contract, a turn that fails with a bare error (no `TurnResult`) rolls back: its partial mutations stay in the live state but must not become a resume point ("a bare error keeps the discard-the-turn contract the spec pins"). The non-detached path honors that — `emitTurnEnd` skips the snapshot because "the live state holds mutations nothing can be resumed from." The detach finalize path doesn't: a detached invocation whose last turn fails uncommitted gets those exact mutations persisted on the failed row.

Before this PR that was harmless because `resumeSessionFrom` rejected failed rows outright. Now it admits them, so the one path that persists discarded-turn state is also newly resumable. `TestAgent_Detach_FlowErrorsBecomesError` exercises resume-from-failed-detached but with an agent that mutates nothing before failing, so the state leak isn't observable there.

Suggested fix: on `fnErr != nil` without a committed turn, finalize the row with the last committed state (what `lastGoodState` / the last turn snapshot holds) rather than the live state — mirroring what `failedOutput` reports for the non-detached failure. A test where the failing turn mutates state before returning a bare error would pin it.

go/ai/exp/agent.go
```diff
@@ -2767,7 +2814,17 @@ func agentLoop[State any](r api.Registry, prompt ai.Prompt, defaultInput any) Ag
...
+				// [ai.Generate]), so it is a conversation the caller can
+				// continue. Fold it into the session and commit the turn as a
+				// resume point: the [TurnResult] beside the error is what says
+				// so; see [SessionRunner.Run]. Without a partial the call
+				// never reached the model, and the turn rolls back instead.
+				if modelResp == nil || modelResp.Request == nil {
+					return nil, fmt.Errorf("generate: %w", err)
+				}
+				sess.SetMessages(turnSessionMessages(modelResp.History()))
+				return &TurnResult{FinishReason: AgentFinishReasonFailed}, fmt.Errorf("generate: %w", err)
```
huangjeff5, line 2838: The body names this behavior ("a restarted tool that interrupts again now commits its interrupt tip … so the new interrupt state stays answerable with Resume") — the commit itself is clearly intended. The issue is how it's labeled: `agentLoop` classifies every generate error that produced a partial the same way, `TurnResult{FinishReason: AgentFinishReasonFailed}`, which `snapshotTurnEnd` writes as `SnapshotStatusFailed`. The re-interrupted restart is in that bucket — #6173 hands it back as a FAILED_PRECONDITION error whose partial has `FinishReason` interrupted and an answerable interrupt tip — so the snapshot a client reads says failed/failed even though the right next move is Resume.

The mislabel makes the documented recovery recipe wrong for this class: the README/sample recipe for a failed snapshot is the empty-input re-attempt, which here re-runs the turn (re-executing the restarted tool) instead of answering the waiting interrupt. A first-run interrupt commits as completed/interrupted; the re-interrupt should presumably look the same, with the error still on the row if you want it recorded.

There's also no test exercising the restart-re-interrupts-again path through `agentLoop` — the named behavior in the body currently rides on the #6173 plumbing alone.

go/ai/exp/agent.go
```diff
@@ -2734,6 +2770,17 @@ func agentLoop[State any](r api.Registry, prompt ai.Prompt, defaultInput any) Ag
 			// conversation does not ride along into tools and prompts
 			// invoked inside the generate loop.
 			history := sess.Messages()
+
+			// An input with no payload of its own runs the turn on the
+			// conversation as it stands, which is how a failed turn is
+			// re-attempted: the failed snapshot holds the messages the turn
+			// committed, and the model is called on them again. There has to
+			// be something to continue.
+			if !hasInputPayload(input) && len(history) == 0 {
```
huangjeff5, line 2779: The body is precise that the relaxed check "reject[s] an empty input only when the session has no messages to continue" — so empty input on a healthy completed conversation running a fresh model call on the same history is the documented, intended semantics, not an accident. But the five new spec cases only cover the failed-tip and empty-session shapes; nothing pins the completed-tip case.

Since JS and Python will implement `resumable-failures` from the spec rather than from the Go source, the completed-tip case is exactly where they'll diverge silently (rejecting it is the natural conservative reading of "re-attempt a failed turn"). A sixth spec case asserting empty-input-on-completed-tip runs a turn would close it.

go/ai/exp/agent.go
```diff
@@ -412,13 +449,18 @@ func (s *SessionRunner[State]) snapshotTurnEnd(ctx context.Context, finishReason
...
+	}
+	saved, err := s.store.SaveSnapshot(context.WithoutCancel(ctx), s.turnSnapshotID,
 		func(_ *SessionSnapshot[State]) (*SessionSnapshot[State], error) {
 			return &SessionSnapshot[State]{
 				SessionID:    sessionID,
 				ParentID:     parentID,
-				Status:       SnapshotStatusCompleted,
+				Status:       snapStatus,
 				FinishReason: finishReason,
+				Error:        convertKeepText(cause),
```
huangjeff5, line 463: Failed rows now persist `convertKeepText(fnErr)` — deliberately preserving the full error text — and `WithStateTransform` redacts only `State` on the way out, so `snap.Error.Message` is a new persistence surface for raw internal error strings (tool failures wrapping infrastructure details, etc.) that the existing redaction hook can't touch. Related to a #6173 comment: `FinishMessage` there takes the unwrapped `cause.Error()` even for public errors, and this PR is what makes that text durable. Worth deciding before merge whether the state transform should also see the error, or whether public-error wording should win when available.

docs/agents-conformance-testing.md
```diff
@@ -26,11 +26,21 @@ tests:
   - name: <string>              # Human-readable test name
     description: <string>       # Optional description
     agent: <string>             # Name of the harness-provided agent
+    requires: [<capability>]    # Optional capability gate; see below
```
huangjeff5, line 30: All three harnesses skip a test whose `requires` names anything outside their local supported set. That's the right forward-compat behavior for real capabilities, but it means a typo (`resumable-failuers`) silently skips the test in every SDK forever — nobody fails, nobody notices. Cheap guard: keep a known-capabilities list beside the spec (the doc already has the table) and have harnesses fail on a name outside it while still skipping known-but-unsupported ones.

huangjeff5 (changes requested):
Comment 1 is a real defect in the new contract; the rest are ship-with-changes grade.

Every comment below was verified directly against the PR diff and the head-branch source (`ap/go-agent-recoverable-turns`), not inferred, and cross-checked against the PR body and the author's own code comments so none restate something the PR already anticipates. For context: I also ran the conformance claims to ground — Go 44/44, JS 39+5 skip, Python 53+5 skip all reproduce exactly.

go/ai/exp/agent.go
apascal07: Correct, and fixed one PR up in #6213. `finalizePendingSnapshot` there resolves the row through `SessionRunner.committedState`, which falls through the live state, the copy `captureLastGood` took at the last committed turn, the last turn-end snapshot, and the state the invocation began with.

That last fallback closes a hole the narrower fix would still have had: a fresh session that detaches on its first input, whose turn never commits, has no last-good copy in memory and no parent row to read, so it kept the very mutations the rollback was meant to shed.

go/ai/exp/agent.go
apascal07: Real, and a follow-up rather than this PR. `convertKeepText` and the failed detached row are both on `main` today, so what lands here is another call site, not the surface itself: attached failed turns now write one too, which changes how often raw text is stored rather than what is stored.

Both candidate seams cost more than this PR should carry. Widening `StateTransform` to also take the error breaks its signature for anyone already on it, and a second transform option is new public surface on a package whose point is to stay small while it settles. Filing it separately so the shape can be argued on its own.

huangjeff5 (approved):

PR #6209, opened 2026-08-27
fix(js/plugins/a2ui): reconstruct prior surfaces as a2ui blocks in history

pavelgj:
Replace surface summarization with canonical a2ui block reconstruction when replaying inbound envelopes. Previously, assistant surfaces were summarized to a sentinel string like `[rendered UI surface]`, which taught the model to emit that literal text instead of real blocks.

Now consecutive surface envelopes are grouped into a single fenced a2ui block, and concrete surface ids are rewritten back to the placeholder so the parser mints a fresh id per turn rather than reusing prior surfaces. Client-synthesized action envelopes remain short text summaries.

### DIFF
--- a/js/plugins/a2ui/src/middleware.ts
+++ b/js/plugins/a2ui/src/middleware.ts
@@ -46,6 +46,7 @@ import type {
 } from 'genkit/model';
 import {
   DEFAULT_CATALOG_ID,
+  SURFACE_ID_PLACEHOLDER,
   renderCatalogInstructions,
   type A2uiCatalog,
 } from './catalog.js';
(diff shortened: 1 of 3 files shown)


js/plugins/a2ui/src/middleware.ts
```diff
@@ -386,29 +387,94 @@ function sanitizeInboundA2ui(req: GenerateRequest): GenerateRequest {
...
+    if (pendingSurface.length === 0) return;
+    // Rewrite concrete surface ids back to the placeholder the model originally
+    // wrote (the middleware swapped in a real id on the way out). Replaying the
+    // real id would let the model copy it into a fresh `createSurface`, and the
+    // parser passes an explicit pre-existing id through unchanged — so a brand
+    // new answer would reuse (and overwrite, in place) the prior surface instead
+    // of rendering a new one. Using the placeholder makes the parser mint a
+    // fresh id per turn, so each turn is a distinct surface.
+    const placeheld = pendingSurface.map(withPlaceholderSurfaceId);
+    out.push('```a2ui\n' + JSON.stringify(placeheld) + '\n```');
```
apascal07: Replayed history is now unbounded. Each prior surface used to cost ~22 bytes (`[rendered UI surface]`); it now costs the full envelope JSON, re-sent on every subsequent model call, so the total grows quadratically with turn count. An 8-turn session rendering a ~6 KB surface per turn carries ~42 KB of a2ui JSON in the 8th request alone.

The trade is worth it for the most recent surface, which is the one the model imitates. Older ones buy much less. Worth replaying only the last N surfaces verbatim and summarizing the rest?

js/plugins/a2ui/src/middleware.ts
```diff
@@ -386,29 +387,94 @@ function sanitizeInboundA2ui(req: GenerateRequest): GenerateRequest {
...
+}
+
+/**
+ * Returns a copy of `env` with any concrete `surfaceId` replaced by the
+ * {@link SURFACE_ID_PLACEHOLDER}. Used when reconstructing prior surfaces for
+ * history so a replayed surface reads exactly as the model first wrote it (with
+ * the placeholder), and the parser mints a fresh id for it on the next turn
+ * rather than the model copying a real id and reusing the old surface.
+ */
+function withPlaceholderSurfaceId(env: A2uiEnvelope): A2uiEnvelope {
```
apascal07: This rewrites every `surfaceId` in a group to the same placeholder, so envelopes that target different surfaces merge into one block the model cannot tell apart.

It is reachable: `normalizeEnvelope`/`swapSurfaceId` in `parser.ts` only substitute when the id is undefined, empty, or the placeholder, so a model that writes its own concrete ids keeps them. One part can then hold `createSurface: "chart"` plus `updateComponents: "legend"`, and the replay reads as a single surface. The same rewrite turns a standalone `deleteSurface` into `deleteSurface: { surfaceId: "SURFACE_ID" }`, which names nothing.

Grouping by concrete `surfaceId` (one block per distinct id, placeholder applied per group) keeps them separate.

js/plugins/a2ui/src/middleware.ts
```diff
@@ -386,29 +387,94 @@ function sanitizeInboundA2ui(req: GenerateRequest): GenerateRequest {
...
+ *   surface as history therefore shows the model its own UI output in the exact
+ *   format it is asked to produce, reinforcing correct behavior. (Summarizing it
+ *   to a sentinel like `[rendered UI surface]` instead taught the model to emit
+ *   that literal string in place of a real block.)
+ * - Client-synthesized `action` envelopes never had a block form, so they
+ *   become a short text summary the model can reason about.
+ *
+ * Consecutive surface envelopes are grouped into a single block (one surface is
+ * usually several envelopes: create + update(s)). Unknown envelope shapes are
+ * dropped; the caller keeps the message non-empty when everything drops.
```
apascal07: "the caller keeps the message non-empty when everything drops": the caller does not do this. `sanitizeInboundA2ui` (line 378) just skips the push, so a message whose only part is an a2ui part with an empty or all-unrecognized envelope array goes downstream with `content: []`. `a2uiPart([])` reaches this, since `isA2uiPart` only checks `Array.isArray(data.envelopes)`. Gemini and Vertex reject empty content.

Either drop such messages in `sanitizeInboundA2ui` or fix the comment.

js/plugins/a2ui/src/middleware.ts
```diff
@@ -386,29 +387,94 @@ function sanitizeInboundA2ui(req: GenerateRequest): GenerateRequest {
...
+      flushSurface();
       const a = env.action;
       const ctx =
         a.context && Object.keys(a.context).length
           ? ` context=${JSON.stringify(a.context)}`
           : '';
-      lines.push(`[UI action "${a.name}" on surface ${a.surfaceId}${ctx}]`);
-    } else if ('createSurface' in env && env.createSurface) {
-      lines.push(`[UI surface ${env.createSurface.surfaceId} created]`);
+      out.push(`[UI action "${a.name}" on surface ${a.surfaceId}${ctx}]`);
```
apascal07, line 460: The action summary keeps the concrete `surfaceId` while the surface block above is placeheld, so the two branches of this function apply opposite id policies.

If the model copies `s1` from here into an `updateComponents`, `finalizeBlock` takes the "explicit, pre-existing surface id" branch: no synthesized `createSurface`, no `validateRoot`. That is the id-reuse path the placeholder rewrite exists to prevent, and it bypasses the root check the catalog instructions assume.

apascal07 (approved):

js/plugins/a2ui/src/middleware.ts
pavelgj: Agreed. This is really a general context-compression concern and shouldn't be solved in the a2ui middleware; a dedicated context-compression pass is the right place to bound replayed history. Leaving replay verbatim here.

js/plugins/a2ui/src/middleware.ts
pavelgj: Fixed by dropping the placeholder rewrite entirely. Replayed history now keeps the real surface ids verbatim, so envelopes targeting different surfaces stay distinguishable (and a standalone deleteSurface keeps a real id). Reuse is prevented at the parser instead: finalizeBlock now forces every envelope in a createSurface-bearing block onto a single freshly-minted id (forceSurfaceId), so a copied id can never overwrite a prior surface.

js/plugins/a2ui/src/middleware.ts
pavelgj: Good catch. Fixed the code rather than the comment: sanitizeInboundA2ui now drops a message that sanitizing empties out (its only part was an a2ui part whose envelopes all summarized to nothing) instead of forwarding content: []. Corrected the summarizeA2uiPart doc comment to match, and added a regression test ("drops a message emptied by sanitizing instead of sending empty content").

js/plugins/a2ui/src/middleware.ts
pavelgj: Resolved by the same redesign. Both branches now keep real ids (surface blocks are no longer placeheld), so there are no longer two opposite id policies. The id-reuse path this worried about is now blocked at the parser: any block containing a createSurface has every envelope forced onto a freshly-minted id, and validateRoot still runs for it. An update-only block targeting an explicit existing id remains the intended incremental-update path that action correlation depends on. Covered by the new "a new render never reuses a surface id copied from history" test.

PR #6213, opened 2026-08-27
feat(go/ai/exp): make an aborted agent run a resume point

apascal07:
Third of the train, on top of [#6173](https://github.com/genkit-ai/genkit/pull/6173) and [#6196](https://github.com/genkit-ai/genkit/pull/6196). Where #6196 made a failed turn a resume point, this does the same for a run the caller stopped.

## One rule: the caller stopped it, or it broke

`aborted` used to mean one thing, a detached run someone called `Abort` on. It now means every way a caller ends a run, and `failed` means every way one breaks. `callerStopped` is the single decision:

```go
func callerStopped(ctx context.Context, cause error) bool {
	if ctx.Err() != nil {
		return true
	}
	s, ok := status.Classified(cause)
	return ok && (s == status.Cancelled || s == status.DeadlineExceeded || s == status.Aborted)
}
```

The context ending covers the caller's own cancel, the transport closing, a deadline it set, and the abort action, which cancels the work context on the status flip. The ABORTED status covers a limit it set: `ai.ErrMaxTurnsExceeded` is an `ErrAborted` subtype, so a turn that propagates a max-turns error unchanged reports stopped without having to say so in a `TurnResult`.

```go
ctx, cancel := context.WithCancel(ctx)
out, err := chatAgent.Run(ctx, &aix.AgentInput{Message: msg})  // cancel() elsewhere
// out.FinishReason == aix.AgentFinishReasonAborted, snapshot status "aborted"
```

It decides all three terminal writers: `Run`'s failure arm, `failedOutput`, and the detach finalize. `terminalStatus` derives the snapshot status from the turn's finish reason rather than re-deriving the rule at each writer, so the two cannot disagree, and the caller stopping wins over a `TurnResult` the way a late abort already wins over the terminal a detached turn was about to land. `Run` derives the reason from the error it is handed, so the prompt loop's `TurnResult` on a generate failure carries none: forwarding generate's own reason would land a `stop` on a failed row whenever post-processing rejected a response the loop completed. The same rule lands in `ai.failurePartial` on #6173, so the generate loop and the agent snapshot agree.

## An aborted snapshot carries the work it kept

`resumeSessionFrom` rejected `aborted`, and lifting that alone would have resumed an empty conversation: the detach handler writes the pending row with no state, and `Agent.Abort` only flips its status. The state has to come from the finalize, which needs the conversation as of the last turn that finished rather than the live one holding an unfinished turn's mutations.

That is a `SessionRunner` property, since the runner owns every source:

```go
func (s *SessionRunner[State]) committedState(ctx context.Context) *SessionState[State] {
	if s.lastTurnCommitted {
		return s.State()
	}
	if s.lastGoodState != nil {
		return s.lastGoodState
	}
	if snapped := s.snapshotState(ctx, s.lastSnapshotID); snapped != nil {
		return snapped
	}
	return s.initialState
}
```

It always resolves, which is what the finalize needs and what a chain of fallbacks does not give. `lastGoodState` covers the turns committed since a detach, and `captureLastGood` now runs once snapshots are suspended to fill it, since detach suspends the per-turn snapshots that would otherwise hold them. The frozen `lastSnapshotID` covers the turns before the detach. `initialState` is the floor: a fresh session that detaches on its first input and never commits has neither of the other two, and without it the row keeps exactly the mutations it exists to shed.

Resume then accepts the row for the reason it accepts a `failed` one: the conversation ends at a turn seam. It still rejects an aborted row carrying no state, which is one caught between the abort's status flip and the finalize that stamps the conversation on.

## A stopped run hands back its resume point

`Run` returned `(nil, err)` on a cancel, so the caller learned nothing about where the run stopped. It now returns both, as `ai.Generate` hands back a partial response beside the error that ended it.

Getting there needed `AgentConnection.Output` to wait: on a cancel, the drain and the core `Output` both take their cancellation arms while the runtime is still unwinding towards the output that names the snapshot. The wait is bounded by `settleGrace`, so an agent function that ignores its context keeps the caller's escape hatch (`TestAgent_OutputUnblocksOnCancel`) and pays only the delay, having produced no snapshot to wait for.

## One behavior a caller could be reading

A turn that ends on a cancelled context, an expired deadline, or a caller-set limit used to write `failed` and now writes `aborted`. Code branching on `SnapshotStatusFailed` to catch those falls through. Everything else is a relaxation: resume accepts a status it used to reject, and `Run` returns an output where it used to return nil. No exported symbol is added, renamed, or removed.

## What this does not do

- **No abort action for an attached run.** `abortPendingSnapshot` only flips a `pending` row, and only detach writes one. A remote client's lever is still closing the transport.
- **Tool side effects inside the discarded turn are not preserved.** The unfinished round goes whole, so a tool that already ran loses its response and runs again on resume. Answering the unfinished requests with an aborted marker would keep the seam and the completed work, but it breaks the partial-response contract #6173 documents, so it is a follow-up.
- **An expired orphan is still a dead end.** Its raw row is `pending`, which resume rejects, and it carries no state for the same reason an aborted one used to carry none. Same problem class, separate change.
- **JS and Python skip the new conformance cases.** They gate behind `resumable-aborts`, alongside `resumable-failures` from #6196.

### DIFF
--- a/go/ai/exp/agent.go
+++ b/go/ai/exp/agent.go
@@ -65,6 +65,14 @@ const (
 	defaultHeartbeatTimeout = 60 * time.Second
 )
 
+// settleGrace is how long [AgentConnection.Output] waits for a cancelled
+// invocation to produce its result before falling back to reporting the bare
+// cancellation. An agent function that honors its context unwinds in well
+// under this, and the wait is what lets the caller see the snapshot the
+// aborted run stopped at. One that ignores its context has nothing to hand
+// back however long the wait, so the grace expires and the caller escapes.
+const settleGrace = 2 * time.Second
+
 // isHeartbeatExpired reports whether snap is a pending (detached, in-flight)
 // snapshot whose heartbeat is older than timeout, i.e. its background worker is
 // presumed dead. A pending snapshot that has not yet written a first heartbeat
@@ -141,11 +149,12 @@ type SessionRunner[State any] struct {
 
 	// lastGoodState is a deep copy of the session state as of the most
 	// recent committed turn (or the initial state when no turn has committed
-	// yet), kept only for client-managed agents (no store). The
-	// client-managed failure path returns it inline so the caller resumes
-	// from the last committed turn, excluding a later turn that failed
-	// before committing. Nil and unused for server-managed agents, whose
-	// failure path returns the last turn snapshot instead.
+	// yet). It is kept for a client-managed agent (no store), whose failure
+	// path returns it inline, and for a detached one, whose finalize writes
+	// it to the pending row; both then resume from the last committed turn,
+	// excluding a later turn that failed before committing. Nil and unused
+	// for an attached server-managed agent, whose failure path returns the
+	// last turn snapshot instead. See captureLastGood.
 	lastGoodState *SessionState[State]
 }
 
@@ -162,6 +171,15 @@ func (s *SessionRunner[State]) suspendSnapshots() (parentID string) {
 	defer s.snapMu.Unlock()
 	s.snapshotsSuspended = true
 	return s.lastSnapshotID
+}
+
+// snapshotsAreSuspended reports whether a detach has stopped turn-end
+// snapshot writes. Read by captureLastGood in the fn goroutine, so it takes
+// snapMu against the detach handler's write.
+func (s *SessionRunner[State]) snapshotsAreSuspended() bool {
+	s.snapMu.Lock()
+	defer s.snapMu.Unlock()
+	return s.snapshotsSuspended
 }
 
 // TurnResult is the optional return value of a [SessionRunner.Run] per-turn
@@ -319,6 +337,13 @@ func (s *SessionRunner[State]) Run(ctx context.Context, fn func(ctx context.Cont
 			if failedResult != nil && failedResult.FinishReason != "" {
 				reason = failedResult.FinishReason
 			}
+			// The caller stopping the run wins over whatever the turn
+			// reported. The detached path settles the same race the same
+			// way: an abort that lands while a turn is settling keeps the
+			// aborted terminal.
+			if callerStopped(ctx, err) {
+				reason = AgentFinishReasonAborted
+			}
 			s.endTurn(ctx, reason, err, failedResult != nil)
 			return err
 		}
@@ -353,16 +378,19 @@ func (s *SessionRunner[State]) endTurn(ctx context.Context, reason AgentFinishRe
 	s.turnIndex++
 }
 
-// captureLastGood deep-copies the committed session state as the
-// client-managed failure fallback: the state a failed invocation returns
-// inline (see failedOutput), excluding the partial mutations of a later
-// turn that failed before committing. Called once at session start (the
-// initial state is the fallback until a turn commits) and after every
-// committed turn, failed or not. It is a no-op for server-managed agents,
-// whose failure path returns the last turn-end snapshot instead, so they
-// pay no per-turn copy.
+// captureLastGood deep-copies the committed session state as the failure
+// fallback, excluding the partial mutations of a later turn that failed
+// before committing. Called once at session start (the initial state is the
+// fallback until a turn commits) and after every committed turn, failed or
+// not.
+//
+// It is kept for a client-managed agent, whose failed invocation returns it
+// inline (see failedOutput), and for a detached one, whose finalize has no
+// per-turn snapshot to fall back on because detach suspended them. An
+// attached server-managed agent needs neither and pays no per-turn copy: its
+// last turn snapshot is the resume point.
 func (s *SessionRunner[State]) captureLastGood() {
-	if s.store != nil {
+	if s.store != nil && !s.snapshotsAreSuspended() {
 		return
 	}
 	s.mu.RLock()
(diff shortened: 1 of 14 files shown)


go/ai/exp/agent.go
```diff
@@ -1508,9 +1616,51 @@ func convertKeepText(cause error) *status.Error {
...
+// Or the run reached a limit the caller set, which [ai] reports with an
+// ABORTED status ([ai.ErrMaxTurnsExceeded]). A turn that propagates such an
+// error unchanged is stopped, not broken, and reports so without having to
+// say it in a [TurnResult].
+func callerStopped(ctx context.Context, cause error) bool {
+	if ctx.Err() != nil {
+		return true
+	}
+	s, ok := status.Classified(cause)
+	return ok && (s == status.Cancelled || s == status.DeadlineExceeded || s == status.Aborted)
```
huangjeff5: The body pastes this function and is explicit about both arms: context ending covers cancel / transport / deadline / `Abort`, and "the ABORTED status covers a limit it set: `ai.ErrMaxTurnsExceeded` is an `ErrAborted` subtype." It also says this is the same rule as #6173's `failurePartial`, so generate and the agent snapshot agree. None of that is in dispute.

The remaining gap is that the code does not test "is this `ErrMaxTurnsExceeded`." It tests `status.Classified(cause) == Aborted` (or `DeadlineExceeded`, or `Cancelled`). That is a larger set than "a limit the caller set." This repo already classifies things the caller did not set that way: HTTP 409 → `ABORTED`, HTTP 504 → `DEADLINE_EXCEEDED` (`go/core/status`), and `wrapAPIError` in googlegenai stamps the *server's* status so retry middleware can see it. That middleware retries `Aborted` and `DeadlineExceeded` by default. A Gemini 504 or a tool's gRPC contention abort, with the client still connected (`ctx.Err() == nil`), therefore lands as `SnapshotStatusAborted` — "you stopped this run" — which is the class a retry loop is most likely to skip.

On #6173 that only stamped an in-memory `FinishReason`. Here the same map decides a persisted snapshot status.

If the intent is still "same map as generate, on purpose," one sentence in the body that a retry client must not treat `aborted` as "do not retry" would close it. If the intent is the comment as written ("a limit the caller set"), narrow the second arm to `errors.Is(cause, ai.ErrMaxTurnsExceeded)` and leave provider/tool `Aborted` on `failed`.

go/ai/exp/agent.go
```diff
@@ -1902,21 +2060,30 @@ func loadSession[State any](
...
 	case SnapshotStatusAborted:
-		return nil, nil, status.Errorf(status.ErrFailedPrecondition,
-			"snapshot %q was aborted", snap.SnapshotID)
+		// An aborted row is the one terminal shape written twice: the abort
+		// flips the pending row, which carries no state, and the finalize
+		// that follows stamps the state onto it. A row still between the two
+		// (the process died, or the finalize write failed) holds nothing, and
+		// resuming it would silently hand back an empty session in place of
+		// the conversation the caller asked to continue.
+		if snap.State == nil {
```
huangjeff5, line 2108: The body already says resume "still rejects an aborted row carrying no state, which is one caught between the abort's status flip and the finalize that stamps the conversation on." The comment on the reject names the two situations you had in mind: "the process died, or the finalize write failed." The harness `waitUntilCompleted` now waits for write 2 (`FinishReason` as the marker) so the spec never reads the half-written row. All of that is designed, not missed.

What those two situations share is that write 2 will never come. A live client is a third case: `Agent.Abort` returns as soon as the status flip lands, finalize is still running, `GetSnapshot` shows `aborted` with `State == nil`, and `resumeSessionFrom` rejects with "resume from an earlier snapshot."

That advice matches the crash case you wrote the reject for. It is the wrong next move while finalize is in flight — the earlier snapshot is the last turn-end row, not the pending row about to receive the committed conversation. Following the error forks away the work this PR exists to keep. There is also no marker that distinguishes "write 2 is coming" from "write 2 is dead" (`HeartbeatAt` is still whatever the pending row had).

Suggested fix, pick one: keep the row `pending` until finalize commits the aborted terminal (abort stays a signal, not a visible status); or reject the stateless aborted row with "this snapshot is still being finalized, wait and retry the same ID" — and keep "resume from an earlier snapshot" for a row that has gone stale, the way expired orphans stay a dead end.

huangjeff5 (changes requested):
The product works: cancel / deadline / detach-abort land an aborted snapshot of committed turns, and resume accepts it.

Two leftovers that still change what a caller sees, plus a stack-landing note. None of these restate the two-write window, the `committedState` arms, or the #6173 alignment — those are already in the body.

huangjeff5 (approved):

go/ai/exp/agent.go
apascal07: Fixed, and wider than the narrowing you suggested. The `DeadlineExceeded` arm has the same defect as `Aborted`: a provider 504 maps to it exactly as a real deadline does, so narrowing only the second arm would have left the same mislabel one HTTP code away.

All three status arms are gone. `callerStopped` now tests the context plus the sentinels: `ctx.Err() != nil`, `errors.Is(cause, context.Canceled)`, `errors.Is(cause, context.DeadlineExceeded)`, `errors.Is(cause, ai.ErrMaxTurnsExceeded)`. The two sentinel arms keep what the status test was actually for, a context that died somewhere this `ctx` cannot see, without catching an HTTP code that merely lands on the same name.

`ai.Generate`'s copy has the identical defect and is fixed in the same commit, with `failurePartial` taking a context so it can ask the same question. The snapshot status and the partial's finish reason stay one rule rather than diverging. A service's ABORTED now reports failed, its classification preserved on the row for whoever judges it; pinned by a test in each package.

go/ai/exp/agent.go
apascal07: Fixed, with a marker that turned out to already be there.

`abortPendingSnapshot` copies the pending row and writes only `Status` and `UpdatedAt`, so `HeartbeatAt` rides through the flip untouched, and `finalizePendingSnapshot` clears it as it stamps the state on. A beat inside `defaultHeartbeatTimeout` therefore means a live worker is between the two writes, and that case now says to retry this same snapshot ID. A quiet or absent one keeps "resume from an earlier snapshot", which is the case the wording was written for.

That way `Abort` keeps reporting what it reported, no row lingers pending, and no field is added. Both halves are pinned by tests that build the stateless row directly with a live and a stale beat.

PR #6218, opened 2026-08-28
fix(google-genai): revert back to non-otel httpclient

MichaelDoyle:
All other runtimes/languages _and all other Go plugins_ default to plain http clients and consider additional non-Genkit telemetry to be optional. 

This restores to the status quo, and will improve the Dev UI experience by reducing noise. The developer has the option to enable an instrumented client if desired.

We _could_ offer a flag to turn telemetry on/off. But that's worth considering more broadly across the runtime and all plugins. In other runtimes OTel can do "magic" auto instrumentation that is not possible in Go.


apascal07 (approved):
Thanks!

PR #6219, opened 2026-08-28
feat(go/plugins/middleware): conform `Skills` to the Agent Skills specification

apascal07:
Rebuilds the `Skills` middleware against the [Agent Skills specification](https://agentskills.io), so a skill directory authored for any compliant agent works unchanged in Genkit, and Genkit's own skills are portable out. The specification's [client guide](https://agentskills.io/client-implementation/adding-skills-support) defines three tiers of progressive disclosure; the middleware implemented one and a half of them.

Two of the defects below are security-relevant and both are reachable on `main` today: a symlinked `SKILL.md` reads any file the process can open into the model's context, and a `SKILL.md` that is a named pipe hangs every `Generate` call for the life of the process.

## Defects fixed

**A symlinked `SKILL.md` reads arbitrary files into the prompt.** Discovery joined the path and called `os.ReadFile`, which follows links. A skills tree is instruction data, often from a checked-out repository, and git restores mode `120000` symlinks on clone. Discovery now requires `DirEntry.Type().IsRegular()`, and the read re-checks with `os.Lstat` in case the file was swapped after the scan.

**A named pipe named `SKILL.md` hangs the request forever.** `os.Open` on a FIFO blocks until a writer appears, and nothing cancels it. The mode has to be checked *before* the open, since a check on the opened file never runs. The same hole existed in the new bundled-file reader; `os.Root` confines the path but says nothing about what kind of file it lands on.

**`description: Use when: the user asks` silently loses the name too.** goccy rejects the unquoted colon, the error was discarded, and `parseFrontmatter` returned both fields empty. JS never hits this because it line-matches instead of parsing YAML, so Go and JS already produced different catalogs from the same directory. The parser now falls back to a line scan and reports the error for logging.

**A `----` line inside valid frontmatter truncated it.** The closing fence was found with `strings.Index(rest, "\n---")`, with no check that the match ends a line, so a quoted scalar spanning a dashed line cut the block and lost every field after it. The fence is now line-anchored.

**`use_skill` was registered even with zero skills discovered.** Every call then failed, with no catalog explaining the tool. The specification requires that a client neither show an empty catalog nor register the tool.

**An unknown skill name destroyed the generation.** The tool returned an error, which the loop turns into `ErrToolFailed`. One hallucinated name killed a multi-turn run. A `WrapTool` hook now converts a failure into a model-visible message listing the valid names, matching the sibling `Filesystem` middleware and the Python runtime. Interrupts still propagate, so composing with `ToolApproval` is unaffected.

**`skill.md` loaded on macOS and not on Linux.** Reading the joined path accepts any casing on a case-insensitive volume, so the same library discovered a different set of skills depending on the developer's machine. `SKILL.md` is now matched case-exactly from the directory listing.

**A description containing `</skills>` closed the catalog block.** Catalog text is unconditional, so this injected system-level instructions without the model calling anything. Author text is now folded to one line and tag-like `<` is escaped; `values < 10` is left alone.

Shadowed skills were also silent (now a warning naming both paths), `SKILL.md` had no size cap, and a description over the specification's 1024-rune limit went into every request whole.

## New surface

```go
// Added
middleware.SkillToolName              // "use_skill"
middleware.SkillResourceToolName      // "read_skill_file"
middleware.SkillActivationMetadataKey // "skillActivation"

// Skills gains three fields; nothing removed or renamed
Skills.Preload             []string
Skills.AllowResourceAccess bool
Skills.ToolNamePrefix      string
```

The tool-name constants are not cosmetic. `ToolApproval.AllowedTools` is matched by exact string, so the moment `Skills` can register a second tool a caller needs a compile-checked name to put in that list.

## Tier 3: the files a skill bundles

A skill that says "see `references/api.md`" previously handed the model a dead path. `Filesystem` cannot cover it, since it roots at its own `RootDir` and knows nothing about skill directories.

```go
ai.WithUse(&middleware.Skills{
    SkillPaths:          []string{"./skills"},
    AllowResourceAccess: true,
})
```

Activation then appends a bounded listing, and `read_skill_file` reads one file, rooted per skill by `os.Root` so `..`, an absolute path, or a symlink out is refused:

```
<skill_content name="limerick" path="/app/skills/limerick">
...SKILL.md verbatim, frontmatter included...

Relative paths in this skill resolve against /app/skills/limerick

<skill_resources>
Paths are relative to the skill directory above; read them with the read_skill_file tool.
 - references/form.md
</skill_resources>
</skill_content>
```

Default off: enabling it grants file-read reach the middleware does not otherwise have, and adds a tool name that `ToolApproval.AllowedTools` must list.

## Activated skills are tracked in the transcript, not in the middleware

The old prompt asked the model to self-police with "Only use them once to load the context", which is a request, not a mechanism. Each activation now stamps `SkillActivationMetadataKey` on the tool-response part, and `WrapGenerate` rebuilds the loaded set from `params.Request.Messages` every turn. Deriving it from the transcript rather than holding it means the middleware can never claim a skill is loaded after a compaction pass dropped the part carrying it, and context-management middleware gets a documented key to find skill instructions by.

Two Go details this turns on. The loop runs one goroutine per tool request in a turn, so the check and the set have to be one critical section: a split check-then-mark let two `use_skill` calls for the same skill both return the full body, in roughly half of real runs. And `ai.NewMultipartTool` does not infer an output schema from a type parameter the way `ai.NewTool` does, while `Definition()` prefers `metadata["originalOutputSchema"]`, so the switch to a multipart tool needs an explicit `ai.WithOutputSchema(map[string]any{"type": "string"})` or the tool silently starts advertising the multipart envelope where it advertised a string.

## Behavior changes for existing callers

- **Default scan paths widen** to `{".agents/skills", "skills"}`. Only affects callers who leave `SkillPaths` unset. `skills` is scanned last, so it keeps winning collisions.
- **`use_skill` no longer returns the raw file.** It returns the wrapper above. Byte-exact eval snapshots will move.
- **The catalog gains one line** naming the activation tool.
- **An unknown skill name no longer fails `Generate`.** Code handling that error sees success instead.
- **Skills that used to load and now do not:** lowercase `skill.md` on macOS and Windows, a symlinked `SKILL.md`, a `SKILL.md` over 1 MiB, and a `SKILL.md` that is not a regular file.
- **`Skills{[]string{...}}` as an unkeyed literal no longer compiles**, since the struct went from one field to four. Keyed literals are unaffected.

The catalog part's metadata now carries the activation tool name rather than `true`, which is what lets two `Skills` instances on one call keep separate catalogs. A bare `true` is still accepted and upgraded on the first turn, so a conversation persisted by an older build, or started in JS or Python, is refreshed rather than gaining a second stale catalog.

Go now diverges from JS and Python on catalog bytes, activation payload, and default paths. Those three want a follow-up in the other runtimes.

## What this deliberately does not do

- **No enum on `skillName`.** The specification suggests constraining the parameter to the valid names. `ai.WithInputSchema` forces the input type parameter to `any`, which would make this the only hand-decoded tool handler in the package, to save one recoverable turn that the soft-fail already answers with the full list.
- **No recursive scanning, XDG lookup, or ancestor walk to the git root.** All optional in the specification. `SkillPaths` covers grouped layouts in one line of caller code, and recursion would stop the skill name being unambiguously the directory basename, exactly when that name starts rooting an `os.Root`.
- **No user-home or organization scope.** A deployed process absorbing instructions from the deploying user's home directory would change the system prompt between a laptop and Cloud Run.
- **No `license`, `compatibility`, `metadata`, or `allowed-tools` parsing.** Nothing consumes them, and activation returns the file whole, so they already reach the model. Wiring `allowed-tools` to an approval bypass is the last thing to build speculatively.
- **No script execution.** The package ships no exec tool and `Skills` is the wrong place to introduce one. The `path` attribute gives a caller with their own exec tool the correct working directory.
- **No trust gate on project-level skills.** Genkit has no trust model to hook into, and a flag everyone sets to `true` is not one. The type doc says plainly that a skills tree is instruction text and should be treated like source code.

### DIFF
--- a/go/plugins/middleware/skills.go
+++ b/go/plugins/middleware/skills.go
@@ -326,7 +326,7 @@ func (s *Skills) newUseSkillTool(info map[string]skillInfo, act *activationSet, 
 	return ai.NewMultipartTool(
 		s.toolName(SkillToolName),
 		"Load a skill's instructions by name.",
-		func(_ *ai.ToolContext, in useSkillInput) (*ai.MultipartToolResponse, error) {
+		func(tc *ai.ToolContext, in useSkillInput) (*ai.MultipartToolResponse, error) {
 			si, ok := lookupSkill(info, in.SkillName)
 			if !ok {
 				return &ai.MultipartToolResponse{
@@ -339,7 +339,7 @@ func (s *Skills) newUseSkillTool(info map[string]skillInfo, act *activationSet, 
 					Output: fmt.Sprintf(skillAlreadyLoadedStub, si.Name),
 				}, nil
 			}
-			content, err := s.renderSkill(si)
+			content, err := s.renderSkill(tc, si)
 			if err != nil {
 				release()
 				return nil, err
@@ -356,14 +356,14 @@ func (s *Skills) newUseSkillTool(info map[string]skillInfo, act *activationSet, 
 // renderSkill reads a skill and returns the instructions to place in the
 // conversation. Both entry points go through it, so an activation and a
 // preload deliver byte-identical content.
-func (s *Skills) renderSkill(si skillInfo) (string, error) {
+func (s *Skills) renderSkill(ctx context.Context, si skillInfo) (string, error) {
 	data, err := readSkillFile(si.Path)
 	if err != nil {
 		return "", err
 	}
 	var resources string
 	if s.AllowResourceAccess {
-		resources = listSkillResources(si.Dir, s.toolName(SkillResourceToolName))
+		resources = listSkillResources(ctx, si.Dir, s.toolName(SkillResourceToolName))
 	}
 	return wrapSkillContent(si, string(data), resources), nil
 }
@@ -905,14 +905,20 @@ func wrapSkillContent(si skillInfo, body, resources string) string {
 // they exist without any of them being read. The optional directory names in
 // the specification are conventions, not a closed set, so everything the skill
 // ships is listed.
-func listSkillResources(dir, toolName string) string {
+func listSkillResources(ctx context.Context, dir, toolName string) string {
 	root := filepath.Clean(dir)
 	var (
 		files     []string
 		truncated bool
 	)
+	// The walk never fails: every entry error is reported and swallowed below,
+	// so WalkDir has nothing left to return. An unreadable corner of a skill
+	// directory costs the model that listing, not the activation, which is why
+	// this degrades rather than propagating.
 	_ = filepath.WalkDir(root, func(p string, d fs.DirEntry, err error) error {
 		if err != nil {
+			logger.Debug(ctx, "skill resource could not be listed, skipping",
+				"path", p, "error", err)
 			return nil //nolint:nilerr // an unreadable entry is skipped, not fatal
 		}
 		rel, relErr := filepath.Rel(root, p)
@@ -1010,7 +1016,7 @@ func (s *Skills) injectSkills(ctx context.Context, req *ai.ModelRequest, catalog
 		if !preload[name] || present[name] {
 			continue
 		}
-		content, err := s.renderSkill(info[name])
+		content, err := s.renderSkill(ctx, info[name])
 		if err != nil {
 			// The skill stays loadable through the activation tool, so this
 			// degrades preloading rather than losing the skill.
--- a/go/plugins/middleware/skills.go
+++ b/go/plugins/middleware/skills.go
@@ -18,43 +18,118 @@ package middleware
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
+	"io/fs"
+	"maps"
 	"os"
 	"path/filepath"
-	"sort"
+	"slices"
+	"strconv"
 	"strings"
+	"sync"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/core/logger"
 	"github.com/goccy/go-yaml"
 )
 
-// defaultSkillsPath is the directory scanned when Skills.SkillPaths is unset.
-const defaultSkillsPath = "skills"
+// Tool names registered by [Skills]. When [Skills.ToolNamePrefix] is set, the
+// registered name is the prefix followed by the constant.
+//
+// Use them to name the tools in [ToolApproval.AllowedTools]: that field is
+// matched by exact string, so a skills tool missing from the list is held for
+// approval rather than run.
+const (
+	// SkillToolName loads a skill's instructions by name. The name matches the
+	// JS and Python runtimes so prompts and evaluations port between them.
+	SkillToolName = "use_skill"
 
-// skillsMarker marks the system prompt part injected by this middleware so it
-// can be refreshed on later tool-loop iterations instead of duplicated.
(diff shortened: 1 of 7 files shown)


go/plugins/middleware/skills.go
```diff
@@ -65,87 +140,409 @@ const useSkillToolName = "use_skill"
...
+// [os.Root] on the one skill's directory, which rejects any path resolving
+// outside it, including via "..", an absolute path, or a symbolic link.
+func (s *Skills) newReadSkillFileTool(info map[string]skillInfo, available string) ai.Tool {
+	return ai.NewTool(
+		s.toolName(SkillResourceToolName),
+		"Read a file bundled inside a skill directory, such as a reference document or a script.",
+		func(_ *ai.ToolContext, in readSkillFileInput) (string, error) {
+			si, ok := lookupSkill(info, in.SkillName)
+			if !ok {
+				return "", errors.New(unknownSkillMessage(in.SkillName, available))
```
huangjeff5, line 469: The model can misspell a skill on `read_skill_file` the same way it can on `use_skill`. `use_skill("nonexistent")` comes back as one sentence: `Unknown skill "nonexistent". Available skills: "javascript", "python".`

`read_skill_file` on that same name comes back as:
`Tool "read_skill_file" failed: … Unknown skill "nonexistent". Available skills: "javascript", "python".. Available skills: "javascript", "python".`

Can `read_skill_file` give the model that first sentence once, the way `use_skill` does?

go/plugins/middleware/skills.go
```diff
@@ -65,87 +140,409 @@ const useSkillToolName = "use_skill"
...
+	available := availableSkillsSentence(info)
+
+	// Registering the activation tool with nothing left to activate would offer
+	// the model a tool whose every input is a dead end, which happens when
+	// Preload names every discovered skill.
+	loadable := loadableNames(info, preload)
+	catalog := s.buildSkillsPrompt(info, loadable)
+
+	var tools []ai.Tool
+	if len(loadable) > 0 {
```
huangjeff5, line 232: `Preload` puts skill bodies in the prompt so the model does not have to call `use_skill` first. If I preload every skill and the files are there, I don't need that tool — every call would be a stub.

If I preload the only skill and that file can't be read, I get no body, no catalog, and no `use_skill`. The model has nothing. The comment says it can still load through the tool — that only works if another skill kept the tool around.

Can `use_skill` still be there when a preload never landed, so the model can load that skill?

go/plugins/middleware/skills.go
apascal07: Done in 833b504ec. An unknown name is now a normal answer, not an error, so the list shows once. The test checks the exact text for both tools.

go/plugins/middleware/skills.go
apascal07: I fixed this at the source, not with a fallback. The scan already reads every SKILL.md. A preload now uses those same bytes, so a skill that the scan found always preloads, and every turn gets the same text. If SKILL.md cannot be read at scan time, the scan does not find the skill: `Preload` logs the name and skips it. Keeping `use_skill` as a fallback would help only if the file changed in the short time between the scan and a second read. The new test uses your case: one skill, preloaded, with its file deleted after the scan (fba19b786).

PR #6221, opened 2026-08-28
feat(go): add `continue_task`, the `aborting` snapshot status, and metadata-only snapshot reads

apascal07:
Third PR of the train, on #6119 (background delegation) and #6118 (`AgentHandle`). Adds one shared `continue_task` tool to the `Agents` middleware, so an orchestrator continues a sub-agent run instead of re-buying it from zero: the runtime already finalizes failed and aborted runs as resumable snapshots (#6196, #6213), and this gives those snapshots a model-facing handle and a spender. Two core changes make the tool's status decisions honest and cheap: the abort protocol gets a real `aborting` status, and snapshot reads get a metadata-only projection down to the store.

## Every settled server-managed delegation carries a handle

```
delegate_to_researcher {task, name?, background?}
  -> {response, taskId: "researcher:<snapshotId>", status: "completed" | "failed" | ..., name?}
continue_task          {taskId, instructions?, background?}
  -> a delegation result, or a fresh pending handle when background
```

A synchronous result from a server-managed sub-agent now names the run's last committed snapshot as `taskId` with its settled `status`, the currency background launches already return, and background reports for `failed`, `aborted`, and `expired` tasks name the tool. Two results stay handle-less: a client-managed delegation (nothing durable behind it) and an interrupt (continuing past one means answering it). `name` is a caller-chosen label echoed on results and reports next to the `taskId`, a reading aid rather than an identifier. Every fold of a server-managed run namespaces its artifacts by snapshot ID, so a later check re-merges over identical names.

## One continue tool, gated by what it can do

- **Failed or aborted**: continues from the last saved progress. Empty `instructions` re-attempts the committed turn as it stood; non-empty ones ride in as a fresh user message.
- **Completed**: accepts follow-up `instructions` inside the sub-agent's session, and refuses without them, since an empty input would re-run the finished turn.
- **Interrupted**: refused as a dead end. "Resume" already names interrupt resumption in Genkit, the one thing this tool cannot do, which is why it is not called that.
- **Expired** (worker stopped heartbeating): the row is abort-fenced first, so a merely slow worker cannot race the recovery, then one metadata re-read decides: a settled row continues, an `aborting` row is refused with "retry once it settles", a still-expired row falls back to its parent snapshot, and a run that never committed a turn is refused with "delegate again".
- **`background: true`** (with `Async`) returns a fresh pending handle in the same session; an empty background retry rides a synthesized continue message, because a payload-less detached input runs no turn.

The tool registers only when some configured sub-agent can leave a handle behind, so an all-client-managed configuration gets no dead tool, and `ToolPrefix` namespaces it like the background-task tools. A continuation spends a `MaxDelegations` slot; a refusal that names a retry which can succeed refunds it, a dead end keeps it.

## The abort window is a status, not a shaping rule

The abort protocol writes twice: the flip that stops the work, and the finalize that stamps the state on. The row between them was `aborted` with no state, hidden by shaping it as `pending`. It is now `aborting`, a shared wire status (TS, Go, and Python types regenerated):

| Write | From | To |
|---|---|---|
| abort flip | `pending` | `aborting`, `heartbeatAt` untouched; idempotent on `aborting`, no-op on a settled row |
| heartbeat | `pending`, `aborting` | same status, `heartbeatAt` refreshed |
| finalize | `pending` | `completed`, `failed`, or `aborted`, with state |
| finalize | `aborting` | `aborted`, with state |

Every transition is one `SaveSnapshot` mutator guarded by its from-status. Read shaping keys on status and heartbeat alone: a `pending` or `aborting` row with a stale beat reads as `expired`. The worker keeps heartbeating while it drains, bounded by `windDownHeartbeatBudget` (5 minutes), so a wedged drain goes stale and becomes recoverable. `Terminal()` excludes `aborting`, so `WaitForSnapshot` rides the window and returns the finalized row. The in-memory store also captures a row's status before running the mutator, closing a gap where an in-place edit hid the flip from the worker's subscription.

**Behavior change worth noting.** The `abort` companion action answers `aborting` for a pending row where it answered `aborted`; the settled row is one `waitForSnapshot` away. The shared spec's two abort cases now await the settled row, and the JS and Python runtimes, which settle their abort in one write, pass unchanged.

## Status decisions read metadata only, down to the store

```go
// Added
const SnapshotStatusAborting SnapshotStatus = "aborting"
type SnapshotReadOption
func WithMetadataOnly() SnapshotReadOption

// Optional store capability; without it the runtime reads in full and drops the state.
// Implemented by localstore.InMemorySessionStore, localstore.FileSessionStore, and
// firebase/exp.FirestoreSessionStore.
type SnapshotMetadataReader[State any] interface {
    GetSnapshotMetadata(ctx, snapshotID string) (*SessionSnapshot[State], error)
    GetLatestSnapshotMetadata(ctx, sessionID string) (*SessionSnapshot[State], error)
}

// Widened with opts ...SnapshotReadOption, on the typed and the untyped surface alike
func (a *Agent[State]) GetSnapshot(ctx, snapshotID string, opts ...SnapshotReadOption)
func (a *Agent[State]) GetLatestSnapshot(ctx, sessionID string, opts ...SnapshotReadOption)
func (h *AgentHandle) GetSnapshot(ctx, snapshotID string, opts ...SnapshotReadOption)
func (h *AgentHandle) GetLatestSnapshot(ctx, sessionID string, opts ...SnapshotReadOption)
func (t *DetachedTask[State]) Poll(ctx, opts ...SnapshotReadOption)

// Wire
GetSnapshotRequest.MetadataOnly bool `json:"metadataOnly,omitempty"`
```

The continue ladder and the background-task reports dispatch on status and finish reason, so their pre-reads skip the conversation history and the run's own load is the only full read. The projection is a read option rather than a second method, so it composes with both addressings and with a task poll, and `GetSnapshotRequest` stays a wire type outside every handle signature. The store side is optional the way `io.Copy` takes `io.WriterTo` when it is there: an out-of-tree store keeps compiling and merely pays for a history it will not use. Firestore answers with one document read instead of the checkpoint shards and diff chain a full read materializes; the file store leaves the state as raw bytes; the in-memory store copies the row without it.

- **No interrupt resume.** Answering a sub-agent's interrupt from the orchestrator is its own feature with its own tool contract.
- **No mid-run persistence of detached progress.** A background turn commits only at its seam, so a hard crash loses the in-flight turn. Checkpointing inside the generate loop is its own feature, and the recovery ladder picks it up automatically if it lands.
- **No config knob on `Agents`.** The tool registers wherever a handle can exist; a TODO records the intended control (disable, or scope per agent).
- **Only Go emits `aborting`.** `genkit-ui` shows a Go task in wind-down as settled until it gains an arm, and JS and Python keep their single-write abort.


go/plugins/middleware/exp/agents_continue.go
```diff
@@ -0,0 +1,364 @@
...
+		return delegationResult{Response: fmt.Sprintf(
+			"Error: could not fence task %q before recovering it (%v). Try again later.", in.TaskID, err)}, nil
+	}
+	cur, err := agent.GetSnapshot(ctx, snapshotID, aix.WithMetadataOnly())
+	if err != nil {
+		logger.Debug(ctx, "continue fence re-read failed", "agent", ref.Name, "taskId", in.TaskID, "error", err)
+		a.releaseDelegation(st)
+		return delegationResult{Response: fmt.Sprintf(
+			"Error: could not read task %q after fencing it (%v). Try again later.", in.TaskID, err)}, nil
+	}
```
pavelgj, line 221: The fence abort and the re-read both `releaseDelegation` unconditionally, but every sibling read arm keys the refund on `deadEndRead` and keeps the slot on a dead end (e.g. `continueFromStore` at ~128-142, `continueFromParent`). So an `Abort` or `GetSnapshot` that comes back NotFound/InvalidArgument/FailedPrecondition here refunds the slot and lets the model retry a call that can't succeed.

The doc comment says "Fence and re-read failures are transient," but a classified dead end isn't. Could we route both through `deadEndRead` like the other arms, or add a line explaining why the expired path always refunds? Just want the budget accounting to stay consistent across the ladder.

go/ai/exp/localstore/inmemory.go
```diff
@@ -87,10 +129,19 @@ func (s *InMemorySessionStore[State]) GetLatestSnapshot(_ context.Context, sessi
...
+
+// snapshotMetadata returns snap without its state: a shallow copy with State
+// nil and its own HeartbeatAt value.
+func snapshotMetadata[State any](snap *exp.SessionSnapshot[State]) *exp.SessionSnapshot[State] {
+	meta := *snap
+	meta.State = nil
+	if snap.HeartbeatAt != nil {
+		hb := *snap.HeartbeatAt
+		meta.HeartbeatAt = &hb
 	}
```
pavelgj, line 143: `meta := *snap` shallow-copies, and we nil `State` and deep-copy `HeartbeatAt`, but `meta.Error` still aliases the stored row's `*status.Error`. `copySnapshot` (the full-read path) hands back a fully-owned deep copy, so this diverges. Not a live race today since rows are only ever replaced wholesale, but a caller mutating `snap.Error` off a metadata read would reach into the store's row. Worth copying `Error` too so metadata reads match the full-read ownership contract.

pavelgj (approved):

PR #6225, opened 2026-08-31
fix(py/plugins/openai): classify embedder and list_actions API errors

### LINKED ISSUE #6189: fix(py/plugins/openai): no HTTP status to Genkit status mapping
There is no HTTP status to Genkit status mapping anywhere in `src/genkit_openai/`: no `APIError` handling, no status-code branching, so 429s, 5xx, and auth failures all surface as unwrapped SDK exceptions instead of typed GenkitErrors.

Go maps via `status.FromHTTPCode(apiErr.StatusCode)` (`go/plugins/compat_oai/errors.go:49`).

Fix: catch the SDK error types and map status codes to Genkit statuses on generate, stream, image, audio, and embedder paths.

hilariie:
Resolves #6189

## Why

#6098 routed chat, streaming, image, TTS, and STT through reraise_openai_error, but the embedder and dynamic model listing still leak raw SDK exceptions. Callers cannot branch on a Genkit status for those paths, and a failed model listing surfaces as a raw openai error rather than a GenkitError like every other action.

## Changes

- embed_fn wraps only the embeddings.create call in except APIStatusError. A dimensions option that is not an int raises INVALID_ARGUMENT before the call; ValueErrors from the SDK or from response conversion propagate unchanged so they are not mistaken for bad caller input.
- list_actions wraps only the models.list() call and catches APIStatusError alone, since no caller input reaches it. The action cache is assigned on success only, so a failed listing is retried rather than served from cache; a test pins that.
- Connection failures and timeouts still propagate unchanged.

### DIFF
--- a/py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
+++ b/py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
@@ -455,26 +455,33 @@ class OpenAI(Plugin):
 
         async def embed_fn(request: EmbedRequest) -> EmbedResponse:
             """Embedder function that calls OpenAI embeddings API."""
-            try:
-                # Extract text from document content
-                texts = []
-                for doc in request.input:
-                    doc_text = ''.join(  # type: ignore[arg-type]
-                        part.root.text for part in doc.content if hasattr(part.root, 'text') and part.root.text
-                    )
-                    texts.append(doc_text)
+            # Extract text from document content
+            texts = []
+            for doc in request.input:
+                doc_text = ''.join(  # type: ignore[arg-type]
+                    part.root.text for part in doc.content if hasattr(part.root, 'text') and part.root.text
+                )
+                texts.append(doc_text)
 
-                # Get optional parameters (omit when None; OpenAI create() uses Omit, not None)
-                dimensions: int | None = None
-                encoding_format: Literal['base64', 'float'] | None = None
-                if request.options:
-                    if dim_val := request.options.get('dimensions'):
+            # Get optional parameters (omit when None; OpenAI create() uses Omit, not None)
+            dimensions: int | None = None
+            encoding_format: Literal['base64', 'float'] | None = None
+            if request.options:
+                if dim_val := request.options.get('dimensions'):
+                    try:
                         dimensions = int(dim_val)
-                    enc_val = request.options.get('encodingFormat')
-                    if enc_val in ('float', 'base64'):
-                        encoding_format = cast(Literal['base64', 'float'], enc_val)
+                    except (TypeError, ValueError) as e:
+                        raise GenkitError(
+                            status='INVALID_ARGUMENT',
+                            message=f'dimensions must be an int, got {dim_val!r}',
+                            cause=e,
+                        ) from e
+                enc_val = request.options.get('encodingFormat')
+                if enc_val in ('float', 'base64'):
+                    encoding_format = cast(Literal['base64', 'float'], enc_val)
 
-                # Call with only non-None optional params to satisfy strict typings
+            # Call with only non-None optional params to satisfy strict typings
+            try:
                 if dimensions is not None and encoding_format is not None:
                     response = await self._runtime_client().embeddings.create(
                         model=clean_name,
@@ -499,12 +506,12 @@ class OpenAI(Plugin):
                         model=clean_name,
                         input=texts,
                     )
-
-                # Convert OpenAI response to Genkit format
-                embeddings = [Embedding(embedding=item.embedding) for item in response.data]
-                return EmbedResponse(embeddings=embeddings)
-            except (APIStatusError, ValueError) as e:
+            except APIStatusError as e:
                 reraise_openai_error(e)
+
+            # Convert OpenAI response to Genkit format
+            embeddings = [Embedding(embedding=item.embedding) for item in response.data]
+            return EmbedResponse(embeddings=embeddings)
 
         return embedder(
             name,
--- a/py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
+++ b/py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
@@ -20,7 +20,7 @@
 import enum
 from typing import Any, Literal, TypeAlias, cast
 
-from openai import AsyncOpenAI
+from openai import APIStatusError, AsyncOpenAI
 from openai.types import Model
 
 from genkit import Embedding, EmbedRequest, EmbedResponse, GenkitError, ModelInfo, ModelRequest, ModelResponse, Supports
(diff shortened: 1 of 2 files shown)


py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
```diff
@@ -439,53 +440,56 @@ def _create_embedder_action(self, name: str) -> Action:
...
+                    response = await self._runtime_client().embeddings.create(
+                        model=clean_name,
+                        input=texts,
+                        encoding_format=encoding_format,
+                    )
+                else:
+                    response = await self._runtime_client().embeddings.create(
+                        model=clean_name,
+                        input=texts,
+                    )
```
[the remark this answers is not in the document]
hilariie: This branching predates the PR. The diff only moves it inside a try, so restructuring it belongs in its own change, not here.

The suggested patch also does not pass this repo's checks: pyrefly rejects unpacking a plain dict into embeddings.create, because the dict's value type is checked against every keyword parameter of the call (e.g. "Unpacked keyword argument int is not assignable to parameter extra_query with type Mapping[str, object] | None")

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
```diff
@@ -454,53 +455,56 @@ def _create_embedder_action(self, name: str) -> Action:
...
 
-            # Convert OpenAI response to Genkit format
-            embeddings = [Embedding(embedding=item.embedding) for item in response.data]
-            return EmbedResponse(embeddings=embeddings)
+                # Convert OpenAI response to Genkit format
+                embeddings = [Embedding(embedding=item.embedding) for item in response.data]
+                return EmbedResponse(embeddings=embeddings)
+            except (APIStatusError, ValueError) as e:
```
cabljac: The `try` covers the SDK call and the `Embedding(...)` construction on L503 as well as `int(dim_val)`, so any `ValueError` from the response side goes through `reraise_openai_error` and comes out as INVALID_ARGUMENT. From what I can see that includes `ValueError("No embedding data received")` from the SDK's embeddings parser, `binascii.Error` on bad base64 (a `ValueError` subclass), and pydantic `ValidationError` when `item.embedding` isn't a `list[float]`. Those are provider or plugin faults, not caller input, so INVALID_ARGUMENT would point users the wrong way.

Can you narrow the `ValueError` catch to just the conversion and keep `APIStatusError` around the `create` call? Roughly:

```python
if dim_val := request.options.get('dimensions'):
    try:
        dimensions = int(dim_val)
    except ValueError as e:
        raise GenkitError(status='INVALID_ARGUMENT', message=f'dimensions must be an int, got {dim_val!r}', cause=e) from e
```

That also drops most of the re-indent. Worth a test pinning that a `ValueError` raised by `create` does not become INVALID_ARGUMENT.

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
```diff
@@ -454,53 +455,56 @@ def _create_embedder_action(self, name: str) -> Action:
...
+                    )
+                    texts.append(doc_text)
+
+                # Get optional parameters (omit when None; OpenAI create() uses Omit, not None)
+                dimensions: int | None = None
+                encoding_format: Literal['base64', 'float'] | None = None
+                if request.options:
+                    if dim_val := request.options.get('dimensions'):
+                        dimensions = int(dim_val)
+                    enc_val = request.options.get('encodingFormat')
```
cabljac: nit: not this PR, but from what I can see `encodingFormat: 'base64'` can't work here. The SDK returns the raw base64 string when `encoding_format` is passed explicitly, and `Embedding.embedding` is `list[float]`, so L503 fails validation. With the broad catch above it would surface as INVALID_ARGUMENT, which is another reason to narrow it. Probably a follow-up issue rather than a fix here.

cabljac (changes requested):
Classification itself looks right and lines up with #6189 (https://github.com/genkit-ai/genkit/issues/6189). One blocking point on the `except ValueError` scope, inline. Rest is nits.

One thing worth verifying on the motivation: from what I can see `Retry` only implements `wrap_model`, and `embed()` in `_ai/_aio.py` runs the action directly with no middleware hook, so embedder errors never reach retry either way (and the SDK's own retry already honours Retry-After). The change still stands on its own for status classification, but can you double check and adjust the description if that holds.

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
hilariie: Opened #6317 to track it separately.

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
hilariie: Narrowed the catch to the dimensions conversion, kept APIStatusError around create only, and added a test that a ValueError from create is not classified as INVALID_ARGUMENT.

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
```diff
@@ -467,36 +468,46 @@ async def embed_fn(request: EmbedRequest) -> EmbedResponse:
             encoding_format: Literal['base64', 'float'] | None = None
             if request.options:
                 if dim_val := request.options.get('dimensions'):
-                    dimensions = int(dim_val)
+                    try:
+                        dimensions = int(dim_val)
+                    except (TypeError, ValueError) as e:
+                        raise GenkitError(
+                            status='INVALID_ARGUMENT',
+                            message=f'dimensions must be an int, got {dim_val!r}',
```
cabljac, line 476: nit: the message says `dimensions must be an int`, but from what I can see `int(dim_val)` still accepts a bit more than that: `'256'` passes, `256.9` truncates to 256, `True` becomes 1, and the truthiness check on the line above drops `dimensions: 0` rather than rejecting it. All pre-existing, so fine to leave for this PR. If you want the message to match, `isinstance(dim_val, int) and not isinstance(dim_val, bool)` with `is not None` on the gate would do it, or we could track it as a follow-up alongside #6317.

cabljac (approved):
Approving, thanks for the rework. Narrowing the except to just `int(dim_val)` and wrapping `create` on its own is the right shape, and `test_embedder_propagates_unclassified_errors` pinning that a plain ValueError escapes unclassified is exactly the test I wanted to see. One small nit inline, nothing blocking.

py/packages/genkit-openai/src/genkit_openai/openai_plugin.py
hilariie: Gate is `is not None` now and the check is `isinstance(dim_val, int) and not isinstance(dim_val, bool)`, so '256', 256.9 and True are rejected rather than coerced, and dimensions: 0 reaches the API instead of being
dropped. Tests cover all four.

PR #6255, opened 2026-09-02
fix(py): generate returns completed history instead of raising when a turn dies

huangjeff5:
Once `ai.generate` has started a model request, it always returns a `ModelResponse`. `.messages` is the conversation you can send again. The unanswered turn is omitted (`.message is None` except an interrupt pause). Why is on `response.error`.

If generate never started a chat, it still raises. Cancelling through `abort_signal` always returns a boxed abort — before the first model call or mid-flight.

```python
# never started
try:
    await ai.generate(model='nope/x')
except GenkitError as e:
    e.reason  # MODEL_NOT_FOUND

# abort_signal — always a response
abort.set()
response = await ai.generate(prompt=prompt, abort_signal=abort)
# finish_reason is ABORTED; .messages is the prompt they passed in

# started
response = await ai.generate(prompt=prompt, tools=['lookup'])
if response.error:
    response = await ai.generate(messages=response.messages)
```

A tool author publishes the sentence on `response.error.message` with `PublicError`. A bare exception is `internal error`. Task cancel and `asyncio.wait_for` raise again; `abort_signal` is the boxed abort.

```python
@ai.tool()
def lookup_order(order_id: str) -> str:
    raise PublicError('NOT_FOUND', f'No order {order_id}')  # this sentence is on the response

@ai.tool()
def lookup_order(order_id: str) -> str:
    raise RuntimeError('connection to postgres://user:pw@host refused')
# r.error.message == 'internal error'

# cancel
await asyncio.wait_for(ai.generate(...), timeout=0.3)  # TimeoutError
task.cancel(); await task  # CancelledError
# abort_signal still returns a boxed ABORTED response
```

## Decisions

- PublicError is how a tool author publishes the sentence on `response.error.message`. A bare exception is `internal error`.
- `task.cancel()` / `asyncio.wait_for` raise again. `abort_signal` always returns a boxed abort, including before the model is called.
- Bad JSON through `define_prompt_agent` is not a dead turn (same as generate: STOP + model text).
- A missing resource name is `NOT_FOUND`, same as an unmatched URI.
- Empty `messages` / `ai.generate()` with no prompt is a valid start; the model speaks first.
- Raise before any chat request: unknown model, tool, or format; colliding tool names (including the same short name); unmatched resource (`NOT_FOUND`); `max_turns < 0`; unknown middleware; no model or a config that does not belong to the model; `resume=` on a background model; a middleware `tools()` name already on `tools=`.
- After a request has started, a dead tool, model, hook, or plugin; bad tool args; unserializable tool output; `resume=` that is not a tool turn; a missing tool the model named; or hitting `max_turns` returns a `ModelResponse` with closed history and `.error`. First-turn `wrap_generate` / `wrap_model` deny, dict, `RuntimeError`, Action validation, and `config={'temperature':'high'}` box the same way.
- The model and tools you named are the ones that run — including on `resume_restart`. Middleware cannot swap them after the door.
- `generate_operation()` after a billed ticket still raises; a start that already failed raises that error, not "did not return an operation." `generate()` keeps `.operation`.
- `max_turns=0` means zero turns. Default is 50.
- `ai.generate_stream`, `prompt.stream`, and `action.stream` have no `timeout=`. Cancel the task to stop waiting.
- Invalid structured output is not a dead turn: `finish_reason` stays `STOP`, `.text` is the raw model text, `.output` is `None`, `error.reason` is `INVALID_OUTPUT`.
- A plugin rejecting the output in strict mode *is* a dead turn: `FAILED`, `.message is None`, `messages` ends at the user turn so a retry does not feed the bad output back. a2ui `validate='strict'` and an unregistered catalog id both land here. The split from the bullet above is deliberate — raw text you can still read is worth keeping, an unrenderable surface is not. Go routes this through `failurePartial`, Dart through `_failedResponse`.
- A plugin error that subclasses `GenkitError` keeps its sentence on `error.message`; a bare exception is still `internal error`. `A2uiParseError` and the new `A2uiCatalogError` subclass it, and stay `ValueError` so existing catch sites work.
- A restarted tool that pauses again returns `INTERRUPTED` they can answer.
- Session and snapshot helpers still raise. After `conn.close()`, `conn.send` raises `CONNECTION_CLOSED`; `conn.output()` is still readable.


py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -797,6 +828,63 @@ def _persist_threaded_conversation(response: ModelResponse, messages: list[Messa
...
+
+def closed_round_from_exc(
+    *,
+    response: ModelResponse | None,
+    messages: list[Message],
+    exc: BaseException,
+    caller_stopped: bool,
+) -> ModelResponse:
+    if isinstance(exc, StreamingCallbackError):
+        raise exc.cause
```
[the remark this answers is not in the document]
huangjeff5: Good catch. This was valid: the resumed/initial chunk boundary could leak the internal StreamingCallbackError. This follow-up sends callback failures from that boundary through the same unwrapping path as later model and tool chunks, and adds coverage confirming callers receive the original exception consistently.

py/packages/genkit/src/genkit/_core/_error.py
```diff
@@ -69,6 +72,49 @@ class StatusCodes(IntEnum):
...
 
+
+class RuntimeErrorReason(StrEnum):
+    """Stable reasons for framework-classified runtime failures."""
+
+    INVALID_SCHEMA = 'INVALID_SCHEMA'
+    INVALID_INPUT = 'INVALID_INPUT'
+    INVALID_OUTPUT = 'INVALID_OUTPUT'
+    ACTION_NOT_FOUND = 'ACTION_NOT_FOUND'
+    PANIC = 'PANIC'
```
huangjeff5: exclude go-specific like PANIC

huangjeff5: Dropped. `RuntimeErrorReason` is only the tokens helpers actually write. `PANIC` is gone, and so are `ACTION_COMPLETED` and the stream-* tokens — Python does not classify those.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -778,412 +897,758 @@ def box_background_start(
...
-            turn_options = await apply_resources(registry, turn_options, run_ctx.abort_signal)
-        assert_valid_tool_names(turn_tools)
+                ctx=ctx,
+                next_fn=run_turn,
+                extra=lambda p: {'iteration': p.iteration},
+                after_result=remember_response,
+            ),
+            name=resolved.model.name,
+        )
+    except (Exception, asyncio.CancelledError) as exc:
```
apascal07, line 1236: `except (Exception, asyncio.CancelledError)` catches cancellation and returns a response. That stops cancellation from working.

I ran this on the branch with a model that sleeps 5s:

- `await asyncio.wait_for(ai.generate(model='slow', prompt='x'), timeout=0.3)` returns a `ModelResponse`. No `TimeoutError`.
- `t = create_task(...)`, then `t.cancel()`, then `await t` returns a `ModelResponse`. No `CancelledError`.

The same clause is at lines 1316 and 1574. It bites harder now that the stream APIs have no `timeout=` and the guidance is to cancel the task instead, because cancelling the task is the thing that stopped working.

One way out: box the response when `ctx.abort_signal` is set, since that is our own abort, and re-raise `CancelledError` otherwise.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -778,412 +897,758 @@ def box_background_start(
...
+            registry=registry,
+            resolved=resolved,
+            call=call,
+            mw_pipeline=mw_pipeline,
+            current_turn=current_turn,
+        )
 
-            runner = cast(Callable[[ModelHookParams, GenerateMiddlewareContext], Awaitable[ModelResponse]], run_next)
-        return await runner(params, ctx)
+    def remember_response(result: object) -> None:
```
apascal07, line 1215: `after_result` fires after every `wrap_generate` hook, so `call.remember` runs once per hook. A middleware that returns a changed response adds its message on top of the one already remembered. `history_with_closed_turn` only skips a duplicate when the message is equal.

Repro: inner middleware returns `r.model_copy(update={'message': redacted})`, outer middleware raises after `next()`. `response.messages` comes back as:

```
user  ['hi']
model ['secret']
model ['REDACTED']
```

Two problems. The text from before the middleware is in the history we tell people to resend, and most providers reject two model turns in a row.

Suggest replacing the last remembered message when it came from the same turn, instead of appending.

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
```diff
@@ -1053,12 +1057,20 @@ def on_chunk(chunk: ModelResponseChunk) -> None:
...
+    # Max-turns abort has no model message (the refused round was
+    # dropped) but request.messages still has the closed tool turns.
+    await persist_turn_messages(
+        session_runner=session_runner,
+        history=history,
+        response_message=response.message,
+        response=response,
+    )
+
+    if response.error is not None:
```
apascal07, line 1069: The PR says invalid structured output is not a dead turn: `finish_reason` stays `STOP` and `.text` still holds the model text. This raise treats every `response.error` the same, so agents do not follow that rule.

Checked with a prompt that has `output_schema=Recipe` and a model that answers `not json at all`:

- `ai.generate` gives `finish_reason=stop`, `error.reason=INVALID_OUTPUT`, and readable text.
- The same prompt through `ai.define_prompt_agent` raises `AgentError('Model output was not valid JSON for the requested schema')`. The turn dies and the snapshot fails, even though the model finished fine.

Suggest skipping the raise when `response.error.reason is RuntimeErrorReason.INVALID_OUTPUT`.

py/packages/genkit/src/genkit/_ai/_generate.py
```diff
@@ -778,412 +897,758 @@ def box_background_start(
...
+        details = None
+    if reason is not None and not caller_stopped:
+        details = dict(details) if isinstance(details, Mapping) else {}
+        details['reason'] = reason.value
+    return box_dead_turn(
+        response=response,
+        messages=messages,
+        finish_reason=FinishReason.ABORTED if caller_stopped else FinishReason.FAILED,
+        finish_message=finish_message,
+        error=GenkitRuntimeError(status=status, message=finish_message, details=details),
```
apascal07, line 1060: `finish_message` is the raw `str(exc)`, and it lands on `response.error.message`. Internal failure text now travels on a normal response instead of on an exception.

A tool raising `RuntimeError('connection to postgres://user:pw@host refused')` used to leave generate as an exception, and the handler could tell a `PublicError` from a normal one (the rule written at `_core/_context.py:53`) and answer with a generic 500. Now a flow that returns the response writes that string into the 200 body.

Worth a decision: keep the raw text, or swap non-`PublicError` text for something generic here.

py/packages/genkit/src/genkit/_ai/_resource.py
```diff
@@ -80,7 +85,11 @@ async def lookup_resource_by_name(registry: Registry, name: str) -> Action:
         or await registry.resolve_action(ActionKind.RESOURCE, f'/dynamic-action-provider/{name}')
     )
     if not resource:
-        raise ValueError(f'Resource {name} not found')
+        raise GenkitError(
+            status='INVALID_ARGUMENT',
+            message=f'Resource {name} not found',
```
apascal07, line 90: A resource name that is not registered raises `INVALID_ARGUMENT`. The unmatched-URI path raises `NOT_FOUND`, `generate_test.py:6036` asserts `NOT_FOUND`, and the PR body lists an unmatched resource as `NOT_FOUND`.

So `ai.generate(resources=['missingResource'], ...)` gives status `INVALID_ARGUMENT` with the message "Resource missingResource not found". Anyone branching on `NOT_FOUND` misses this one.

py/packages/genkit/src/genkit/_core/_channel.py
```diff
@@ -67,18 +62,13 @@ async def __anext__(self) -> T_co:
 
         pop_task = asyncio.ensure_future(self._pop())
         if not self._close_future:
-            return await wait_for(pop_task, timeout=self._timeout)
+            return await pop_task
```
apascal07, line 65: This was the last caller of `_compat.wait_for`. With the timeout gone, `wait_for_310` and the `wait_for` alias in `_core/_compat.py` (lines 38-49) have no users left. Worth deleting them here so the version-gated branch does not sit around untested.

py/packages/genkit/src/genkit/_ai/_prompt.py
```diff
@@ -720,17 +712,11 @@ def coerce_prompt_template_input(template_input: Any) -> dict[str, Any]:  # noqa
     return cast(dict[str, Any], template_input)
 
 
-async def to_generate_request(registry: Registry, options: GenerateActionOptions) -> ModelRequest:
+async def to_model_request(*, registry: Registry, options: GenerateActionOptions) -> ModelRequest:
```
apascal07, line 715: There are now two public `to_model_request` functions with different keyword-only signatures: this one takes `(registry, options)`, and `_generate.py:1925` takes `(options, tools, model)`. `_prompt.py` already imports six names from `_generate` at line 37.

If someone adds `to_model_request` to that import block it silently shadows this one, and the call at line 503 fails with a confusing "missing 'tools'" error rather than a name error. Renaming one of them (maybe `to_prompt_model_request`) closes the trap.

apascal07 (approved):

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: We re-raise `CancelledError` unless `ctx.abort_signal` is set. `wait_for` and `task.cancel()` raise again; abort_signal still boxes ABORTED.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: `remember()` now replaces the last same-turn message when a later hook rewrites it, so history is one model turn (the rewritten one), not secret then REDACTED.

py/packages/genkit/src/genkit/_ai/_agents/_runtime.py
huangjeff5: We skip the raise when `response.error.reason is INVALID_OUTPUT`. send() returns STOP + the model text; other `response.error` cases still fail the snapshot.

py/packages/genkit/src/genkit/_ai/_generate.py
huangjeff5: We went with generic unless PublicError. A tool `RuntimeError` is `internal error` on the returned response. `PublicError` is how the tool author publishes the sentence. Stream-sink wording is unchanged.

py/packages/genkit/src/genkit/_ai/_resource.py
huangjeff5: A missing resource name is `NOT_FOUND` now, same as an unmatched URI.

py/packages/genkit/src/genkit/_core/_channel.py
huangjeff5: `wait_for_310` and the `wait_for` alias are deleted.

py/packages/genkit/src/genkit/_ai/_prompt.py
huangjeff5: The prompt one is now `to_prompt_model_request`.

PR #6258, opened 2026-09-02
fix(py/plugins/openai): handle completions with no choices

### LINKED ISSUE #6232: fix(py/plugins/openai): empty choices raises IndexError on the non-streaming path
`_generate` indexes `choices[0]` without checking the list is non-empty, so a completion with no choices raises `IndexError` instead of anything a caller can handle. The debug log one line above already guards for it.

The trigger is a well-formed 200: a gateway returning a usage-only completion, or a content filter that drops the choice rather than annotating it. `_generate_stream` is unaffected, since it skips chunks that carry no choices.

The two references disagree, so this needs a decision rather than a straight port. JS returns the usage-only response with no message; Go errors with `ErrInvalidOutput`.

hilariie:
Fixes #6232

## Summary

A completion with no choices now raises `GenkitError(status='INTERNAL')` from the plugin instead of IndexError. INTERNAL matches Go's `ErrInvalidOutput`.

The streaming path gets the same treatment: a stream in which no chunk carried a choice raises the same error instead of returning an empty message. The check keys off whether a choice was seen, not off accumulated content: a tool-call-only stream and OpenAI's usual empty first and last deltas carry choices but no content, and must still succeed.

## Verification

| case | before | after |
| --- | --- | --- |
| non-streaming, `choices: []` | IndexError | GenkitError, INTERNAL |
| stream where no chunk carries a choice | empty message | GenkitError, INTERNAL |
| tool-call-only stream | ok | ok |
| stream of empty deltas | ok | ok |

### DIFF
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -399,14 +399,18 @@ class OpenAIModel:
         openai_config.pop('stream_options', None)
         logger.debug('OpenAI generate request', model=self._model, streaming=False)
         response = await self._openai_client.chat.completions.create(**openai_config)
+        if not response.choices:
+            raise GenkitError(
+                status='INTERNAL',
+                message='No choices in completion.',
+                details={'usage': _usage_from_completion(response.usage).model_dump(exclude_none=True)},
+            )
+
         logger.debug(
             'OpenAI raw API response',
             model=self._model,
-            finish_reason=str(response.choices[0].finish_reason) if response.choices else None,
+            finish_reason=str(response.choices[0].finish_reason),
         )
-
-        if not response.choices:
-            raise GenkitError(status='INTERNAL', message='No choices in completion.')
 
         result = ModelResponse(
             request=request,
@@ -494,7 +498,11 @@ class OpenAIModel:
                 callback(ModelResponseChunk(role=Role.MODEL, content=content))
 
         if not saw_choice:
-            raise GenkitError(status='INTERNAL', message='No choices in completion.')
+            raise GenkitError(
+                status='INTERNAL',
+                message='No choices in completion.',
+                details={'usage': _usage_from_completion(usage).model_dump(exclude_none=True)},
+            )
 
         if tool_calls:
             message = MessageConverter.to_genkit(
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -27,6 +27,7 @@ from openai.types import CompletionUsage
 from openai.types.completion_usage import CompletionTokensDetails, PromptTokensDetails
 
 from genkit import (
+    GenkitError,
     Message,
     ModelRequest,
     ModelResponse,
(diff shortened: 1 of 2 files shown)


py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -403,6 +404,9 @@ async def _generate(self, request: ModelRequest) -> ModelResponse:
             finish_reason=str(response.choices[0].finish_reason) if response.choices else None,
         )
 
+        if not response.choices:
+            raise GenkitError(status='INTERNAL', message='No choices in completion.')
```
cabljac: The provider charged for this request and told us the token count, but the raise drops it. `response.usage` is right there, and same story at :496 where `usage` is already captured before the `continue`. From what I can see `GenkitError` takes `details`, so something like:

```python
raise GenkitError(
    status='INTERNAL',
    message='No choices in completion.',
    details={'usage': _usage_from_completion(response.usage)},
)
```

Can you check that `details` is the right place to carry it, and whether the tracing/telemetry side picks it up from there?

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -403,6 +404,9 @@ async def _generate(self, request: ModelRequest) -> ModelResponse:
             finish_reason=str(response.choices[0].finish_reason) if response.choices else None,
```
cabljac: nit: the `if response.choices else None` guard only exists to survive the case the next block raises on. Move the `if not response.choices` check above the log and this can go back to a plain `response.choices[0].finish_reason`.

py/packages/genkit-openai/tests/openai_model_test.py
```diff
@@ -398,6 +444,54 @@ def callback(chunk: ModelResponseChunk) -> None:
...
+@pytest.mark.asyncio
+async def test__generate_stream_no_choices(sample_request: ModelRequest) -> None:
+    """A stream carrying only the usage chunk fails rather than returning an empty response."""
+    mock_client = MagicMock()
+    mock_client.chat.completions.create = _mock_stream([_usage_chunk()])
+
+    model = OpenAIModel(model='gpt-4', client=mock_client)
+
+    with pytest.raises(GenkitError) as exc_info:
+        await model._generate_stream(sample_request, lambda _: None)
```
cabljac, line 473: nit: both no-choices tests call the private `_generate`/`_generate_stream`, so nothing pins that `GenkitError` escapes the `except (APIStatusError, ValueError)` in `generate()` (model.py:527). It does today (plain `Exception` subclass), but a one-line test through `generate()` would stop a future `except Exception` from swallowing it.

cabljac (approved):
Fix looks right on both paths and matches Go (`ErrInvalidOutput` on non-streaming at generate.go:726 and via the accumulator on streaming at :595). The two no-choices tests fail on main and pass here.

One ask, inline at model.py:408: the provider-returned usage gets dropped on the error path in both `_generate` and `_generate_stream`. Worth saying Go drops it too, so this would be Python going a step past both references rather than a parity gap. It's already in scope at both raise sites, so attaching it via `details` is a one-liner. Happy either way if you'd rather keep it out, just say so here.

The rest are nits.

PR #6259, opened 2026-09-02
fix(py/plugins/openai): report finish reason, finish message, and refusals

### LINKED ISSUE #6230: fix(py/plugins/openai): finish_reason, finish_message, and refusals never set
Neither chat path sets `finish_reason` or `finish_message`, so every response reports `None` no matter how the model stopped. `_generate` reads the finish reason only to write a debug log and then discards it, while `_generate_stream` never reads it at all.

Refusals are worse than absent: nothing in the plugin reads `refusal`, so a refusal-only message (no content, no tool calls) raises `ValueError: Unable to determine content part` instead of returning a blocked response.

Go maps the reason with a provider-aware switch that also covers xAI's `end_turn` and `model_context_window_exceeded` and DeepSeek's `insufficient_system_resource`, and sets the finish message from `refusal` and from the error object a gateway attaches to a failing choice. JS maps the reason but does not read `refusal` either (tracked for JS in #6165).

hilariie:
Closes #6230

## Summary

Both chat paths now map the provider's finish reason and set the finish message.

The reason table covers what no OpenAI model sends but the compatible providers do: xAI's end_turn, z.AI's sensitive, model_context_window_exceeded and network_error, DeepSeek's insufficient_system_resource, and a gateway's error. An unrecognized or absent reason is unknown.

The finish message comes from the error object a gateway attaches to a failing choice. A gateway that loses its upstream part way through a generation has already committed a 200, so the failure rides on the choice next to whatever text was produced first, and reading it is the only thing that tells a failed request from a complete answer. A choice that stopped cleanly reports no finish message even when an error-shaped extra rides beside it.

`refusal` is read on both paths. The response is blocked and the refusal is the finish message. A refusal wins over a gateway failure message, since the model did reply, having declined. A refusal-only message converts to an empty-content message instead of raising.

A message with no content at a blocked or truncated stop is the answer and keeps its finish reason; only an empty message at a clean stop is treated as unreadable.

### DIFF
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -27,6 +27,7 @@ from openai.types import CompletionUsage
 from openai.types.completion_usage import CompletionTokensDetails, PromptTokensDetails
 
 from genkit import (
+    FinishReason,
     GenkitError,
     Message,
     ModelRequest,
@@ -146,6 +147,69 @@ def _usage_from_completion(usage: CompletionUsage | None) -> ModelUsage:
         ),
         custom=custom or None,
     )
+
+
+def _failure_message(extras: dict[str, Any] | None) -> str | None:
+    """The message of an error object a provider attached to a choice.
+
+    A gateway whose upstream fails part-way through a generation answers with
+    a 200 — status and headers are already committed — and reports the failure
+    as an ``error`` object on the choice, beside whatever text was produced
+    before it.
+
+    Args:
+        extras: The fields a choice carried that are not in OpenAI's schema.
+
+    Returns:
+        The failure message, or None when the choice carries no error object.
+    """
+    failure = extras.get('error') if extras else None
+    if not isinstance(failure, dict):
+        return None
+    message = failure.get('message')
+    return message if isinstance(message, str) and message else None
+
+
+_FINISH_REASONS: dict[str, FinishReason] = {
+    'content_filter': FinishReason.BLOCKED,
+    'end_turn': FinishReason.STOP,
+    'error': FinishReason.OTHER,
+    'function_call': FinishReason.OTHER,
+    'insufficient_system_resource': FinishReason.OTHER,
+    'length': FinishReason.LENGTH,
+    'model_context_window_exceeded': FinishReason.LENGTH,
+    'network_error': FinishReason.OTHER,
+    'sensitive': FinishReason.BLOCKED,
+    'stop': FinishReason.STOP,
+    'tool_calls': FinishReason.STOP,
+}
+
+
+def _finish_state(
+    reason: str | None, *, refusal: str | None, failure_message: str | None
+) -> tuple[FinishReason, str | None]:
+    """The finish reason and finish message to report for a choice.
+
+    A refusal is the more specific answer and wins over a failure message: the
+    model replied, having declined. A choice that stopped cleanly reports no
+    finish message even when an error-shaped extra rides beside it.
+
+    Args:
+        reason: The ``finish_reason`` the provider sent, or None when it sent
+            none.
+        refusal: The refusal the model replied with, or None.
+        failure_message: The message of an error object on the choice, or None.
+
+    Returns:
+        The mapped finish reason, UNKNOWN when the reason is unrecognized or
+        absent, and the finish message, None when there is none to report.
+    """
+    if refusal:
+        return FinishReason.BLOCKED, refusal
+    finish_reason = _FINISH_REASONS.get(reason or '', FinishReason.UNKNOWN)
+    if finish_reason is FinishReason.STOP:
+        return finish_reason, None
+    return finish_reason, failure_message
 
 
 class OpenAIModel:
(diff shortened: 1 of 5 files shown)


py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -403,9 +469,19 @@ async def _generate(self, request: ModelRequest) -> ModelResponse:
             finish_reason=str(response.choices[0].finish_reason) if response.choices else None,
         )
 
+        choice = response.choices[0]
```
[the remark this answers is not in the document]
hilariie: Pre-existing, not from this change. The empty choices case is fixed in #6258 (both chat paths). Will stack this PR on #6258

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -148,6 +149,71 @@ def _usage_from_completion(usage: CompletionUsage | None) -> ModelUsage:
...
+    """
+    failure = extras.get('error') if extras else None
+    if not isinstance(failure, dict):
+        return None
+    message = failure.get('message')
+    return message if isinstance(message, str) and message else None
+
+
+# Includes the reasons no OpenAI model sends: end_turn and
+# model_context_window_exceeded are xAI's, insufficient_system_resource DeepSeek's.
```
cabljac: From what I can see this attribution is off: `model_context_window_exceeded` and `sensitive` are z.AI's values, not xAI's or DeepSeek's. They came into Go via #5828 (z.AI support) and docs.z.ai lists `stop, tool_calls, length, sensitive, model_context_window_exceeded, network_error`. Worth a check on your side. Either way I'd drop the comment: the parametrized test already pins each value, and provider provenance belongs in the PR body where it can be checked.

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -148,6 +149,71 @@ def _usage_from_completion(usage: CompletionUsage | None) -> ModelUsage:
...
+        refusal: The refusal the model replied with, or None.
+        failure_message: The message of an error object on the choice, or None.
+
+    Returns:
+        The mapped finish reason, UNKNOWN when the reason is unrecognized or
+        absent, and the finish message, None when there is none to report.
+    """
+    if refusal:
+        return FinishReason.BLOCKED, refusal
+    finish_reason = _FINISH_REASONS.get(reason, FinishReason.UNKNOWN) if reason else FinishReason.UNKNOWN
```
cabljac: nit: `dict.get(None, UNKNOWN)` already returns UNKNOWN, so the `if reason else` branch is dead. `_FINISH_REASONS.get(reason or '', FinishReason.UNKNOWN)` or widen the key type.

py/packages/genkit-openai/tests/openai_model_test.py
```diff
@@ -492,6 +547,222 @@ async def test__generate_stream_empty_deltas(sample_request: ModelRequest) -> No
...
+    assert response.message.content == []
+
+
+@pytest.mark.asyncio
+async def test__generate_stream_reports_the_failure_message_on_a_choice(sample_request: ModelRequest) -> None:
+    """A stream whose upstream failed reports the message the gateway sent."""
+    mock_client = MagicMock()
+    mock_client.chat.completions.create = _mock_stream([
+        _chunk(content='Partial '),
+        _chunk(finish_reason='error', error={'message': 'upstream timed out', 'code': 504}),
```
cabljac, line 763: nit: both streaming failure tests put the error object on the same chunk as `finish_reason='error'`. The accumulation at model.py:533-534 is last-non-empty-wins, but nothing pins the case where a gateway attaches the error mid-stream and then closes with `stop` (which should drop the message, same as the non-streaming `clean_stop` test). One extra case would cover it.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
```diff
@@ -480,7 +508,7 @@ def to_genkit(cls, message: ChatCompletionMessageAdapter) -> Message:
             if message.content:
                 content.append(cls.text_part_to_genkit(message.content))
 
-        if not content:
+        if not content and not message.refusal:
             raise ValueError('Unable to determine content part')
```
cabljac: From what I can see `_generate` computes the finish state first and then this raise discards it: a non-stream choice with `content: null` and no `refusal` ends up as `INTERNAL` whatever the reason was. The SDK types `content` as `Optional[str]` and documents `content_filter` as content omitted, so that looks like a real shape, and the same applies to `length`. The stream path returns `BLOCKED` / `LENGTH` with an empty message for the same events, and Go/JS do the same, so the two paths currently disagree.

Can you check that with `content=None, finish_reason='content_filter'` on the non-stream path? If it reproduces, I'd move the empty check into `_generate` and only raise when the reason is `STOP`, plus a test for that case. `test_generate_classifies_an_unreadable_message` (content None + stop) can stay as is.

nit: the stream path has no test for "error object on an early chunk, then clean stop" or "error field is a string". Both behave right from what I can see, just not pinned.

cabljac (approved):

py/packages/genkit-openai/src/genkit_openai/models/model.py
hilariie: comment dropped and the PR body corrected

py/packages/genkit-openai/src/genkit_openai/models/model.py
hilariie: resolved in [563370c](https://github.com/genkit-ai/genkit/commit/563370c1f7987874010ff3d7ff186eecaf6454a0)

py/packages/genkit-openai/tests/openai_model_test.py
hilariie: resolved in [563370c](https://github.com/genkit-ai/genkit/commit/563370c1f7987874010ff3d7ff186eecaf6454a0)

py/packages/genkit-openai/src/genkit_openai/models/utils.py
hilariie: Moved the empty check into _generate, raising only at stop. content_filter and length now return an empty message with their reason, tests also added

PR #6261, opened 2026-09-02
fix(py/plugins/openai): raw and custom response metadata never set

### LINKED ISSUE #6231: fix(py/plugins/openai): raw and custom response metadata never set
Neither chat path sets `raw` or `custom`, so every piece of response metadata outside the generated message is dropped. `system_fingerprint`, `id`, and `model` are never read on either path.

Non-standard fields have nowhere to land either. xAI answers a live search with a `citations` field the SDK does not model, and a gateway whose upstream fails mid-generation returns a 200 with an error object on the choice. A caller that asked for citations has no way to read them back.

Go collects the fingerprint, model, id, citations, and error object into one map and assigns it to both `Raw` and `Custom`. JS sets `raw` to the whole response.

hilariie:
Closes #6231

## Summary

Both chat paths now collect the response metadata the message does not carry and set it on `custom` and `raw`: `systemFingerprint`, `model`, `id`, the `citations` xAI returns for a live search, and the `error` object a gateway attaches to a failing choice. Absent fields are omitted rather than written as null. The fenced-JSON cleanup copies the response instead of rebuilding it field by field.

`model` and `id` are reported whenever present, not only next to a `system_fingerprint`: DeepSeek, xAI, OpenRouter and Ollama never send one. `raw` holds the same map as `custom` on both paths; the streaming path never materialises a full completion to dump.

### DIFF
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -46,6 +46,7 @@ from genkit_openai.models.utils import (
     DictMessageAdapter,
     MessageAdapter,
     MessageConverter,
+    extract_response_metadata,
     reraise_openai_error,
     strip_markdown_fences,
 )
@@ -380,6 +381,7 @@ class OpenAIModel:
                 latency_ms=response.latency_ms,
                 usage=response.usage,
                 custom=response.custom,
+                raw=response.raw,
             )
         return response
 
@@ -489,12 +491,15 @@ class OpenAIModel:
         if not genkit_message.content and finish_reason is FinishReason.STOP:
             raise ValueError('Unable to determine content part')
 
+        metadata = extract_response_metadata(response)
         result = ModelResponse(
             request=request,
             message=genkit_message,
             finish_reason=finish_reason,
             finish_message=finish_message,
             usage=_usage_from_completion(response.usage),
+            custom=metadata or None,
+            raw=metadata or None,
         )
         return self._clean_json_response(result, request)
 
@@ -521,12 +526,14 @@ class OpenAIModel:
 
         tool_calls: dict[int, Any] = {}
         accumulated_content: list[Part] = []
+        metadata: dict[str, Any] = {}
         usage: CompletionUsage | None = None
         saw_choice = False
         raw_finish_reason: str | None = None
         refusal_fragments: list[str] = []
         failure_message: str | None = None
         async for chunk in stream:  # type: ignore
+            metadata.update(extract_response_metadata(chunk))
             # Usage rides on a final chunk that carries no choices.
             if chunk.usage is not None:
                 usage = chunk.usage
@@ -615,6 +622,8 @@ class OpenAIModel:
             finish_reason=finish_reason,
             finish_message=finish_message,
             usage=_usage_from_completion(usage),
+            custom=metadata or None,
+            raw=metadata or None,
         )
         return self._clean_json_response(result, request)
 
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -373,16 +373,7 @@ class OpenAIModel:
                 cleaned_parts.append(part)
 
         if changed:
-            return ModelResponse(
-                request=request,
-                message=Message(role=response.message.role, content=cleaned_parts),
-                finish_reason=response.finish_reason,
-                finish_message=response.finish_message,
-                latency_ms=response.latency_ms,
-                usage=response.usage,
-                custom=response.custom,
-                raw=response.raw,
-            )
+            return response.model_copy(update={'message': Message(role=response.message.role, content=cleaned_parts)})
         return response
 
     @staticmethod
--- a/py/packages/genkit-openai/src/genkit_openai/models/utils.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/utils.py
@@ -23,7 +23,8 @@ import re
 from collections.abc import Callable
 from typing import Any, NoReturn
 
-from openai import APIStatusError
+from openai import APIStatusError, BaseModel
+from openai.types.chat import ChatCompletion, ChatCompletionChunk
 
 from genkit import (
     GenkitError,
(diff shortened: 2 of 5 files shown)


py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -481,12 +483,15 @@ async def _generate(self, request: ModelRequest) -> ModelResponse:
...
 
+        metadata = extract_response_metadata(response)
         result = ModelResponse(
             request=request,
             message=MessageConverter.to_genkit(message),
             finish_reason=finish_reason,
             finish_message=finish_message,
             usage=_usage_from_completion(response.usage),
+            custom=metadata or None,
+            raw=metadata or None,
```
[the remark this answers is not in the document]
hilariie: Pydantic already validates each field into its own dict, so the two references are independent; a shallow copy would not separate the nested values either. Matches Go, which assigns the same map to Custom and Raw.

cabljac, line 493: nit: on this path we have the full `ChatCompletion`, so `raw` could be `response.model_dump()` rather than the same curated map as `custom`. That's what JS does (`raw: response`, js/plugins/compat-oai/src/model.ts:707), and it means the Dev UI shows any provider field we didn't think to extract, which is the case `raw` is for. I get that the stream path can't do the same since it only ever sees chunks, so the current shape there is fine by me. Happy to take this as a follow-up if you'd rather keep the two paths symmetric here.

hilariie: Taking the follow-up option because dumping the full completion into raw needs four existing tests to stop feeding _generate bare mocks, so I will do both in a small PR on top of this one.

cabljac: Makes sense. Can you file it as an issue so it doesn't get lost, with the note about the four tests feeding _generate bare mocks? Happy for it to land after this one.

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -380,6 +381,7 @@ def _clean_json_response(self, response: ModelResponse, request: ModelRequest) -
                 latency_ms=response.latency_ms,
                 usage=response.usage,
                 custom=response.custom,
+                raw=response.raw,
```
cabljac: This rebuild is why `raw` went missing in the first place: it lists fields by hand, so anything added to `ModelResponse` elsewhere has to be remembered here. `operation` and `candidates` already fall off it (both None on this path, so nothing breaks yet).

`response.model_copy(update={'message': Message(role=response.message.role, content=cleaned_parts)})` cannot drop a field. I checked it carries `custom`, `raw` and `latency_ms` through the copy.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
```diff
@@ -202,6 +203,60 @@ def extract_config_dict(request: ModelRequest) -> dict[str, Any]:
...
+        A dict of the metadata fields the response actually carries, empty
+        when it carries none.
+    """
+    metadata: dict[str, Any] = {}
+    for key, value in (
+        ('systemFingerprint', response.system_fingerprint),
+        ('model', response.model),
+        ('id', response.id),
+    ):
+        if isinstance(value, str) and value:
```
cabljac, line 244: nit: nothing pins the `and value` half. I dropped it locally and all 250 tests still pass. Worth either a case for an empty-string `model`/`id`, or dropping the guard, since `construct` fills a missing required field with None and the `isinstance` already catches that.

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -521,12 +526,14 @@ async def _generate_stream(
...
         tool_calls: dict[int, Any] = {}
         accumulated_content: list[Part] = []
+        metadata: dict[str, Any] = {}
         usage: CompletionUsage | None = None
         saw_choice = False
         raw_finish_reason: str | None = None
         refusal_fragments: list[str] = []
         failure_message: str | None = None
         async for chunk in stream:  # type: ignore
+            metadata.update(extract_response_metadata(chunk))
```
cabljac, line 527: nit: `update` is last-write-wins per key, which is what you want for `error` and for `id`/`model`. For `citations` it assumes xAI sends the complete list in one chunk. Can you confirm that? If it ever streams them in slices, only the last slice would survive.

cabljac (commented):
Looks good. The fixture pair is what stands out: building the responses through `.construct()` is what the SDK itself does off the wire (`_base_client._process_response_data` -> `construct_type` -> `.construct`), so `model_extra` behaves in the tests exactly as it does in production. That is why the citations and choice-error cases are trustworthy rather than merely green.

Tiny thing, gemini's `getattr(source, 'model_extra', None)` on utils.py:219 is not worth taking. `source` is a `BaseModel` by signature, and a MagicMock returns a MagicMock there rather than raising, so the `isinstance` already handles it. The only caller that would raise is one passing a plain dict, which is exactly the case you want to hear about.

One thing worth digging into inline on `_clean_json_response`, then two nits.

cabljac (approved):
lgtm with nits, and filing the follow-up as an issue and tracking it on the project board.

py/packages/genkit-openai/src/genkit_openai/models/model.py
hilariie: Confirmed that xAI sends citations as one complete list on the final chunk, so last-write-wins is what we want. Vercel's xai provider reads them the same way and Go's stream path overwrites too.

py/packages/genkit-openai/src/genkit_openai/models/model.py
hilariie: Switched to model_copy and added a test that sets operation as well, which fails against the old rebuild.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
hilariie: Kept the guard and pinned it with a test for empty id, model and fingerprint.

py/packages/genkit-openai/src/genkit_openai/models/model.py
hilariie: opened #6335

PR #6267, opened 2026-09-02
feat(js/middleware/context-compression): Add tool response deduplication

ssbushi:
- `deduplicateToolResponses`: Prunes repeated tool outputs, replacing them with a replacement notice while preserving recent responses.
- Input matching: Supports matching by tool name and exact input arguments (`name-and-input`) or tool name only (`name-only`).

### DIFF
--- a/js/plugins/middleware/README.md
+++ b/js/plugins/middleware/README.md
@@ -252,8 +252,9 @@ Compresses conversation context when it grows too large, reducing token usage an
 **Strategies applied:**
 
 1. **Safety cap**: Hard-truncates any single oversized tool response (`maxToolResponseChars`, default: 400,000 chars).
-2. **Tool response truncation**: Truncates tool responses exceeding a character limit (`toolResponses`), preserving the most recent responses.
-3. **Message count cap**: Drops older non-system messages when exceeding `maxMessages`, optionally inserting a notice.
+2. **Deduplication**: Replaces duplicate tool responses with a short notice (`deduplicateToolResponses`).
+3. **Tool response truncation**: Truncates tool responses exceeding a character limit (`toolResponses`), preserving the most recent responses.
+4. **Message count cap**: Drops older non-system messages when exceeding `maxMessages`, optionally inserting a notice.
 
 ```typescript
 import { genkit } from 'genkit';
@@ -268,6 +269,7 @@ const response = await ai.generate({
   use: [
     contextCompression({
       maxInputTokens: 80000,
+      deduplicateToolResponses: { matchBy: 'name-and-input' },
       toolResponses: { maxChars: 2000, preserveRecent: 2 },
       maxMessages: 20,
     }),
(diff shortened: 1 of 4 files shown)


js/plugins/middleware/src/context-compression.ts
```diff
@@ -330,6 +382,114 @@ export const contextCompression: GenerateMiddleware<
...
+            !hasCompressionFlag(part, 'deduplicated')
+          ) {
+            deduplicatedCount++;
+            changed = true;
+            return {
+              ...part,
+              metadata: withCompressionMetadata(part, { deduplicated: true }),
+              toolResponse: {
+                ...part.toolResponse,
+                output: dedupNotice,
```
apascal07: Only `output` is replaced; `toolResponse.content` stays. A screenshot tool called twice with the same input keeps sending the older image while the notice says the response was removed. Go's replacement (`dropContent`) clears multipart content for this reason.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -51,6 +51,38 @@ export const ToolResponsesOptionsSchema = z.object({
...
+    .describe(
+      'Match by tool name and arguments ("name-and-input") or name only ("name-only"). Default: "name-and-input".'
+    ),
+
+  /**
+   * Number of most recent responses to leave untouched per tool/args group.
+   * Older duplicates are replaced with `notice`.
+   * @default 1
+   */
+  keepRecent: z
```
apascal07: `keepRecent` has no `.int().nonnegative()` like the options in #6266. With `0`, every occurrence is replaced, including a tool's only response that the model has not read yet, and the notice points the model at "the most recent call", which is itself. Negative values replace more.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -330,6 +382,114 @@ export const contextCompression: GenerateMiddleware<
...
+                : prevParts.find(
+                    (p) => p.toolRequest?.name === part.toolResponse?.name
+                  );
+            if (positionalPart?.toolRequest) {
+              toolInput = positionalPart.toolRequest.input;
+            }
+          }
+
+          const key =
+            dedupMatchBy === 'name-only'
```
apascal07: `name-only` groups calls with different inputs. `fetchReport(101)`, `(102)`, `(103)` leaves only 103, and the notices on 101 and 102 tell the model to look at the most recent call, which returned a different report. The option deletes distinct data, and the docs do not say so.

js/plugins/middleware/README.md
```diff
@@ -284,6 +286,7 @@ const response = await ai.generate({
 | `maxInputTokens` | `number` | `Infinity` | Triggers compression when token count exceeds this threshold. |
 | `preserveSystem` | `boolean` | `true` | Always keep system instructions intact. |
 | `maxToolResponseChars` | `number` | `400000` | Hard cap on any single tool response size in characters. Set negative to disable. |
+| `deduplicateToolResponses` | `object` | — | Deduplication settings for repeated tool calls. |
```
apascal07: The table has no rows for `matchBy`, `keepRecent`, or `notice`, so the defaults (`name-and-input`, `1`) and the `name-only` mode are only in the source.

PR #6269, opened 2026-09-02
feat(js/middleware/context-compression): Add LLM summarization

ssbushi:
- LLM Summarization: Condenses older history into a summary message using a designated model while preserving the last N messages untouched.
- Custom prompt support
- Cost optimization: Skips the LLM summarization call if cheaper strategies save enough context.
- Adaptive overshoot scaling

### DIFF
--- a/js/plugins/middleware/src/context-compression.ts
+++ b/js/plugins/middleware/src/context-compression.ts
@@ -16,11 +16,13 @@
 
 import {
   generateMiddleware,
+  ModelReferenceSchema,
   z,
   type GenerateMiddleware,
   type MessageData,
   type Part,
 } from 'genkit';
+import { logger } from 'genkit/logging';
 import { AsyncLocalStorage } from 'node:async_hooks';
 
 interface CompressionExecutionState {
@@ -87,6 +89,33 @@ export const DeduplicateToolResponsesOptionsSchema = z.object({
     .string()
     .optional()
     .describe('Replacement text for deduplicated tool responses.'),
+});
+
+export const SummarizeOptionsSchema = z.object({
+  /**
+   * Model to use for summarization. A model reference, model name string,
+   * or ModelAction, e.g. `{ name: 'googleai/gemini-flash-lite-latest' }`.
+   */
+  model: ModelReferenceSchema.describe('Model to use for summarization.'),
+
+  /**
+   * Number of most recent messages to keep un-summarized.
+   * Everything before this window is replaced with a summary.
+   * @default 6
+   */
+  preserveRecent: z
+    .number()
+    .optional()
+    .describe('Keep last N messages un-summarized. Default: 6.'),
+
+  /**
+   * Custom summarization prompt. The string `{conversation}` will be
+   * replaced with a text rendering of the messages to summarize.
+   */
+  prompt: z
+    .string()
+    .optional()
+    .describe('Custom summarization prompt. Use {conversation} placeholder.'),
 });
 
 export const ContextCompressionOptionsSchema = z
@@ -157,6 +186,27 @@ export const ContextCompressionOptionsSchema = z
       .describe('Hard cap on message count. Drop oldest beyond this.'),
 
     /**
+     * Use an LLM to summarize older messages into a condensed form.
+     * The summary replaces the original messages, preserving recent context.
+     */
+    summarize: SummarizeOptionsSchema.optional().describe(
+      'Summarize older messages using an LLM.'
+    ),
+
+    /**
+     * If cheap strategies (deduplication + tool truncation) reduce estimated
+     * context by at least this fraction, skip the LLM summarization step.
+     * Set to `0` to always summarize when configured.
+     * @default undefined (always summarize when configured)
+     */
+    skipSummarizationThreshold: z
+      .number()
+      .optional()
+      .describe(
+        'Skip summarization if cheap strategies save at least this fraction of context. E.g. 0.3 = 30%.'
+      ),
+
+    /**
      * Insert a notice message when messages are dropped during message
      * truncation, so the model knows context was removed.
      * @default true
@@ -190,6 +240,15 @@ const DEFAULT_DEDUP_KEEP_RECENT = 1;
 const DEFAULT_DEDUP_NOTICE =
   '[Deduplicated: This tool response has been removed to save context. ' +
   'See the most recent call of this tool for current output.]';
+const DEFAULT_PRESERVE_RECENT = 4;
+const DEFAULT_SUMMARIZE_PRESERVE_RECENT = 6;
+const SUMMARY_PREFIX = '[Conversation Summary]';
+const DEFAULT_SUMMARIZE_PROMPT = `Summarize the following conversation concisely. Capture key facts, decisions made, tool calls and their results, and the current state of the conversation so that the assistant can continue helping the user effectively.
+
+Conversation:
+{conversation}
+
+Summary:`;
 const DEFAULT_TRUNCATION_NOTICE =
   '[NOTE] Some earlier messages in this conversation have been removed to stay within ' +
   'context limits. The most recent messages are preserved. Pay close attention to the ' +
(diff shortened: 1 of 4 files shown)


js/plugins/middleware/src/context-compression.ts
```diff
@@ -714,6 +869,69 @@ export const contextCompression: GenerateMiddleware<
...
+
+      if (nonSystemMessages.length <= summaryPreserveRecent) {
+        return { messages, summarized: false, tailCount: 0 };
+      }
+
+      const toSummarize = nonSystemMessages.slice(
+        0,
+        nonSystemMessages.length - summaryPreserveRecent
+      );
+      const toKeep = nonSystemMessages.slice(-summaryPreserveRecent);
```
apascal07: `toKeep` is a plain slice, so it can start with a `tool` message whose `toolRequest` was summarized away. With the example's `summarize.preserveRecent: 2` at overshoot 1.5 to 2, `adjustForOvershoot` gives 1 and `toKeep` is `[lastToolMessage]`. Anthropic and OpenAI reject a tool result without a preceding tool use, and Gemini rejects a `functionResponse` without its `functionCall`. Go moves the boundary back past tool messages.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -150,6 +189,27 @@ export const ContextCompressionOptionsSchema = z.object({
...
+    'Summarize older messages using an LLM.'
+  ),
+
+  /**
+   * If cheap strategies (deduplication + tool truncation) reduce estimated
+   * context by at least this fraction, skip the LLM summarization step.
+   * Set to `0` to always summarize when configured.
+   * @default undefined (always summarize when configured)
+   */
+  skipSummarizationThreshold: z
```
apascal07: The skip decision compares the fraction of chars saved against a threshold, not the result against `maxInputTokens`. At 2x budget, a 30% saving with a threshold of 0.3 skips the summary while the request is still 1.4x over budget, and the model call fails on context length. Go removed this option because the cheap strategies always run first.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -83,6 +85,33 @@ export const DeduplicateToolResponsesOptionsSchema = z.object({
...
+   * or ModelAction, e.g. `{ name: 'googleai/gemini-flash-lite-latest' }`.
+   */
+  model: ModelReferenceSchema.describe('Model to use for summarization.'),
+
+  /**
+   * Number of most recent messages to keep un-summarized.
+   * Everything before this window is replaced with a summary.
+   * @default 6
+   */
+  preserveRecent: z
```
apascal07: `summarize.preserveRecent: 0` makes `slice(-0)` return the whole array, so `toSummarize` and `toKeep` are both every message and the result is `[system, summary, ...all]`: larger than the input, after a paid summarizer call. This option, the top-level `preserveRecent`, and `skipSummarizationThreshold` (0..1) have none of the validation added in #6266.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -714,6 +869,69 @@ export const contextCompression: GenerateMiddleware<
...
+        const response = await ai.generate({
+          model: summaryModelRef as any,
+          config: summaryModelRef?.config,
+          prompt,
+        });
+
+        return {
+          messages: buildSummarizedMessages(
+            systemMessages,
+            response.text,
```
apascal07: `response.text` is used without checking `finishReason` or emptiness, and the call sets no output limit. A safety block or max-tokens stop replaces the whole prefix with `[Conversation Summary]` and nothing, or half a sentence. The task context is gone with no warning.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -714,6 +869,69 @@ export const contextCompression: GenerateMiddleware<
...
+      }
+
+      const toSummarize = nonSystemMessages.slice(
+        0,
+        nonSystemMessages.length - summaryPreserveRecent
+      );
+      const toKeep = nonSystemMessages.slice(-summaryPreserveRecent);
+
+      try {
+        const conversationText = renderMessages(toSummarize);
```
apascal07: The summarizer input is the prefix of a context that is already over budget, with no cap or chunking. With a 100k budget, a 250k context, and a 128k summarizer, the summary call fails, the error becomes a warning, and the main call then fails on context length too.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -714,6 +869,69 @@ export const contextCompression: GenerateMiddleware<
...
+        const prompt = summaryPromptTemplate.replaceAll(
+          '{conversation}',
+          () => conversationText
+        );
+
+        const response = await ai.generate({
```
apascal07: `ctx.abortSignal` and `ctx.context` are not passed. An aborted agent turn keeps the summarizer call running and billing, and model plugins that read per-request context do not get it here.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -820,15 +1055,54 @@ export const contextCompression: GenerateMiddleware<
...
+                        (basePreserveRecent - adjustedPreserveRecent)
+                    )
+                  )
+                : undefined;
+
+              if (
+                effectiveMaxMessages &&
+                messages.length > effectiveMaxMessages
+              ) {
+                const msgResult = applyMessageTruncation(
```
apascal07: Truncation runs after summarization and treats the summary as the oldest ordinary message. With `summarize.preserveRecent` at its default of 6 and `maxMessages: 6`, the result has 8 messages, truncation keeps the last 5, and the summary is dropped right after it was paid for. The anchor path can also pick the summary as the anchor user message.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -820,15 +1055,54 @@ export const contextCompression: GenerateMiddleware<
...
+                    messages,
+                    adjustedSummaryPreserveRecent
+                  );
+                  messages = sumResult.messages;
+                  isSummarized = sumResult.summarized;
+                }
+              }
+
+              // 5. Message truncation (as fallback or hard cap)
+              const effectiveMaxMessages = maxMessages
```
apascal07: The new top-level `preserveRecent` does not preserve anything. It only lowers `effectiveMaxMessages` by `base - adjusted` (at most 2 with defaults), and only when `maxMessages` is set. That makes three `preserveRecent` options with three meanings, and without `summarize` nothing here targets `maxInputTokens`.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -304,6 +394,57 @@ function lastReportedInputTokens(messages: MessageData[]): number | undefined {
...
+ * Inject conversation summary as a dedicated user message preceding preserved messages.
+ */
+function buildSummarizedMessages(
+  systemMessages: MessageData[],
+  summaryText: string,
+  toKeep: MessageData[]
+): MessageData[] {
+  const summaryPrefix = `${SUMMARY_PREFIX}\n${summaryText}`;
+  const summaryMessage: MessageData = {
+    role: 'user',
```
apascal07: The summary is built from tool output, which is untrusted, and goes to the model as a `user` message with only a `[Conversation Summary]` prefix. Instructions in a fetched page that the summarizer restates now carry user authority. Go frames its summary message as a record of a prior session.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -255,6 +324,27 @@ function withCompressionMetadata(
...
+  return messages
+    .map((m) => {
+      const parts = m.content
+        .map((p) => {
+          if (p.text) return p.text;
+          if (p.toolRequest)
+            return `[Tool call: ${p.toolRequest.name}(${JSON.stringify(p.toolRequest.input)})]`;
+          if (p.toolResponse)
+            return `[Tool response: ${p.toolResponse.name} → ${stringifyOutput(p.toolResponse.output)}]`;
+          return '[other content]';
```
apascal07: Media, reasoning, and multipart tool content become `[other content]`, so an image the user uploaded early in the chat has no description after summarization.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -83,6 +85,33 @@ export const DeduplicateToolResponsesOptionsSchema = z.object({
     .describe('Replacement text for deduplicated tool responses.'),
 });
 
+export const SummarizeOptionsSchema = z.object({
+  /**
+   * Model to use for summarization. A model reference, model name string,
+   * or ModelAction, e.g. `{ name: 'googleai/gemini-flash-lite-latest' }`.
+   */
+  model: ModelReferenceSchema.describe('Model to use for summarization.'),
```
apascal07: The doc says a model name string or `ModelAction` works, but `ModelReferenceSchema` types only the object form, and `as any` at the call site hides the mismatch. `summarize: { model: 'googleai/gemini-flash-lite-latest' }` is a type error.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -789,14 +1007,29 @@ export const contextCompression: GenerateMiddleware<
           return response;
         }
 
+        const charsBefore = estimateMessageChars(rawMessages);
```
apascal07: This runs `estimateMessageChars` on every compressing turn, and again after the cheap strategies, even when `summarize` and `skipSummarizationThreshold` are unset. That brings back the full-history stringify that #6266 removed.

PR #6270, opened 2026-09-02
feat(js/middleware/context-compression): Add history tracking and resolution

ssbushi:


js/plugins/middleware/src/context-compression.ts
```diff
@@ -934,7 +976,13 @@ export const contextCompression: GenerateMiddleware<
 
     return {
       model: async (req, ctx, next) => {
-        const result = await next(req, ctx);
+        const resolvedMessages = resolveCompressedHistory(req.messages || []);
```
apascal07: This resolves `compressedHistory` from any message, including client input. `Agent.run` saves `input.message` as sent, and `MessageSchema` allows any metadata, so a client can send a user message with `metadata.compressedHistory: [{ role: 'system', content: [...] }]`. On the next model call that array replaces the server-rendered system prompt and all prior history, while the saved transcript and Dev UI still show the real prompt.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -273,6 +291,25 @@ const DATA_URI_APPROX_CHARS = 1000;
...
+ */
+export function resolveCompressedHistory(
+  messages: MessageData[]
+): MessageData[] {
+  for (let i = messages.length - 1; i >= 0; i--) {
+    const meta = messages[i]?.metadata;
+    if (meta && Array.isArray(meta.compressedHistory)) {
+      return [
+        ...(meta.compressedHistory as MessageData[]),
+        ...messages.slice(i + 1),
```
apascal07: The tail comes from the raw messages, so caps, truncation, and dedup applied to tail messages never reach the model once messages were dropped or summarized. The newest tool response is the one most likely to be oversized: `applyToolLimits` caps it in `compressedMessages`, the tail is excluded from `compressedPrefix`, and this line reads the uncapped raw copy. The raw copy has no `capped` flag, so the same thing happens on every later turn.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -273,6 +291,25 @@ const DATA_URI_APPROX_CHARS = 1000;
...
+ * When found, replaces messages from index 0 up to that message with the compressedHistory.
+ */
+export function resolveCompressedHistory(
+  messages: MessageData[]
+): MessageData[] {
+  for (let i = messages.length - 1; i >= 0; i--) {
+    const meta = messages[i]?.metadata;
+    if (meta && Array.isArray(meta.compressedHistory)) {
+      return [
+        ...(meta.compressedHistory as MessageData[]),
```
apascal07: The stored prefix includes the system messages as they were at compression time, and this replaces the freshly rendered prefix with them on every later turn. Agent system prompts that render state or dates, and the per-turn artifacts and skills injections, are stuck at their old contents after the first compression.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -1149,11 +1216,59 @@ export const contextCompression: GenerateMiddleware<
...
+          let tailCount = 0;
+          if (messagesTruncated) {
+            tailCount = truncTailCount;
+          } else if (summarized) {
+            tailCount = summaryTailCount;
+          } else {
+            tailCount = 0;
+          }
+
+          const cutIndex =
```
apascal07: Compression ran on `activeMessages`, but `cutIndex` and the tail are computed against `rawMessages`, and the two differ once an earlier `compressedHistory` exists. If the new tail reaches back into the old compressed prefix (for example the anchor user message or the summary, which exist only in `compressedHistory`), `rawMessages.length - tailCount - 1` lands in the raw region that was dropped before. The model then gets previously dropped raw messages, possibly starting with an orphan tool response.

js/plugins/middleware/src/context-compression.ts
```diff
@@ -1149,11 +1216,59 @@ export const contextCompression: GenerateMiddleware<
...
+                )
+              : compressedMessages;
+
+          outgoingMessages = rawMessages.map((m, idx) => {
+            if (idx === cutIndex) {
+              return {
+                ...m,
+                metadata: {
+                  ...m.metadata,
+                  compressedHistory: structuredClone(compressedPrefix),
```
apascal07: Go stores a `{summary, stats}` boundary stamp under `contextCompression` and derives the view from it. This stores a full clone of the prefix under a new top-level `compressedHistory` key. Shared or Dev UI history now has two compaction formats, stored sessions roughly double in size, and the copy is recorded again in every generate span input.

PR #6282, opened 2026-09-04
fix(py/plugins/openai): emit content, reasoning, and tool-call deltas that share a chunk

### LINKED ISSUE #6233: fix(py/plugins/openai): streaming loop drops content, reasoning, or tool-call deltas that share a chunk
`_generate_stream` handles content, reasoning, and tool-call deltas in an `if / elif / elif` chain, so only the first matching branch runs and a chunk carrying two of them loses one. A reasoning model interleaving `reasoning_content` with `content`, or a provider emitting text alongside a tool-call delta, is enough to trigger it.

The loss is silent as a dropped tool-call fragment never reaches the accumulator, so the aggregate response is short too and the caller gets no signal. The non-streaming path has a narrower version of the same problem, returning tool calls or reasoning-plus-content, never both.

Go handles all three from the same chunk with independent `if`s and appends each to one outgoing chunk. JS combines reasoning with content but still makes tool calls exclusive of both, so it is only a partial reference.

hilariie:
Closes #6233

## Summary

The streaming loop now reads content, reasoning and tool-call deltas independently and appends them to a single outgoing chunk, in reasoning, text, tool-call order. The chunk is emitted only when it carries something. A tool-call fragment that shares a chunk with text reaches the accumulator, so the aggregate response is complete as well.

`MessageConverter.to_genkit` gets the same treatment on the non-streaming path: reasoning, text, then one part per tool call, rather than tool calls to the exclusion of both.

A text delta converts straight to a text part instead of going through `to_genkit`, which would otherwise return the reasoning and tool-call parts riding on the same delta a second time.

### DIFF
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -516,7 +516,8 @@ class OpenAIModel:
         stream = await self._openai_client.chat.completions.create(**openai_config)
 
         tool_calls: dict[int, Any] = {}
-        accumulated_content: list[Part] = []
+        reasoning_parts: list[Part] = []
+        text_parts: list[Part] = []
         metadata: dict[str, Any] = {}
         usage: CompletionUsage | None = None
         saw_choice = False
@@ -546,18 +547,16 @@ class OpenAIModel:
             if delta.refusal:
                 refusal_fragments.append(delta.refusal)
 
-            # Reasoning content (DeepSeek R1 / reasoner models). Pydantic
-            # models raise AttributeError for unknown fields, so the lookup
-            # goes through MessageAdapter.
+            # Reasoning content (DeepSeek R1 / reasoner models).
             if reasoning_text := MessageAdapter(delta).reasoning_content:
                 reasoning_part = Part(root=ReasoningPart(reasoning=reasoning_text))
-                accumulated_content.append(reasoning_part)
+                reasoning_parts.append(reasoning_part)
                 parts.append(reasoning_part)
 
             # Text content
             if delta.content:
                 text_part = MessageConverter.text_part_to_genkit(delta.content)
-                accumulated_content.append(text_part)
+                text_parts.append(text_part)
                 parts.append(text_part)
 
             # Tool calls (partial function calls)
@@ -585,6 +584,7 @@ class OpenAIModel:
                 details={'usage': _usage_from_completion(usage).model_dump(exclude_none=True)},
             )
 
+        accumulated_content: list[Part] = [*reasoning_parts, *text_parts]
         if tool_calls:
             message = MessageConverter.to_genkit(
                 DictMessageAdapter({'tool_calls': tool_calls.values(), 'role': Role.MODEL})
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -540,37 +540,28 @@ class OpenAIModel:
                 failure_message = failure
 
             delta = choice.delta
+            parts: list[Part] = []
 
             # A refusal streams in fragments and is reported as the finish message, not content.
             if delta.refusal:
                 refusal_fragments.append(delta.refusal)
 
-            # Text content chunk
-            if delta.content:
-                message = MessageConverter.to_genkit(MessageAdapter(delta))
-                accumulated_content.extend(message.content)
-                callback(
-                    ModelResponseChunk(
-                        role=Role.MODEL,
-                        content=message.content,
-                    )
-                )
-
-            # Reasoning content chunk (DeepSeek R1 / reasoner models).
-            # Note: Pydantic models raise AttributeError for unknown fields,
-            # so getattr() with a default doesn't work. Use try-except.
-            elif reasoning_text := MessageAdapter(delta).reasoning_content:
+            # Reasoning content (DeepSeek R1 / reasoner models). Pydantic
+            # models raise AttributeError for unknown fields, so the lookup
+            # goes through MessageAdapter.
+            if reasoning_text := MessageAdapter(delta).reasoning_content:
                 reasoning_part = Part(root=ReasoningPart(reasoning=reasoning_text))
                 accumulated_content.append(reasoning_part)
-                callback(
-                    ModelResponseChunk(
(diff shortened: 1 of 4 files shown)


py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -580,14 +571,15 @@ async def _generate_stream(
                         existing = tool_calls[tool_call.index]
                         if hasattr(existing, 'function') and existing.function and tool_call.function:
                             existing.function.arguments += tool_call.function.arguments
```
cabljac: `ChoiceDeltaToolCallFunction.arguments` is `Optional[str]`, so a fragment with `arguments: None` raises a bare `TypeError` here, and `generate()` only catches `APIStatusError` / `ValueError`. Pre-existing, but since this loop was rewritten here, worth folding in:

```python
existing.function.arguments = (existing.function.arguments or '') + (tool_call.function.arguments or '')
```

Can you add a case with a `None` fragment while you're in there.

hilariie: Folded in. The null fragment now also emits an empty input instead of the accumulated string, which args_segment=None used to fall back to. Added case test__generate_stream_tolerates_a_null_arguments_fragment.

cabljac: Thanks. One more case in the same spot: the tolerance stops at concatenation, and the final assembly still parses whatever accumulated. A zero-argument tool call arrives as `arguments: ''`, and `tool_call_to_genkit` hands that straight to `json.loads`, so it raises JSONDecodeError. I get the same from `_generate`, so it is not stream-only.

```python
args_input = args_parser(args_input) if args_input else {}
```

The callback path is unaffected since it passes no parser, so this only touches final assembly. Worth a zero-argument case alongside the null-fragment one.

Do you know whether OpenAI itself sends `''` here rather than `'{}'`? I have only checked the code, not a live provider, so how much this actually bites depends on that.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
```diff
@@ -520,9 +520,8 @@ def to_openai(cls, message: Message) -> list[dict]:
     def to_genkit(cls, message: ChatCompletionMessageAdapter) -> Message:
         """Converts an OpenAI-style message into a Genkit `Message` object.
 
-        Handles tool calls, reasoning content (from DeepSeek R1 / reasoner),
-        and regular text content. Matches the JS canonical implementation
-        in fromOpenAIChoice().
+        Emits a reasoning part, a text part, and one tool request part per
+        tool call, in that order, for whichever of them the message carries.
```
cabljac, line 524: This docstring is the first place that order is written down, and I think it is the right call. The stream loop cannot always produce it though: tool parts go onto `accumulated_content` after the loop, so a stream carrying text in one chunk and reasoning in a later one ends up `[text, reasoning]` where `to_genkit` would give `[reasoning, text]`.

DeepSeek sends all its reasoning before any content, so I doubt it is reachable today. But the order is a promise now and nothing pins it. Worth a test over a multi-chunk mixed stream asserting the final message order, so the two paths cannot drift apart quietly?

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -549,53 +549,43 @@ async def _generate_stream(
...
-                    )
-                )
-
-            # Reasoning content chunk (DeepSeek R1 / reasoner models).
-            # Note: Pydantic models raise AttributeError for unknown fields,
-            # so getattr() with a default doesn't work. Use try-except.
-            elif reasoning_text := MessageAdapter(delta).reasoning_content:
+            # Reasoning content (DeepSeek R1 / reasoner models). Pydantic
+            # models raise AttributeError for unknown fields, so the lookup
+            # goes through MessageAdapter.
```
cabljac: nit: `getattr(delta, 'reasoning_content', None)` returns None here rather than raising, so the reason given is not the real one. Bare attribute access is what raises. The reason to keep MessageAdapter is that it is the same lookup the non-stream path uses, which the code already shows. Worth cutting back to `# Reasoning content (DeepSeek R1 / reasoner models).`

cabljac (commented):
Nice, this is the right shape: three independent `if`s, one callback per chunk, and a single order that `to_genkit` and the stream loop now share. The `if parts:` guard is the detail I like most. Refusal-only and heartbeat deltas emitted nothing under the old if/elif chain by accident, and a naive rewrite would have started sending empty chunks; you kept it deliberately.

I ran the suite on the branch (257 pass) and broke each behaviour the PR claims to check the tests bite. They all do, including the ordering and the null-fragment case.

One thing worth digging into on the null-arguments thread, then two inline.

py/packages/genkit-openai/src/genkit_openai/models/model.py
hilariie: Done, cut back to the first sentence

py/packages/genkit-openai/src/genkit_openai/models/utils.py
hilariie: The stream loop now collects reasoning and text separately and assembles the final message as reasoning, text, then tool calls, the same order to_genkit gives. Fragments keep their arrival order within each group. Pinned in test__generate_stream_final_message_orders_parts_like_to_genkit, which feeds text, reasoning, tool call, text and asserts the streamed order is untouched while the final message is grouped.

py/packages/genkit-openai/src/genkit_openai/models/model.py
hilariie: Folded in: tool_call_to_genkit now parses an empty arguments string to {} when a parser is given, so both _generate and the stream assembly get an empty input. Added a zero-argument case for the stream, for _generate, and for to_genkit directly.

On the live question: checked with a forced call to a parameterless tool. gpt-4o-mini and gpt-4.1 send '{}' non-stream and ['', '{}'] streamed, so the accumulated string is '{}'. Ollama (qwen3.5:4b via /v1) sends '{}' both ways in a single fragment. So neither leaves it at ''. The only way to reach json.loads('') that I can see is a stream that stops after the '' opener, or a provider I have not checked. Cheap to guard, and the tests pin it, but I would not call it a bug people are hitting.

cabljac (approved):

PR #6298, opened 2026-09-07
fix(py): use async Google GenAI model discovery

### LINKED ISSUE #6186: fix(py/plugins/google-genai): init() blocks the event loop with a sync models.list()
`_list_genai_models` iterates a synchronous `client.models.list()` (`google.py:194`) and is called directly from `async def init` (`google.py:650` GoogleAI, `google.py:1018` VertexAI) plus the resolve/list_actions paths. This blocks the event loop on a network call at startup.

Fix: use the async client surface or `asyncio.to_thread`.

mikemikimike:
## Summary

- Use `client.aio.models.list()` for Google GenAI model discovery in asynchronous plugin entry points.
- Keep model discovery on one asynchronous path and remove the unused synchronous discovery helpers.
- Update the Google GenAI plugin tests to use reusable async pager mocks that provide a fresh iterator for each discovery call.

Fixes #6186

## Tests

- `pytest -p no:cacheprovider -o addopts='' py/packages/genkit-google-genai/test py/packages/genkit-google-genai/tests -q` — 738 passed
- `ruff check` on the changed Google GenAI source and test files — passed
- `python -m compileall` on the Google GenAI source and test packages — passed
- `git diff --check` on the changed files — passed

## Compatibility / Known limitations

- Model discovery now consistently uses `client.aio.models.list()` for both Google AI and Vertex AI plugin entry points.
- The removed synchronous/private discovery helpers were only used by tests and are replaced by the single asynchronous discovery path.


py/packages/genkit-google-genai/src/genkit_google_genai/google.py
```diff
@@ -251,6 +252,17 @@ def _list_genai_models(client: genai.Client, is_vertex: bool) -> GenaiModels:
...
 
+async def _list_genai_models_async(client: genai.Client, is_vertex: bool) -> GenaiModels:
+    """Discover models without blocking the event loop.
+
+    The Google GenAI SDK exposes model discovery through a synchronous client
+    surface. Keep the synchronous helper for callers that already run outside
+    an event loop, but move its network I/O to the default executor for
+    asynchronous plugin entry points.
+    """
+    return await asyncio.to_thread(_list_genai_models, client, is_vertex)
```
hilariie: We should use client.aio.models.list() here instead of the thread hop. Every other SDK call in this plugin goes through client.aio, and the SDK exposes model listing as an async pager, so making _list_genai_models itself async would keep discovery on the same path as generation and drop the need for a second helper. If there is a reason to_thread is the better fit here, I'm happy to hear it.

hilariie: Thanks for the switch. One follow-up: _list_genai_models_sync now only serves the _list_known_* methods, which nothing outside the tests calls. Can we drop those and the sync helper too, so discovery has one path? If they need to stay, the detailed docstring should move onto the async function, since that is the one production uses.

hilariie (approved):

PR #6301, opened 2026-09-08
fix(py/plugins/google-genai): TTS fallback ModelInfo no longer claims constrained support

### LINKED ISSUE #6183: fix(py/plugins/google-genai): TTS fallback ModelInfo declares Constrained.ALL, defeating the constrained gate
TTS model ids are absent from `SUPPORTED_MODELS`, so they resolve via `GENERIC_TTS_MODEL` (`gemini.py:720-729`), which declares `constrained=Constrained.ALL`. The model-capability gate at `gemini.py:1811-1826` therefore never blocks constrained JSON for TTS models, which do not support it.

Fix: TTS fallback ModelInfo should not claim constrained support (Go gates TTS by leaving `Constrained` empty).

hilariie:
## Summary

Closes #6183.

`GENERIC_TTS_MODEL` now declares `constrained=Constrained.NONE`, so the gate in `_genkit_to_googleai_cfg` no longer sets `response_mime_type` or `response_schema` for TTS models. The fallback also sets `system_role=False` and `output=['media']`, matching the Go and JS TTS model info.

Adds a TTS row to `test_google_model_info` and two gate tests: a TTS model gets neither field for JSON output, a standard Gemini model still gets both.

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
@@ -731,8 +731,9 @@ GENERIC_TTS_MODEL = ModelInfo(
         media=False,
         tools=False,
         tool_choice=False,
-        system_role=True,
-        constrained=Constrained.ALL,
+        system_role=False,
+        constrained=Constrained.NONE,
+        output=['media'],
     ),
 )
 
--- a/py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
+++ b/py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
@@ -65,6 +65,7 @@ from genkit import (
     TextPart,
     ToolDefinition,
 )
+from genkit._core._model import OutputConfig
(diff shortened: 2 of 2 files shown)


cabljac (approved):
Approving. The gate itself is the right fix: routing TTS through GENERIC_TTS_MODEL and letting Constrained.NONE fall through to supports_constrained is a cleaner place to put it than special-casing at the config-build site, and the test drives the real _genkit_to_googleai_cfg rather than asserting on a mock, which is what makes it worth having.

One passing note, not blocking: system_role=False and output=['media'] go a bit past what #6183 asked for. From what I can see nothing in core or the plugin currently reads supports.output or system_role, so they're declarative only and harmless, but worth confirming that's still true on the Python side before merge.

Separately, and not for this PR: js/plugins/google-genai/src/googleai/gemini.ts:446 still has GENERIC_TTS_MODEL declaring constrained: 'all', so the JS side looks to have the same bug this fixes. Worth an issue if you agree with that read.

https://github.com/genkit-ai/genkit/issues/6183

PR #6302, opened 2026-09-08
fix(py/plugins/google-genai): camelCase embedder options and request chunking

### LINKED ISSUE #6187: fix(py/plugins/google-genai): embedder options are snake_case only and requests are unbatched
Two embedder gaps vs JS/Go:

- Options are read as snake_case only (`task_type`, `output_dimensionality`; `embedder.py:400-402`, `:234`) with no camelCase aliases and no typed schema, so users following the JS docs get silent no-ops.
- No batching: one `embed_content` call carries all contents (`embedder.py:180-184`), so large batches fail rather than split (Go chunks at 250, `go/plugins/googlegenai/embedder.go:18-21`); the multimodal path rejects more than one document (`embedder.py:226-229`).

Fix: typed options schema with camelCase aliases, and request chunking.

hilariie:
Fixes #6187

## Summary

Embedder options are validated through a typed schema that accepts both snake_case and camelCase keys (`taskType`/`task_type`, `outputDimensionality`/`output_dimensionality`, `title`, `version`), tolerates unknown keys, and is advertised as `customOptions` per backend. Vertex embedders additionally accept `mimeType` and `autoTruncate`. Invalid options raise `GenkitError` with `INVALID_ARGUMENT` naming the field.

Requests are chunked and reassembled in input order: 100 contents per call on the Gemini API, 250 on Vertex text-embedding models, and 1 on Vertex `gemini-*` models, which accept a single input text per request. The multimodal path sends one `:predict` call per document instead of rejecting multi-document requests.

`EmbeddingTaskType` gains `CODE_RETRIEVAL_QUERY`.

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
@@ -16,9 +16,11 @@
 
 """Google-Genai embedder model."""
 
+import asyncio
 import json
 import sys
-from typing import Any, cast
+from collections.abc import Coroutine
+from typing import Any, Protocol, TypeVar, cast
 
 if sys.version_info < (3, 11):
     from strenum import StrEnum
@@ -126,6 +128,10 @@ class VertexEmbeddingConfigSchema(EmbeddingConfigSchema):
 GOOGLEAI_EMBED_BATCH_SIZE = 100
 VERTEXAI_EMBED_BATCH_SIZE = 250
 
+# Caps simultaneous embedding calls; large batches otherwise trip rate limits or
+# exhaust the client's connection pool.
+EMBED_CONCURRENCY_LIMIT = 10
+
 
 # Static dimensions for known embedders. Keys are version-suffix free
 # (e.g. 'multimodalembedding', not 'multimodalembedding@001') because model
@@ -200,6 +206,69 @@ def get_embedder_info(name: str, label: str, is_vertex: bool = False) -> Embedde
     )
 
 
+# One call's result: the embeddings decoded from a single API response.
+_T = TypeVar('_T')
+
+
+async def _run_bounded(calls: list[Coroutine[Any, Any, _T]]) -> list[_T]:
+    """Run the calls concurrently, cancelling the rest on the first failure.
+
+    The semaphore is built per call rather than once per module: it binds to the
+    running loop, and the Dev UI reflection server runs a second one.
+
+    Args:
+        calls: Coroutines to run, in the order their results are wanted.
+
+    Returns:
+        The results, in the order the calls were given.
+
+    Raises:
+        BaseException: The failure of the earliest failing call in that order,
+            so the reported error does not depend on completion order.
+    """
+    if not calls:
+        return []
+    semaphore = asyncio.Semaphore(EMBED_CONCURRENCY_LIMIT)
+
+    async def _bounded(call: Coroutine[Any, Any, _T]) -> _T:
+        try:
+            async with semaphore:
+                return await call
+        except asyncio.CancelledError:
+            # Cancelled while queued on the semaphore, so nothing awaited it.
+            call.close()
+            raise
+
+    tasks = [asyncio.create_task(_bounded(call)) for call in calls]
+    _, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)
+    for task in pending:
+        # Without this the rest of the batch still bills one call each.
+        task.cancel()
+    # Also retrieves the cancelled outcomes, so none resurfaces later as
+    # "Task exception was never retrieved".
+    results = await asyncio.gather(*tasks, return_exceptions=True)
+    for result in results:
+        # Call order, not completion order, so the reported failure is
+        # deterministic across the calls that were left to run.
+        if isinstance(result, BaseException) and not isinstance(result, asyncio.CancelledError):
+            raise result
+    return cast(list[_T], results)
+
+
+class _HttpResponse(Protocol):
+    """The google-genai HTTP response surface used to read a JSON body."""
+
+    body: str | bytes | None
+
+
+class _AsyncRequester(Protocol):
+    """The google-genai transport surface used to POST to ``:predict``."""
+
+    async def async_request(self, *, http_method: str, path: str, request_dict: dict[str, Any]) -> _HttpResponse:
+        """Send an authenticated request and return the HTTP response."""
+        ...
+
+
 class Embedder:
     """Embedder for Google-Genai."""
 
@@ -225,8 +294,8 @@ class Embedder:
         """Generate embeddings for a given request.
 
         Requests with more documents than the endpoint accepts per call are
-        split into sequential batches; the response carries one embedding per
-        document, in input order.
+        split into batches, sent with bounded concurrency; the response carries
+        one embedding per document, in input order.
 
         Args:
             request: Genkit embed request.
@@ -253,25 +322,49 @@ class Embedder:
         config = self._genkit_to_googleai_cfg(options)
         batch_size = self._embed_batch_size(model)
 
+        batches = [contents[start : start + batch_size] for start in range(0, len(contents), batch_size)]
+        results = await _run_bounded([self._embed_batch(model, batch, config) for batch in batches])
+
         embeddings: list[Embedding] = []
-        for start in range(0, len(contents), batch_size):
-            batch = contents[start : start + batch_size]
-            response = await self._client.aio.models.embed_content(
-                model=model,
-                contents=cast(genai_types.ContentListUnion, batch),
-                config=config,
-            )
-            returned = response.embeddings or []
-            if len(returned) != len(batch):
-                raise GenkitError(
-                    status='INTERNAL',
-                    message=(
-                        f'{model} returned {len(returned)} embeddings for {len(batch)} documents; '
-                        'embeddings cannot be aligned with the input.'
-                    ),
-                )
-            embeddings.extend(Embedding(embedding=em.values or []) for em in returned)
+        for batch_embeddings in results:
+            embeddings.extend(batch_embeddings)
         return EmbedResponse(embeddings=embeddings)
+
+    async def _embed_batch(
+        self,
+        model: str,
+        batch: list[genai.types.Content],
+        config: genai.types.EmbedContentConfig | None,
+    ) -> list[Embedding]:
+        """Embed one batch of contents in a single ``embed_content`` call.
+
+        Args:
+            model: API model id.
+            batch: Contents to send in this call.
+            config: Google-genai embed config, or None.
+
+        Returns:
+            One embedding per content in the batch, in batch order.
+
+        Raises:
+            GenkitError: INTERNAL when the endpoint returns a different number
+                of embeddings than documents were sent.
+        """
+        response = await self._client.aio.models.embed_content(
+            model=model,
+            contents=cast(genai_types.ContentListUnion, batch),
+            config=config,
+        )
+        returned = response.embeddings or []
+        if len(returned) != len(batch):
+            raise GenkitError(
+                status='INTERNAL',
+                message=(
+                    f'{model} returned {len(returned)} embeddings for {len(batch)} documents; '
+                    'embeddings cannot be aligned with the input.'
+                ),
+            )
+        return [Embedding(embedding=em.values or []) for em in returned]
 
     def _parse_options(self, options: object) -> EmbeddingConfigSchema:
         """Validate raw request options against this backend's option schema.
@@ -333,9 +426,9 @@ class Embedder:
         """Embed text/image/video via the Vertex multimodal ``:predict`` endpoint.
 
         ``multimodalembedding@001`` accepts one instance per ``:predict`` call,
-        so every document is sent as its own request and the resulting
-        embeddings are concatenated in document order. All documents are
-        validated before the first request is made.
+        so every document is sent as its own request, with bounded concurrency,
+        and the resulting embeddings are concatenated in document order. All
+        documents are validated before the first request is made.
 
         Args:
             request: Genkit embed request.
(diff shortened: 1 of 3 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -133,9 +202,73 @@ def get_embedder_info(name: str, label: str, is_vertex: bool = False) -> Embedde
...
+        try:
+            async with semaphore:
+                return await call
+        except asyncio.CancelledError:
+            # Cancelled while queued on the semaphore, so nothing awaited it.
+            call.close()
+            raise
+
+    tasks = [asyncio.create_task(_bounded(call)) for call in calls]
+    _, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_EXCEPTION)
```
cabljac: asyncio.wait won't cancel these if the outer task is cancelled. Cancelling `tasks` in a BaseException handler and gathering with return_exceptions before re-raising should cover it.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -133,9 +202,73 @@ def get_embedder_info(name: str, label: str, is_vertex: bool = False) -> Embedde
...
+    The semaphore is built per call rather than once per module: it binds to the
+    running loop, and the Dev UI reflection server runs a second one.
+
+    Args:
+        calls: Coroutines to run, in the order their results are wanted.
+
+    Returns:
+        The results, in the order the calls were given.
+
+    Raises:
```
cabljac, line 233: nit: docstring says "the earliest failing call in that order", but a call that fails later can get cancelled as pending, so a later-ordered failure gets reported instead. The inline comment at :251 has it right.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -65,10 +71,68 @@ class EmbeddingTaskType(StrEnum):
...
 
 
+class EmbeddingConfigSchema(BaseModel):
+    """Options accepted by Google GenAI embedders on both backends.
+
+    Keys may be written in snake_case or camelCase (``task_type`` or
+    ``taskType``). Unknown keys are kept and ignored.
+    """
+
+    model_config = ConfigDict(extra='allow', populate_by_name=True, alias_generator=to_camel)
```
cabljac, line 89: nit: when both taskType and task_type are present the camelCase one wins and the snake_case value stays in model_extra. Fine as behaviour, but worth a test pinning the precedence so it doesn't drift.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -385,21 +625,30 @@ async def _build_contents(self, request: EmbedRequest) -> list[genai.types.Conte
...
-
-        return cfg
+        fields: dict[str, Any] = {}
+        if options.task_type is not None:
+            fields['task_type'] = options.task_type.value
+        if options.title is not None:
+            fields['title'] = options.title
+        if options.output_dimensionality is not None:
+            fields['output_dimensionality'] = options.output_dimensionality
+        if isinstance(options, VertexEmbeddingConfigSchema):
```
cabljac, line 688: nit: the docstring at :631 says the Gemini API rejects mimeType and autoTruncate, but the code strips them before the API ever sees them, and the advertised schema has additionalProperties true, so the user gets no signal at all.

cabljac (changes requested):
The :predict request_dict assertions and the reverse-completion ordering test are the right shape. One fix, rest inline.

_run_bounded doesn't cancel its children when the outer generate() is cancelled. asyncio.wait doesn't propagate, so on abort or timeout the in-flight calls keep running and billing. The FIRST_EXCEPTION path is fine, it's just the outer-cancel case. BaseException handler, cancel pending, gather with return_exceptions, re-raise. Floor is 3.10 so no TaskGroup. Or drop the concurrency, which #6187 doesn't ask for anyway, and it goes away.

One that can't anchor inline: google.py:533 still types config as dict[str, object] while this PR exports the two schemas, so passing a typed instance is a pyright error even though _parse_options handles it. The dict typing predates this (#6104), the mismatch doesn't.

Also, not this PR: Go batches gemini-embedding-001 at 250 (embedder.go:35) against the Vertex doc's "each request can only include a single input text". Your 1 is right, Go's wrong.

https://github.com/genkit-ai/genkit/issues/6187

py/packages/genkit-google-genai/test/models/googlegenai_embedder_test.py
```diff
@@ -87,6 +99,310 @@ async def test_options_version_is_the_api_model(mocker: MockerFixture) -> None:
...
+) -> genai.types.EmbedContentResponse:
+    """Return one embedding per content whose value is the text of that content."""
+    values: list[float] = []
+    for content in contents:
+        part = (content.parts or [])[0]
+        values.append(float(part.text or 'nan'))
+    return genai.types.EmbedContentResponse(embeddings=[genai.types.ContentEmbedding(values=[v]) for v in values])
+
+
+def _numbered_docs(count: int) -> list[Document]:
```
cabljac: nit: `_numbered_docs` and `_numbered_media_docs` return `list[Document]`, `EmbedRequest.input` is `list[DocumentData]`, and list is invariant, so pyright flags every call site (13 in this file). `-> list[DocumentData]` on both helpers clears it. CI only runs pyright on `src`, but `just lint` covers `packages/` and fails.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -160,60 +310,170 @@ def __init__(
...
+
+        Args:
+            options: Validated embedding options.
+        """
+        if self._is_vertex:
+            return
+        extra = options.model_extra or {}
+        dropped = [name for name in VERTEX_ONLY_OPTION_NAMES if name in extra]
+        if dropped:
+            logger.warning(
```
cabljac, line 428: nit: this fires on every `generate()`. An indexing loop with `mimeType` in the ref config emits one warning per call. Warn-once per embedder instance would be quieter. Fine to leave if you prefer.

cabljac (changes requested):
The cancellation fix and its test are right, and the dropped-option warning is what I was after. One item from the last round is still open, it wasn't inline-anchored so easy to miss. Two nits inline.

google.py:533 still types `config` as `dict[str, object] | None`. Passing an `EmbeddingConfigSchema` instance is a pyright error even though `_parse_options` accepts it. `EmbedderRef.config` is `Any`, so widening to `EmbeddingConfigSchema | dict[str, object] | None` is enough. Same type on `family_embedder_ref` in `_model_refs.py:110`.

https://github.com/genkit-ai/genkit/issues/6187

py/packages/genkit-google-genai/test/models/googlegenai_embedder_test.py
hilariie: Done in `d6d7e7ce8`

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
hilariie: Leaving this one. _create_embedder_action builds a fresh Embedder inside _run on every request, so a per-instance flag would still warn every call. Happy to add a process-wide once if you would rather have it.

hilariie:
Sorry, missed that one last round. Done in `429d8781b`: embedding() and family_embedder_ref take EmbeddingConfigSchema | dict[str, object] | None, with a test passing an instance through.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -65,10 +74,73 @@ class EmbeddingTaskType(StrEnum):
...
+# Per-request input limits of the embedding endpoints. The Gemini API's
+# batchEmbedContents endpoint accepts up to 100 contents per call; the Vertex AI
+# :predict endpoint accepts up to 250 input texts per call for the
+# text-embedding models and a single input text for gemini-* models.
+GOOGLEAI_EMBED_BATCH_SIZE = 100
+VERTEXAI_EMBED_BATCH_SIZE = 250
+
+# Caps simultaneous embedding calls; large batches otherwise trip rate limits or
+# exhaust the client's connection pool.
+EMBED_CONCURRENCY_LIMIT = 10
```
cabljac, line 137: A single embed call now fans out to up to 10 concurrent requests. Is 10 the right ceiling for the Gemini API rate limits? Worth naming the number in the `generate()` docstring either way.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -160,60 +310,170 @@ def __init__(
...
+    def _embed_batch_size(self, model: str) -> int:
+        """Maximum number of documents sent in one ``embed_content`` call.
+
+        Vertex AI serves gemini-* and MaaS embedding models one input per
+        request; other Vertex models accept 250 and the Gemini API 100.
+        """
+        if not self._is_vertex:
+            return GOOGLEAI_EMBED_BATCH_SIZE
+        lowered = model.lower()
+        if 'gemini' in lowered or 'maas' in lowered:
```
cabljac: The `'maas'` branch is copied from Go, but from what I can see nothing on the Python side registers a MaaS embedder and no test covers it. Drop it, or add a test if you know of a path that reaches it.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
```diff
@@ -133,9 +210,82 @@ def get_embedder_info(name: str, label: str, is_vertex: bool = False) -> Embedde
...
+    for task in pending:
+        # Without this the rest of the batch still bills one call each.
+        task.cancel()
+    # Also retrieves the cancelled outcomes, so none resurfaces later as
+    # "Task exception was never retrieved".
+    results = await asyncio.gather(*tasks, return_exceptions=True)
+    for result in results:
+        # Call order, not completion order, so the reported failure is
+        # deterministic across the calls that were left to run.
+        if isinstance(result, BaseException) and not isinstance(result, asyncio.CancelledError):
```
cabljac, line 272: nit: if a single call raises `CancelledError` on its own (no outer cancel), this filter skips it and the exception object ends up in `results`, so `embeddings.extend` at :346 gets a `CancelledError` and raises `TypeError`. Rare, but the skip branch should re-raise when no other exception was found.

py/packages/genkit-google-genai/test/models/googlegenai_embedder_test.py
```diff
@@ -87,6 +100,310 @@ async def test_options_version_is_the_api_model(mocker: MockerFixture) -> None:
...
+    count = GOOGLEAI_EMBED_BATCH_SIZE + 50
+    client = mocker.AsyncMock()
+    client.aio.models.embed_content.side_effect = _indexed_embed_content
+    embedder = Embedder(GeminiEmbeddingModels.GEMINI_EMBEDDING_001, client)
+
+    response = await embedder.generate(EmbedRequest(input=_numbered_docs(count)))
+
+    calls = client.aio.models.embed_content.call_args_list
+    assert [len(call.kwargs['contents']) for call in calls] == [GOOGLEAI_EMBED_BATCH_SIZE, 50]
+    assert [e.embedding for e in response.embeddings] == [[float(i)] for i in range(count)]
```
cabljac, line 347: nit: batch tests assert per-call lengths and ordered outputs. One direct assertion on chunk contents would make composition explicit.

cabljac (changes requested):
The cancellation handling and the reverse-completion ordering tests hold up well, and every point from the last round is either done or declined for a sound reason. One thing to dig into, which is on me, then a couple of small ones.

The typed config widening I asked for type-checks but crashes on first use: `Ai.embed()` does `{**embedder.config}` and a `BaseModel` is not a mapping. With a fake key, `GoogleAI.embedding('gemini-embedding-001', config=EmbeddingConfigSchema(task_type=...))` raises `TypeError: 'EmbeddingConfigSchema' object is not a mapping` before any network call (with `version=` set it fails one line earlier on item assignment). Two options: normalise in core `embed()` with `model_dump(exclude_none=True)`, or dump to a dict inside `family_embedder_ref`. Which would you prefer? Either way it needs a test that goes through `ai.embed`, not just `ref.config is config`.

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
hilariie: Dropped in `e54a4519c`

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
hilariie: Resolved in `9fe4340c0`

py/packages/genkit-google-genai/test/models/googlegenai_embedder_test.py
hilariie: Added in `ef594c111`

py/packages/genkit-google-genai/src/genkit_google_genai/models/embedder.py
hilariie: Named in the `generate()` and `_generate_multimodal` docstrings in `2d33cb790`. Both quota pages express limits as per-minute requests or tokens and list concurrency quotas only for batch jobs, so 10 is a burst and connection-pool cap rather than a match to any quota as JS sends one call per document uncapped while Go is sequential. Happy to lower it or expose it as an option if you would rather.

hilariie:
Went with the plugin-local dump in `dac44f45d`, with a test through ai.embed() asserting task type, dimensionality and version reach the SDK call. Snake_case keys keep per-call options winning in either spelling, since camelCase only wins when both are present.

Issue #5898, opened 2026-08-03
[plugins] OpenAI chat model table sync across JS, Go, Python


cabljac:
The three OpenAI compat plugins have drifted on chat model coverage. No SDK has the full set; each is missing models another one ships.

- [x] Go (go/plugins/compat_oai/openai/openai.go): done in #5976 - table now carries the gpt-5 generation through gpt-5.6, o3, o4-mini. o3-pro deliberately excluded (Responses-API-only, documented in the file), gpt-4.5 retired by OpenAI. Last nit: gpt-5-chat-latest is not in the curated table - it resolves dynamically but misses the tools-disabled/text-only pinning JS and Py carry.
- [ ] Python (py/packages/genkit-openai/src/genkit_openai/models/model_info.py): gpt-5.4, gpt-5.4-mini, gpt-5.4-nano, gpt-5.5. Stops at 5.3-codex.
- [ ] JS (js/plugins/compat-oai/src/openai/gpt.ts): o3-pro, gpt-oss-120b, gpt-oss-20b, gpt-5.1-codex, gpt-5.1-codex-max, gpt-5.2-chat, gpt-5.2-pro, gpt-5.3-codex, gpt-4.1-nano.

JS gpt-5.x additions landed in #5389; use those capability groupings as the reference. Note gpt-5-chat-latest has tools disabled and text-only output in JS and Py; Go should carry that over.

Part of #5903.

cabljac:
Adding `gpt-6-astra` in three drafts, one per SDK: #6303 (JS), #6304 (Go), #6306 (Python). Recording what a live run against `/v1/chat/completions` showed, since it shapes the catalog entry and will matter for anyone syncing the tables:

- Function tools are rejected at every `reasoning_effort`, and `reasoning_effort: 'none'` is itself rejected, so no value lets tools through on this endpoint. The 400 points at `/v1/responses`. All three entries declare no tool support.
- `reasoning_effort` accepts `low`, `medium`, `high`, `xhigh` only; `max` and `none` 400.
- `temperature` other than 1 and `top_p` 400.
- Text, streaming, `json_schema` output without `strict`, image input, system role and multiturn work. Strict structured output is unverified (Go sends `strict: true`); see #6307.

Catalog drift as of today: Go stops at `gpt-5.6-*`, JS at `gpt-5.5`, Python at `gpt-5.2`. The astra PRs do not touch the 5.6 gap.

PR #6304, opened 2026-09-08
feat(go/plugins/compat_oai): add gpt-6-astra to the OpenAI model catalog

cabljac:
`gpt-6-astra` was only reachable through the dynamic resolution path in the Go OpenAI plugin, which describes it with the generic multimodal defaults. Those claim tool support the model does not have on this endpoint, so the Dev UI and `supports` metadata were wrong for it. This is the Go counterpart of #6303; Python is #6306.

The entry takes a new `multimodalNoTools` supports set: multiturn, media and system role stay on, `Tools` and `ToolChoice` are false. The live run in #6303 against `/v1/chat/completions` with a single function tool returned:

```
400 Function tools with reasoning_effort are not supported for gpt-6-astra in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.
```

The same error came back at `low` and `xhigh`, and `reasoning_effort: 'none'` was itself rejected, so no value lets tools through on this endpoint. The model page lists function calling as supported without an endpoint caveat, so this may lift; the Responses API for Go is open in #5346. Unlike JS, Go core enforces `supports`: a request with tools now fails early with `ErrUnsupportedByModel` rather than reaching OpenAI. The README catalog sentence names the model and the constraint.

`TestGPT6AstraCatalogEntry` pins the label, the single version, the exact `ai.ModelSupports` and that `Init` registers the model without tools advertised. `TestGPT6AstraRejectsToolsBeforeSend` pins the early failure offline through `genkit.Generate` with a tool. The existing `TestSupportedModelCatalog` and `TestConstrainedSupport` table checks cover the entry too. In `go/`: `gofmt -l` clean, `go vet ./plugins/compat_oai/...` clean, `go test -count=1 ./plugins/compat_oai/...` passes all 9 packages.

Caveats: no Go live smoke was run. `Constrained` stays at the catalog default of `All`, verified live from Go: `WithOutputType` sent `json_schema` with `strict: true` and astra returned schema-conformant JSON that parsed. A Go live run also confirmed text, streaming, image input, system role, multiturn, `max_completion_tokens`, and that tools fail early with `ErrUnsupportedByModel`; the existing `TestPluginLive` passes 16/16 with a key. Go passes `reasoning_effort` through verbatim, so `max` and `none` will 400 on this endpoint, and `xhigh` needs a raw string since openai-go only defines low, medium and high constants. The model also rejects `temperature` other than 1 and `top_p`. Related: #5898.

### DIFF
--- a/go/plugins/compat_oai/openai/openai_test.go
+++ b/go/plugins/compat_oai/openai/openai_test.go
@@ -434,6 +434,9 @@ func TestGPT6AstraCatalogEntry(t *testing.T) {
 		Output:      []string{"text", "json"},
 		Constrained: ai.ConstrainedSupportAll,
 	}
+	if opts.Supports == nil {
+		t.Fatal("gpt-6-astra has no Supports")
+	}
 	if diff := cmp.Diff(want, *opts.Supports); diff != "" {
 		t.Errorf("supports mismatch (-want +got):\n%s", diff)
 	}
--- a/go/plugins/compat_oai/openai/openai_test.go
+++ b/go/plugins/compat_oai/openai/openai_test.go
@@ -18,6 +18,7 @@ package openai
 
 import (
 	"context"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -27,6 +28,7 @@ import (
 	"github.com/firebase/genkit/go/ai"
 	"github.com/firebase/genkit/go/core/api"
 	"github.com/firebase/genkit/go/genkit"
+	"github.com/google/go-cmp/cmp"
 	openaiGo "github.com/openai/openai-go"
 	"github.com/openai/openai-go/option"
 	"github.com/openai/openai-go/shared"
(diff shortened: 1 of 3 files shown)


go/plugins/compat_oai/openai/openai_test.go
```diff
@@ -407,6 +409,74 @@ func TestSupportedModelCatalog(t *testing.T) {
...
+	want := ai.ModelSupports{
+		Multiturn:   true,
+		Tools:       false,
+		SystemRole:  true,
+		Media:       true,
+		ToolChoice:  false,
+		Output:      []string{"text", "json"},
+		Constrained: ai.ConstrainedSupportAll,
+	}
+	if diff := cmp.Diff(want, *opts.Supports); diff != "" {
```
adesinah, line 440: let's  add `if opts.Supports == nil { t.Fatal(...) } ` before the dereferencing.

adesinah (changes requested):
Looks pretty good - I added a comment

go/plugins/compat_oai/openai/openai_test.go
cabljac: Added in a01eafcd6: the test now fails fast with `gpt-6-astra has no Supports` instead of panicking on a nil dereference.

adesinah (approved):
Looks good to me!

Issue #6185, opened 2026-08-26
chore(py/plugins/google-genai): SpeechConfigSchema drops fields, ImagenConfigSchema is empty, Lyria is dead code but exported


cabljac:
Three dead or silently lossy config surfaces:

- `SpeechConfigSchema` (`gemini.py:450-459`) has no `model_config`, so pydantic's default `extra='ignore'` silently drops `multiSpeakerVoiceConfig` and `languageCode`, both supported by JS (`js/plugins/google-genai/src/googleai/gemini.ts:250,319`).
- `ImagenConfigSchema` is an empty `extra='allow'` class (`models/imagen.py:151-154`), typing nothing.
- Lyria is dead code but publicly exported: `'lyria'` is in `UNROUTABLE_FAMILIES` (`_routing.py:57`) yet `LyriaConfig`/`LyriaVersion` are exported from `__init__.py:97,126-127`.

Fix: type the speech config fully, populate or remove `ImagenConfigSchema`, and drop the unreachable Lyria exports (or route Lyria). Public surface should be frozen before v1.

PR #6305, opened 2026-09-08
chore(py/plugins/google-genai): retire the Imagen catalog

### LINKED ISSUE #6176: chore(py/plugins/google-genai): retire the Imagen catalog (Imagen is shut down)
This issue originally asked to move the curated Imagen list from Imagen 3 to Imagen 4. That is no longer the right fix, because Imagen 4 is shut down too.

- Vertex AI discontinued every Imagen 3 and Imagen 4 endpoint on June 30, 2026 and points users to `gemini-2.5-flash-image`: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/image/generate-images
- The Gemini API shut down `imagen-4.0-generate-001`, `imagen-4.0-fast-generate-001` and `imagen-4.0-ultra-generate-001` on August 17, 2026 and points users to `gemini-3.1-flash-image`: https://ai.google.dev/gemini-api/docs/deprecations

The Python plugin still curates `imagen-3.0-generate-002` and `imagen-3.0-fast-generate-001` in `ImagenVersion`, `KnownImagen` and `SUPPORTED_MODELS`, and ships the `vertexai-imagen` sample. All of it targets endpoints that no longer exist. The replacement models are already curated and available through `gemini_image_model`.

Notes:
- `ImagenVersion`, `KnownImagen` and `ImagenConfigSchema` are public exports, so this is a breaking change.
- Overlaps #6185, which asks to populate or remove `ImagenConfigSchema`.
- JS and Go still list the same Imagen 4 ids and need the same cleanup.

hilariie:
Closes #6176

## Summary

Removes Imagen from the Python google-genai plugin. `models/imagen.py` is deleted along with `ImagenVersion`, `KnownImagen`, `ImagenConfigSchema`, `ImagenModel`, and the `imagen_model` constructors on `GoogleAI` and `VertexAI`. This is a breaking change to the package's public exports.

An `imagen-` id now fails closed on both plugins: `ai.generate` reports the model as not found instead of falling through to a Gemini text model, and every typed ref constructor rejects it with a hint pointing at `gemini_image_model`. Model discovery never registers `imagen-*` entries the API may still list.

The `vertexai-imagen` sample and its workspace entry are removed. The image flow in `google-genai-media` now targets `googleai/gemini-2.5-flash-image`. Removing the workspace member invalidated `uv.lock`, so `uv lock` re-emitted a few redundant environment markers in their current form; no dependency versions changed.

Also resolves the `ImagenConfigSchema` bullet of #6185. JS and Go still list the Imagen 4 ids and are out of scope here.

The new fail-closed and discovery tests were run against the parent commit's source and fail there.

Also, the Vertexai sample folder is still named `vertexai-imagen`. Renaming it would touch the root workspace config, the samples index, and possibly CI matrices, which is wider than this change, so I left it (can be a follow-up issue).


py/packages/genkit-google-genai/src/genkit_google_genai/models/_model_refs.py
```diff
@@ -66,6 +65,8 @@ def wrong_family_error(*, plugin_class: str, method: str, family: str, local: st
         )
     elif actual == 'unsupported':
         hint = f"'{local}' is not a supported model."
+        if local.lower().startswith('imagen-'):
+            hint = f"'{local}' is not a supported model; for image generation use {plugin_class}.gemini_image_model()."
```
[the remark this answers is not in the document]
hilariie: resolved in `db4f1c7`

py/packages/genkit-google-genai/src/genkit_google_genai/models/_model_refs.py
```diff
@@ -65,7 +64,10 @@ def wrong_family_error(*, plugin_class: str, method: str, family: str, local: st
             else f"'{local}' has no ref constructor in this plugin."
         )
     elif actual == 'unsupported':
-        hint = f"'{local}' is not a supported model."
+        if local.lower().startswith('imagen-'):
```
cabljac: nit: this re-derives the `imagen-` prefix. A predicate next to `is_unsupported_image_model_name` in `_routing.py` would keep it in one place.

cabljac (changes requested):
The fail-closed routing holds up well: `imagen-*` is classified unsupported before the Gemini fallback in both plugins, and the typed constructors raising with the `gemini_image_model()` hint is a nice touch. One thing worth digging into, then a small one.

This conflicts with main now (five files, including modify/delete on `imagen.py` and its test, which main touched after this branch deleted them). #6298 also made `_list_genai_models` async, so when rebasing, `test_list_genai_models_googleai_skips_imagen` needs to `await` it and mock `client.aio.models.list`, and `test_list_actions_never_advertise_imagen` needs its mock moved to `mock_client.return_value.aio.models.list`. Main already converted the vertex substring test, so that one should just take main's side. Can you rebase and re-run those?

#6176 also lists the `vertexai-imagen` sample. From what I can see the only references are `py/pyproject.toml:160`, `uv.lock`, and the folder itself, so the rename looks like a three-file change. Would you rather fold it in here, or file a follow-up issue and link it?

hilariie:
I've rebased onto main. The two discovery tests now mock client.aio.models.list, and the vertex substring test took main's side. Also folded the sample rename in since you're happy to review it all in this PR.

py/packages/genkit-google-genai/src/genkit_google_genai/models/_model_refs.py
hilariie: Resolved in `6ec209c23`

PR #6306, opened 2026-09-08
feat(py/plugins/openai): add gpt-6-astra to the OpenAI model catalog

cabljac:
`gpt-6-astra` only reaches the Python OpenAI plugin through dynamic resolution, which registers it with `Supports(multiturn=True)`, so the Dev UI hides media, system role and JSON output. This adds a catalog entry, the Python counterpart of #6303 (JS) and #6304 (Go).

The entry declares `multiturn`, `media`, `system_role` and `json_mode`/`text`, with `tools=False`. The live run in #6303 against `/v1/chat/completions` returned:

```
400 Function tools with reasoning_effort are not supported for gpt-6-astra in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.
```

`none` is itself rejected, so no value lets tools through. Python core does not enforce `supports.tools`, so the flag is advisory. Listing `json_mode` is wire-neutral: unlisted ids already send `json_object` and strict `json_schema`. Both verified live on astra, along with text, streaming, image input, system role and multiturn.

Tests pin the entry, registration via `init()` and `resolve()`, `json_object` selection, and the unlisted default. `pytest` 191 passed; ruff, pyright and ty clean.

Live smoke run from Python. Astra rejects `max_tokens` (`Use 'max_completion_tokens' instead`), which `OpenAIConfig(max_tokens=...)` still sends; pre-existing for every reasoning model, fix pending in a separate change since `model.py` is in flight. Related: #5898.

### DIFF
--- a/py/packages/genkit-openai/tests/openai_plugin_test.py
+++ b/py/packages/genkit-openai/tests/openai_plugin_test.py
@@ -24,14 +24,12 @@ from typing import Any, cast
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
-from genkit_openai.models import OpenAIModel
 from genkit_openai.models.model_info import SUPPORTED_OPENAI_MODELS
 from genkit_openai.openai_plugin import OpenAI, openai_model
 from genkit_openai.typing import SupportedOutputFormat
 from openai.types import Model
 
-from genkit import Message, ModelRequest, Part, Role, Supports, TextPart
-from genkit._core._model import OutputConfig
+from genkit import Supports
 from genkit.plugin_api import ActionKind, ActionMetadata, loop_local_client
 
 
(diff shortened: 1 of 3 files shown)


py/packages/genkit-openai/tests/openai_plugin_test.py
```diff
@@ -77,6 +82,69 @@ async def test_openai_plugin_resolve_action(kind: ActionKind, name: str) -> None
...
+
+    action = await plugin.resolve(ActionKind.MODEL, 'openai/gpt-6-nova')
+
+    assert action is not None
+    assert action.metadata is not None
+    model_meta = cast(dict[str, Any], action.metadata['model'])
+    assert model_meta['supports'] == {'multiturn': True}
+
+
+def test_gpt_6_astra_json_mode_uses_json_object() -> None:
```
hilariie: this test builds an OpenAIModel rather than the plugin, and openai_model_test.py already has test_unknown_chat_id_json_mode_uses_json_object right next to where it would live. Can we move it there?

hilariie (approved):

PR #6309, opened 2026-09-08
feat(py): drop Part(root=); one Part through generate

huangjeff5:
Callers already build with `Part.from_text` / `Part(text=...)`. This change deletes `.root`. `Part(root=...)` raises. The part they put on `Message` / `Document` / `ai.generate` is the same `Part` they read `.text` from.

```python
# before
msg = Message(role='user', content=[Part(root=TextPart(text='hi'))])
print(msg.content[0].root.text)

# after
msg = Message(role='user', content=[Part.from_text('hi')])
# or Part(text='hi')
print(msg.content[0].text)
```

`Part.from_data(None)` is a data part. Empty `{}` or metadata-only parts in inbound message content are dropped when the message is read; `Part()` with no kind still raises.

```python
Part.from_data(None)  # data part, same as 0 or []

# inbound junk is skipped; the rest of the message loads
Message(role='model', content=[{}, {'text': 'hello'}]).text  # 'hello'
```

## Decisions

- Rewrite construction sites to `Part.from_*` or `Part(text=...)`. `Part(root=...)` raises.
- Kind classes stay off `from genkit import`. Inner objects (`ToolRequest`, `ToolResponse`, `Media`) stay public.
- Two exclusive kinds still raise. `custom` may ride on another kind, or be the kind when it is the only payload.
- JSON null is a real data payload. `Part.from_data(None)` dumps and reloads as a data part.
- An inbound `{}` or metadata-only part is dropped when reading a message so a junk empty part does not fail generate. Building `Part()` with no kind still raises.
- A thought whose summary includes an image replays as that same thought step.
- A typed plugin config on `ModelRequest` stays that class on the response. `/generate` still requires `messages`.
- A bad `pendingContent` part reports the kind rule (`exactly one of text, media, ...`) instead of a generic "must be a part".


py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -248,9 +241,11 @@ class Part(PartData):
     @model_validator(mode='before')
     @classmethod
     def _exactly_one_kind(cls, value: object) -> object:
-        if isinstance(value, PartData):
-            return value.root
-        if not isinstance(value, dict) or 'root' in value:
+        if isinstance(value, Part):
+            return value
+        if isinstance(value, BaseModel):
+            value = value.model_dump(fallback=lambda x: x)
```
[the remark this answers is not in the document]
huangjeff5: Not applying. Pydantic 2.11+ added `fallback` on `BaseModel.model_dump`. We keep it so a kind model that is not already a dict still becomes a mapping for the one-kind check.

py/packages/genkit-amazon-bedrock/src/genkit_amazon_bedrock/embedders.py
```diff
@@ -233,10 +231,10 @@ def image_from_document(document: DocumentData) -> tuple[str, str]:
         GenkitError: INVALID_ARGUMENT when an image part holds a remote URL.
     """
     for part in document.content:
-        if not isinstance(part.root, MediaPart):
+        if not part.media is not None:
```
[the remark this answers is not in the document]
huangjeff5: Agreed; equivalent and clearer. Applying.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -272,7 +263,7 @@ def split_system_instruction(messages: list[Message]) -> tuple[str | None, list[
         if message.role != 'system':
             turns.append(message)
             continue
-        if any(not isinstance(part.root, TextPart) for part in message.content):
+        if any(not part.text is not None for part in message.content):
```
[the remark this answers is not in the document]
huangjeff5: Agreed. Applying as a drive-by on the follow-up commit.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -169,33 +163,31 @@ def ensure_tool_ids(messages: list[Message]) -> list[Message]:
...
+                request_ids.append(part.tool_request.ref)
 
     # A response that already named its call isn't in the pairing pool.
     claimed = {
-        part.root.tool_response.ref
+        part.tool_response.ref
         for message in new_messages
         for part in message.content
-        if isinstance(part.root, ToolResponsePart) and part.root.tool_response and part.root.tool_response.ref
+        if part.tool_response is not None and part.tool_response and part.tool_response.ref
```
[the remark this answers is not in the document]
huangjeff5: Agreed. Applying as a drive-by on the follow-up commit.

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
```diff
@@ -610,55 +594,40 @@ def from_function_result_step(step: FunctionResultStep) -> Part:
...
     summary = step.summary or []
     reasoning = '\n'.join(item.text or '' if isinstance(item, TextContent) else '[Image]' for item in summary)
-    return Part(
-        ReasoningPart(
-            reasoning=reasoning,
-            metadata={THOUGHT_SIGNATURE: step.signature},
-            custom={THOUGHT_CUSTOM: step.model_dump(mode='python')},
-        )
-    )
+    return Part.from_reasoning(reasoning, metadata={THOUGHT_SIGNATURE: step.signature})
```
apascal07, line 630: `from_thought_step` no longer keeps `custom={THOUGHT_CUSTOM: step.model_dump(mode='python')}`.

That breaks verbatim replay. `to_thought_step` (line 379) starts with `thought_from_custom(part.custom)`, and `to_standalone_step` (line 452) reads the same key. After this change nothing in the package writes `THOUGHT_CUSTOM`, so both reads are dead.

Example: a reply holds a `ThoughtStep` whose summary has an image item. The part now keeps only `"see this\n[Image]"` and the signature. On the next turn the converter sends a rebuilt thought step, and the image summary item is gone.

`Part` now permits `custom` next to `reasoning`, so the payload can still ride along:

```suggestion
    return Part(
        reasoning=reasoning,
        metadata={THOUGHT_SIGNATURE: step.signature},
        custom={THOUGHT_CUSTOM: step.model_dump(mode='python')},
    )
```

Note that `test_signed_multimodal_thought_replays_original_step` was changed to build the custom part by hand, and `test_thought` dropped its `'custom'` assertion, so no test covers this now.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -789,6 +1031,27 @@ class ModelResponse(GenkitModel, Generic[OutputT]):
...
+    @field_validator('message', mode='before')
+    @classmethod
+    def _wrap_message(cls, v: object) -> object:
+        if v is None:
+            return v
+        return as_message(v)
+
+    @field_validator('request', mode='before')
+    @classmethod
+    def _wrap_request(cls, v: object) -> object:
```
apascal07, line 1043: `as_model_request` dumps and revalidates on the unparameterized class, so a typed plugin config is downgraded to a plain dict.

```python
req = ModelRequest[MyCfg](messages=[...], config=MyCfg(my_knob='x'))
ModelResponse(request=req).request.config   # {'myKnob': 'x'}, not MyCfg
```

Before this change pydantic accepted the subclass instance unchanged (`revalidate_instances='never'`), so `response.request.config` stayed a `MyCfg`. A `wrap_model` middleware that reads `response.request.config.my_knob` now gets an `AttributeError`, and `config['my_knob']` also fails because the key is camelCased. `_generate.py:1025` reassigns the original request straight after, so only the middleware window is affected, but that window is real.

Suggest keeping an instance as it is, the same way `as_output_config` does:

```python
def as_model_request(value: object) -> ModelRequest:
    if isinstance(value, ModelRequest):
        return value
    return ModelRequest.model_validate(value)
```

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -479,75 +508,98 @@ def _wrap_parts(cls, v: object) -> object:
...
+        )
+    return Candidate.model_validate(value)
+
+
+class GenerateActionOptions(GenkitModel):
     """Generate options with messages as list[Message] for type-safe use with ai.generate()."""
 
-    messages: list[Message]
+    model: str | None = None
+    messages: list[Message] = Field(default_factory=list)
```
apascal07, line 560: `messages` was required on the old veneer (`messages: list[Message]`). It is optional now.

A `/generate` payload with no messages used to fail with `messages: Field required`. It now builds with `messages=[]` and reaches the model with an empty conversation. Every other veneer field matches its generated twin, so this one looks unintended.

```suggestion
    messages: list[Message]
```

py/packages/genkit/src/genkit/_ai/_tools.py
```diff
@@ -124,22 +104,16 @@ def normalize_pending_content(pending_content: object, *, tool_name: str) -> lis
         )
     out: list[dict[str, Any]] = []
     for i, item in enumerate(pending_content):
-        try:
-            part = coerce_part(item)
-            if part is None and isinstance(item, dict):
+        part = coerce_part(item)
+        if part is None and isinstance(item, dict):
```
apascal07, line 108: This retry cannot succeed. `coerce_part` now calls `as_part`, and `as_part(dict)` is exactly `Part.model_validate(dict)`. If the first call raised, the second raises the same error.

The cost is the message. The old code reported the pydantic detail, for example `Tool 'x' pendingContent[0]: a part must have exactly one of text, media, ...`. Now both branches give the generic `must be a part, got dict.`, so the caller does not learn which rule was broken.

Suggest deleting lines 108-116 and putting `e.errors()[0]['msg']` back into the message at line 120.

py/packages/genkit/src/genkit/_core/_model.py
```diff
@@ -1064,16 +1326,66 @@ def output(self) -> OutputT | None:
...
+
+    @field_validator('artifact', mode='before')
+    @classmethod
+    def _wrap_artifact(cls, v: object) -> object:
+        if v is None:
+            return v
+        return as_artifact(v)
+
+
+_VENEEER_NS = {**vars(typing_mod), **globals()}
```
apascal07, line 1364: These twelve types are hand-written copies of `_typing` classes now, not subclasses, and nothing checks that the field sets stay equal.

Because `GenkitModel` sets `extra='forbid'`, a field added to `genkit-schema.json` and regenerated into `_typing.py` will make a newer peer's payload fail with `Extra inputs are not permitted` instead of round-tripping. The copies already differ today: `GenerateActionOptions.messages` lost its required flag.

A small test that compares `model_fields` between each veneer and its `_typing` twin would hold the line.

Minor: the constant on this line reads `_VENEEER_NS`.

apascal07 (approved):

py/packages/genkit-google-genai/src/genkit_google_genai/_interactions/converters.py
huangjeff5: Restored `custom={THOUGHT_CUSTOM: step.model_dump(...)}` on `from_thought_step`. A thought with an image summary replays as that same step.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: `as_model_request` keeps the `ModelRequest` (typed config stays). Messages are still walked so a planted two-kind part still fails.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: `GenerateActionOptions.messages` is required again.

py/packages/genkit/src/genkit/_ai/_tools.py
huangjeff5: Deleted the dead retry. pendingContent now surfaces the pydantic kind message.

py/packages/genkit/src/genkit/_core/_model.py
huangjeff5: Added a pin that each veneer still has every generated-twin field. Renamed `_VENEEER_NS` → `_VENEER_NS`.

PR #6311, opened 2026-09-08
fix(go/ai/exp): settle an `aborting` snapshot by how its work ended

apascal07:
Fixes the finalize of a detached run so an abort that lands as the turn is ending no longer overwrites the outcome. The abort flips the snapshot to `aborting`, and the finalize stamped `aborted` on every such row, even when the turn had already finished without an error or failed on its own. The row then held a fully committed turn labelled as stopped, so a client resumed it and ran the turn twice, and a real failure landed as `aborted` with its error dropped. JS keeps `aborted` winning on purpose (`abortAwareMutator`); this is a Go-only change, since the JS abort flips straight to `aborted` and carries no state to settle.

The abort is best effort: it stops the turn in flight and drops the inputs queued behind it, and the row records how that turn ended, on the rule attached turns already use (`callerStopped`):

- **A turn that finishes without an error is `completed`**, whatever the abort asked for in the meantime. A partial reply returned beside `ctx.Err()` is an error, not a finish: it commits and lands `aborted`, as today.
- **A turn that ends in an error is `aborted` when the caller's stop had reached the run by the time `fn` returned**, and `failed` otherwise. The error text is kept either way.
- **A stop that lands between turns ends the run before the next queued input starts.** The finished turn's outcome stands, so an empty continuation is refused rather than re-running it; the queued inputs never run. A run stopped before any turn ran is `aborted`.
- **A function that returns neither a result nor an error after a failed turn takes the turn's outcome.** The status follows the reason `Run` recorded and the turn's error is kept, on the attached output and the detached row alike. Returning an `AgentResult` completes the invocation on the function's own terms.

The attached cancel follows the same rule: a turn that finishes without an error under the disconnect returns the completed output, and only a turn cut short, or a run stopped before any turn ran, returns `aborted`. The "had reached the run" check is the work context's `Err` captured in the fn goroutine the moment `fn` returns (`fnDoneResult.ctxErr`): a cancel that lands after the run ended on its own did not stop it, and a flag set by the store-subscription goroutine only reflected scheduling. One `invocationOutcome` classifies both the attached output and the detached finalize, so the two cannot drift. The sub-agent middleware's abort tool and its winding-down report state the same rule to the model.

**Exported surface is unchanged.** The runtime changes are the ones above: an `aborting` row can settle `completed` or `failed`, a custom agent that swallows a turn error no longer reports `completed`, and an attached cancel under a finishing turn reports `completed`. Waits key on terminal status, and a second `Abort` reports the settled status. The package is experimental.


huangjeff5 (commented):
First bullet: “A turn that returns a result is `completed`.” A prompt-backed turn can also return a result **and** `ctx.Err()` — a partial reply next to a cancel. That case is `aborted` today, which is right.

Can the description distinguish a completed reply (successful return, no error) from a partial reply (text plus a cancel)? Otherwise “result” reads as “the model said anything.”

huangjeff5 (approved):

PR #6312, opened 2026-09-09
feat(py): add A2UI Surfaces middleware

huangjeff5:
Experimental `genkit-a2ui` middleware: add `Surfaces()` to `use=[...]` on `ai.generate` or `define_agent`. The model may emit a ```a2ui fence. On a finished turn the middleware rewrites that fence into `application/a2ui+json` data parts. `envelopes_from_parts(response.message.content)` is how you pull the UI payload for a renderer.

```python
from genkit_a2ui import Surfaces, envelopes_from_parts

response = await ai.generate(
    model='googleai/gemini-2.5-flash',
    prompt='Show me the weather in Tokyo',
    use=[Surfaces()],
)
envelopes = envelopes_from_parts(response.message.content)
```

Streaming rewrites chunks as fences close. The streamed card and `response.message` share the same surface ids.

On the next generate, those data parts become text again so the model can see what was painted: a prior surface is replayed as a ```a2ui fence, and a button click becomes `[UI action "…" on surface …]`.

A stopped turn (blocked / aborted / interrupted / failed / unknown / other) keeps the raw fence. The stop is the result; the middleware does not turn it into a card.

A length-stopped turn that ends inside a fence keeps the unfinished fence as text (`validate='warn'` / `'off'`). It is not dropped and it is not turned into a card.

Install with `pip install 'genkit[a2ui]'` or `pip install genkit-a2ui`.

## Catalogs

`Surfaces()` and `Surfaces(catalog='basic')` use the bundled basic catalog (Text, Card, Button, and the rest). You do not have to register basic to generate with it.

For your own components, register a catalog and pass its id. Registering a catalog does not change `Surfaces()` — you have to pass `catalog=<id>`.

```python
from genkit_a2ui import Surfaces, A2uiCatalog, A2uiCatalogComponent, load_catalog

catalog = A2uiCatalog(
    id='https://my-app.org/catalogs/custom.json',
    components=(
        A2uiCatalogComponent(name='Banner', description='A prominent alert.', props='title: string.'),
    ),
)
load_catalog(ai, catalog)

response = await ai.generate(
    model='googleai/gemini-2.5-flash',
    prompt='Show a warning banner',
    use=[Surfaces(catalog=catalog.id)],
)
```

`catalog=` is a registry id string, not a catalog object. An unknown id fails before the model runs.

`load_catalog_file(ai, './my-catalog.json')` registers from disk. A file that is not UTF-8, or JSON that is not a catalog (missing id, components, or a component name), fails with a message that names the path. `register_basic_catalog(ai)` puts the bundled catalog on the registry so the Developer UI can list it next to custom ones (`GET /api/values?type=a2ui-catalog`).

Loading the same id twice is fine. A different catalog under an already-used id is ignored; the first stays. If that id already holds a value that is not a catalog, `load_catalog` raises.

## Options

- `validate` (`'warn'` default): `'warn'` drops a bad fence or unknown component and keeps the rest of the turn; `'strict'` raises `A2uiParseError`; `'off'` skips the check. This checks envelope shape and component type names only, not prop values.
- `instructions` (`'system'` default): appends the catalog's capabilities to the system prompt so the model knows how to emit a fence. `'none'` injects nothing.
- `surface_id` / `surfaceId`: a fixed id reused for every surface. Omit it for a fresh UUID per surface.
- `version`: protocol version stamped on emitted envelopes (`'v0.9'` default; `'v0.9.1'` is also accepted). A typo is rejected so a renderer is not handed an unknown version.

### DIFF
--- a/py/packages/genkit-a2ui/pyproject.toml
+++ b/py/packages/genkit-a2ui/pyproject.toml
@@ -1,0 +1,81 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[project]
+authors = [
+  { name = "Google" },
+]
+classifiers = [
+  "Development Status :: 3 - Alpha",
+  "Environment :: Console",
+  "Environment :: Web Environment",
+  "Framework :: AsyncIO",
+  "Framework :: Pydantic",
+  "Framework :: Pydantic :: 2",
+  "Intended Audience :: Developers",
+  "Operating System :: OS Independent",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+  "Topic :: Scientific/Engineering :: Artificial Intelligence",
+  "Topic :: Software Development :: Libraries",
+  "Typing :: Typed",
+  "License :: OSI Approved :: Apache Software License",
+]
+dependencies = [
+  "genkit>=0.10.0",
+  "pydantic>=2.10.5",
+]
+description = "A2UI (Agent-to-UI) middleware for Genkit."
+keywords = [
+  "genkit",
+  "ai",
+  "llm",
+  "a2ui",
+  "middleware",
+]
+license = "Apache-2.0"
+name = "genkit-a2ui"
+readme = "README.md"
+requires-python = ">=3.10"
+version = "0.11.0"
+
+[project.optional-dependencies]
+dev = [
+  "pytest>=8.3.4",
+  "pytest-asyncio>=0.25.2",
+  "pytest-cov>=6.0.0",
+  "pytest-xdist>=3.6.1",
+]
+
+[project.urls]
+"Bug Tracker"   = "https://github.com/genkit-ai/genkit/issues"
+Changelog       = "https://github.com/genkit-ai/genkit/releases"
+"Documentation" = "https://genkit.dev/docs/get-started/?lang=python"
+"Homepage"      = "https://github.com/genkit-ai/genkit"
+"Repository"    = "https://github.com/genkit-ai/genkit/tree/main/py"
+
+[build-system]
+build-backend = "hatchling.build"
+requires      = ["hatchling"]
+
+[tool.hatch.build.targets.wheel]
+only-include = ["src/genkit_a2ui"]
+sources      = ["src"]
--- a/py/packages/genkit-a2ui/pyproject.toml
+++ b/py/packages/genkit-a2ui/pyproject.toml
@@ -40,7 +40,7 @@ classifiers = [
   "License :: OSI Approved :: Apache Software License",
 ]
 dependencies = [
-  "genkit>=0.10.0",
+  "genkit>=0.9.0",
   "pydantic>=2.10.5",
 ]
 description = "A2UI (Agent-to-UI) middleware for Genkit."
--- a/py/packages/genkit-a2ui/src/genkit_a2ui/_loader.py
+++ b/py/packages/genkit-a2ui/src/genkit_a2ui/_loader.py
@@ -1,0 +1,82 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Register A2UI catalogs so generate and the Developer UI can find them."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+from genkit._core._logger import get_logger
+from genkit._core._protocols import GenkitLike, RegistryLike
+
+from ._catalog import BASIC_CATALOG, A2uiCatalog
+from ._types import A2UI_CATALOG_VALUE_TYPE, BASIC_CATALOG_ID, DEFAULT_CATALOG_ID
+
+logger = get_logger(__name__)
+
+
+def load_catalog(ai: GenkitLike, catalog: A2uiCatalog) -> A2uiCatalog:
+    if not catalog.id:
+        raise ValueError('a2ui: load_catalog: catalog has no id')
+    existing = ai.registry.lookup_value(A2UI_CATALOG_VALUE_TYPE, catalog.id)
+    if existing is not None:
+        current = A2uiCatalog.from_value(existing)
+        if current is not None and current != catalog:
+            logger.warning(
+                'a2ui: load_catalog: a different catalog is already registered under this id; keeping the existing one'
+            )
+        return current or catalog
+    ai.registry.register_value(A2UI_CATALOG_VALUE_TYPE, catalog.id, catalog.as_value())
+    return catalog
+
+
+def load_catalog_file(ai: GenkitLike, path: str) -> A2uiCatalog:
+    return load_catalog(ai, read_catalog_file(path=path))
+
+
+def register_basic_catalog(ai: GenkitLike) -> A2uiCatalog:
+    return load_catalog(ai, BASIC_CATALOG)
+
+
+def read_catalog_file(*, path: str) -> A2uiCatalog:
+    try:
+        raw = json.loads(Path(path).read_text(encoding='utf-8'))
+    except OSError as exc:
+        raise ValueError(f'a2ui: failed to read catalog file {path!r}: {exc}') from exc
+    except json.JSONDecodeError as exc:
+        raise ValueError(f'a2ui: catalog file {path!r} is not valid JSON: {exc}') from exc
+    catalog = A2uiCatalog.from_value(raw)
+    if catalog is None:
+        raise ValueError(f'a2ui: catalog file {path!r} must have an "id" and a "components" array')
+    return catalog
+
+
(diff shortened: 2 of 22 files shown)


py/packages/genkit-a2ui/pyproject.toml
```diff
@@ -0,0 +1,81 @@
...
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+  "Topic :: Scientific/Engineering :: Artificial Intelligence",
+  "Topic :: Software Development :: Libraries",
+  "Typing :: Typed",
+  "License :: OSI Approved :: Apache Software License",
+]
+dependencies = [
+  "genkit>=0.10.0",
```
huangjeff5: 0.11

py/packages/genkit-a2ui/src/genkit_a2ui/_middleware.py
```diff
@@ -0,0 +1,308 @@
...
+            rewritten = True
+            text = summarize_envelopes(envelopes=envelopes_from_parts([part]))
+            if text:
+                content.append(Part(TextPart(text=text)))
+        if not rewritten:
+            messages.append(message)
+            continue
+        changed = True
+        if content:
+            messages.append(message.model_copy(update={'content': content}))
```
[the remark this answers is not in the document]
huangjeff5: If a prior turn was only A2UI parts and those parts summarize to empty text, we drop the whole message. That can put two user turns next to each other, and some providers 400. I'll keep a `[UI]` text part on that message so the role stays in history.

py/packages/genkit-a2ui/src/genkit_a2ui/_middleware.py
```diff
@@ -0,0 +1,308 @@
...
+    *,
+    response: ModelResponse,
+    catalog: A2uiCatalog,
+    validate: ValidateMode,
+    version: str,
+    surface_id: Callable[[], str],
+) -> ModelResponse:
+    message = response.message
+    if message is None and response.candidates:
+        message = Message(response.candidates[0].message)
```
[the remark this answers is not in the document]
huangjeff5: `candidates[0].message` is already a Message. I'll assign it directly instead of wrapping it in `Message(...)`.

py/packages/genkit-a2ui/pyproject.toml
huangjeff5: Pinning the genkit floor to `>=0.9.0` (minimum version providing agent middleware support).

py/packages/genkit-a2ui/src/genkit_a2ui/_middleware.py
```diff
@@ -0,0 +1,316 @@
...
+from ._parser import A2uiParseError, Segment, StreamParser
+from ._part import a2ui_part, envelopes_from_parts, has_a2ui_mime
+from ._types import DEFAULT_VERSION, SURFACE_KEYS, Envelope, SupportedVersion, ValidateMode
+
+ABNORMAL_FINISH_REASONS = frozenset({
+    FinishReason.BLOCKED,
+    FinishReason.ABORTED,
+    FinishReason.INTERRUPTED,
+    FinishReason.OTHER,
+})
```
apascal07, line 45: `FinishReason` also has `FAILED` and `UNKNOWN`. Neither is in this set, so a failed turn falls through to `transform_response` and its fence becomes a card. I checked with a `ProgrammableModel`: `finish=failed` and `finish=unknown` both give 2 envelopes, while `finish=blocked` keeps the raw fence.

The class docstring says "blocked / cancelled / aborted / other", but there is no `cancelled` reason; the code uses `INTERRUPTED`.

```suggestion
ABNORMAL_FINISH_REASONS = frozenset({
    FinishReason.BLOCKED,
    FinishReason.ABORTED,
    FinishReason.INTERRUPTED,
    FinishReason.FAILED,
    FinishReason.OTHER,
    FinishReason.UNKNOWN,
})
```

py/packages/genkit-a2ui/src/genkit_a2ui/_middleware.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def replay_next(self) -> str:
+        if self.cursor < len(self.recorded):
+            value = self.recorded[self.cursor]
+            self.cursor += 1
+            return value
+        return self.next()
+
+
+def part_text(*, part: Part) -> str | None:
```
apascal07, line 174: `root.text` of `''` makes this return `None`, so `rewrite_parts` sends the part down the `flush_nontext` path and calls `parser.flush()` while a fence is still open.

Content of `[TextPart('```a2ui\n{first half'), TextPart(''), TextPart('second half}\n```')]` gives 0 envelopes and puts the raw JSON tail in `response.text`. A `ReasoningPart` in that slot does the same.

Two changes fix it: return `root.text` here whenever the root is a `TextPart`, and test `if text is not None:` at line 191.

py/packages/genkit-a2ui/src/genkit_a2ui/_parser.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def take_closed_block(self, *, final: bool) -> ClosedBlock:
+        match = CLOSE_FENCE_RE.search(self.buffer[self.block_scan :])
+        if match is None:
+            if not final:
+                nl = self.buffer.rfind('\n')
+                if nl + 1 > self.block_scan:
+                    self.block_scan = nl + 1
+                return ClosedBlock(need_more=True)
+            batch = self.finalize_block(raw=self.buffer)
```
apascal07: When the turn ends inside a fence, `finalize_block` gets incomplete JSON, `reject()` warns, and the raw text is thrown away. This is the normal shape of a `finish_reason='length'` turn.

I ran it: `response.text` goes from `'Here is the weather:\n```a2ui\n[{"createSurface"...'` to just `'Here is the weather:\n'`. The user loses the card and the text.

In `warn` mode, please return the raw block as prose instead of dropping it.

py/packages/genkit-a2ui/src/genkit_a2ui/_parser.py
```diff
@@ -0,0 +1,304 @@
...
+        if self.require_root(envelopes=out) is None:
+            return None
+        catalog_id = self.catalog.id if self.catalog is not None else ''
+        create: Envelope = {
+            'version': self.version,
+            'createSurface': {'surfaceId': surface_id, 'catalogId': catalog_id},
+        }
+        return [create, *out]
+
+    def require_root(self, *, envelopes: list[Envelope]) -> list[Envelope] | None:
```
apascal07, line 226: `require_root` returns `None` on failure in every mode, so `validate='off'` still drops the block. Line 239 guards `validate_components` with `self.validate != 'off'`, but this check has no such guard.

In `off` mode `reject()` is a no-op, so the user gets no card, no text, and no log line. The README says `off` "skips the check". Please add the same guard here.

py/pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
   "genkit-google-cloud",
   "genkit-google-genai",
   "genkit-middleware",
+  "genkit-a2ui",
```
apascal07, line 32: The package is now a workspace member at version 0.11.0, but `.github/workflows/publish_python.yml` still lists only the older packages. `py/packages/genkit-a2ui` needs to go in that `PACKAGES` array, or `pip install genkit-a2ui` will fail after the next release.

Worth considering an `a2ui = ["genkit-a2ui"]` extra in `packages/genkit/pyproject.toml` too, next to the existing `middleware` extra.

py/packages/genkit-a2ui/src/genkit_a2ui/_loader.py
```diff
@@ -0,0 +1,82 @@
...
+    if not catalog.id:
+        raise ValueError('a2ui: load_catalog: catalog has no id')
+    existing = ai.registry.lookup_value(A2UI_CATALOG_VALUE_TYPE, catalog.id)
+    if existing is not None:
+        current = A2uiCatalog.from_value(existing)
+        if current is not None and current != catalog:
+            logger.warning(
+                'a2ui: load_catalog: a different catalog is already registered under this id; keeping the existing one'
+            )
+        return current or catalog
```
apascal07: If the id is already held by a value that is not a catalog, `from_value` returns `None` and this line returns the caller's catalog. Nothing was registered, but the caller sees a catalog come back and assumes success.

Every later `Surfaces(catalog=<id>)` then raises "registry value ... is not a catalog" from `resolve_catalog`. Raising here instead would surface the clash at load time.

py/packages/genkit-a2ui/src/genkit_a2ui/_loader.py
```diff
@@ -0,0 +1,82 @@
...
+    return load_catalog(ai, read_catalog_file(path=path))
+
+
+def register_basic_catalog(ai: GenkitLike) -> A2uiCatalog:
+    return load_catalog(ai, BASIC_CATALOG)
+
+
+def read_catalog_file(*, path: str) -> A2uiCatalog:
+    try:
+        raw = json.loads(Path(path).read_text(encoding='utf-8'))
```
apascal07, line 60: `Path.read_text` raises `UnicodeDecodeError` on a non-UTF-8 file. That is a `ValueError`, not an `OSError`, so it escapes both handlers and reaches the caller with no file name in the message. Adding `UnicodeDecodeError` to the first `except` fixes it.

Separately, line 65 says the file "must have an `id` and a `components` array", but `from_value` also returns `None` when a component row has no `name`, so that message points at the wrong problem.

apascal07 (approved):

py/packages/genkit-a2ui/src/genkit_a2ui/_middleware.py
huangjeff5: Added `FAILED` and `UNKNOWN`. A failed or unknown turn now keeps the raw fence.

py/packages/genkit-a2ui/src/genkit_a2ui/_middleware.py
huangjeff5: Empty text is still a text part, so a split fence is not flushed. The card still lands.

py/packages/genkit-a2ui/src/genkit_a2ui/_parser.py
huangjeff5: `warn` / `off` now keep the unfinished fence as prose instead of dropping it.

py/packages/genkit-a2ui/src/genkit_a2ui/_parser.py
huangjeff5: `validate='off'` now skips the root check the same way it skips component validation. The surface still lands.

py/pyproject.toml
huangjeff5: `genkit-a2ui` is in the publish list, and `pip install 'genkit[a2ui]'` is the extra.

py/packages/genkit-a2ui/src/genkit_a2ui/_loader.py
huangjeff5: `load_catalog` now raises when that id already holds a value that is not a catalog.

py/packages/genkit-a2ui/src/genkit_a2ui/_loader.py
huangjeff5: Non-UTF-8 files now raise with the path in the message. A file that is JSON but not a catalog (missing id, components, or a component name) says that in the error.

PR #6324, opened 2026-09-09
fix(py/plugins/openai): map max_tokens for reasoning models

### LINKED ISSUE #6320: [py/plugins/openai] max_tokens is sent verbatim and rejected by reasoning models
`OpenAIConfig(max_tokens=...)` is sent to `/v1/chat/completions` as `max_tokens`, which
reasoning models reject:

    Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not
    supported with this model. Use 'max_completion_tokens' instead.", 'type':
    'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}

Repro on gpt-6-astra: `ai.generate(prompt='Say hi.', config={'max_tokens': 32})`. The Dev UI
model config panel exposes Max tokens too, so it fails the same way from there.

Cause: `_openai_create_kwargs` in `models/model.py` forwards every non-None `OpenAIConfig`
field under its own name. `typing.py` already marks `max_tokens` deprecated but nothing
translates it.

Surfaced while testing #6306. Related: #5898, #4425.

0x90000:
## Summary

Fixes #6320.

`OpenAIConfig(max_tokens=...)` was forwarded as `max_tokens` for reasoning models, which OpenAI rejects. This maps the deprecated option to `max_completion_tokens` for reasoning-model IDs (`o1`, `o3`, `o4`, `gpt-5*`, and `gpt-6*`) while preserving `max_tokens` for legacy models.

When both options are supplied, the explicit `max_completion_tokens` value wins so mutually exclusive fields are not sent together.

## Tests

- Added async regression coverage for reasoning models, legacy models, and explicit precedence.
- `ruff check` and `ruff format --check` pass on changed files.
- Python compilation passes.
- The focused test file runs 49 passing tests in an isolated dependency environment; one unrelated existing usage assertion fails against the installed OpenAI SDK because `image_tokens` is omitted from its `PromptTokensDetails` model.
- The full workspace environment cannot be resolved on Windows because `dotpromptz-handlebars` currently publishes no Windows wheel.

Google CLA requirements in `CONTRIBUTING.md` noted.

### DIFF
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -61,7 +61,16 @@ def _uses_max_completion_tokens(model: str | None) -> bool:
     if not model:
         return False
     model_id = model.rsplit('/', 1)[-1].lower()
-    return model_id.startswith(('o1', 'o3', 'o4', 'gpt-5', 'gpt-6'))
+    # Fine-tuned OpenAI model ids are prefixed with ``ft:`` and custom
+    # deployment names may put the base model after an arbitrary prefix.
+    if model_id.startswith('ft:'):
+        parts = model_id.split(':')
+        if len(parts) > 1:
+            model_id = parts[1]
+    reasoning_prefixes = ('o1', 'o3', 'o4', 'gpt-5', 'gpt-6')
+    return model_id.startswith(reasoning_prefixes) or any(
+        f'-{prefix}' in model_id for prefix in reasoning_prefixes
+    )
 
 
 def _openai_create_kwargs(*, config: OpenAIConfig, model: str | None = None) -> dict[str, Any]:
--- a/py/packages/genkit-openai/src/genkit_openai/models/model.py
+++ b/py/packages/genkit-openai/src/genkit_openai/models/model.py
@@ -56,7 +56,15 @@ logger = structlog.get_logger(__name__)
 _GENKIT_ONLY = frozenset({'api_key', 'top_k', 'version', 'max_output_tokens', 'stop_sequences'})
 
 
-def _openai_create_kwargs(*, config: OpenAIConfig) -> dict[str, Any]:
+def _uses_max_completion_tokens(model: str | None) -> bool:
+    """Whether a model requires the reasoning-model token-limit parameter."""
+    if not model:
+        return False
+    model_id = model.rsplit('/', 1)[-1].lower()
+    return model_id.startswith(('o1', 'o3', 'o4', 'gpt-5', 'gpt-6'))
+
+
+def _openai_create_kwargs(*, config: OpenAIConfig, model: str | None = None) -> dict[str, Any]:
     """Kwargs for chat.completions.create().
 
     Peel Genkit-only keys. ``stop_sequences`` becomes ``stop`` when ``stop``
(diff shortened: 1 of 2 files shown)


py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -71,6 +86,15 @@ def _openai_create_kwargs(*, config: OpenAIConfig) -> dict[str, Any]:
             continue
         value = getattr(config, name)
         if value is not None:
+            if name == 'max_tokens':
+                # OpenAI reasoning models reject the deprecated max_tokens
+                # field. Keep the explicit max_completion_tokens value when
+                # both knobs are supplied so the request remains valid.
+                if config.max_completion_tokens is not None:
+                    continue
+                if _uses_max_completion_tokens(model) or config.reasoning_effort is not None:
```
hilariie, line 97: The reasoning_effort branch is not covered. Can we add one case with a model id the heuristic does not match (say 'my-deployment') and reasoning_effort set, asserting max_completion_tokens goes out?

py/packages/genkit-openai/src/genkit_openai/models/model.py
```diff
@@ -71,6 +86,15 @@ def _openai_create_kwargs(*, config: OpenAIConfig) -> dict[str, Any]:
             continue
         value = getattr(config, name)
         if value is not None:
+            if name == 'max_tokens':
```
hilariie, line 91: Nit: This docstring still says everything else goes out under its field name. Can we add a line that max_tokens becomes max_completion_tokens for reasoning models?

py/packages/genkit-openai/tests/openai_model_test.py
```diff
@@ -131,6 +131,52 @@ async def test_get_openai_config_peels_genkit_keys_and_passes_the_rest() -> None
     assert 'version' not in body
 
 
+@pytest.mark.asyncio
+async def test_get_openai_config_uses_max_completion_tokens_for_reasoning_models() -> None:
+    """Reasoning models reject the deprecated max_tokens request field."""
+    for model_name in ('gpt-6-astra', 'ft:o1-mini:my-org:custom', 'my-o1-mini-deployment'):
```
hilariie: Nit: This file already uses pytest.mark.parametrize; can we use it here so each id shows up as its own case?

0x90000:
Addressed the requested review feedback in commit `4df4c376d78`:

- Added `my-deployment` with `reasoning_effort=high` coverage, verifying `max_completion_tokens` is emitted even when the model ID does not match the heuristic.
- Converted the reasoning-model cases to `pytest.mark.parametrize`, so each model ID is reported independently.
- Documented in `_openai_create_kwargs` that `max_tokens` maps to `max_completion_tokens` for reasoning models.

Verification:
- Focused review cases: `4 passed, 49 deselected`
- `ruff check --preview`: passed
- `ruff format --check --preview`: passed
- `compileall`: passed

The full package test run is `52 passed, 1 failed`; the existing failure is `test__generate_reports_extra_token_counts`, where the installed OpenAI SDK drops `image_tokens` from `PromptTokensDetails`, unrelated to this change.

hilariie (approved):

PR #6328, opened 2026-09-10
feat(py/plugins/google-genai): curate stable Gemini 2.5, flash-lite-latest, TTS, and Gemma 4

### LINKED ISSUE #6178: chore(py/plugins/google-genai): stable Gemini 2.5, flash-lite-latest, TTS, and Gemma 4 are missing from the curated catalog
Python's curated map in `gemini.py` (`SUPPORTED_MODELS`, populated via `_add_model`) covers the previews, the `-latest` aliases, and the 3.x families, but not the stable `gemini-2.5-pro` / `gemini-2.5-flash` names or the `gemini-flash-lite-latest` alias (absent from Python entirely). Those resolve through the generic dynamic fallback with default metadata, while JS curates all three. TTS names resolve via a `'tts' in name` substring check onto one shared generic entry instead of per-name curated entries like JS's `gemini-2.5-flash-preview-tts` / `gemini-2.5-pro-preview-tts` / `gemini-3.1-flash-tts-preview`. And Gemma is a generation behind: the enum stops at `gemma-3`/`gemma-3n` while JS curates `gemma-4-26b-a4b-it` and `gemma-4-31b-it`.

hilariie:
Closes #6178

## Summary

Curates the ids the issue lists: `gemini-2.5-pro`, `gemini-2.5-flash`, the `gemini-flash-lite-latest` alias, per-name entries for `gemini-2.5-flash-preview-tts`, `gemini-2.5-pro-preview-tts` and `gemini-3.1-flash-tts-preview`, and `gemma-4-26b-a4b-it` / `gemma-4-31b-it`. The new ids autocomplete on the typed ref constructors. TTS entries, including the generic fallback, no longer advertise `system_role`.

Two fixes in the TTS request path:

- `SpeechConfigSchema` kept only `voiceConfig` and only in its camelCase form. It now allows extra fields and populates by field name, so `language_code` and `multi_speaker_voice_config` reach the SDK and the snake_case config in the media sample applies. An unknown nested key now fails with `INVALID_ARGUMENT` instead of being dropped.
- `generate()` overwrote a caller-supplied `response_modalities` on TTS and image models. The default is applied only when the caller left it unset.

## Verification

Ran every curated id from the Dev UI against the Gemini API. All resolve to their curated metadata and generate, with two exceptions:

- `gemini-2.5-pro` returns 404 "no longer available to new users" on an account that never called it. Google lists it as GA with no shutdown date and says the 2.5 models are not deprecated (https://discuss.ai.google.dev/t/176380). The gate is per account, so the entry stays stage-less.
- `gemini-3.1-flash-tts-preview` returns 400 `INVALID_ARGUMENT` for a plain prompt, the same as on `main`.

Speech config with the media sample's snake_case config (voice Kore):

| input | before | after |
|---|---|---|
| `voice_config` (snake_case) | dropped | applied |
| `languageCode` | dropped | applied |
| unknown key `languageCodes` | dropped | `INVALID_ARGUMENT` |

### DIFF
--- a/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
+++ b/py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
@@ -557,6 +557,16 @@ GEMINI_2_5_FLASH_PREVIEW_04_17 = ModelInfo(
     ),
 )
 
+GEMINI_2_5_PRO = ModelInfo(
+    label='Google AI - Gemini 2.5 Pro',
+    supports=GEMINI_TEXT_SUPPORTS,
+)
+
+GEMINI_2_5_FLASH = ModelInfo(
+    label='Google AI - Gemini 2.5 Flash',
+    supports=GEMINI_TEXT_SUPPORTS,
+)
+
 GEMINI_2_5_FLASH_LITE = ModelInfo(
     label='Google AI - Gemini 2.5 Flash Lite',
     supports=Supports(
@@ -672,6 +682,11 @@ GEMINI_3_1_FLASH_LITE = ModelInfo(
     ),
 )
 
+GEMINI_FLASH_LITE_LATEST = ModelInfo(
+    label='Google AI - Gemini Flash Lite Latest',
+    supports=GEMINI_TEXT_SUPPORTS,
+)
+
 GEMINI_IMAGE_SUPPORTS = Supports(
     multiturn=True,
     media=True,
@@ -711,6 +726,50 @@ GEMINI_2_5_FLASH_IMAGE_PREVIEW = ModelInfo(
     supports=GEMINI_IMAGE_SUPPORTS,
 )
 
+GEMINI_TTS_SUPPORTS = Supports(
+    multiturn=False,
+    media=False,
+    tools=False,
+    tool_choice=False,
+    system_role=False,
+    constrained=Constrained.ALL,
+)
+
+GEMINI_2_5_FLASH_PREVIEW_TTS = ModelInfo(
+    label='Google AI - Gemini 2.5 Flash Preview TTS',
+    supports=GEMINI_TTS_SUPPORTS,
+)
+
+GEMINI_2_5_PRO_PREVIEW_TTS = ModelInfo(
+    label='Google AI - Gemini 2.5 Pro Preview TTS',
+    supports=GEMINI_TTS_SUPPORTS,
+)
+
+GEMINI_3_1_FLASH_TTS_PREVIEW = ModelInfo(
+    label='Google AI - Gemini 3.1 Flash TTS Preview',
+    supports=GEMINI_TTS_SUPPORTS,
+)
+
+GEMMA_SUPPORTS = Supports(
+    multiturn=True,
+    media=True,
+    tools=True,
+    tool_choice=True,
+    system_role=True,
+    constrained=Constrained.ALL,
+    output=['text', 'json'],
+)
+
+GEMMA_4_26B_A4B_IT = ModelInfo(
+    label='Google AI - Gemma 4 26B A4B IT',
+    supports=GEMMA_SUPPORTS,
+)
+
+GEMMA_4_31B_IT = ModelInfo(
+    label='Google AI - Gemma 4 31B IT',
+    supports=GEMMA_SUPPORTS,
+)
+
 GENERIC_GEMINI_MODEL = ModelInfo(
     label='Google AI - Gemini',
     supports=Supports(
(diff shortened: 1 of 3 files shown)


py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -711,6 +744,50 @@ class GemmaConfigSchema(GeminiConfigSchema):
     supports=GEMINI_IMAGE_SUPPORTS,
 )
 
+GEMINI_TTS_SUPPORTS = Supports(
```
cabljac, line 746: This is the definition main's #6301 fix collides with (the textual conflict itself lands on `GENERIC_TTS_MODEL` at :804). Suggest: keep the five `False` fields, `constrained=Constrained.NONE`, `output=['media']`, and keep `GENERIC_TTS_MODEL` on it.

py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
```diff
@@ -612,6 +613,87 @@ def test_vertexai_gemini_3_x_text_models_register_real_capabilities(model_name:
...
+    """Each TTS id carries its own label instead of sharing the generic TTS entry."""
+    model_info = google_model_info(model_name)
+
+    assert model_info.label == expected_label
+    assert model_info.supports is not None
+    assert model_info.supports.multiturn is False
+    assert model_info.supports.media is False
+    assert model_info.supports.tools is False
+    assert model_info.supports.tool_choice is False
+    assert model_info.supports.output is None
```
cabljac: This asserts `output is None`, which is the opposite of what main pins. Worth asserting the full `Supports` here (and for the Gemma 4 entries) so `constrained` is covered.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -778,6 +840,7 @@ class VertexAIGeminiVersion(StrEnum, metaclass=Deprecations):  # pyrefly: ignore
     GEMINI_2_5_FLASH_LITE = 'gemini-2.5-flash-lite'
     GEMINI_2_5_FLASH_PREVIEW_TTS = 'gemini-2.5-flash-preview-tts'
     GEMINI_2_5_PRO_PREVIEW_TTS = 'gemini-2.5-pro-preview-tts'
+    GEMINI_3_1_FLASH_TTS_PREVIEW = 'gemini-3.1-flash-tts-preview'
```
cabljac, line 843: nit: the Vertex enum gains `gemini-3.1-flash-tts-preview` and the gemma-4 ids. JS vertexai curates neither TTS nor Gemma. Consistent with the 2.5 TTS and gemma-3 ids already here, just flagging in case you have seen them served on Vertex.

cabljac (changes requested):
The text-model `Supports` match JS field for field, and defaulting `response_modalities` only when unset is the right call. One thing to dig into, then a couple of small ones.

This branch predates #6301, and `GEMINI_TTS_SUPPORTS` reintroduces `constrained=ALL` with no `output` for every TTS id including the fallback. From what I can see that re-enables `response_mime_type` / `response_schema` on TTS requests, which is what #6301 was merged to stop. After a rebase main's `test_google_model_info` TTS case will also go red on the new label, constrained, and output. Can you rebase, set the TTS entries to `constrained=NONE` / `output=['media']`, and re-run?

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
hilariie: Vertex serves gemini-3.1-flash-tts-preview via generateContent ([source](https://docs.cloud.google.com/text-to-speech/docs/gemini-tts)), so that stays. Gemma 4 on Vertex is only the 26B, under gemma-4-26b-a4b-it-maas ([source](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/maas/google/gemma-4-26b-a4b-it)), so both gemma-4 ids are dropped from the Vertex enum in `cc4cf6af9`

py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
hilariie: Resolved in `e2ae140e1`

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
hilariie: Rebased onto main and resolved in `67aa0ab3c`

PR #6339, opened 2026-09-14
fix(go/plugins/mcp): honor StreamableHTTPConfig.HTTPClient in the Str…

### LINKED ISSUE #6329: [Go] `StreamableHTTPConfig.HTTPClient` is silently ignored by the Streamable HTTP transport
# Summary
`StreamableHTTPConfig.HTTPClient` (`go/plugins/mcp/client.go`) is a dead field:
Genkit pins `mcp-go v0.29.0`, whose Streamable HTTP transport exposes no option
to inject a custom `http.Client`, so the field is silently ignored.

The SSE transport already honors its `HTTPClient` field via `WithHTTPClient`;
Streamable HTTP does not, even though it's the newer, recommended transport.

**Why a custom `http.Client` matters**

A custom client is the only way to set the transport-level behavior that a
remote MCP server often requires:

- **mTLS / client certificates** (`TLSClientConfig.Certificates`)
- **Private CA / custom trust** (`TLSClientConfig.RootCAs`)
- **Corporate proxies** (`http.Transport.Proxy`)
- **Connection pooling / keep-alive tuning** (`MaxIdleConns`,
  `MaxIdleConnsPerHost`, `IdleConnTimeout`)
- **Dial and TLS handshake timeouts** (`DialContext`, `TLSHandshakeTimeout`)
- **Custom `RoundTripper`** (tracing, retries, rate limiting)

# Reproducing the Issue
Pass a custom `http.Client` to the Streamable HTTP transport and observe that it
is never used:

```go
package main

import (
    "net/http"

    "github.com/firebase/genkit/go/plugins/mcp"
)

// A round tripper that records whether it is ever called. It isn't.
type recordingTransport struct{ called bool }
func (t *recordingTransport) RoundTrip(*http.Request) (*http.Response, error) {
    t.called = true
    return nil, nil
}

func main() {
    recorder := &recordingTransport{}
    customClient := &http.Client{Transport: recorder}

    _, _ = mcp.NewGenkitMCPClient(mcp.MCPClientOptions{
        Name: "my-mcp-server",
        StreamableHTTP: &mcp.StreamableHTTPConfig{
            BaseURL:    "https://mcp.example.com",
            HTTPClient: customClient, // <-- silently ignored
        },
    })

    // recorder.called is false: the transport used its own default client.
}
```

**Expected behavior**

The provided HTTPClient should be used for all requests made by the Streamable
HTTP transport, so that callers can control TLS (mTLS, private CAs), proxy,
connection pooling / keep-alive, and transport-level timeouts — exactly as the
SSE transport already does via WithHTTPClient.

**Runtime:**
 - OS: MacOS
 - Version 26.6

**Go version**
go version go1.27.0 darwin/arm64

**_Edit: I tried opening a PR, but I believe there are access restrictions, so I incremented this issue with a "How to Fix" section._**

## How to Fix

The custom-client option already exists upstream: `mcp-go` added
`transport.WithHTTPBasicClient` in `v0.32.0`. The fix is a dependency bump plus
wiring the field. Some minor **internal** API compatibility changes must be made as well, which would not impact the user.

### 1. Bump `mcp-go`

In `go/go.mod`:

```diff
-	github.com/mark3labs/mcp-go v0.29.0
+	github.com/mark3labs/mcp-go v0.32.0
```

`v0.32.0` is the minimal version with `WithHTTPBasicClient`. It declares `go 1.23`, so it does **not** force Genkit's `go` directive (currently `1.25.0`).

### 2. Wire the field

In `go/plugins/mcp/client.go`, inside `createTransport`'s Streamable HTTP branch
(the `if options.StreamableHTTP != nil { ... }` block), apply the custom client
**before** `WithHTTPTimeout`:

```go
if options.StreamableHTTP != nil {
    var streamableHTTPOptions []transport.StreamableHTTPCOption
    // Apply the custom client first so a later WithHTTPTimeout sets its
    // Timeout field rather than being lost when the client is replaced.
    if options.StreamableHTTP.HTTPClient != nil {
        streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPBasicClient(options.StreamableHTTP.HTTPClient))
    }
    if options.StreamableHTTP.Headers != nil {
        streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPHeaders(options.StreamableHTTP.Headers))
    }
    if options.StreamableHTTP.Timeout > 0 {
        streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPTimeout(options.StreamableHTTP.Timeout))
    }
    // ...
}
```

The ordering matters: `WithHTTPTimeout` mutates the `Timeout` field of whichever client is currently installed, so it must run after `WithHTTPBasicClient`. The purpose of this is to not impact the current `Genkit` interface.

### 3. Migrate the request literals (API change)

`mcp-go v0.32.0` renamed the anonymous `Params` structs to named types, so the request construction in `prompts.go`, `tools.go`, and `resources.go` must change, e.g.:

```go
// before (v0.29.0)
req := mcp.GetPromptRequest{
    Params: struct {
        Name      string            `json:"name"`
        Arguments map[string]string `json:"arguments,omitempty"`
    }{Name: promptName, Arguments: args},
}

// after (v0.32.0)
req := mcp.GetPromptRequest{
    Params: mcp.GetPromptParams{Name: promptName, Arguments: args},
}
```

The full list of replacements:

| File | Old (anonymous) | New (named) |
|---|---|---|
| `prompts.go` | `GetPromptRequest.Params` struct | `mcp.GetPromptParams` |
| `prompts.go` | `ListPromptsRequest.PaginatedRequest.Params` struct | `mcp.PaginatedParams` |
| `tools.go` | `ListToolsRequest.PaginatedRequest.Params` struct | `mcp.PaginatedParams` |
| `tools.go` | `CallToolRequest.Params` struct | `mcp.CallToolParams` |
| `resources.go` | `ReadResourceRequest.Params` struct | `mcp.ReadResourceParams` |
| `resources.go` | `ListResourcesRequest.PaginatedRequest.Params` struct | `mcp.PaginatedParams` |
| `resources.go` | `ListResourceTemplatesRequest.PaginatedRequest.Params` struct | `mcp.PaginatedParams` |

### 4. Add a test

Add a unit test (e.g. `client_test.go`) asserting `createTransport` actually uses the custom client, a recording `RoundTripper` that the transport must hit, plus a check that the custom client's `Timeout` is set (proving the ordering in step 2):

```go
customClient := &http.Client{Transport: recordingRoundTripper{...}}
c := &GenkitMCPClient{}
tr, err := c.createTransport(MCPClientOptions{
    StreamableHTTP: &StreamableHTTPConfig{
        BaseURL:    "http://example.com/mcp",
        HTTPClient: customClient,
        Timeout:    7 * time.Second,
    },
})
// assert: customClient.Timeout == 7s, and driving tr.SendRequest(...)
// hits the recordingRoundTripper.
```

### 5. Verify

```bash
cd go
go build ./...
go vet ./plugins/mcp/...
go test ./plugins/mcp/...
gofmt -l plugins/mcp
```

vllst-io:
## Summary

The Streamable HTTP transport ignored `StreamableHTTPConfig.HTTPClient`. When a
caller supplied a custom `*http.Client`, it was silently dropped and the MCP
connection used `mcp-go`'s internal default client instead. This PR makes the
Streamable HTTP transport honor the caller's custom client.

## Problem

In `createTransport`, the Streamable HTTP transport options were applied in this
order:

```go
if options.StreamableHTTP.Headers != nil {
    streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPHeaders(...))
}
if options.StreamableHTTP.Timeout > 0 {
    streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPTimeout(...))
}
```

`StreamableHTTPConfig.HTTPClient` existed on the config struct, but was never
passed to the transport. Worse, applying a custom client *after* `WithHTTPTimeout`
would have been lossy: `WithHTTPTimeout` mutates the client currently installed on
the transport, so the timeout would land on the internal default client and then be
discarded when the client was replaced.

This meant callers could not bring their own transport (custom TLS roots, proxies,
connection pooling, etc.) for Streamable HTTP connections, unlike the SSE transport
which already honored `SSEConfig.HTTPClient`.

## Solution

Apply the custom client **first**, before `WithHTTPTimeout`, so the timeout is set
on the client that is actually used:

```go
if options.StreamableHTTP.HTTPClient != nil {
    streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPBasicClient(options.StreamableHTTP.HTTPClient))
}
if options.StreamableHTTP.Headers != nil {
    streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPHeaders(...))
}
if options.StreamableHTTP.Timeout > 0 {
    streamableHTTPOptions = append(streamableHTTPOptions, transport.WithHTTPTimeout(...))
}
```

The `mcp-go` dependency is bumped from `v0.29.0` to `v0.32.0`, which is required to
construct the request params with the named types (`GetPromptParams`,
`PaginatedParams`, `ReadResourceParams`, `CallToolParams`) that the newer library
versions expect. The inline anonymous structs were replaced accordingly. That was the minimal version bump to avoid breaking changes.

## Changes

- `go/plugins/mcp/client.go` — honor `StreamableHTTPConfig.HTTPClient` by applying
  `WithHTTPBasicClient` before `WithHTTPTimeout`.
- `go/plugins/mcp/prompts.go`, `resources.go`, `tools.go` — replace inline anonymous
  params structs with named `mcp-go` types.
- `go/go.mod` / `go/go.sum` — bump `github.com/mark3labs/mcp-go` to `v0.32.0`.
- `go/plugins/mcp/client_test.go` (new) — `TestCreateTransportHonorsStreamableHTTPClient`
  verifies that the custom client is used by the transport and that the timeout is
  applied to it.

## Testing

- `go build ./plugins/mcp/...` — passes
- `go test ./plugins/mcp/...` — passes
- `go test ./plugins/mcp/ -run TestCreateTransportHonorsStreamableHTTPClient -v` — passes
- `go vet ./plugins/mcp/...` and `gofmt` — clean

## Related

Fixes #6329


vllst-io:
Hey there, @hilariie! Could you take a look at this PR? I can't formally assign, maybe access restrictions. Thanks a lot, great lib

adesinah (approved):
@vllst-io thanks for working on this. I pushed a few changes that addressed some missed points.

vllst-io:
> @vllst-io thanks for working on this. I pushed a few changes that addressed some missed points.

Thanks a lot for the review and commits. Hope I can contribute more!

apascal07 (approved):

PR #6341, opened 2026-09-14
fix(js/plugins/google-genai): TTS ModelInfo no longer claims constrained support

hilariie:
## Why

TTS models don't support constrained output, but `GENERIC_TTS_MODEL` declared `constrained: 'all'`. Every `KNOWN_TTS_MODELS` entry and the uncatalogued-TTS branch of `model()` spread that same `info`, so the framework skipped simulateConstrainedGeneration for every TTS model and the plugin sent responseMimeType: 'application/json' alongside responseModalities: ['AUDIO'], which the API rejects with 500 INTERNAL. Go's `TTSSupports` leaves constrained unset; Python fixed the same bug in #6301

## Changes

`GENERIC_TTS_MODEL` declares `constrained: 'none'` and `output: ['media']`. With the simulator running, the `request.output?.constrained && jsonMode` guard no longer fires and a TTS request carries just `responseModalities`.

Adds a TTS request-formation test asserting the whole `generationConfig`, two positive controls for the native constrained path (plain and `legacyResponseSchema`), and TTS supports assertions for known, preview, and uncatalogued model names.

## Verification

`pnpm test` and `pnpm check` in `js/plugins/google-genai`. The new tests fail with `constrained: 'all'` restored. Confirmed live against gemini-2.5-flash-preview-tts: the pre-fix payload returns 500, the post-fix payload returns a clear 400.


adesinah (approved):
Looks good to me.

ifielker (approved):

PR #6343, opened 2026-09-14
refactor(go): pluggable, OpenTelemetry-decoupled telemetry instrumentation

pavelgj:
Pluggable, OpenTelemetry-decoupled telemetry instrumentation

## Summary

Reworks Genkit Go's telemetry so instrumentation is a pluggable abstraction
instead of a hardcoded OpenTelemetry (OTel) dependency. `RunInNewSpan` becomes a
thin dispatcher over one or more configured `Instrumentation` providers whose
span hook behaves like middleware (it wraps `next`), keeping all
backend-independent Genkit semantics (path building, isRoot, state/output, debug
logging) in one place so providers only encode.

The default stays instrumented via `OTelInstrumentation` for back-compat, but it
no longer owns the global OTel state: it reads whatever provider the application
registered with `otel.SetTracerProvider` (OTel's no-op provider otherwise) and
never installs one itself. In the dev environment a self-contained
`DirectTelemetryInstrumentation` is auto-injected when a telemetry server is
configured; it feeds the Developer UI by posting spans straight to
`/api/traces`, with no OTel SDK on that path.

This mirrors the direction of the Genkit Dart (#348) and JS (#6336) work,
adapted for Go idioms.

## Instrumentation vs collection

The rework separates two concerns that were previously entangled:

- **Instrumentation** (span creation) is owned by `RunInNewSpan` and the
  pluggable `Instrumentation` providers.
- **Collection/export** (OTel SDK, exporters, prod backends) stays with the
  explicit `WriteTelemetry*` helpers and the GCP / Firebase plugins.

`TracerProvider()` no longer auto-registers the telemetry-server span processor,
and `OTelInstrumentation` no longer creates a global provider. Genkit stops
mutating global OTel state on the span-creation path.

## Example: a minimal custom instrumentation

A provider implements `Instrumentation.RunInNewSpan`: it wraps `next` and
returns its result. This one prints a span's labels before the call, its output
after, and any error if it fails. It mints no ids, so it hands `next` a span
that reports empty trace ids.

```go
import (
	"context"
	"fmt"

)

type printingSpan struct{}

func (printingSpan) TraceInfo() tracing.TraceInfo  { return tracing.TraceInfo{} }
func (printingSpan) SetMetadata(map[string]string) {}

type PrintingInstrumentation struct{}

func (PrintingInstrumentation) RunInNewSpan(ctx context.Context, info *tracing.SpanInfo, next tracing.NextFunc) (any, error) {
	fmt.Printf("-> span labels: %v\n", info.Labels)
	out, err := next(ctx, printingSpan{})
	if err != nil {
		fmt.Printf("!! error: %v\n", err)
		return out, err
	}
	fmt.Printf("<- output: %v\n", out)
	return out, nil
}

// Replace the implicit OTel default with your own provider.
tracing.ConfigureInstrumentation(PrintingInstrumentation{})
```

(An exported, read-only view of the running span metadata (name, type, input,
...) beyond `Labels` follows in the Go GenAI plugin PR, #6345.)

## What's new

- `Instrumentation`, `Span`, `NextFunc`, `SpanInfo` and a provider registry
  (`ConfigureInstrumentation`, `ResetInstrumentation`); a composite span
  resolves ids as first-non-empty across the chain and fans metadata out to all
  providers.
- `OTelInstrumentation` (implicit default): the historical OTel body, now
  reading `otel.GetTracerProvider()`; yields empty ids when nothing is
  configured.
- `DirectTelemetryInstrumentation`: mints its own ids, builds the same
  `SpanData`/`Data` the OTel exporter produced, and posts start+end to the dev
  telemetry server. Dev UI traces are byte-for-byte identical to before.
- `SetCustomMetadataAttributes`: backend-independent way to annotate the current
  span, fanning out to every active provider. Replaces direct OTel span writes
  (agent.go's `agent:sessionId` / `agent:snapshotId` now go through it).

## Behavior change

With no OTel provider configured outside dev, OTel-backed spans now carry no
trace ids and export nothing, instead of Genkit lazily creating an (inert) SDK
provider. Configure OpenTelemetry yourself, or use the GCP / Firebase plugins.
`TracerProvider`, `Tracer`, and `WriteTelemetry*` are deprecated.

## Other

- Log export correlates via Genkit's own span ids (composite, Direct-wins in
  dev) instead of the OTel active context.
- Tests assert against the captured `SpanData` via an in-memory client rather
  than OTel side effects; OTel remains only where it is the subject
  (`OTelInstrumentation`, the span processor/exporter) and in plugins.
- Adds a model-free `basic-flows` sample used to verify trace-shape parity.


go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,214 @@
...
+	start := time.Now()
+	if realtimeTelemetryEnabled() {
+		d.exportStart(info, span, parent, start)
+	}
+
+	ctx = directParentKey.NewContext(ctx, &directParent{traceID: traceID, spanID: spanID})
+	out, err := next(ctx, span)
+	d.exportEnd(info, span, parent, start, time.Now(), err)
+	return out, err
+}
```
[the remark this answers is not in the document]
pavelgj: Good catch, taken in 38339c309. Deferred exportEnd with named returns so a panic in next still finalizes the span, matching the OTel provider.

go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	if parent == nil {
+		td.DisplayName = info.metadata.Name
+		td.StartTime = ToMilliseconds(start)
+		if !end.IsZero() {
+			td.EndTime = ToMilliseconds(end)
+		}
+	}
+	return td
+}
```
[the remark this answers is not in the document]
pavelgj: Declining: SpanInfo.metadata is unexported and the only construction site is the dispatcher (tracing.go), which always sets a non-nil metadata. Out-of-package providers receive that same *SpanInfo and cannot reconstruct it with nil metadata, so this path is unreachable by construction. If we later promote metadata to an exported, out-of-package-settable view, we will add guards then.

go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,214 @@
...
+	sm := info.metadata
+	attrs := attributesToMap(sm.attributes())
+	for k, v := range info.Labels {
+		attrs[k] = v
+	}
+	span.mu.Lock()
+	for k, v := range span.extra {
+		attrs[attrPrefix+":metadata:"+k] = v
+	}
+	span.mu.Unlock()
```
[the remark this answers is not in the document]
pavelgj: Declining: SpanInfo.metadata is unexported and the only construction site is the dispatcher (tracing.go), which always sets a non-nil metadata. Out-of-package providers receive that same *SpanInfo and cannot reconstruct it with nil metadata, so this path is unreachable by construction. If we later promote metadata to an exported, out-of-package-settable view, we will add guards then.

go/core/tracing/otel_instrumentation.go
```diff
@@ -0,0 +1,104 @@
...
+	// run-determined output/state). Registered after span.End so it runs first.
+	defer func() { span.SetAttributes(sm.attributes()...) }()
+
+	out, err := next(ctx, &otelSpan{span: span})
+	if err != nil && !isErrorAlreadyMarked(err) {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+	}
+	return out, err
+}
```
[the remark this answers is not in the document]
pavelgj: Declining: SpanInfo.metadata is unexported and the only construction site is the dispatcher (tracing.go), which always sets a non-nil metadata. Out-of-package providers receive that same *SpanInfo and cannot reconstruct it with nil metadata, so this path is unreachable by construction. If we later promote metadata to an exported, out-of-package-settable view, we will add guards then.

go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return td
+}
+
+// exportStart ships the span as in progress (endTime 0), asynchronously, so
+// span creation never blocks on telemetry I/O. A live preview is best-effort:
+// a dropped start export is corrected by the synchronous end export.
+func (d *DirectTelemetryInstrumentation) exportStart(info *SpanInfo, span *directSpan, parent *directParent, start time.Time) {
+	td := d.buildData(info, span, parent, start, time.Time{}, nil)
+	go func() {
```
apascal07, line 138: The start export runs on a goroutine while the end export is synchronous, so a start `Save` can land after the end `Save` for the same span. `TestOnlyTelemetryClient.Save` replaces `existing.Spans[id]` unconditionally (telemetry.go:97), so the completed span is overwritten by the in-progress copy: no `EndTime`, no `genkit:output`, empty `genkit:state`. The JS file store keeps a completed span over an incoming in-progress one; the Go client does not.

Reproduces deterministically: `GENKIT_ENABLE_REALTIME_TELEMETRY=true go test ./ai/` fails `TestWrapToolShortCircuitEmitsToolSpan` with `genkit:state = "", want "success"`. `genkit start` exports that variable to the child process, so `genkit start -- go test ./...` breaks go/ai, go/ai/exp, go/core and go/plugins/middleware.

go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,217 @@
...
+		if err := d.client.Save(context.Background(), td); err != nil {
```
apascal07, line 139: Nothing tracks this goroutine. The realtime processor this replaces kept a WaitGroup and a closed flag so ForceFlush/Shutdown could drain in-flight start saves; here a process that exits right after a run drops them, and a high-fan-out turn spawns one unbounded POST per span with no backpressure. The exported constructor also accepts a nil client, and the first span then panics here where `telemetryServerExporter` and `realtimeSpanProcessor` both returned quietly.

go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+// directParent carries the enclosing Direct span's ids down the context, so
+// child spans share the trace id and point their parentSpanId at it.
+type directParent struct {
+	traceID string
+	spanID  string
+}
+
+var directParentKey = base.NewContextKey[*directParent]()
```
apascal07, line 63: This key is shared by every `DirectTelemetryInstrumentation` instance, so two Direct providers in one chain corrupt each other's parentage. The inner instance reads the outer's `{traceID, spanID}` from ctx, reuses that trace id, stamps `ParentSpanID` with a span its own client never receives, and `buildData` skips `DisplayName`/`StartTime`/`EndTime` because `parent != nil`. The inner client ends up with a trace that has no root.

That chain is easy to get: `genkit start` exports `GENKIT_TELEMETRY_SERVER`, so `resolveDirectLocked` prepends an HTTP Direct, and anything that calls `ConfigureInstrumentation(NewDirectTelemetryInstrumentation(client))` (the documented use, and every test package in this PR) runs `[Direct(env), Direct(own)]`. `GENKIT_TELEMETRY_SERVER=http://127.0.0.1:1 go test ./core/` fails `TestRunWithContextNesting`: `inner-nested parent = 1ac8c2..., want the with-context step 229b1a...`.

Same root cause from the other side: when Direct is enabled mid-process (the CLI handshake arriving after a long-lived agent root span started), children find no `directParent` in ctx, mint fresh trace ids, and show in the Dev UI as unrelated roots. The dispatcher already knows the parent span; if it owned parentage, no provider would need a private context key.

go/core/tracing/instrumentation.go
```diff
@@ -0,0 +1,214 @@
...
+// (explicitly set, or read once from GENKIT_TELEMETRY_SERVER). The env fallback
+// mirrors the pre-refactor TracerProvider bootstrap. Caller holds
+// instrumentationMu.
+func resolveDirectLocked() Instrumentation {
+	if direct != nil {
+		return direct
+	}
+	url := devServerURL
+	if url == "" {
+		url = os.Getenv("GENKIT_TELEMETRY_SERVER")
```
apascal07, line 155: This env read runs on every span while the server is unset, because only a non-empty result is cached, and it runs inside `instrumentationMu`. Every action, step, model and tool span in a production process now pays a `syscall.Getenv` environ scan under a global mutex plus a chain slice allocation; `realtimeTelemetryActive` a few lines away is cached at init for exactly this reason.

It also means `ResetInstrumentation` cannot produce a clean chain: the next span re-reads the env. `GENKIT_TELEMETRY_SERVER=http://127.0.0.1:65123 go test ./core/tracing/` fails `TestRunInNewSpan_ExposesCompositeTraceInfo` (composite id is Direct's minted one, not the fake's) and `TestActiveInstrumentations_PrependsDirectWhenServerConfigured` (chain length 2, want 1), and every test span in go/ai, go/ai/exp, go/core and go/plugins/middleware POSTs to whatever URL the shell has.

`devServerURL` is dead state on the same path: it is only non-empty when `direct` is non-nil, and this function only runs when `direct` is nil, so `url := devServerURL` never observes a value. Reading the env once at init into `devServerURL` (or in the genkit dev bootstrap next to `EnableLogExport`), letting Reset clear it, and caching the resolved chain in an `atomic.Pointer` covers all three.

go/core/tracing/tracing.go
```diff
@@ -256,95 +231,78 @@ func RunInNewSpan[I, O any](
...
-	// debug, or the Dev UI export sink).
-	start := time.Now()
-	logDebug := logger.FromContext(ctx).Enabled(ctx, slog.LevelDebug)
-	if logDebug {
-		startArgs := []any{"name", metadata.Name}
-		if metadata.Type != "" {
-			startArgs = append(startArgs, "type", metadata.Type)
+		// Fire the telemetry callback the moment ids are known. Omit blank ids
+		// so an uninstrumented run does not look like a broken exporter.
+		if cb := telemetryCallback(ctx); cb != nil && sm.TraceInfo.TraceID != "" {
```
apascal07, line 247: Gating the callback on a non-empty id turns "no provider minted ids" into "the reflection server never learns about the run". `telemetryCb` in reflection.go is the only place `activeActions.Set` and the `X-Genkit-Trace-Id` / `X-Genkit-Span-Id` headers happen, so with the default chain over OTel's no-op provider the Dev UI Cancel button gets 404 from `handleCancelAction`, the early headers are absent, and `ActionRunResult.Telemetry.TraceID` and `details.traceId` come back empty.

That default chain is reachable from this PR train's own feature: `genkit start --useOtel` withholds `GENKIT_TELEMETRY_SERVER` and the handshake `telemetryServerUrl`, and Go has no autoexport wiring, so a Go app under it has no ids unless the user configured OTel by hand. Before this change `Tracer()` lazily installed an SDK provider, so ids always existed. reflection_test.go's new TestMain forces Direct package-wide, so `TestEarlyTraceIDTransmission` and `TestActionCancellation` no longer cover this path.

The ids drive framework mechanics (cancel registry, log correlation, error details), not only export. Minting Genkit ids in the dispatcher when no provider supplies them would keep those working regardless of backend. Either way, the cancel and header consequences belong in the PR body's behavior-change section.

go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+}
+
+// buildSpan reproduces the SpanData shape telemetryServerExporter.convertSpan
+// produced, so before/after Dev UI traces are structurally identical. A zero
+// end leaves EndTime at 0, which the telemetry server treats as in progress.
+func (d *DirectTelemetryInstrumentation) buildSpan(info *SpanInfo, span *directSpan, parent *directParent, start, end time.Time, runErr error) *SpanData {
+	sm := info.metadata
+	attrs := attributesToMap(sm.attributes())
+	for k, v := range info.Labels {
```
apascal07, line 161: `buildSpan` encodes the same span differently from the OTel exporter in three places, so the "byte-for-byte identical" claim does not hold and the Dev UI (Direct) and an OTel export of one run disagree:

- Labels are written after `attributesToMap(sm.attributes())`, so a `TelemetryLabels` key such as `genkit:state` (they arrive verbatim from the reflection request body) clobbers the real value here. In the OTel provider the deferred `SetAttributes(sm.attributes()...)` reasserts the genkit keys last.
- The start export uses the full `attributes()`, so an in-progress span ships `genkit:state: ""` where the old realtime `OnStart` (`startAttributes` + `inputAttributes`) emitted no state key at all.
- `exception.type` (line 198) is `%T`, but OTel's `RecordError` uses `typeStr`, which yields `github.com/x/pkg.MyErr` for a named value error type where `%T` yields `pkg.MyErr`.

One `SpanData` builder shared by `convertSpan` and this path would remove the drift instead of maintaining two writers of the wire shape.

go/core/action_test.go
```diff
@@ -201,25 +200,24 @@ func TestActionStreaming(t *testing.T) {
...
+	// The package-wide client captures every test's spans, so filter to this
+	// action's uniquely named span. Expect exactly one.
+	var got int
+	for _, s := range tc.Spans() {
+		if s.DisplayName == name {
+			got++
 		}
 	}
-	t.Fatalf("did not find trace named %q", name)
+	if got != 1 {
```
apascal07, line 218: The package-wide client is never reset, so this asserts on every run of the test in the binary. `go test -count=2 ./core/` fails with `got 2 spans named "test/TestTracing-inc", want 1`; the pre-change version returned on the first match and was count-insensitive. A per-run unique action name or a reset of `testTelemetryClient` fixes it. The rewrite also dropped the old trace-envelope assertion (`td.DisplayName == name`), so `buildData`'s root-only `DisplayName`/`StartTime`/`EndTime` logic now has no test.

go/core/tracing/instrumentation.go
```diff
@@ -0,0 +1,214 @@
...
+// directly. A future change promotes an exported, read-only view once the shape
+// settles.
+type SpanInfo struct {
+	// Labels are the raw TelemetryLabels set directly as span attributes.
+	Labels map[string]string
+
+	// metadata is the dispatcher-owned running metadata. In-package providers
+	// (OTel, Direct) encode it via its attributes()/startAttributes() methods.
+	// It is fully populated (state, output, error) only after next returns.
+	metadata *spanMetadata
```
apascal07, line 48: With `metadata` unexported, an out-of-package `Instrumentation` gets `Labels` and nothing else: no name, path, type, input, output, state or custom metadata. The package's own fake reads `fake.spans[0].metadata.Name`, which no external implementation can do, so the exported extension point cannot be used for a real provider yet.

Shipping `Instrumentation`, `SpanInfo` and `Span` as public API in a minor commits to a shape that #6345 must then extend on a type users already depend on. Either the read-only view lands here, or the interface stays unexported until it does. JS exposes `InstrumentationSpanInfo.metadata` as the exported `SpanMetadata` plus `metadataToAttributes`; Go diverges here without a Go-idiom reason.

go/core/tracing/tracing.go
```diff
@@ -50,17 +48,6 @@ func (e *markedError) Unwrap() error {
...
-	if errors.As(err, &me) {
-		me.marked = true
-		return me
-	}
-
-	return &markedError{error: err, marked: true}
-}
-
 // isErrorAlreadyMarked checks if an error has already been marked as a failure source
 func isErrorAlreadyMarked(err error) bool {
```
apascal07, line 52: `markErrorAsHandled` is gone and nothing else constructs a `markedError`, so `isErrorAlreadyMarked` is constant false. Both providers still carry the `!isErrorAlreadyMarked(err)` gate (otel_instrumentation.go:99, direct_instrumentation.go:189), and the Direct comment says it "mirrors the OTel provider exactly", so the next provider will copy a branch that cannot fire. Deleting `markedError`, `isErrorAlreadyMarked` and both guards leaves nothing behind; if first-failure-only marking is wanted (JS's `ignoreFailedSpan`), it belongs once in `runBody` where `IsFailureSource` is set.

go/core/tracing/otel_instrumentation.go
```diff
@@ -0,0 +1,104 @@
...
+		opts = append(opts, trace.WithAttributes(sm.inputAttributes()...))
+	}
+
+	// Read the provider the caller configured; do not install or manage one.
+	// With nothing configured this is OTel's no-op provider (empty ids).
+	tracer := otel.GetTracerProvider().Tracer("genkit-tracer", trace.WithInstrumentationVersion("v1"))
+	ctx, span := tracer.Start(ctx, sm.Name, opts...)
+	defer span.End()
+	// The deferred end write reasserts the full attribute set (including the
+	// run-determined output/state). Registered after span.End so it runs first.
```
apascal07, line 95: This marshals input, init and output (`base.JSONString` per key) into the span whether or not the span records. With the new default of no provider, `otel.GetTracerProvider()` is the no-op provider and every attribute is discarded, so a production process pays a full JSON marshal of the model request and response per model span, at start and again here, as garbage. In dev the `[Direct, OTel]` chain marshals the same payloads a second time in `buildSpan`. Gating the OTel work on `span.IsRecording()`, and marshaling once on `spanMetadata` for every provider, removes both.

go/genkit/reflection.go
```diff
@@ -648,18 +648,14 @@ func handleCancelAction(activeActions *activeActionsMap) func(w http.ResponseWri
...
-		// `genkit start` sets GENKIT_ENABLE_REALTIME_TELEMETRY so traces stream to
-		// the dev UI as spans start, not just when they end (which, for a
-		// long-lived agent connection, is only when it closes).
-		realtime := os.Getenv("GENKIT_ENABLE_REALTIME_TELEMETRY") == "true"
-		if realtime {
-			tracing.WriteTelemetryRealtime(client)
-		} else {
-			tracing.WriteTelemetryImmediate(client)
-		}
+		// Feed the Dev UI through the Direct instrumentation (axis 1) rather
```
apascal07, line 651: "axis 1" and "axis 2" are labels from the PR write-up and appear nowhere in the code, and "rather than registering an OTel span processor" describes the deleted version. Same pattern at tracing.go:81, direct_instrumentation.go:39, instrumentation.go:103 and 147 ("mirrors the pre-refactor TracerProvider bootstrap"), and the package doc of go/samples/basic-flows/main.go, which explains that the sample verifies this refactor rather than what a user learns from it. A reader in six months has no axes to look up and no memory of the old wiring; the comments should describe the final state.

go/core/tracing/direct_instrumentation.go
```diff
@@ -0,0 +1,217 @@
+// Copyright 2025 Google LLC
```
apascal07, line 1: nit: the six new files carry `Copyright 2025` but were created in 2026; go/GEMINI.md's header template says to update the year, and every other file added to go/ this year uses 2026. go/samples/basic-flows/main.go also lacks the `SPDX-License-Identifier` line the template ends with.

apascal07 (changes requested):

PR #6344, opened 2026-09-14
feat(js/plugins/google-genai): New model support and Interactions API support for some gemini models

ifielker:
Support for some models (not all the new ones, just from July, newer ones will be added in future PRs). This adds Interactions API support to newer and -latest googleai/gemini models. Older models will still use generateContent to prevent unexpected breakages with different outputs for "custom/raw" fields. 

Similar support for vertexai/GEAP will follow in the next PR.


js/plugins/google-genai/tests/googleai/gemini_test.ts
```diff
@@ -284,11 +284,13 @@ describe('Google AI Gemini', () => {
             urlContext: {},
           },
         };
+
         await model.run(request);
 
         const apiRequest: GenerateContentRequest = JSON.parse(
           fetchStub.lastCall.args[1].body
         );
+        console.log('apiRequest: ' + apiRequest);
```
pavelgj, line 293: Stray debug log, let's drop it. Also fwiw `'apiRequest: ' + apiRequest` would just print `[object Object]`.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -701,6 +785,7 @@ export function defineModel(
 
       const modelVersion = request.config?.version || extractVersion(ref);
       const isGemma = isGemmaModelName(modelVersion);
+      const useInteractions = modelInteractionsMap.get(modelVersion) ?? true;
```
pavelgj, line 788: Is defaulting unknown models to the Interactions API (`?? true`) intentional? Any model string not in the known map (user-specified or a fresh preview name) will route through interactions, so if that model doesn't support the endpoint it'd fail at runtime. Just want to confirm that's deliberate vs falling back to generateContent.

js/plugins/google-genai/src/googleai/interaction-converters.ts
```diff
@@ -240,18 +559,34 @@ export function toInteractionSteps(messages: MessageData[]): Step[] {
...
         ) {
-          output = { result: output };
+          result = { result: result };
         }
+
         steps.push({
           type: 'function_result',
           name: part.toolResponse.name,
-          result: RecordUnknownOrStringSchema.optional().parse(output),
+          result: RecordUnknownOrStringOrArraySchema.parse(result ?? {}),
```
pavelgj, line 589: When `toolResponse.content` is present we map each part via `toInteractionContent` (which can produce audio/video/document), but the array branch of `RecordUnknownOrStringOrArraySchema` only allows image/text, so a tool returning e.g. a PDF part would fail the zod parse and throw here. Might want to broaden the array schema or filter to image/text before pushing.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -859,6 +1016,97 @@ export function defineModel(
...
+        }
+
+        const req: CreateInteractionRequest = {
+          system_instruction: interactionsSystemInstruction,
+          model: extractVersion(ref),
+          tools: interactionsTools.length ? interactionsTools : undefined,
+          generation_config: interactionGenerationConfig,
+          stream: streamingRequested,
+          input: toInteractionSteps(newMessages),
+          service_tier: serviceTier as ServiceTier,
```
pavelgj, line 1050: The `as ServiceTier` cast hides any drift between the config enum and the interactions `ServiceTier`. If the config zod enum already produces the right union, we could reconcile the types and drop the cast.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -562,31 +629,48 @@ const KNOWN_GEMMA_MODELS = {
...
 } as const;
 export type KnownGemmaModels = keyof typeof KNOWN_GEMMA_MODELS;
 export type GemmaModelName = `gemma-${string}`;
 export function isGemmaModelName(value: string): value is GemmaModelName {
   return value.startsWith('gemma-');
 }
 
+const DEPRECATED_MODELS = {
+  // When models are < 1 month from shutdown, move them here instead.
+  // They will still be instatiated with the correct options,
```
pavelgj, line 650: nit: typo `instatiated` -> `instantiated` (same comment is duplicated in imagen.ts).

js/plugins/google-genai/src/googleai/interaction-converters.ts
```diff
@@ -106,6 +121,274 @@ export function toInteractionTool(tool: ToolDefinition): InteractionTool {
...
+  const restConfig = { ...gs };
+  delete restConfig.searchTypes;
+  delete restConfig.search_types;
+
+  Object.assign(result, toSnakeCaseObj(restConfig));
+
+  return result;
+}
+
+export function toInteractionResponseModalities(
```
pavelgj, line 344: nit: return is annotated `ResponseModality[]` but `m.toLowerCase()` yields `string[]`. Type-checks today only because nothing narrows it; a validated map (or explicit cast) would be more honest.

js/plugins/google-genai/src/googleai/utils.ts
```diff
@@ -265,3 +265,40 @@ export function applyContextOverrides(
...
 }
+
+/**
+ * Converts a string in camelCase to lowercase snake_case.
+ */
+export function camelToSnakeCase(str: string): string {
+  return str.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`);
+}
+
+export function convertObjectKeysToSnakeCase(
```
pavelgj, line 276: The return type `Record<string, unknown> | unknown` collapses to just `unknown`, which pushes casting onto callers (e.g. `as Record<string, unknown>` in toInteractionGenerationConfig). Consider returning `unknown` plainly, or overloads, so the intent is clearer.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -420,11 +440,16 @@ type ConfigSchemaType =
...
+const modelInteractionsMap = new Map<string, boolean>();
+
 function commonRef(
   name: string,
   info?: ModelInfo,
-  configSchema: ConfigSchemaType = GeminiConfigSchema
+  configSchema: ConfigSchemaType = GeminiConfigSchema,
+  useInteractions: boolean = true
 ): ModelReference<ConfigSchemaType> {
+  modelInteractionsMap.set(name, useInteractions);
```
pavelgj, line 451: Setting `modelInteractionsMap` as a side effect of building a ref, then reading it later from a module global, is a bit surprising. Attaching the flag to the model ref/metadata would be more self-contained, though I get the current shape is expedient.

js/plugins/google-genai/src/googleai/interaction-converters.ts
```diff
@@ -346,6 +681,128 @@ export function toInteractionSteps(messages: MessageData[]): Step[] {
...
+        {
+          toolRequest: {
+            name: delta.name,
+            ref: delta.id,
+            input: delta.arguments || {},
+            partial: true,
+          },
+        },
+      ];
+    case 'arguments_delta':
```
pavelgj, line 730: For streamed function calls, `arguments_delta` returns `[]` and the `function_call` delta emits `partial: true` with empty input, so partial args never reach `sendChunk` (they only land in the final assembled response). Flagging in case surfacing progressive args during streaming was intended.

pavelgj (commented):
Nice work wiring up Interactions. Left a bunch of mostly-questions inline, nothing blocking. One general thing: for the models that now go through the Interactions API the `raw`/`custom` payload changes shape (interaction object with `steps` instead of `candidates[].groundingMetadata`, as the testapp edits show). Is it worth a short README note on which models use Interactions + what `raw` looks like now, plus a release-note callout, and maybe a matching docsite update?

js/plugins/google-genai/src/googleai/gemini.ts
ifielker: Yes deliberate. New models should all go through Interactions API. generateContent is only for known models that have been going through generateContent since the beginning (to avoid breaking changes). We are phasing out generateContent as those known models are deprecated. Eventually it will only be Interactions for everything.

js/plugins/google-genai/tests/googleai/gemini_test.ts
ifielker: done

js/plugins/google-genai/src/googleai/interaction-converters.ts
```diff
@@ -197,7 +481,42 @@ export function toInteractionRole(role: MessageData['role']): string {
...
+
+const FunctionResultArraySchema = z.array(
+  z.union([ImageContentSchema, TextContentSchema])
+);
+
+const RecordUnknownOrStringOrArraySchema = z.union([
+  RecordUnknownSchema,
+  z.string(),
+  FunctionResultArraySchema,
+]);
```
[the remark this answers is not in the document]
ifielker: I've looked in the backend and anything other than text and image arrays for FunctionResults are not allowed. It's all nice for us to support plain arrays for tools in Genkit, but if Gemini doesn't, then... that's not useful.
The other types are for things like CodeExecutionResult or GoogleSearchResult.

js/plugins/google-genai/src/googleai/interaction-converters.ts
ifielker: The API only accepts image/text. If they are trying to use a tool that returns a PDF, I think throwing is the correct thing to do. As a developer I would much rather get "Invalid tool output error" than some made up response from the model that *doesn't take the PDF into consideration* and has no errors, because I would think that it did read the PDF and it's response is based on that... which is much worse.

js/plugins/google-genai/src/googleai/interaction-converters.ts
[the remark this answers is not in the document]
ifielker: Plain arrays are not supported by the API we are calling.

js/plugins/google-genai/src/googleai/gemini.ts
ifielker: The zod type is basically the three strings that are currently valid + string. The 'as ServiceTier' handles "the API added a new valid serviceTier and we haven't released since then" case (like passthrough). So we could either just forget ServiceTier altogether and go with 'string'? Or leave it as is? Or enforce the three valid ones? Other?

js/plugins/google-genai/src/googleai/gemini.ts
ifielker: done.

js/plugins/google-genai/src/googleai/interaction-converters.ts
ifielker: string[] is assignable to ResponseModality[] though... it's defined like: 
export declare type ResponseModality =
  | 'text'
  | 'image'
  | 'audio'
  | (string & {});  // <- This allows string[] to be ResponseModality[]... 
It was my understanding that we wanted the open ResponseModality type to allow other strings when new ResponseModalities show up in the API...?

js/plugins/google-genai/src/googleai/utils.ts
ifielker: done (overloads).

js/plugins/google-genai/src/googleai/gemini.ts
ifielker: I think I did it this way because I didn't want to surface the 'useInteractions' in the visible options, and I think that's the only way to get it into the metadata. The reason for not surfacing it is two fold: 1) People should not need to worry about how they are calling the model. We will default to the correct thing for them. We will also not break backwards compatibility randomly. 2) I don't want the old generateContent to be hanging around longer than necessary just because "someone might have specified it, so we need to keep it". So we define useInteractions internally instead so new models get the new functionality and old models keep the old functionality and when all the old models are shut down, we can remove generateContent code safely.

js/plugins/google-genai/src/googleai/interaction-converters.ts
ifielker: I did not intend to stream progressive args. If we want that we can change it later.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -860,6 +1030,97 @@ export function defineModel(
...
+              previousInteractionId = prevId;
+              newMessages = messages.slice(i + 1);
+              break;
+            }
+          }
+        }
+
+        const req: CreateInteractionRequest = {
+          system_instruction: interactionsSystemInstruction,
+          model: extractVersion(ref),
```
pavelgj, line 1059: The interactions request ignores `config.version`. A few lines up we compute `const modelVersion = request.config?.version || extractVersion(ref)` and use that both to pick the code path and for the generateContent call, but here we send `extractVersion(ref)`. So `googleAI.model('gemini-flash-latest')` with `config: { version: 'gemini-3.6-flash' }` decides "interactions" based on one name and then calls the API with the other.

```suggestion
          model: modelVersion,
```

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -860,6 +1030,97 @@ export function defineModel(
...
+              typeof prevId === 'string'
+            ) {
+              previousInteractionId = prevId;
+              newMessages = messages.slice(i + 1);
+              break;
+            }
+          }
+        }
+
+        const req: CreateInteractionRequest = {
```
pavelgj, line 1057: `safetySettings`, `toolConfig` and `retrievalConfig` are destructured out of the config above but only ever make it into `generateContentRequest`, so on the interactions path they're silently dropped. Concretely: `config.safetySettings` on `gemini-flash-latest` now does nothing, no error. The testapp hits this too, the `combine tools and builtins` flow sets `toolConfig: { includeServerSideToolInvocations: true }` and now runs on `gemini-3.6-flash`, so that toolConfig is a no-op.

If the Interactions API has no equivalent, can we throw (same as the non-text system message check) or at least `logger.warn`? Silently ignoring safety settings feels like the one we really don't want.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -720,17 +805,37 @@ export function defineModel(
       // systemInstructions to be provided as a separate input. The first
       // message detected with role=system will be used for systemInstructions.
       let systemInstruction: GeminiMessage | undefined = undefined;
+      let interactionsSystemInstruction: string | undefined = undefined;
       const systemMessage = messages.find((m) => m.role === 'system');
       if (systemMessage) {
         messages.splice(messages.indexOf(systemMessage), 1);
         systemInstruction = toGeminiSystemInstruction(systemMessage);
+        if (useInteractions) {
+          if (systemMessage.content.some((c) => !c.text)) {
```
pavelgj, line 814: `!c.text` is also true for `{ text: '' }`, so an empty-string text part throws here. `c.text === undefined` is what we want. Also the nested `if (useInteractions) { if (...) }` could collapse into one condition.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -860,6 +1030,97 @@ export function defineModel(
...
+          service_tier: serviceTier as ServiceTier,
+        };
+
+        if (jsonMode) {
+          req.response_format = {
+            type: 'text',
+            mime_type: 'application/json',
+          };
+          if (request.output?.constrained) {
+            req.response_format.schema = cleanSchema(request.output.schema);
```
pavelgj, line 1073: The generateContent path only runs `cleanSchema` when `pluginOptions.legacyResponseSchema` is set, otherwise it passes the schema through untouched as `responseJsonSchema`. Here we always clean, which strips `additionalProperties` and collapses nullable unions. If `response_format.schema` takes full JSON Schema, can we pass `request.output.schema` straight through?

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -860,6 +1030,97 @@ export function defineModel(
...
+        }
+
+        const req: CreateInteractionRequest = {
+          system_instruction: interactionsSystemInstruction,
+          model: extractVersion(ref),
+          tools: interactionsTools.length ? interactionsTools : undefined,
+          generation_config: interactionGenerationConfig,
+          stream: streamingRequested,
+          input: toInteractionSteps(newMessages),
+          service_tier: serviceTier as ServiceTier,
```
pavelgj, line 1064: Re your question on the earlier thread: my vote is to make `ServiceTier` open the same way `ResponseModality` already is, `'flex' | 'standard' | 'priority' | (string & {})`, and drop the cast. You keep autocomplete and forward compat without the type lying.

Related: `isServiceTier` is exported from `interaction-types.ts` but never used anywhere, so either it gets used here or it can go.

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -836,8 +974,40 @@ export function defineModel(
...
+      const sanitizedConfigOptions = {
         ...removeClientOptionOverrides(restOfConfigOptions),
+      };
+
+      const interactionGenerationConfig: ModelGenerationConfig =
+        toInteractionGenerationConfig(sanitizedConfigOptions);
+
+      if (useInteractions) {
+        if (functionCallingConfig) {
+          interactionGenerationConfig.tool_choice = {
```
pavelgj, line 986: Question: `functionCallingConfig.mode?.toLowerCase()` can produce `'mode_unspecified'`, and a plain `toolChoice: 'none'` passes straight through below. Does `allowed_tools.mode` accept those values, or should `MODE_UNSPECIFIED` mean omitting `tool_choice` entirely?

js/plugins/google-genai/src/googleai/gemini.ts
```diff
@@ -752,6 +857,8 @@ export function defineModel(
         tools: toolsFromConfig,
         retrievalConfig,
         serviceTier,
+        previousInteractionId: previousInteractionIdFromConfig,
```
pavelgj, line 860: nit: `previousInteractionId` isn't declared in `GeminiConfigSchema`, it only works because the schema is `.passthrough()`. That means it's invisible in the DevUI and untyped for callers. `deep-research.ts` and `antigravity.ts` both declare it explicitly, worth matching.

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -347,6 +367,258 @@ export function cleanSchema(schema: JSONSchema): JSONSchema {
...
+    throw new Error('Error processing stream because response.body not found');
+  }
+  const inputStream = response.body.pipeThrough(
+    new TextDecoderStream('utf8', { fatal: true })
+  );
+  const responseStream = getInteractionResponseStream(inputStream);
+  const [stream1, stream2] = responseStream.tee();
+  return {
+    stream: generateInteractionResponseSequence(stream1),
+    response: getInteractionResponsePromise(stream2),
```
pavelgj, line 383: `processStream` right below deliberately attaches a no-op catch to the teed response promise (#5435 / #6170): if the caller only consumes `stream` and it errors before `response` is awaited, the rejection is unhandled and takes the process down. That's exactly the shape in `gemini.ts`, the `for await` throws on an `error` event and `result.response` is never awaited.

```suggestion
  const responsePromise = getInteractionResponsePromise(stream2);
  // See processStream: avoid an unhandled rejection if the caller only
  // consumes `stream` and it errors before `response` is ever awaited.
  responsePromise.catch(() => {});
  return {
    stream: generateInteractionResponseSequence(stream1),
    response: responsePromise,
  };
```

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -347,6 +367,258 @@ export function cleanSchema(schema: JSONSchema): JSONSchema {
...
+      let currentText = '';
+      return pump();
+      function pump(): Promise<(() => Promise<void>) | undefined> {
+        return reader
+          .read()
+          .then(({ value, done }) => {
+            if (done) {
+              reader.releaseLock();
+              if (currentText.trim()) {
+                controller.error(new Error('Failed to parse stream'));
```
pavelgj, line 402: `parseStreamErrorText` lives a few lines up in this same file and exists for exactly this case: the API sometimes returns HTTP 200 with a plain JSON error body instead of SSE frames, and we want a `GenkitError` with the right status so retry middleware can react. A generic `Error` here loses that.

```suggestion
                controller.error(parseStreamErrorText(currentText));
```

pavelgj (commented):
A few comments/questions

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -347,6 +367,258 @@ export function cleanSchema(schema: JSONSchema): JSONSchema {
...
+        if (partialArgumentsMap.has(value.index)) {
+          const step = interaction.steps?.[value.index];
+          if (step && step.type === 'function_call') {
+            try {
+              const argStr = partialArgumentsMap.get(value.index);
+              if (argStr) {
+                step.arguments = JSON.parse(argStr);
+              }
+            } catch (e) {
+              console.warn(
```
pavelgj, line 607: `logger.warn` from `genkit/logging`, everything else in the plugin uses the logger.

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -27,6 +27,12 @@ import {
   z,
 } from 'genkit';
 import { GenerateRequest } from 'genkit/model';
+import { toJsonSchema } from 'genkit/schema';
+import {
+  GeminiInteraction,
+  InteractionSseEvent,
+  InteractionStreamResult,
+} from '../googleai/interaction-types.js';
```
pavelgj, line 35: `common/` now depends on a provider-specific module, and `vertexai/` has its own `interaction-types.ts` with divergent shapes. Since the GEAP PR is next, this is going to need untangling right away. Either keep the SSE/interaction stream plumbing in `googleai/`, or lift the shared interaction types up into `common/` first.

js/plugins/google-genai/src/common/utils.ts
```diff
@@ -347,6 +367,258 @@ export function cleanSchema(schema: JSONSchema): JSONSchema {
   return out;
 }
 
+export function interactionProcessStream(
```
pavelgj, line 370: There aren't any tests for `interactionProcessStream`, which is the most intricate new code in the PR: frame splitting, `step.delta` assembly into `steps[index]`, `arguments_delta` accumulation then JSON parse at `step.stop`, and the `error` event. `common/utils_test.ts` has pretty thorough `processStream` coverage to model on. Even just three cases would help: happy path with text deltas, an `error` event, and one `arguments_delta` reassembly.

js/plugins/google-genai/src/googleai/interaction-converters.ts
```diff
@@ -346,6 +681,128 @@ export function toInteractionSteps(messages: MessageData[]): Step[] {
...
+      };
+      if (
+        (delta.type === 'image' || delta.type === 'video') &&
+        delta.resolution !== undefined
+      ) {
+        part.metadata = { resolution: delta.resolution };
+      }
+      return [part];
+    }
+    case 'thought_summary':
```
pavelgj, line 710: `fromInteractionContent` on a `TextContent` returns `{ text }`, but the final assembled response maps the `thought` step through `fromThoughtContent` into `{ reasoning }`. So with `includeThoughts: true` the chain of thought streams out as `chunk.text` and then disappears from the final message text. These two should agree, the delta should emit `{ reasoning }` (plus signature metadata) too.

js/plugins/google-genai/src/googleai/interaction-converters.ts
pavelgj, line 589: Following up on the earlier thread: I'm sold on throwing rather than silently dropping a PDF tool output. But a bare Zod parse error here won't tell the developer anything actionable. Could we catch and rethrow as a `GenkitError({ status: 'INVALID_ARGUMENT', message: 'Tool output for <name> may only contain text or image content.' })`? Then the error actually says what you argued it should say.

js/plugins/google-genai/src/googleai/interaction-types.ts
```diff
@@ -63,7 +74,36 @@ export declare interface InteractionMcpServerTool {
   name?: string;
   url?: string;
   headers?: Record<string, string>;
-  allowed_tools?: string[];
+  allowed_tools?: InteractionAllowedTools[];
```
pavelgj, line 77: This went from `string[]` to an array of `{ mode, tools }`. Is that actually the MCP wire shape, or did it get swept along with the `tool_choice` work?

js/plugins/google-genai/src/googleai/embedder.ts
```diff
@@ -111,8 +120,8 @@ export function model(
 ): EmbedderReference<ConfigSchemaType> {
   const name = checkModelName(version);
 
-  if (isKnownKey(name, KNOWN_MODELS)) {
-    const known = KNOWN_MODELS[name];
+  if (isKnownKey(name, ALL_MODELS)) {
```
pavelgj, line 123: `model()` resolves against `ALL_MODELS` now, but `listKnownModels` below still iterates `KNOWN_MODELS`, so `gemini-embedding-001` is resolvable but not pre-defined at init. `gemini.ts` and `imagen.ts` both switched to `ALL_MODELS` for that, worth making consistent.

Separately, `gemini-embedding-001` is still used across the testapps (`rag`, `evals`, `flow-simple-ai`). If it's a month from shutdown those are worth bumping to `gemini-embedding-2` here.

js/plugins/google-genai/tests/googleai/gemini_test.ts
```diff
@@ -422,6 +423,152 @@ describe('Google AI Gemini', () => {
...
+        mockFetchResponse(defaultApiResponse);
+        const request: GenerateRequest<typeof GeminiImageConfigSchema> = {
+          ...minimalRequest,
+          config: {
+            google_search: {},
+          },
+        };
+        await model.run(request);
+
+        const apiRequest: any = JSON.parse(fetchStub.lastCall.args[1].body);
```
pavelgj, line 440: nit: the existing tests in this file type this as `GenerateContentRequest`. `CreateInteractionRequest` would work the same way for the new ones and keep `any` out (same in the other new Interactions tests below).

js/plugins/google-genai/src/common/utils.ts
ifielker: Actually, the 're-unify googleai and vertexai (or whatever we are calling it today)' PR is next, then the vertex/GEAP/GE gets Interactions one.

js/plugins/google-genai/src/googleai/gemini.ts
ifielker: done

PR #6345, opened 2026-09-14
feat(go/plugins/otel): add OpenTelemetry GenAI plugin and upgrade otel deps

pavelgj:
## What

Adds `github.com/firebase/genkit/go/plugins/otel`, an OpenTelemetry GenAI
semantic-conventions instrumentation provider for Genkit Go, and upgrades the
OpenTelemetry dependencies to make it work. This is the Go counterpart of the
JS `@genkit-ai/otel` plugin (#6340).

It plugs into the pluggable instrumentation abstraction from the base PR via
`tracing.ConfigureInstrumentation(...)`, composing with the built-in dev
instrumentation that feeds the Developer UI.

```go
import (
	"go.opentelemetry.io/otel"
	genaiotel "github.com/firebase/genkit/go/plugins/otel"
)

// The app owns the OTel SDK (point exporters at your OTLP collector).
otel.SetTracerProvider(/* ... */)

tracing.ConfigureInstrumentation(genaiotel.NewGenAiInstrumentation(genaiotel.GenAiInstrumentationOptions{
	ContentCapturingMode: genai.SpanOnly, // default NoContent; env var honored when unset
	EmitToolSpans:        true,
}))

g := genkit.Init(ctx, genkit.WithPlugins(&googlegenai.GoogleAI{}))
```

## What it emits

Maps Genkit actions to the [OTel GenAI semantic conventions](https://github.com/open-telemetry/semantic-conventions-genai):

- Model actions become `chat <model>` CLIENT spans with `gen_ai.*` attributes
  (provider, model, request config, response finish reasons, token usage
  including reasoning/cache tokens).
- Tool actions optionally become `execute_tool <name>` spans (`EmitToolSpans`).
- Every other action type becomes a generic span tagged `genkit.action.type`,
  so the trace tree stays connected.
- The two spec metrics `gen_ai.client.token.usage` and
  `gen_ai.client.operation.duration` per model call.
- Optional message-content capture (`gen_ai.input.messages` /
  `gen_ai.output.messages` / `gen_ai.system_instructions`), gated by
  `ContentCapturingMode` (spec env var `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`
  honored when unset). SPAN / EVENT / SPAN_AND_EVENT modes; EVENT emits the
  `gen_ai.client.inference.operation.details` log record.

The app owns the OTel SDK; when no SDK is configured the provider is inert
(no-op spans/instruments), matching the JS plugin's contract.

## Extension surface

`SpanInfo` gains read-only accessors (`Name`/`Type`/`Subtype`/`Input`) so an
out-of-package provider can encode a span without reaching the unexported
metadata. Output is deliberately read from the value `next` returns.

## Dependency upgrade

OpenTelemetry Go `v1.36.0 -> v1.46.0` (+ `otel/log` `v0.22.0`, `sdk/log`),
adds `contrib/exporters/autoexport` and the OTLP exporters. The bump is
required because `--use-otel` drives the app SDK with
`OTEL_EXPORTER_OTLP_*_PROTOCOL=http/json`, which the Go OTLP HTTP exporter only
supports as of v1.46.0. The only code impact from the bump was the `otel/log`
value-constructor API change; all 48 Go packages build and test clean.

## Sample

`go/samples/otel-genai` shows the wiring with an env-driven SDK setup (the Go
analog of JS `new NodeSDK()`), runnable against the local Jaeger + collector
stack (`scripts/local-telemetry.ts`) or `genkit start --use-otel`.

## Tests

Unit tests for the pure mappers (attributes, messages) and the instrumentation
itself (in-memory `SpanRecorder` / `ManualReader` driven through
`RunInNewSpan`): model/tool/generic spans, content modes, action IO, metrics,
error path, and nil-safety of the new accessors/mappers.

## Stack

- Base: #6343 (pluggable, OTel-decoupled telemetry instrumentation)
- JS counterpart: #6340


go/core/tracing/instrumentation.go
```diff
@@ -48,6 +47,24 @@ type SpanInfo struct {
...
+func (i *SpanInfo) Name() string { return i.metadata.Name }
+
+// Type is the Genkit span type ("action", "flowStep", "util", ...).
+func (i *SpanInfo) Type() string { return i.metadata.Type }
+
+// Subtype is the finer categorization ("model", "tool", "flow", ...), or "".
+func (i *SpanInfo) Subtype() string { return i.metadata.Subtype }
+
+// Input is the raw Genkit input the action was invoked with, before next runs.
+func (i *SpanInfo) Input() any { return i.metadata.Input }
```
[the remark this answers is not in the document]
pavelgj: Done in c6b49ef1a. These are exported methods on an exported type now, so a zero-value *SpanInfo (from an out-of-package provider or a test) should not panic. Added nil-receiver + nil-metadata guards to Name/Type/Subtype/Input, plus a TestSpanInfoAccessorsNilSafe.

go/plugins/otel/instrumentation.go
```diff
@@ -0,0 +1,185 @@
...
+	return g.metrics
+}
+
+// RunInNewSpan dispatches on the Genkit action type and encodes the span using
+// the OTel GenAI conventions.
+func (g *GenAiInstrumentation) RunInNewSpan(ctx context.Context, info *tracing.SpanInfo, next tracing.NextFunc) (any, error) {
+	actionType := info.Subtype()
+	if actionType == "" {
+		actionType = info.Type()
+	}
```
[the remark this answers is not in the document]
pavelgj: Done in c6b49ef1a: `if info == nil { return next(ctx, nil) }` at the top, so a nil info keeps the chain intact rather than panicking.

go/plugins/otel/genai/messages.go
```diff
@@ -0,0 +1,138 @@
...
+			m["id"] = tr.Ref
+		}
+		return m
+	case part.IsToolResponse():
+		tr := part.ToolResponse
+		m := map[string]any{"type": "tool_call_response", "response": tr.Output}
+		if tr.Ref != "" {
+			m["id"] = tr.Ref
+		}
+		return m
```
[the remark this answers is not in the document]
pavelgj: Good catch, taken in c6b49ef1a. MapPart maps potentially malformed/external content, so a part whose kind is tool-request/response but whose pointer is nil now returns a safe empty shape instead of panicking. Added TestMapPartNilPointers.

go/plugins/otel/attributes.go
```diff
@@ -0,0 +1,89 @@
...
+
+// resolveFinishReasons maps the response finish reason to the GenAI vocabulary.
+// A turn ending in tool calls reports tool_calls, following the OpenAI GenAI
+// profile: it is the more informative signal for consumers.
+func resolveFinishReasons(response *ai.ModelResponse, failed bool) []string {
+	if msg := resolveMessage(response); msg != nil && genai.HasToolRequestPart(msg.Content) {
+		return []string{"tool_calls"}
+	}
+	return []string{genai.MapFinishReason(string(response.FinishReason), failed)}
+}
```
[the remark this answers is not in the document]
pavelgj: Done in c6b49ef1a. Added a nil-response guard returning MapFinishReason("", failed). Callers currently always pass non-nil, but the helper is now self-safe.

go/plugins/otel/content.go
```diff
@@ -0,0 +1,117 @@
...
+	var input *genai.NormalizedMessages
+	if request != nil {
+		nm := genai.NormalizeMessages(request.Messages)
+		input = &nm
+	}
+	var outputMessages []map[string]any
+	if msg := resolveMessage(response); msg != nil {
+		reason := resolveFinishReasons(response, false)[0]
+		outputMessages = append(outputMessages, genai.MapOutputMessage(msg, reason))
+	}
```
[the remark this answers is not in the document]
pavelgj: Covered by the nil-response guard on resolveFinishReasons (it always returns exactly one element), so the index is provably safe. Added a comment noting the invariant rather than a redundant length check.

go/plugins/otel/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+	b, err := json.Marshal(config)
+	if err != nil {
+		return nil
+	}
+	var c ai.GenerationCommonConfig
+	if err := json.Unmarshal(b, &c); err != nil {
+		return nil
+	}
+	return &c
+}
```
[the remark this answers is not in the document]
pavelgj: Going to keep the JSON round-trip here. The suggestion assumes provider configs embed ai.GenerationCommonConfig, but the primary provider config is genai.GenerateContentConfig (from google.golang.org/genai), which does not embed it: it is a separate struct with matching JSON tags and *float32 fields. So the FieldByName("GenerationCommonConfig") fast-path would never match the real provider and would always fall through to JSON anyway. The round-trip is what actually makes extraction work across providers, and it runs once per generate call right before an LLM network round-trip, so the marshal cost is negligible.

go/plugins/otel/instrumentation.go
```diff
@@ -0,0 +1,189 @@
...
+		return next(ctx, nil)
+	}
+	actionType := info.Subtype()
+	if actionType == "" {
+		actionType = info.Type()
+	}
+	switch actionType {
+	case "model":
+		return g.runModelSpan(ctx, info, next)
+	case "tool":
```
apascal07, line 179: `case "tool"` never matches. `DefineTool` registers actions as `api.ActionTypeToolV2` (`ai/tools.go:363`), `core/action.go:298` sets the span subtype to `string(a.desc.Type)`, and `ai/generate.go:1049` uses the same value for the short-circuit span, so every real tool arrives here as `"tool.v2"` and falls through to `runGenericSpan`. With `EmitToolSpans: true` no `execute_tool <name>` span is emitted and there is no `gen_ai.tool.name`, contrary to the README. `googlecloud.go:389` already matches both spellings. The test at `instrumentation_test.go:275` passes only because it fabricates `Subtype: "tool"`, which no Genkit code path produces. Switching on `string(api.ActionTypeTool)` and `string(api.ActionTypeToolV2)` (`core/api` does not import `tracing`) also turns a future rename into a compile error.

go/plugins/otel/spans.go
```diff
@@ -0,0 +1,128 @@
...
+	start := time.Now()
+	ctx, span := g.tracer().Start(ctx, genai.OperationChat+" "+model,
+		oteltrace.WithSpanKind(oteltrace.SpanKindClient),
+		oteltrace.WithAttributes(attrs...))
+	defer span.End()
+	g.maybeWarnNotRecording(span)
+
+	out, err := next(ctx, spanHandle(span))
+	if err != nil {
+		g.recordError(span, err)
```
apascal07, line 66: The error branch discards `out`, but `next` can return a result alongside the error: `core/action.go:281` returns the output together with the output-validation error, and `Generate` documents the partial `(resp, err)` contract. The base dispatcher records `sm.Output` on the error path for exactly that case. Here a Gemini call that spends 2100 tokens and then fails output-schema validation gets `error.type` on the span and on the duration histogram, but no `gen_ai.usage.*` attributes, no finish reason, no content, and `recordModelMetrics` receives `response=nil`, so `gen_ai.client.token.usage` never sees the billed tokens. `runToolSpan` and `runGenericSpan` skip `maybeCaptureActionIO` on error too, so `CaptureActionIO` yields `genkit.input` but no `genkit.output` on the spans someone is actually debugging. Side effect: `addResponseAttributes` is only ever called with `failed=false`, so the `error` branch of `genai.MapFinishReason` is unreachable.

go/plugins/otel/instrumentation.go
```diff
@@ -0,0 +1,189 @@
...
+//
+// Wire it up with [tracing.ConfigureInstrumentation]:
+//
+//	tracing.ConfigureInstrumentation(otel.NewGenAiInstrumentation(otel.GenAiInstrumentationOptions{
+//		ContentCapturingMode: otel.SpanOnly,
+//		EmitToolSpans:        true,
+//	}))
+//	g := genkit.Init(ctx, genkit.WithPlugins(&googlegenai.GoogleAI{}))
+//
+// It composes with the built-in dev instrumentation, which feeds the Developer
```
apascal07, line 33: This holds for the dev instrumentation but not for the default one. `activeInstrumentations` (`core/tracing/instrumentation.go:172-183`) uses the configured provider instead of `defaultOTel`, and only `OTelInstrumentation` writes `genkit:type`, `genkit:path`, `genkit:state`, `genkit:input` and `genkit:output`. `googlecloud.go:355-364` returns early when `genkit:type` is absent, so an app that adds the README one-liner to an existing googlecloud or firebase deployment loses every feature, action, generate, path and engagement metric with no error or warning. The redaction pass there (`googlecloud.go:405-437`) only rewrites `genkit:input` / `genkit:output`, so with `CaptureActionIO` or a `SPAN_*` mode the raw `genkit.input` and `gen_ai.input.messages` payloads reach Cloud Trace on a configuration whose stated purpose is to keep them out. Either the chain keeps the default provider alongside a configured one (a base-PR change), or this provider re-emits the `genkit:*` contract, or the README and this paragraph say the two are mutually exclusive.

go/plugins/otel/spans.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+// runModelSpan opens a gen_ai chat CLIENT span for a model action, records the
+// request config and (after next) the response attributes, content, and metrics.
+func (g *GenAiInstrumentation) runModelSpan(ctx context.Context, info *tracing.SpanInfo, next tracing.NextFunc) (any, error) {
+	prefix, model := genai.SplitModelName(info.Name())
+	provider := genai.DeriveProviderName(prefix)
+	request := asModelRequest(info.Input())
+
+	attrs := []attribute.KeyValue{
```
apascal07, line 37: `info.Labels` is never read in this package. Labels arrive through the reflection server (`genkit/reflection.go:813` via `tracing.WithTelemetryLabels`) and through `tracing.TelemetryLabelsFromContext` in `core/action.go:301` and `ai/generate.go:1051`, and `OTelInstrumentation` sets them as span-start attributes (`otel_instrumentation.go:70-76`). None of the three span builders here do, so a request tagged with telemetry labels exports none of them, and the `SpanInfo.Labels` doc ("raw TelemetryLabels set directly as span attributes") is unmet for this provider.

go/plugins/otel/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+	if t == nil {
+		return "error"
+	}
+	return t.String()
+}
+
+// maybeWarnNotRecording warns once if the SDK is not collecting: when no
+// TracerProvider is registered, OTel returns a non-recording span, so telemetry
+// is silently dropped. Surface that instead of failing quietly.
+func (g *GenAiInstrumentation) maybeWarnNotRecording(span oteltrace.Span) {
```
apascal07, line 139: `IsRecording()` is also false for a correctly configured SDK that sampled the span out (`parentbased_traceidratio` or any head sampler), so the first dropped call in a production app logs a warning telling the operator to initialize an SDK that is already exporting. It also contradicts the README and the package doc, which both promise the provider is inert when no SDK is configured. If the warning stays, the real condition is checkable once at construction (`otel.GetTracerProvider()` still being the no-op provider). Either way, `go/GEMINI.md` "Logging" asks for `logger.Warn(ctx, ...)` when a ctx is available (every call site has one) and for lowercase messages without trailing punctuation.

go/go.mod
```diff
@@ -34,118 +34,148 @@ require (
 	github.com/weaviate/weaviate v1.30.0
 	github.com/weaviate/weaviate-go-client/v5 v5.1.0
 	github.com/xeipuuv/gojsonschema v1.2.0
-	go.opentelemetry.io/otel v1.36.0
-	go.opentelemetry.io/otel/metric v1.36.0
-	go.opentelemetry.io/otel/sdk v1.36.0
-	go.opentelemetry.io/otel/sdk/metric v1.36.0
-	go.opentelemetry.io/otel/trace v1.36.0
+	go.opentelemetry.io/contrib/exporters/autoexport v0.71.0
```
apascal07, line 37: `go mod why -m go.opentelemetry.io/contrib/exporters/autoexport` reports one consumer: `go/samples/otel-genai`. Because the sample lives in the library module, every app that imports any part of `github.com/firebase/genkit/go` now carries `prometheus/client_golang`, `prometheus/common`, `otlptranslator`, `procfs`, `contrib/bridges/prometheus`, `exporters/prometheus`, `grpc-gateway/v2` and the five OTLP exporter modules in its `go.sum`, MVS resolution and vulnerability-scan surface, for a demo it never builds. `go/samples/mcp-git-pr-explainer` already has its own `go.mod` for the same reason; alternatively the sample can wire `otlptracehttp` and friends directly and drop autoexport.

go/plugins/otel/attributes.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// requestConfigAttributes maps a model request's config to gen_ai.request.*
+// attributes. It reads the config through the common shape; provider-specific
+// configs that embed GenerationCommonConfig are also handled.
+func requestConfigAttributes(request *ai.ModelRequest) []attribute.KeyValue {
+	var attrs []attribute.KeyValue
+	cfg := asCommonConfig(request.Config)
+	if cfg != nil {
+		if cfg.Temperature != 0 {
```
apascal07, line 34: `!= 0` cannot tell "unset" from "deliberately 0", and 0 is the standard deterministic temperature. A googlegenai config with `Temperature` pointing at 0 survives the round trip as `"temperature":0` and is then dropped here, so the span has no `gen_ai.request.temperature` and a consumer cannot distinguish deterministic sampling from the provider default. The JS plugin gates on `!= null` (`gen-ai-instrumentation.ts:352-363`) and records the 0, so the same workload reports different attribute sets per language. Same at lines 38, 41 and 43, and in `content.go:91-98` a legitimate 0-token count is never recorded. Decoding the round-tripped JSON into a pointer-field struct (or a map) preserves presence for every config except a bare `GenerationCommonConfig` value, which cannot express it.

go/plugins/otel/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+// recordError sets the error status, error.type attribute, and exception event.
+func (g *GenAiInstrumentation) recordError(span oteltrace.Span, err error) {
+	span.SetStatus(codes.Error, err.Error())
+	span.SetAttributes(attribute.String(genai.AttrErrorType, errorTypeOf(err)))
+	span.RecordError(err)
+}
+
+// errorTypeOf reports the error type for the error.type attribute: the concrete
+// Go type name (e.g. "*errors.errorString"), mirroring JS using the error name.
+func errorTypeOf(err error) string {
```
apascal07, line 125: `reflect.TypeOf(err).String()` varies with how the error was constructed, not with what went wrong. Genkit wraps action errors with `fmt.Errorf("...: %w")` and status errors, so an invalid-input rejection and an upstream 429 both land on the span and on `gen_ai.client.operation.duration` as `error.type="*fmt.wrapError"` or the status wrapper's type name, and a refactor from `errors.New` to `fmt.Errorf` silently changes the attribute value. The comment says this mirrors JS, but JS emits `e.name` (`"GenkitError"`, `"TypeError"`), so a cross-language alert on `error.type` splits into disjoint buckets. The Genkit status name (the discriminator googlecloud already uses) or `semconv.ErrorType(err)` gives the low-cardinality value `error.type` is specified for.

go/plugins/otel/instrumentation_test.go
```diff
@@ -0,0 +1,294 @@
...
+	span := findSpan(t, sr, "chat gemini-flash-latest")
+	if attrMap(span)[genai.AttrErrorType].AsString() == "" {
+		t.Error("expected error.type attribute on failed span")
+	}
+}
+
+func runTool(t *testing.T, name string) {
+	t.Helper()
+	_, err := tracing.RunInNewSpan(context.Background(),
+		&tracing.SpanMetadata{Name: name, Type: "action", Subtype: "tool"},
```
apascal07, line 275: `Subtype: "tool"` is a value no Genkit code path produces (real tools carry `"tool.v2"`), which is why the dispatch miss at `instrumentation.go:179` ships green. Same gap in `metrics_test.go:62-67`: it asserts only that the two instrument names exist, so recording input tokens under `gen_ai.token.type=output`, passing `OutputTokens` for both, or omitting `gen_ai.request.model` from the points all pass, and nothing covers the error-path duration point or a partial response (`TestModelSpanError` returns a nil response). Driving a `genkit.DefineTool` through the instrumentation and asserting the histogram data points from the `ManualReader` that is already wired up would catch both.

go/plugins/otel/helpers.go
```diff
@@ -0,0 +1,166 @@
...
+	if config == nil {
+		return nil
+	}
+	if c, ok := config.(*ai.GenerationCommonConfig); ok {
+		return c
+	}
+	if c, ok := config.(ai.GenerationCommonConfig); ok {
+		return &c
+	}
+	b, err := json.Marshal(config)
```
apascal07, line 71: The round trip only matches Genkit's camelCase tags. The anthropic plugin's config is the SDK's `MessageNewParams` (`plugins/anthropic/refs.go:25`), whose wire tags are `max_tokens`, `top_p`, `top_k` and `stop_sequences`, against `maxOutputTokens`, `topP`, `topK` and `stopSequences` in `ai/gen.go:163-172`. Only `temperature` survives, so `gen_ai.request.max_tokens`, `top_p`, `top_k` and `stop_sequences` are absent on every Anthropic span; the same applies to any config type that carries the provider's wire names. One type mismatch also drops everything: googlegenai's `GenerateContentConfig.TopK` is `*float32` against `int` here, so a fractional TopK makes `Unmarshal` fail and line 77 returns `nil`, losing temperature, topP and maxOutputTokens as well. The comment above says provider configs "share the common field names/tags", which does not hold for an SDK-native config.

go/plugins/otel/instrumentation.go
```diff
@@ -0,0 +1,189 @@
...
+	return mode
+}
+
+func (g *GenAiInstrumentation) tracer() oteltrace.Tracer {
+	if g.injectedTracer != nil {
+		return g.injectedTracer
+	}
+	// Resolve lazily so the app's SDK setup, which may run after construction,
+	// is picked up. otel.Tracer reads the global provider on each call.
+	return otel.Tracer(g.scopeName, oteltrace.WithInstrumentationVersion(genai.SemConvVersion))
```
apascal07, line 146: The laziness buys nothing and costs a lock per span. A tracer handed out by `otel.Tracer` before `SetTracerProvider` is a `global.tracer` whose delegate is swapped in atomically when the SDK is installed (`otel/internal/global/trace.go:138`), so a tracer resolved once in the constructor still picks up a provider registered later. Resolving here instead runs on every `RunInNewSpan`: the option slice and `NewTracerConfig` allocate per call, and once the SDK is installed `sdktrace.TracerProvider.Tracer` takes the provider-wide `p.mu` for a map lookup (`sdk/trace/provider.go:165`), so concurrent flows serialize on one mutex at every span start. A `sync.Once` around the lookup keeps both the injected-tracer escape hatch and the late-SDK behavior this comment wants.

go/plugins/otel/genai/messages.go
```diff
@@ -0,0 +1,144 @@
...
+		if tr == nil {
+			return map[string]any{"type": "tool_call_response", "response": nil}
+		}
+		m := map[string]any{"type": "tool_call_response", "response": tr.Output}
+		if tr.Ref != "" {
+			m["id"] = tr.Ref
+		}
+		return m
+	case part.IsMedia():
+		m := map[string]any{"type": "media", "content": part.Text}
```
apascal07, line 76: `part.Text` is the whole payload for inline media: `ai.NewMediaPart` stores the data URI in `Text` (`ai/document.go:114`), so a 2 MB base64 image in the request lands verbatim in `gen_ai.input.messages` under `SPAN_ONLY` and in the operation.details log record under `EVENT_ONLY`. The SDK's attribute length limit then truncates the JSON string mid-document, leaving an unparseable value, and an OTLP gRPC export above the receiver's max message size (4 MiB by default) fails with every other span in the batch. Nothing in the package truncates or substitutes a reference (`content_type` plus a length or hash), and a `data:` prefix is cheap to detect.

go/plugins/otel/genai/attributes.go
```diff
@@ -0,0 +1,215 @@
...
+//
+// See the spec:
+// https://github.com/open-telemetry/semantic-conventions-genai
+package genai
+
+import "strings"
+
+// Canonical gen_ai.* attribute names used by this instrumentation.
+const (
+	AttrOperationName = "gen_ai.operation.name"
```
apascal07, line 29: These duplicate typed keys the vendored dependency already ships. `semconv/v1.39.0` in otel v1.46.0 exports `GenAIOperationNameKey`, `GenAIProviderNameKey`, `GenAIRequestModelKey`, `GenAIUsageInputTokensKey`, `GenAITokenTypeKey`, `GenAIToolNameKey`, `GenAIInputMessagesKey`, `GenAIOutputMessagesKey`, `GenAISystemInstructionsKey`, `GenAIResponseFinishReasonsKey` and `GenAIOutputTypeKey`, the enum values (`GenAIOperationNameChat`, `GenAITokenTypeInput`, `GenAIProviderNameGCPGemini`, `GenAIProviderNameAnthropic`, ...), and `genaiconv.NewClientTokenUsage` / `NewClientOperationDuration`, which build the two instruments with the spec's descriptions (`metrics.go` already disagrees: "Duration of a GenAI model operation." vs the generated "GenAI operation duration."). The "no OpenTelemetry imports" rationale in the package doc no longer holds, since `metrics.go` in this package imports `otel/metric` and `otel/attribute`. Hand-written strings drift silently when the spec renames a key, and `SemConvVersion` is already pinned at `1.38.0` while the vendored schema is `1.39.0` (`semconv.SchemaURL`). Only `reasoning.output_tokens` and `cache_read.input_tokens`, which the stable package lacks, need local constants.

go/plugins/otel/instrumentation.go
```diff
@@ -0,0 +1,189 @@
...
+	injectedMeter  metric.Meter
+
+	metricsOnce sync.Once
+	metrics     *genai.Metrics
+
+	warnOnce sync.Once
+}
+
+// NewGenAiInstrumentation builds a GenAiInstrumentation from options.
+func NewGenAiInstrumentation(opts GenAiInstrumentationOptions) *GenAiInstrumentation {
```
apascal07, line 109: The documented wiring is a process-global mutator in `core/tracing` called before `genkit.Init`, while every other app-facing knob in this repo is a `genkit.With*` option or a plugin passed to `genkit.WithPlugins` (`genkit/genkit.go:151-216`). Two `genkit.Init` instances in one process (a server plus a test harness, or two tenants) cannot carry different instrumentation because `configured` is one package-level variable, a test that configures it must remember `tracing.ResetInstrumentation`, and the plugin has no `Init` slot in plugin ordering. Implementing `api.Plugin` here (`genkit.WithPlugins(&genaiotel.GenAI{...})`, with `Init` calling `ConfigureInstrumentation`) gives it the standard surface now and keeps `ConfigureInstrumentation` as the low-level escape hatch; per-instance scoping needs the base PR to hang the configured provider off the registry, which is easier to do before this global example ships in a README.

apascal07 (changes requested):

PR #6346, opened 2026-09-15
fix(py/plugins/openai): decode base64 embeddings to float vectors

### LINKED ISSUE #6317: fix(py/plugins/openai): encodingFormat 'base64' fails embedding validation
The OpenAI embedder passes `encodingFormat` through to `embeddings.create`. When `encoding_format` is given explicitly the SDK returns the response unmodified, so each `item.embedding` is the raw base64 string rather than a list of floats. `Embedding(embedding=item.embedding)` then fails pydantic validation.

Repro:

```python
await ai.embed(
    embedder='openai/text-embedding-3-small',
    content='hi',
    options={'encodingFormat': 'base64'},
)
```

usmamalik345:
## Summary
- When `encodingFormat: 'base64'` is set, OpenAI returns embedding vectors as base64 strings of little-endian float32 values. Genkit's `Embedding` model requires `list[float]`, so construction failed with a Pydantic validation error.
- Decode that wire format in the embedder path (and leave float lists unchanged) before building `Embedding` objects.
- Add unit coverage for the decoder and an embedder regression that mocks a base64 API payload.

Fixes #6317

## Test plan
- [x] Local decode verification (base64 float32 LE → floats; invalid base64 rejected; float list passthrough)
- [x] `ruff check` / `ruff format` on changed files
- [ ] CI: `genkit-openai` unit tests (Windows cannot install `dotpromptz-handlebars`; Linux CI should run the full suite)
- [ ] Manual: `await ai.embed(embedder='openai/text-embedding-3-small', content='hi', options={'encodingFormat': 'base64'})` returns float vectors


hilariie:
Hi @usamas346, @usmamalik345. Please make sure to sign the CLA and if you have, remove the cursor agent co-author from the commit so CLA passes.

usmamalik345:
Thanks @hilariie - CLA is signed under my GitHub account and the Cursor co-author has been removed.

py/packages/genkit-openai/tests/openai_plugin_test.py
```diff
@@ -313,6 +315,31 @@ async def test_embedder_maps_status_errors_for_every_option_shape(options: dict[
...
+    client = MagicMock()
+    item = MagicMock()
+    item.embedding = encoded
+    result = MagicMock()
+    result.data = [item]
+    client.embeddings.create = AsyncMock(return_value=result)
+
+    response = await _run_embedder(client, options={'encodingFormat': 'base64'})
+
+    assert client.embeddings.create.await_args.kwargs['encoding_format'] == 'base64'
```
hilariie, line 336: This is what fails the ty check (await_args is `_Call | None`). Can we use `client.embeddings.create.assert_awaited_once_with(model='text-embedding-3-small', input=['hello'], encoding_format='base64')`? Building the stub through `_embedding_client(embedding=encoded)` instead of inline would also drop the duplicated setup above.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
```diff
@@ -81,6 +83,38 @@ def strip_markdown_fences(text: str) -> str:
     return text
 
 
+def coerce_embedding_vector(value: list[float] | str) -> list[float]:
```
hilariie, line 86: The SDK already requests base64 and decodes it itself when encoding_format is omitted, and only leaves strings when the format is passed explicitly. Mapping the user's 'base64' to "omit" should be enough here. Same request on the wire, no decoder to maintain.

py/packages/genkit-openai/src/genkit_openai/models/utils.py
```diff
@@ -81,6 +83,38 @@ def strip_markdown_fences(text: str) -> str:
...
+    vector as a base64 string of little-endian float32 values instead of a
+    float list. Genkit's ``Embedding`` model requires ``list[float]``, so decode
+    that wire format before constructing the response.
+    """
+    if isinstance(value, list):
+        return value
+    if isinstance(value, str):
+        try:
+            # validate= is 3.11+; keep the call compatible with requires-python >=3.10.
+            raw = base64.b64decode(value)
```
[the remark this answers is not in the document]
hilariie: `96248e4` removes the decoder, so 'base64' now omits encoding_format and the OpenAI SDK decodes the response itself.

cabljac (approved):
lgtm

PR #6243, opened 2026-09-01
feat(py): construct and read Part without .root; one Part through generate API

huangjeff5:
Callers build a part with `Part.from_text` / `Part(text=...)` and read `.text` / `.media` without `.root`. Kind classes are not on `from genkit import`. `resume_respond`, `resume_restart`, and `wrap_tool` take a `Part` of the matching kind. A2UI helpers take that same `Part`.

```python
from genkit import Part

caption = Part.from_text('a ceramic bowl')
# or
caption = Part(text='a ceramic bowl')

image = Part.from_media('https://example.com/bowl.png', content_type='image/png')

await ai.generate(prompt=[caption, image])

print(caption.text)
```

```python
# before
from genkit import TextPart

Part(root=TextPart(text='hi'))

# after
from genkit import Part

Part.from_text('hi')
# or Part(text='hi')
```

```python
from genkit_a2ui import Surfaces, envelopes_from_parts, is_a2ui_part

response = await ai.generate(prompt='Show me the weather in Tokyo', use=[Surfaces()])

envelopes = envelopes_from_parts(response.message.content)
is_a2ui_part(response.message.content[0])
```

## Decisions
- A part answers exactly one of text, media, toolRequest, toolResponse, reasoning, resource, or data. `metadata` and `custom` may ride. `Part.from_custom` is when the vendor blob is the part.
- `{}` or two answers raise, including caption+image or caption+`{data}` on one object. Those are two parts.
- `from genkit import` keeps `Part`, `ToolRequest`, `ToolResponse`, and `Media`. `TextPart`, `MediaPart`, `ToolRequestPart`, `ToolResponsePart`, `CustomPart`, and `ReasoningPart` are not public imports.
- `resume_respond` needs a tool-response part, `resume_restart` and `wrap_tool` need a tool-request part. A text part there raises.
- There is no `from_resource` / `.resource`. A resource on the wire has no public getter.
- `model_dump()` is the Message wire: camelCase, no null siblings, no `root` key.
- `is_a2ui_part`, `envelopes_from_parts`, and the other A2UI part helpers take `Part`. Pass `message.content` or a `Part` from generate — not a dict.

PR #6363, opened 2026-09-17
fix(py/plugins/google-genai)!: send output schemas via response_json_schema

### LINKED ISSUE #6182: fix(py/plugins/google-genai): responseJsonSchema unsupported and responseSchema lossily converted
Constrained output support is half-implemented:

- There is no `responseJsonSchema` path at all; the only mention is an excluded strip-field (`gemini.py:459`). JS defaults to `responseJsonSchema`; Go rejects it loudly rather than half-sending.
- Output schemas go through `_convert_schema_property` (`gemini.py:1195-1271`), which returns `None` for a schema with no top-level `type` and silently drops `anyOf`, `additionalProperties`, `format`, `default`, and numeric bounds.
- No test in `py/` touches `response_mime_type` or `response_schema`, so none of this is pinned.

Fix: support `responseJsonSchema` (or reject unsupported schemas loudly), stop lossy conversion, and pin the request body in tests.

hilariie:
Fixes #6182

## Summary

Constrained JSON output sends the request's output schema verbatim as `responseJsonSchema`, so `anyOf`, `$ref`, `additionalProperties`, `format`, `default` and numeric bounds reach the model as written. `GoogleAI` and `VertexAI` take a `legacy_response_schema` option that sends the older OpenAPI-subset `responseSchema` instead, and a schema that cannot be converted for that field raises `INVALID_ARGUMENT` rather than silently going unconstrained. Enum output sends `text/x.enum` with a string schema.

`_convert_schema_property` no longer drops what the Gemini `Schema` type can represent. It resolves `anyOf`, folding a `null` branch into `nullable`, and carries `title`, `format`, `default`, `minimum`, `maximum`, `minItems`, `maxItems` and `propertyOrdering` through. That also fixes tool declarations, where an `Optional[...]` parameter used to vanish from the function schema entirely. `nullable` is now set only when a null member was actually removed, not for every `type` list.

A behaviour change: `response_mime_type`, `response_schema` and `response_json_schema` passed through plugin config now raise `INVALID_ARGUMENT` in either spelling. They previously reached the API unchecked, because the fields meant to strip them sat on `SpeechConfigSchema` instead of the Gemini config.

Tests pin the `GenerateContentConfig` reaching the client for both wire shapes, for enum and unconstrained output, and for a model that does not support constrained output.


py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1239,45 +1258,82 @@ def _convert_schema_property(
...
+        self._copy_schema_keywords(input_schema, schema)
 
         if 'required' in input_schema:
             schema.required = cast(list[str], input_schema['required'])
 
-        if 'type' in input_schema:
-            raw_type = input_schema['type']
-            if isinstance(raw_type, list):
-                non_null = [t for t in raw_type if t != 'null']
+        raw_type = input_schema['type']
```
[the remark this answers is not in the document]
hilariie: A `KeyError` isn't reachable here: `if 'type' not in input_schema: return None` is the guard immediately above this line.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1829,6 +1880,61 @@ async def _genkit_to_googleai_cfg(self, request: ModelRequest) -> genai_types.Ge
...
+
+    def _apply_output_config(self, cfg: genai_types.GenerateContentConfig, request: ModelRequest) -> None:
+        """Set the response MIME type and schema for the request's output format.
+
+        JSON output sends the schema verbatim as ``response_json_schema``, or as
+        a converted ``response_schema`` when ``legacy_response_schema`` is set.
+        Enum output sends ``text/x.enum`` with a converted ``response_schema``.
+        """
+        schema = request.output_schema
+        json_mode = request.output_format == 'json' or request.output_content_type == 'application/json'
```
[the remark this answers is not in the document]
hilariie: `array` already takes this branch through the second clause. `_ai/_formats/_array.py` sets `content_type='application/json'` and `apply_format` in `_ai/_generate.py` assigns it to the request, so an array-format request arrives with the JSON content type. These are also the same two clauses as `js/plugins/google-genai/src/googleai/gemini.ts:835-837`, so adding `'array'` would make Python diverge from JS without changing any request the framework produces.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1239,45 +1265,87 @@ def _convert_schema_property(
...
+                schema.items = self._convert_schema_property(cast(dict[str, object], items_value), defs)
+
+        if schema_type == genai_types.Type.OBJECT:
+            schema.properties = {}
+            properties_value = input_schema.get('properties', {})
+            if isinstance(properties_value, dict):
+                properties = cast(dict[str, dict[str, object]], properties_value)
+                for key in properties:
+                    nested_schema = self._convert_schema_property(properties[key], defs)
+                    if nested_schema:
```
cabljac: Same path builds tool declarations (`_create_tool` at 1200/1209), so a `oneOf`-typed tool param would vanish here too. `const` looks like it lands as plain STRING rather than `enum=[const]` as well, which would leave `Literal` fields unconstrained on the legacy path. The SDK's own `process_schema` rewrites `const` -> `enum`, so matching that might be the cleanest fix.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1829,6 +1892,61 @@ async def _genkit_to_googleai_cfg(self, request: ModelRequest) -> genai_types.Ge
...
+    _OUTPUT_KEYS = frozenset([
+        'response_mime_type',
+        'response_schema',
+        'response_json_schema',
+        'responseMimeType',
+        'responseSchema',
+        'responseJsonSchema',
+    ])
+
+    def _reject_output_keys(self, config: dict[str, Any]) -> None:
```
cabljac, line 1939: This one is a behaviour change for anyone passing `config={'response_mime_type': ...}` today, since that used to reach the API unchecked. Worth marking the title `fix(py/plugins/google-genai)!` so it shows up as breaking.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1829,6 +1892,61 @@ async def _genkit_to_googleai_cfg(self, request: ModelRequest) -> genai_types.Ge
...
+
+        if json_mode:
+            cfg.response_mime_type = 'application/json'
+            if schema and request.output_constrained:
+                if self._legacy_response_schema:
+                    cfg.response_schema = self._output_response_schema(schema)
+                else:
+                    cfg.response_json_schema = schema
+        elif enum_mode:
+            cfg.response_mime_type = 'text/x.enum'
```
cabljac, line 1970: Enum output sends `text/x.enum` plus a converted `response_schema` regardless of `legacy_response_schema`, so constrained decoding replaces instruction-only for existing enum users. From what I can see JS never sends `x.enum`. Intentional divergence on the Python side?

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1829,6 +1892,61 @@ async def _genkit_to_googleai_cfg(self, request: ModelRequest) -> genai_types.Ge
...
+        json_mode = request.output_format == 'json' or request.output_content_type == 'application/json'
+        enum_mode = request.output_format == 'enum' or request.output_content_type == 'text/enum'
+
+        if json_mode:
+            cfg.response_mime_type = 'application/json'
+            if schema and request.output_constrained:
+                if self._legacy_response_schema:
+                    cfg.response_schema = self._output_response_schema(schema)
+                else:
+                    cfg.response_json_schema = schema
```
cabljac, line 1968: Is `response_json_schema` safe on older models? It's not gated on model or backend from what I can see, so a pre-2.5 model that worked via the converted schema would now get whatever the API returns, and the only escape hatch is plugin-wide `legacy_response_schema`. Have you been able to try it against one?

py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
```diff
@@ -970,6 +986,84 @@ def test_gemini_model__convert_schema_property_raises_exception(
         gemini_model_instance._convert_schema_property(input_schema, defs)
 
 
+def test_gemini_model__convert_schema_property_keeps_keywords(gemini_model_instance: GeminiModel) -> None:
```
cabljac, line 989: nit: two things don't seem pinned yet. Reverting the `nullable` logic to always-true still passes, since every type list in the tests contains `'null'`, and extending `has_output` to cover `output_content_type` also passes when reverted. A `['string', 'integer']` property asserting `nullable is None`, and a content-type-only request, would cover both.

cabljac (commented):
Core swap looks right, and pinning the exact GenerateContentConfig for both wire shapes in `test_generate_json_output_pins_request_body` is a good way to test this.

One thing worth digging into: in legacy mode, does a nested node that fails conversion actually get rejected? `_convert_schema_property` handles `anyOf` but I can't see `oneOf` anywhere, and `gemini.py:1312` skips any property whose conversion returns `None`. So from what I can see a pydantic discriminated union (`Annotated[Union[Cat, Dog], Field(discriminator='kind')]`) emits `oneOf` with no `type`, and that property would drop out of `response_schema` silently rather than raise, leaving the field unconstrained. `README.md:77` says a schema that cannot be converted is rejected. Worth running one through to confirm either way.

Couple of smaller ones inline.

py/packages/genkit-google-genai/tests/google_genai_plugin_test.py
```diff
@@ -175,6 +176,30 @@ async def _get_client() -> MagicMock:
     assert other_loop_client is not first
 
 
+@patch('genkit_google_genai.google.genai.client.Client')
+@pytest.mark.asyncio
+@pytest.mark.parametrize('legacy_response_schema', [False, True])
+async def test_googleai_legacy_response_schema_threads_to_gemini_model(
```
cabljac, line 182: nit: these two don't await anything, so they can be plain `def` without the asyncio marker.

hilariie: _new_gemini fetches the loop-local client, which calls asyncio.get_running_loop() and raises as a plain def, so I kept async.

py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
hilariie: Resolved in [7f64ec6](https://github.com/genkit-ai/genkit/pull/6363/commits/7f64ec66d0e8dc6460863ece91671cd8c15be3bf)

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
hilariie: Enum output already sends a converted response_schema before this change, only without the MIME type the API requires for it. text/x.enum is what Go sends and JS has no enum mode.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
hilariie: Resolved in [7f64ec6](https://github.com/genkit-ai/genkit/pull/6363/commits/7f64ec66d0e8dc6460863ece91671cd8c15be3bf) for string `const`. A non-string `const` keeps its bare type rather than raising like the SDK does, since this path builds every tool declaration.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
hilariie: There is no gate in JS either. As for the live test, I could not try a pre-2.5 model: models.list on the Developer API returns none today and the 2.0 Flash line shut down on 1 June.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1200,13 +1210,48 @@ def _create_tool(self, tool: ToolDefinition) -> genai_types.Tool:
         )
         return genai_types.Tool(function_declarations=[function])
 
+    # JSON Schema keywords copied onto a Gemini Schema: (keyword, attribute, coercion).
+    _SCHEMA_SCALAR_KEYS: tuple[tuple[str, str, Callable[[Any], Any]], ...] = (
+        ('title', 'title', str),
+        ('format', 'format', str),
```
cabljac, line 1215: `format` is now copied onto every node, including tool params, so pydantic `date` / `UUID` / `HttpUrl` go out as `format: 'date'` / `'uuid'` / `'uri'` where main sent a bare `STRING`. The SDK docstring lists a few STRING formats and says "other formats", so I have not found a pinned accepted set, and JS never sends `format` at all. Can you verify those three against both backends before we merge? A whitelist (`date-time`, `int32`/`int64`, `float`/`double`) is the safe shape if any of them 400.

py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
```diff
@@ -1118,39 +1309,237 @@ async def test_gemini_model__unknown_extra_rides_on_extra_body(
...
+)
+
+
+def _output_request(
+    *,
+    format: str | None = None,
+    json_schema: dict[str, Any] | None = None,
+    constrained: bool | None = None,
+    content_type: str | None = None,
+) -> ModelRequest:
```
cabljac, line 1440: Rebase note: #6243 removed `TextPart` from the exports and `Part(root=...)` now raises. This helper, the tests around :1497 / :1542, and the pre-existing `Part(root=TextPart(` uses in this file need `Part.from_text('hi')` and `content[0].text`.

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
```diff
@@ -1239,45 +1284,102 @@ def _convert_schema_property(
...
+                message=f'{self._version}: schema type {raw_type!r} is not supported',
+            )
+        schema_type = genai_types.Type(raw_type)
+        schema.type = schema_type
+
+        const = input_schema.get('const')
+        if schema_type == genai_types.Type.STRING and isinstance(const, str):
+            schema.enum = [const]
+        elif 'enum' in input_schema:
+            schema.enum = cast(list[str], input_schema['enum'])
```
cabljac: nit: non-string enums still land in `Schema.enum` (`Literal[1, 2]` -> `enum=[1, 2]`). Pre-existing, ignore unless cheap while you are here.

py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
```diff
@@ -1118,39 +1309,237 @@ async def test_gemini_model__unknown_extra_rides_on_extra_body(
...
+    cfg = await model._genkit_to_googleai_cfg(request)
+
+    assert cfg is not None
+    assert cfg.response_mime_type == 'text/x.enum'
+    assert cfg.response_schema == genai_types.Schema(type=genai_types.Type.STRING, enum=['a', 'b'])
+    assert cfg.response_json_schema is None
+
+
+@pytest.mark.asyncio
+async def test_gemini_model__constrained_output_skipped_without_model_support() -> None:
```
cabljac, line 1567: nit: this pins that a gate skip sends no MIME type and no schema. Not this PR's problem (Python core has no equivalent of JS `simulateConstrainedGeneration`), I will raise it with the core team separately.

cabljac (changes requested):
The `response_json_schema` swap is right (the SDK identity-maps it on both backends) and the request-body pin test is exactly what #6182 asked for. Thanks for turning the earlier round so quickly. One thing worth digging into, then a couple of small ones.

`_convert_schema_property` now raises for a required property with no type, and `_create_tool` goes through the same converter. So a tool with a `value: Any` param now fails at generate time with `INVALID_ARGUMENT`, where main drops the property silently, and JS still drops silently. Was that intended? If so we should pin it at the `_create_tool` level; if not, a strict flag from `_output_response_schema` keeps the raise on the legacy path only.

py/packages/genkit-google-genai/test/models/googlegenai_gemini_test.py
hilariie: Rebased onto `b8019785b` and the Part migration is folded into the first commit

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
hilariie: Resolved in `4d1fec157`

py/packages/genkit-google-genai/src/genkit_google_genai/models/gemini.py
hilariie: Verified live with gemini-2.5-flash on both backends, on both `responseSchema` and tool parameters: `date-time`, `date`, `uuid`, `uri` and email were all accepted, matching the Developer API discovery doc ('any value is allowed') and Vertex's list of string formats. Kept format verbatim; `4d1fec157` pins the pydantic date, uuid and uri cases.

Issue #6194, opened 2026-08-26
feat(py/plugins/mcp): stdio MCP client exposing one server tools as tool.v2 dynamic actions


cabljac:
First vertical slice of the MCP plugin: the thinnest end-to-end path, forcing the risky decisions (SDK choice, namespacing, tool.v2-only registration, multipart mapping) before any breadth.

Scope:
- Connect to a single stdio MCP server via the official `mcp` Python SDK.
- List its tools and register each through core's dynamic action provider, namespaced `<server>/<tool>`.
- Register as tool.v2 only (multipart tool responses); no tool.v1.
- Map tool result content blocks (text + media) to multipart tool.v2 responses.
- Test against a scripted fake MCP server pinning the registered actions and a round-trip call; verify tools appear and run in the Dev UI.

Out of scope here, potential follow-ups as separate sub-issues once this lands:
- Streamable HTTP transport + auth headers (JS parity, cf. #6144)
- Multi-server host (createMcpHost equivalent: config, enable/disable, reconnect)
- MCP prompts as Genkit prompts
- Server side: expose Genkit tools/flows as an MCP server (genkitx-mcp equivalent)

Excluded entirely: MCP resources (resources support is being removed from py core).

PR #6365, opened 2026-09-17
feat(py/plugins/mcp): add the MCP client package

cabljac:
Python has no MCP client, so Python users cannot call tools hosted by an MCP server. JS and Go both ship one.

This adds the `genkit-mcp` package and wires it into the workspace: root dependency and uv source, type-checker source lists, and the `mcp` dependency moved out of the root dev group into the package that needs it.

`_errors.py` lands here because it is nothing but class definitions. `McpClientError` deliberately does not derive from `RuntimeError`: the MCP SDK reports a result failing the server's own output schema as a bare `RuntimeError`, which the tool wrapper folds into an `isError` envelope, so sharing a base class would send a dead subprocess to the model as a retryable tool error.

`McpStdioServerConfig.env` documented the opposite of what the SDK does, and is corrected.

No public API is exported yet and the package is absent from the publish workflow. Both arrive with the client.

Part of #6194.


py/packages/genkit-mcp/src/genkit_mcp/_tools.py
```diff
@@ -0,0 +1,77 @@
...
+        The envelope ``generate`` expects from a ``tool.v2`` action.
+    """
+    raise NotImplementedError
+
+
+def mcp_tool_to_action(
+    tool: McpTool,
+    server_name: str,
+    call: object,
+) -> Action:
```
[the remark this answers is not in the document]
cabljac: Stale: this was written against ca5d1ff67, and the branch has been rebased since. `call` is now typed `McpToolCall`, an alias declared at the top of the same file as `Callable[[str, dict[str, Any], dict[str, Any] | None], Awaitable[CallToolResult]]`. That is this suggestion with a concrete argument list.

py/packages/genkit-mcp/src/genkit_mcp/_errors.py
```diff
@@ -0,0 +1,46 @@
...
+    ``RuntimeError``: the MCP SDK reports a tool result that fails the server's
+    own output schema as a bare ``RuntimeError``, and that failure is reported
+    to the model as a tool error while these are not.
+    """
+
+
+class McpConnectionClosedError(McpClientError):
+    """The client was shut down by :meth:`McpClient.close`.
+
+    Closing is final. :meth:`McpClient.restart` connects again.
```
hilariie, line 33: Can `restart()` revive a client after `close()`? This line reads both ways, and `McpClient.close` says "for good". Can we make the two docstrings agree on one answer?

py/packages/genkit-mcp/pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+  "Topic :: Scientific/Engineering :: Artificial Intelligence",
+  "Topic :: Software Development :: Libraries :: Application Frameworks",
+]
+dependencies = [
+  "genkit",
+  "mcp>=1.26.0",
```
hilariie, line 34: Is 1.x the intended SDK? The lock stays on 1.26.0, but a fresh install resolves mcp 2.2.0, which renames `isError`/`structuredContent`/`inputSchema` to snake_case. If 1.x is the plan, can we cap it at `<2`?

py/packages/genkit-mcp/src/genkit_mcp/_client.py
```diff
@@ -0,0 +1,84 @@
...
+    *,
+    tool_prefix: str | None = None,
+    cache_ttl_millis: int | None = None,
+    request_timeout_millis: int | None = DEFAULT_REQUEST_TIMEOUT_MILLIS,
+) -> McpClient:
+    """Register an MCP server's tools with Genkit.
+
+    The server's tools become ``tool.v2`` actions behind a dynamic action provider
+    named ``name``. Select them from ``generate`` as ``<name>:tool/*`` for all of
+    them, or ``<name>:tool/<name>_<tool>`` for one. Connection is lazy: the child
```
hilariie, line 60: Nit: The tool name is built from `tool_prefix`, so this should be `<name>:tool/<tool_prefix>_<tool>`. As written it only works when `tool_prefix` is left at its default.

py/packages/genkit-mcp/src/genkit_mcp/_config.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+
+from pydantic import BaseModel
+
+
+class McpStdioServerConfig(BaseModel):
+    """How to launch an MCP server as a child process over stdio.
+
+    Field names match the `mcpServers` entries used by Claude Desktop, Cursor and
+    the JS plugin, so an existing config block can be passed through unchanged.
```
hilariie, line 28: Nit: can we drop the JS plugin reference?

PR #6367, opened 2026-09-17
feat(py/plugins/mcp): manage the stdio connection to an MCP server

cabljac:
Start the server process on first use and own it for the life of the connection.

The MCP SDK opens its transport and session inside a task group, so both must be entered and exited by the same task. A background owner task holds them and callers marshal work to it through a queue. One connection is held per event loop, because the Dev UI reflection server runs its own loop on a separate thread.

The session sets `read_timeout_seconds`, default 60s, which is what the TypeScript SDK applies to every request. Without it one hung `tools/call` stops every later call to that server. A timeout arrives as `McpError` 408, outside the ambiguous `CONNECTION_CLOSED` range, so it fails alone and leaves the connection open.

Closing cancels the owner, interrupts any in-flight call and reaps the child. Tool listing follows pagination cursors, with a repeat-cursor guard and a page cap that JS and Go both lack.

Known gaps: the owner serves one request at a time where JS and Go multiplex, and Windows is untested.

Part of #6194.


py/packages/genkit-mcp/src/genkit_mcp/_connection.py
```diff
@@ -69,11 +113,132 @@ async def call_tool(
...
+        result: asyncio.Future[_T] = loop.create_future()
+        async with self._state_lock:
+            if self._state is _ConnectionState.FAILED:
+                raise self._failed_error()
+            if self._state is _ConnectionState.CLOSED:
+                raise self._closed_error()
+            if self._state is not _ConnectionState.OPEN:
+                raise RuntimeError('MCP connection is not available')
+            await self._queue.put(_Request(operation, result))
+        return await result
```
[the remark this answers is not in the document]
cabljac: Already handled. asyncio cancels the future that a cancelled task is awaiting, and the serve loop skips any request whose future is already cancelled before dispatching it (`if request.result.cancelled(): continue`). That line exists for exactly this case.

py/packages/genkit-mcp/src/genkit_mcp/_connection.py
```diff
@@ -69,11 +113,132 @@ async def call_tool(
...
+            async with stdio_client(parameters) as (read_stream, write_stream):
+                async with ClientSession(read_stream, write_stream) as session:
+                    initialized = await session.initialize()
+                    self._server_name = initialized.serverInfo.name
+                    async with self._state_lock:
+                        if self._state is _ConnectionState.CONNECTING:
+                            self._state = _ConnectionState.OPEN
+                    self._ready.set()
+                    while True:
+                        request = await self._queue.get()
```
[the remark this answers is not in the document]
cabljac: Stale: written against e6ed9135b. That guard is gone. Operations now run as child tasks, so comparing the current task against the owner would no longer recognise them. The check is a context variable that raises `McpClientError` instead.

py/packages/genkit-mcp/tests/mcp_connection_test.py
```diff
@@ -0,0 +1,194 @@
...
+    assert isinstance(failure.value.__cause__, OSError)
+    assert connection._state is _ConnectionState.FAILED
+
+    with pytest.raises(RuntimeError, match='failed'):
+        await asyncio.wait_for(connection.call_tool('echo', {}), timeout=15)
+
+    await connection.close()
+    del connection, failure
+    gc.collect()
+    assert unhandled == []
```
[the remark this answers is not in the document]
cabljac: The premise does not hold here. This suite runs with `asyncio_default_test_loop_scope=function`, so each async test gets its own event loop which is discarded when the test ends. There is no shared handler to leak into another test, and nothing to restore.

py/packages/genkit-mcp/tests/mcp_connection_test.py
```diff
@@ -0,0 +1,796 @@
...
+async def test_server_death_makes_the_connection_terminal(tmp_path: Path) -> None:
+    connection = McpConnection(config(tmp_path))
+    await asyncio.wait_for(connection.list_tools(), timeout=LIMIT)
+    owner = connection._owner_task
+    assert owner is not None
+    os.kill(int((tmp_path / 'server.pid').read_text(encoding='utf-8')), signal.SIGKILL)
+
+    with pytest.raises(McpConnectionFailedError, match='failed') as first:
+        await asyncio.wait_for(connection.call_tool('echo', {}), timeout=LIMIT)
+    assert first.value.__cause__ is not None
```
hilariie: This asserts ordering rather than behaviour. If the child is fully dead before the request's stdin write, the write fails inside the SDK's stdin_writer, the task group cancels the in-flight _execute, and the caller is released with no cause before _run records it, as a 2-10ms stall between os.kill and call_tool reproduces a None cause every time. Can we move this to the failure raised after `await settle(owner)`, where _failure is set in both orderings?

py/packages/genkit-mcp/src/genkit_mcp/_connection.py
```diff
@@ -77,13 +247,198 @@ async def call_tool(
...
+        _serving.set(self)
+        try:
+            try:
+                value = await request.operation(session)
+            except Exception as error:
+                if await _transport_is_dead(session, error):
+                    # Settle now rather than leaving callers waiting the seconds the SDK spends terminating the child.
+                    self._settle(error)
+                    serving.cancel()
+                elif not request.result.done():
```
hilariie: I don't think we have anything in the suite cancels a caller after its request is dispatched, so this guard and the one at 387 never take their False branch; without them the reply's set_result raises InvalidStateError and the connection ends FAILED with no cause

py/packages/genkit-mcp/src/genkit_mcp/_connection.py
```diff
@@ -77,13 +247,198 @@ async def call_tool(
...
+
+        async def call(session: ClientSession) -> CallToolResult:
+            return await _call_tool(session, name, arguments, meta)
+
+        return await self._request(call)
 
     async def close(self) -> None:
         """Shut the session down and reap the child process.
 
         Safe to call when never connected, and safe to call twice. A closed
```
hilariie: Nit: No test closes a connection that never started. A subprocess-free test that closes twice and asserts list_tools raises McpConnectionClosedError would pin both promises.

py/packages/genkit-mcp/src/genkit_mcp/_connection.py
```diff
@@ -77,13 +247,198 @@ async def call_tool(
...
+                    client_info=CLIENT_INFO,
+                ) as session:
+                    await session.initialize()
+                    async with self._state_lock:
+                        if self._state is _ConnectionState.CONNECTING:
+                            self._state = _ConnectionState.OPEN
+                    self._ready.set()
+                    await self._serve(session)
+        except BaseException as error:
+            self._settle(error if isinstance(error, Exception) else None)
```
hilariie: When the SDK's stdio task group crashes from inside (stdin gone while stdout stays open), the in-flight caller has already been released with no cause and this is the only place the real error is seen, unlogged. Should we use a structlog warning here?

PR #6369, opened 2026-09-17
feat(py/plugins/mcp): define MCP servers as dynamic tool providers

cabljac:
`define_mcp_client` registers one MCP server as a dynamic action provider, so `tools=['<name>:tool/*']` binds every tool that server lists. Each action carries its own name under `metadata['name']`, which is what wildcard expansion matches on; without it the selector expands to zero tools and reports nothing.

It is a function rather than a `Plugin` subclass because `Plugin.init()` renames actions to `<plugin>/<action>`, and a provider name cannot contain a slash. Neither JS nor Go registers MCP as a plugin either.

Building a client and registering it are separate: `create_mcp_client` returns one that registers nothing, and `define_mcp_client` registers a provider for it and assigns it back. That is the JS shape, and it is the seam a multi-server host needs, since the host owns one provider across several clients.

Tool names are prefixed with the provider name the caller chose, not the name the server advertises. An advertised name is arbitrary server-controlled text: `@modelcontextprotocol/server-everything` calls itself `mcp-servers/everything`, so prefixing with it produced names holding a slash, which is a 400 from Gemini and OpenAI. Go already prefixes with its own `options.Name`.

`close()` is final, as it is for files, sockets and HTTP clients throughout Python. `restart()` reconnects, and a connection that failed is rebuilt on next use rather than killing the client for the life of the process.

The public wiring withheld from the earlier PRs lands here: publish workflow, the `genkit[mcp]` extra, exports and the smoke import.

A multi-server host is deliberately out of scope, and is tracked as a follow-up on #6194.

Part of #6194.


py/packages/genkit-mcp/src/genkit_mcp/_client.py
```diff
@@ -26,17 +36,98 @@ class McpClient:
...
+        """
+        current = asyncio.get_running_loop()
+        with self._connections_lock:
+            self._closed = True
+            connections = list(self._connections.items())
+        for loop, connection in connections:
+            if loop is current:
+                await connection.close()
+            elif loop.is_running():
+                await asyncio.wrap_future(asyncio.run_coroutine_threadsafe(connection.close(), loop))
```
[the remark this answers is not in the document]
cabljac: Point one is already done: `_disconnect_all` clears `self._connections` under the lock. Point two I am not taking. A bare `except Exception: pass` around each close would hide teardown failures. The risk it targets is already closed, because the branch is guarded on `loop.is_running()`, and a loop that has already stopped cancelled its connection on the way out, taking the server process with it.

py/packages/genkit-mcp/src/genkit_mcp/_client.py
```diff
@@ -16,32 +16,257 @@
...
+            One action per tool, in server order.
+
+        Raises:
+            McpConnectionClosedError: If this client is closed.
+            McpConnectionFailedError: If the server cannot be started.
+        """
+        if self._server.disabled:
+            return []
+        tools = await self._connection().list_tools()
+        return [mcp_tool_to_action(tool, self._tool_prefix, self._call_tool) for tool in tools]
```
huangjeff5: `get_active_tools` builds one action per listed tool and never calls `mcp_tools_to_actions`. A server that lists `look.up` and `look/up` registers both as `bookshop_look_up`. `tools=['bookshop:tool/bookshop_look_up']` runs whichever was listed first, and `tools=['bookshop:tool/*']` dies later with `Cannot provide two tools with the same name: 'bookshop_look_up' and 'bookshop_look_up'`, which does not name `look.up` and `look/up`.

`mcp_tools_to_actions` already raises `MCP tools 'look.up' and 'look/up' both map to the Genkit tool name 'bookshop_look_up'`.

I suggest `get_active_tools` return that list, so the listing fails before `generate`.

py/packages/genkit-mcp/README.md
```diff
@@ -25,6 +22,63 @@ client = define_mcp_client(
...
+`restart()` and `close()` behave as they do for a defined client. `get_active_tools()` lists from the
+server on every call, registered or not, so what it returns is what the server serves now.
+
+## Tool names
+
+Every tool is named `<prefix>_<tool>`, so the tool above is `everything_echo`. The prefix is the
+provider name you registered the server under, `everything` here, and `tool_prefix=` overrides it.
+The separator is an underscore because Gemini and OpenAI reject a `/` in a function declaration
+name, and the prefix is checked against the same rules when the client is defined, so a name a model
+would refuse fails there rather than on your first `generate`.
```
huangjeff5: This says a name a model would refuse fails when the client is defined. A prefix of illegal characters does. A prefix that makes `<prefix>_<tool>` longer than 64 characters is logged and still registered, and the 400 arrives on the first `generate`.

Maybe we can say illegal characters fail at definition, and an over-long name is warned and passed through.

